PR #4, opened 2017-07-08
Networking Enhancements

SunnyBat:
Fix server crashes due to network read/write failures, Socket failures. Remove player model when disconnected.

### DIFF
--- a/NitroxClient/Communication/Client.cs
+++ b/NitroxClient/Communication/Client.cs
@@ -54,7 +54,14 @@ namespace NitroxClient.Communication
                 packetReceiver.PacketReceived(packet);                
             }
 
-            connection.BeginReceive(new AsyncCallback(DataReceived));
+            if (connection.Open)
+            {
+                connection.BeginReceive(new AsyncCallback(DataReceived));
+            } else
+            {
+                Console.WriteLine("Disconnected from server.");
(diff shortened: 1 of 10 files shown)


NitroxClient/Communication/Client.cs
```diff
@@ -54,7 +74,14 @@ private void DataReceived(IAsyncResult ar)
                 packetReceiver.PacketReceived(packet);                
             }
 
-            connection.BeginReceive(new AsyncCallback(DataReceived));
+            if (connection.Open)
+            {
+                connection.BeginReceive(new AsyncCallback(DataReceived));
+            } else
```
Sunrunner37: minor: mixing convention (else on new line) throughout these files. I'll add a style checker later to catch these.

NitroxClient/Communication/Client.cs
```diff
@@ -66,5 +93,25 @@ public void PacketSentSuccessful(IAsyncResult ar)
...
+            if (connection == null)
+            {
+                return false;
+            }
+            return connection.Open;
+        }
+
+        private void outputMessage(String msg)
+        {
+            if (testClient)
```
Sunrunner37: Maybe for the future we can pass in a logger as this looks be the only thing testClient is used for

PR #6, opened 2017-07-09
Fix NRE in HideEntities patch.

MarijnS95:
Checks for nullity on `VoxelandChunk`s returned by `get_chunks`.

### DIFF
--- a/NitroxPatcher/Patches/ClipMapManager_HideEntities_Patch.cs
+++ b/NitroxPatcher/Patches/ClipMapManager_HideEntities_Patch.cs
@@ -4,7 +4,6 @@ using System;
 using System.Collections.Generic;
 using System.Reflection;
 using System.Reflection.Emit;
-using UnityEngine;
 
 namespace NitroxPatcher.Patches
 {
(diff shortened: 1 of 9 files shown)


NitroxTest/Patcher/Patches/ClipMapManager_HideEntities_PatchTest.cs
```diff
@@ -1,29 +1,23 @@
...
-using NitroxModel.Helper;
-using NitroxPatcher;
 using NitroxTest.Patcher.Test;
 
 namespace NitroxTest.Patcher.Patches
 {
     [TestClass]
     public class ClipMapManager_HideEntities_PatchTest
     {
+        // TODO: Pass a valid ILGenerator instead of null.
```
Sunrunner37: Let's fix the unit tests before merging

NitroxPatcher/Patches/ClipMapManager_HideEntities_Patch.cs
```diff
@@ -19,17 +15,33 @@ public class ClipMapManager_HideEntities_Patch : NitroxPatch
...
+            var chunk = generator.DeclareLocal(typeof(VoxelandChunk));
             foreach (CodeInstruction instruction in instructions)
             {
                 if (instruction.opcode.Equals(INJECTION_OPCODE))
                 {
                     /*
-                     * Multiplayer.RemoveChunk(this.chunk.transform.position);
+                     * VoxelandChunk chunk = this.chunk;
+                     * if (chunk != null)
+                     *     Multiplayer.RemoveChunk(chunk.transform.position);
```
Sunrunner37: Would it be better to do Multiplayer.RemoveChunk(chunk) then check for null on the receiving side?  This would reduce the amount of IL that we need to inject.  I'll soon be migrating a lot of the business logic out of Multiplayer but for now it resides there.

Your call on this one.

PR #8, opened 2017-07-09
Unified project config

MarijnS95:
Uses some MSBuild targets to check whether the game is installed in the default Steam location. If not, copies a targets file (whose changes will be ignored) where a developer can enter the alternative location.
(The target file overrides the default location if it exists).

Because the dependencies and projects will end up in a single folder anyway, I merged all the bin folders into one to prevent all these duplicates from existing in different folders. This commit can be removed if this behavior is not desired.

Final commit contains another MSBuild target to copy generated assemblies to the Subnautica managed folder. I'm still checking out automated injection, so let me know if that's something we want.

The wiki setup page should be updated to reflect these changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@
 *.user
 *.userosscache
 *.sln.docstates
+DevVars.targets
 
 # User-specific files (MonoDevelop/Xamarin Studio)
 *.userprefs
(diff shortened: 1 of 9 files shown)


Sunrunner37:
I'll update the wiki docs with these changes some time today.

PR #9, opened 2017-07-12
ClientLogger added, more notifications, packets, server crash fix

SunnyBat:
-Added ClientLogger
--Able to specify whether to print to game, console, and debug messages
-Made PacketSender always deactivate when disconnected
-Remove all Player GameObjects when disconnected from server
-Small unused code cleanup
-Made server drop all packets until Authenticate packet received
-Added Connect packet
--Sent when player first authenticates
-Added fix for server crash when reading data into buffer on disconnected Socket
--Note: May have slightly weird interaction with NitroxServer.Listener.ClientAccepted() last line -- ClientAccepted() may be unnecessarily called again before disconnect is processed
--Hard to replicate without modifying receive code, will work on further


NitroxClient/Logger/ClientLogger.cs
```diff
@@ -0,0 +1,44 @@
...
+namespace NitroxClient.Logger
+{
+    public class ClientLogger
+    {
+        public const int LOG_CONSOLE = 1;
+        public const int LOG_ERRORMESSAGE = 2;
+        public const int LOG_DEBUG = 4;
+        private static int logLocations = LOG_ERRORMESSAGE;
+
+        // Set with combinaition of constants -- setLogLocation(LOG_CONSOLE | LOG_ERRORMESSAGE)
```
Sunrunner37: Honestly, I feel that bitmasking is a little weird in high level languages.  It is not immediately obvious and this is not an environment with space constraints.  It also makes for weird code throughout the implementation, ie: if ((logLocations & LOG_ERRORMESSAGE) != 0).  

Personally, I think a better implementation would be a LogLevel enumeration(None, Debug, Error) and a boolean for enabling console logging.

NitroxModel/Tcp/Connection.cs
```diff
@@ -65,7 +72,7 @@ public void SendPacket(Packet packet, AsyncCallback callback)
                 {
                     Socket.BeginSend(packetData, 0, packetData.Length, 0, callback, Socket);
                 }
-                catch (SocketException se)
+                catch (SocketException)
```
Sunrunner37: Some of these you may want to print the stack trace and message.

NitroxServer/Listener.cs
```diff
@@ -50,27 +50,48 @@ public void ClientAccepted(IAsyncResult ar)
...
-                UpdatePlayerPosition(player, packet);
+            bool authenticated = false;
 
-                if (!loggingPacketBlackList.Contains(packet.GetType()))
+            foreach (Packet packet in connection.GetPacketsFromRecievedData(ar))
+            {
+                if (authenticated)
+                {
+                    ParsePacket(connection, packet);
+                } else
```
Sunrunner37: minor: watch the style... else on a new line

NitroxClient/MonoBehaviours/Chat.cs
```diff
@@ -27,7 +27,7 @@ public void OnConsoleCommand_chat(NotificationCenter.Notification n)
                 }
 
                 Multiplayer.PacketSender.SendChatMessage(text);
-                ErrorMessage.AddMessage("Me: " + text);
+                ErrorMessage.AddMessage("Me: " + text); // Assumption: This method will only be called in-game
```
Sunrunner37: What does this comment mean?  This is a monobehaviour that should only be called in the game.

NitroxClient/Communication/Client.cs
```diff
@@ -43,26 +37,29 @@ public void Start(String ip)
...
             }
             catch (Exception e)
             {
-                outputMessage("Unable to connect to server");
+                ClientLogger.WriteLine("Unforeseen error when connecting: " + e.GetBaseException());
+                throw new InvalidOperationException("Unknown error while connecting");
             }
         }
 
         public void Stop()
```
Enflorence: General question about the changes here... Why move game-centric clean up logic here? This class seems to be concerned with basic socket communication here (i.e. receiving and sending packets). My concern with adding outside clean-up logic here is the obfuscation of the core goal of this class.

NitroxClient/Logger/ClientLogger.cs
SunnyBat: I agree that bitmasking is weird. I was conflicted between a boolean for each option and the bitmasking. I just went with bitmasking because it was a bit more condensed, which obviously isn't always better. My thought on this is to allow any combination of outputs, since you might add more options later.
I'll swap this over to methods for setting each individual option.

NitroxModel/Tcp/Connection.cs
SunnyBat: I don't really see why. These should only occur when there are read/write errors with the Socket. The connection is marked for closure when they occur. There's no real need to see the exception details IMO.

Sunrunner37:
Hey SunnyBat - just pinging you on these.  Would love to have them merged in.

SunnyBat:
Sorry for the long time between updates. I'm hoping to push some fixes to this soon.

NitroxServer/Listener.cs
```diff
@@ -52,29 +52,51 @@ public void ClientAccepted(IAsyncResult ar)
...
+        private void dropPacketsUntilAuthentication(IAsyncResult ar)
         {
             Connection connection = (Connection)ar.AsyncState;
-            
-            foreach(Packet packet in connection.GetPacketsFromRecievedData(ar))
-            {
-                Player player = GetPlayer(packet, connection);
-                connection.PlayerId = player.Id;
-                UpdatePlayerPosition(player, packet);
+            bool authenticated = false;
```
Sunrunner37: Is there way to consolidate this into a single function?  Maybe refactoring authenticated into a member of connection.

NitroxClient/Logger/ClientLogger.cs
```diff
@@ -0,0 +1,49 @@
...
+        {
+            Disabled = 0,
+            InGameMessages = 1,
+            ConsoleMessages = 2,
+            ConsoleDebug = 4
+        }
+        private static LogLevel logLocations = LogLevel.Disabled;
+
+        // Set with combinaition of enum -- setLogLocation(LOG_CONSOLE | LOG_ERRORMESSAGE)
+        public static void SetLogLocation(LogLevel location)
```
Sunrunner37: minor: setLogLevel (same with variable name above)

NitroxClient/Logger/ClientLogger.cs
```diff
@@ -0,0 +1,49 @@
...
+        {
+            ClientLogger.logLocations = location;
+            Console.WriteLine("Log location set to " + ClientLogger.logLocations);
+        }
+
+        public static void WriteLine(String msg)
```
Sunrunner37: Since we don't have the ability to write anything other than a line I would recommend simplifying these method names.  I've seen loggers usually use Info() and Debug().

NitroxClient/Logger/ClientLogger.cs
```diff
@@ -0,0 +1,49 @@
...
+        {
+            if ((logLocations & LogLevel.ConsoleMessages) != 0) // == LogLevel.ConsoleMessage works as well, but is more verbose
+            {
+                Console.WriteLine(msg);
```
Sunrunner37: Let's add the prefix [Nitrox] to anything that is written to the console. (Same below)

NitroxModel/Tcp/Connection.cs
```diff
@@ -17,11 +20,34 @@ public Connection(Socket socket)
...
+            }
+        }
         
         public void BeginReceive(AsyncCallback callback)
         {
-            Socket.BeginReceive(MessageBuffer.ReceivingBuffer, 0, MessageBuffer.RECEIVING_BUFFER_SIZE, 0, callback, this);
+            try
+            {
+                Socket.BeginReceive(MessageBuffer.ReceivingBuffer, 0, MessageBuffer.RECEIVING_BUFFER_SIZE, 0, callback, this);
+            } catch (SocketException)
```
Sunrunner37: minor: watch style, catch on new line.  Also it may be worth logging the exception name.

NitroxServer/Listener.cs
```diff
@@ -52,29 +52,51 @@ public void ClientAccepted(IAsyncResult ar)
             Console.WriteLine("New client connected");
 
             Socket socket = (Socket)ar.AsyncState;
-            
+
             Connection connection = new Connection(socket.EndAccept(ar)); // TODO: Will this throw an error if timed correctly?
+            connection.BeginReceive(new AsyncCallback(dropPacketsUntilAuthentication));
             connection.BeginReceive(new AsyncCallback(DataReceived));
```
Sunrunner37: There are two Async BeginReceives here - is this behaving how you intended?

NitroxClient/GameLogic/PlayerGameObjectManager.cs
```diff
@@ -111,10 +111,23 @@ private GameObject CreateOtherPlayer(string playerId)
...
+                GameObject.DestroyImmediate(player);
+                gameObjectByPlayerId.Remove(playerId);
+            }
+        }
+
+        public void RemoveAllPlayerGameObjects()
+        {
+            String[] playerIds = new String[gameObjectByPlayerId.Keys.Count];
+            gameObjectByPlayerId.Keys.CopyTo(playerIds, 0);
+            foreach (string playerId in playerIds)
```
Sunrunner37: Why not iterate directly over the dictionary keys?

NitroxClient/Communication/Packets/Processors/DisconnectProcessor.cs
```diff
@@ -16,6 +17,7 @@ public DisconnectProcessor(PlayerGameObjectManager playerGameObjectManager)
         public override void Process(Disconnect disconnect)
         {
             playerGameObjectManager.RemovePlayerGameObject(disconnect.PlayerId);
+            ClientLogger.WriteLine(disconnect.PlayerId + " disconnected");
```
Sunrunner37: I'm not sure we should mix messages that need to always go to the player.  If someone had logging turned off then the player would not see this message.  Would recommend having a new method for that.

SunnyBat:
OK, let's see how this commit does. Still getting used to C#, so this still might have a few rough edges.

Sunrunner37:
We have recently refactored the entire server to use a different authentication scheme.  You'll want to pull in the latest changes and probably remove your server changes. Everything else looks fine to me.

PR #10, opened 2017-07-12
Better animations

nesrak1:
* Other players now have separate animations
* Movement swimming animations
* Basic packet sending for swimming / walking animation toggle
* Ani tester command
---
* iTween reference fix
* Pickup fix


NitroxClient/GameLogic/PlayerGameObjectManager.cs
```diff
@@ -17,6 +20,30 @@ public void UpdatePlayerPosition(string playerId, Vector3 position, Quaternion r
...
+
+            switch (type)
+            {
+                case AnimChangeType.Underwater:
+                    controller.SetBool("is_underwater", value);
+                    break;
+            }
+        }
+
         public void HidePlayerGameObject(string playerId)
```
Enflorence: I recommend refactoring this - perhaps breaking this up into helper functions and avoiding magic values.

NitroxClient/Communication/PacketSender.cs
```diff
@@ -58,6 +58,12 @@ public void PickupItem(GameObject gameObject, String techType)
             PickupItem pickupItem = new PickupItem(PlayerId, ApiHelper.Vector3(itemPosition), guid, techType);
             Send(pickupItem);
         }
+        
+        public void PickupItem(Vector3 itemPosition, String guid, String techType)
```
Sunrunner37: minor optimization: you can have the above function call this one.

NitroxClient/GameLogic/PlayerGameObjectManager.cs
```diff
@@ -17,6 +20,30 @@ public void UpdatePlayerPosition(string playerId, Vector3 position, Quaternion r
...
         }
 
+        public void UpdateAnimation(string playerId, AnimChangeType type, AnimChangeState state)
+        {
+            GameObject player = GetPlayerGameObject(playerId);
+            GameObject playerView = player.transform.Find("player_view").gameObject;
+            AnimationController controller = playerView.GetComponent<AnimationController>();
+
+            bool value;
+            if (state == AnimChangeState.Off)
```
Sunrunner37: you can do:  bool value = (state != AnimChangeState.off).  Let's also name 'value' something more descriptive.

NitroxClient/GameLogic/PlayerGameObjectManager.cs
```diff
@@ -45,7 +72,30 @@ private GameObject createOtherPlayer(string playerId)
             body.transform.parent.gameObject.GetComponent<Player>().head.shadowCastingMode = UnityEngine.Rendering.ShadowCastingMode.On;
             GameObject bodyCopy = UnityEngine.Object.Instantiate(body);
             body.transform.parent.gameObject.GetComponent<Player>().head.shadowCastingMode = UnityEngine.Rendering.ShadowCastingMode.ShadowsOnly;
-            bodyCopy.transform.Find("player_view").gameObject.GetComponent<ArmsController>().smoothSpeed = 0; //Disables the other character's move animations
+
+            //Get player
+            GameObject playerView = bodyCopy.transform.Find("player_view").gameObject;
```
Sunrunner37: For the future(not this pull request), we'll want to probably tuck this into a clean builder object.

NitroxClient/MonoBehaviours/AnimationController.cs
```diff
@@ -0,0 +1,55 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using UnityEngine;
+
+namespace NitroxClient.MonoBehaviours
+{
+    public class AnimationController : MonoBehaviour
+    {
+        public static readonly float SMOOTHING_SPEED = 0.95f;
```
Sunrunner37: Will this constant be the same for all animations?  For the future(not this commit) may need to adjust this based on latency.

NitroxClient/MonoBehaviours/AnimationController.cs
```diff
@@ -0,0 +1,55 @@
...
+        public Animator animator;
+        private Dictionary<string, bool> animatorBools = new Dictionary<string, bool>()
```
Sunrunner37: minor: more descriptive name.  animationStatusById

NitroxClient/MonoBehaviours/AnimationController.cs
```diff
@@ -0,0 +1,55 @@
...
+        {
+            foreach (KeyValuePair<string, bool> kvp in animatorBools)
+            {
+                //For whatever reason, attempting to setbool once most of the time won't work
+                //Will investigate soon but this seems to work for now
+                animator.SetBool(kvp.Key, kvp.Value);
+            }
+        }
+
+        Vector3 lastPosition = Vector3.zero;
```
Sunrunner37: Let's keep all local variables at the top of the file.

NitroxClient/GameLogic/PlayerGameObjectManager.cs
nesrak1: In the future, this is going to change a lot. I originally had this, and depending on how the other animations turn out, I may either have to keep this and modify a lot or use this.

NitroxClient/MonoBehaviours/AnimationController.cs
nesrak1: This is a client animation and doesn't have anything to do with latency.

Sunrunner37: Correct me if I'm wrong, if the other player initiated a player animation 65ms in the past, are there an instances were it is prudent to subtract that from the duration to keep the clients in sync?

nesrak1: What will need to be adjusted for latency is the movement helper, but this variable here just moves the legs back down when the player slows down swimming.

NitroxClient/Communication/PacketSender.cs
```diff
@@ -56,6 +56,12 @@ public void PickupItem(GameObject gameObject, String techType)
             String guid = GuidHelper.GetGuid(gameObject);
             Vector3 itemPosition = gameObject.transform.position;
 
+            PickupItem(itemPosition, guid, techType);
```
Sunrunner37: This looks like it will send the message twice. (Once in PickupItem and then on the next line)

NitroxModel/Helper/ReflectionHelper.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace NitroxModel.Helper
+{
+    public static class ReflectionHelper
```
Sunrunner37: I would do Validate.NotNull(var, o.getType() + " does not have a x named" + name) instead of allowing these functions to silently fail by returning null.  Makes the code cleaner(no null check) and is a controlled failure(no NRE in the caller).  Also better naming for mi/fi would be good.

NitroxPatcher/Patches/ArmsController_Update_Patch.cs
```diff
@@ -0,0 +1,45 @@
...
+                leftAim.ReflectionCall("Update", __instance.ikToggleTime);
+                rightAim.ReflectionCall("Update", __instance.ikToggleTime);
+                __instance.ReflectionCall("UpdateHandIKWeights");
+
+                return false;
+            }
+            return true;
+        }
+        
+        /*
```
Sunrunner37: minor: I don't think we need this comment anymore

ClientTester/Commands/DefaultCommands/AniCommand.cs
```diff
@@ -0,0 +1,38 @@
...
+            if (args.Length < 2)
+            {
+                CommandManager.NotEnoughArgumentsMessage(2, Syntax);
+                return;
+            }
+            AnimChangeState state;
+            if (args[1] == "on" || args[1] == "1" || args[1] == "true")
+            {
+                state = AnimChangeState.On;
+            } else if (args[1] == "off" || args[1] == "0" || args[1] == "false")
```
Sunrunner37: minor: mixing styles here - else if/else on a new line

NitroxClient/GameLogic/PlayerGameObjectManager.cs
```diff
@@ -17,6 +20,23 @@ public void UpdatePlayerPosition(string playerId, Vector3 position, Quaternion r
...
+            GameObject player = GetPlayerGameObject(playerId);
+            GameObject playerView = player.transform.Find("player_view").gameObject;
+            AnimationController controller = playerView.GetComponent<AnimationController>();
+
+            bool animationValue;
+
+            switch (type)
+            {
+                case AnimChangeType.Underwater:
+                    animationValue = (state != AnimChangeState.Off);
```
Sunrunner37: should this be on the same line as animationValue's initialization?  bool animationValue = (state != AnimChangeState.Off)

nesrak1: Just thinking for the future

PR #11, opened 2017-07-14
Fixed warnings and messages.

MarijnS95:
Gets rid of all the warnings and 'informative' messages that are shown on the current master branch.

Some are C#6/7 specific, so we should discuss which version we want to support at a minimum. I'll remove those changes once we've settled on a version. Some of the messages are about improper naming, as currently `lowerCamelCasing` is sometimes used for private methods, whilst visual studio is set to `UpperCamelCasing` to default. IMHO it doesn't matter which one we use, as long as we stay consistent. A styleguide would be really helpful (perhaps a config for VS if such a thing is even possible?).

Also, I think we should enforce that PR code is free of warnings. Those on the current master branch are harmless (unused variables), but more serious ones will get lost in a pile of these. And it forces everyone to keep their code clean.

In the end, this PR is not so much about getting these changes in to the master branch but showing which pieces of code VS complains about, and to get a discussion started about this topic.

Some more things which VS doesn't really complain about, but are shown a bit faded/greyed-out, and are generally not that nice to see:

- Superfluous use of `this.`;
- Unnecessary cast;
- Unused using statements;
- Unsorted using statements.

These should be covered in the styleguide as well.

### DIFF
--- a/ClientTester/MultiplayerClient.cs
+++ b/ClientTester/MultiplayerClient.cs
@@ -26,7 +26,7 @@ namespace ClientTester
         public void Start(String ip)
         {
             client.Start(ip);
-            if (client.isConnected())
+            if (client.IsConnected())
             {
                 PacketSender.Active = true;
                 PacketSender.Authenticate();
--- a/ClientTester/Program.cs
+++ b/ClientTester/Program.cs
@@ -16,6 +16,7 @@ namespace ClientTester
 
         private static Vector3 clientPos = new Vector3(-50f, -2f, -38f);
 
+        [STAThread]
         static void Main(string[] args)
         {
             String playerId1 = "sunrunner";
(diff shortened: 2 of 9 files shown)


Sunrunner37:
Thanks for putting this together Marjin.  I've discussed this post with Enflorence and he will be responsible for putting together style guidelines and an automated system for verifying.  Completely agree with pull requests not bringing in any new warning. 

As for this pull request, I'll merge it in.  However, I think we may want to log a few of these errors to the window instead of dropping the exception variable.

MarijnS95:
Good to hear. Once this document is up, how about cleaning up all the code on master so there's a proper base to continue on?

> However, I think we may want to log a few of these errors to the window instead of dropping the exception variable.

I think SunnyBat is working on the logging already so I didn't really touch that, but it's definitely a good idea - there's some places where I was wondering what the error was, but in most cases the try-catch block encapsulates a relatively small piece of code, where a short log is usually enough. Printing the `.Message` field (not the entire stacktrace) is a good inbetweener in those cases.

Also this merge will break the build for everyone not on VS2017/C#7, because there's [pattern matching](https://docs.microsoft.com/en-us/dotnet/csharp/pattern-matching) [here](https://github.com/Sunrunner37/Nitrox/pull/11/files#diff-79714fb20ea7a00843d9747bfab45db9R55).

PR #12, opened 2017-07-19
Multiplayer menu

nesrak1:
* Can add/delete servers in the main menu
* Set username before join


NitroxClient/MonoBehaviours/MultiplayerButton.cs
```diff
@@ -0,0 +1,234 @@
...
+using System.IO;
+using System.Linq;
+using System.Text;
+using UnityEngine;
+using UnityEngine.SceneManagement;
+using UnityEngine.UI;
+using UWE;
+using NitroxClient.Communication;
+
+namespace NitroxClient.MonoBehaviours
```
Enflorence: I recommend moving classes MainMenuMultiplayerPanel and JoinServerScript into their own files to make it easier to maintain down the road.

NitroxClient/MonoBehaviours/Multiplayer.cs
```diff
@@ -17,6 +17,7 @@ public class Multiplayer : MonoBehaviour
 
         public static Logic Logic;
         public static PacketSender PacketSender; //TODO: migrate logic out of and remove as member variable
+        public static Multiplayer main;
```
Enflorence: Is there any other way to reference the Multiplayer component without this static reference? At the very least, I would recommend leveraging a static method to grab an instance.

NitroxClient/MonoBehaviours/MultiplayerButton.cs
Sunrunner37: Yeah maybe put all of this stuff in a new namespace NitroxClient.MonoBehaviours.Gui.MainMenu

NitroxClient/MonoBehaviours/Multiplayer.cs
nesrak1: We had a discussion about this and I didn't really know what to do. The resolution was that subnautica already uses plenty of singletons, but I do agree it is a bit silly.

NitroxClient/MonoBehaviours/MultiplayerButton.cs
```diff
@@ -0,0 +1,234 @@
...
+        public void Awake()
+        {
+            MultiplayerMenuMods();
+        }
+
+        private void MultiplayerMenuMods()
+        {
+            GameObject startButton = GameObject.Find("Menu canvas/Panel/MainMenu/PrimaryOptions/MenuButtons/ButtonPlay");
+            GameObject showLoadedMultiplayer = Instantiate(startButton);
+            Text buttonText = showLoadedMultiplayer.transform.Find("Circle/Bar/Text").gameObject.GetComponent<Text>() as Text;
```
Sunrunner37: minor: do we need to say "as Text" - it looks like you are already using the generic method call in get component;

NitroxClient/MonoBehaviours/MultiplayerButton.cs
```diff
@@ -0,0 +1,234 @@
...
+                HideNewServer();
+            }
+            if (GUI.Button(new Rect(Screen.width / 2 - 250, Screen.height / 2 + 100, 500, 50), "Cancel"))
+            {
+                HideNewServer();
+            }
+        }
+    }
+
+    public class JoinServerScript : MonoBehaviour
```
Sunrunner37: minor: would recommend dropping the word Script as it is implied by the monobehaviour namespace.  Maybe brainstorm a more descriptive postfix(Gui maybe?)

NitroxClient/MonoBehaviours/MultiplayerButton.cs
```diff
@@ -0,0 +1,234 @@
...
+        }
+
+        public void OnGUI()
+        {
+            if (!showingUsername)
+            {
+                return;
+            }
+            username = GUI.TextField(new Rect(Screen.width / 2 - 250, Screen.height / 2 - 25, 500, 50), username);
+            if (GUI.Button(new Rect(Screen.width / 2 - 250, Screen.height / 2 + 25, 500, 50), "Add server"))
```
Sunrunner37: Note: for the future we'll need to research how they accomplish localization.  I'll add an entry into the Trello area.

NitroxPatcher/Patches/uGUI_MainMenu_Start_Patch.cs
```diff
@@ -0,0 +1,27 @@
...
+namespace NitroxPatcher.Patches
+{
+    public class uGUI_MainMenu_Start_Patch : NitroxPatch
+    {
+        public static readonly Type TARGET_CLASS = typeof(uGUI_MainMenu);
+        public static readonly MethodInfo TARGET_METHOD = TARGET_CLASS.GetMethod("Start", BindingFlags.NonPublic | BindingFlags.Instance);
+
+        public static void Postfix(uGUI_MainMenu __instance)
+        {
+            new GameObject().AddComponent<MultiplayerButton>();
```
Sunrunner37: should we use uGui_MainMenu's gameObject instead of creating a new one?

nesrak1: Sure, why not

PR #15, opened 2017-07-25
Minor improvements

MarijnS95:
Many minor improvements.

- Automatic patch discovery;
- Automatic packet+packetprocessor discovery (+ unittests);
- Refactor of player functions into class;
- Player rotation (and animationrotation sync);
- Live reloading of methods (TODO: write a little wiki page detailing this);
- And some other changes.


NitroxPatcher/Main.cs
```diff
@@ -1,49 +1,45 @@
...
             HarmonyInstance.DEBUG = false;
 
             HarmonyInstance harmony = HarmonyInstance.Create("com.nitroxmod.harmony");
 
-            foreach(NitroxPatch patch in patches)
-            {
-                Console.WriteLine("[NITROX] Applying " + patch.GetType());
-                patch.Patch(harmony);
-            }
+            //AppDomain.CurrentDomain.GetAssemblies()
```
Enflorence: Minor: remove commented code

NitroxReloader/Reloader.cs
```diff
@@ -0,0 +1,244 @@
...
+                var handle = GetRuntimeMethodHandle(method);
+                RuntimeHelpers.PrepareMethod(handle);
+                return handle.GetFunctionPointer();
+            }
+
+            public static unsafe void WriteJump(IntPtr original, IntPtr destination)
+            {
+                Console.WriteLine("Reloader: Writing jump to {0} at {1}", destination, original);
+                bool x64 = IntPtr.Size == sizeof(long);
+#if true
```
Enflorence: Is this directive conditional necessary?

NitroxReloader/Reloader.cs
```diff
@@ -0,0 +1,244 @@
...
+
+    public class Reloader
+    {
+        private static readonly HashSet<string> assemblyWhitelist = new HashSet<string>() {
+            "NitroxModel.dll",
+            "NitroxClient.dll",
+            "NitroxPatcher.dll",
+        };
+
+        // TODO: Reloadable attribute for classes? (To define all methods in there to be reloadable)
```
Enflorence: It might be better to define these TODOs as tasks in Trello than in code.

NitroxReloader/Reloader.cs
```diff
@@ -0,0 +1,244 @@
...
+// Modified to remove RimWorld dependency, and cleanups.
+
+namespace NitroxReloader
+{
+    [AttributeUsage(AttributeTargets.Method)]
+    public class ReloadableMethod : Attribute
+    {
+    }
+
+    public class Reloader
```
Enflorence: I would consider refactoring the classes into individual files to make it easier to maintain and evolve.

NitroxReloader/Reloader.cs
MarijnS95: It was to quickly toggle between the two methods of writing the jump, but I couldn't figure out why the second one doesn't do what it's supposed to do so I removed it.

NitroxClient/GameLogic/PlayerManager.cs
```diff
@@ -0,0 +1,46 @@
...
+            set
+            {
+                if (value == null)
+                    RemovePlayer(playerId);
+                else
+                    playersById[playerId] = value;
+            }
+        }
+
+        public RemotePlayer FindPlayer(string playerId, bool createPlayer = false)
```
Sunrunner37: Optional RemotePlayer to encapsulate null OR always instantiate it if it is not found.  I don't think there is really a good reason for a null reference to be floating around.

NitroxClient/GameLogic/PlayerManager.cs
```diff
@@ -0,0 +1,46 @@
...
+            RemotePlayer player;
+            if (!playersById.TryGetValue(playerId, out player) && createPlayer)
+                player = playersById[playerId] = new RemotePlayer(playerId);
+            return player;
+        }
+
+        public void ForPlayer(string playerId, Action<RemotePlayer> action, bool createPlayer = false)
+        {
+            var player = FindPlayer(playerId, createPlayer);
+            if (player != null)
```
Sunrunner37: Please always use curly braces - no one liners like this.  There are a few other places in the PR - do a quick glance through.

NitroxClient/GameLogic/PlayerManager.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        public RemotePlayer FindPlayer(string playerId, bool createPlayer = false)
+        {
+            RemotePlayer player;
+            if (!playersById.TryGetValue(playerId, out player) && createPlayer)
+                player = playersById[playerId] = new RemotePlayer(playerId);
+            return player;
+        }
+
+        public void ForPlayer(string playerId, Action<RemotePlayer> action, bool createPlayer = false)
```
Sunrunner37: Can you make this method name more descriptive.  It isn't immediately obvious what it does

NitroxClient/GameLogic/PlayerManager.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            var player = FindPlayer(playerId, createPlayer);
+            if (player != null)
+                action(player);
+        }
+
+        public void RemovePlayer(string playerId)
+        {
+            playersById[playerId].Destroy();
```
Sunrunner37: Looks like a possible NRE if the player does not exist. (This is the type of situation that optional protects against)

NitroxModel/Packets/VehicleMovement.cs
```diff
@@ -21,7 +19,7 @@ public VehicleMovement(String playerId, Vector3 playerPosition, Quaternion rotat
 
         public override string ToString()
         {
-            return "[VehicleMovement - playerId: " + PlayerId + " position: " + PlayerPosition + " Rotation: " + Rotation + " TechType: " + TechType + " guid: " + Guid + "]";
+            return "[VehicleMovement - TechType: " + TechType + " guid: " + Guid + "]\n\t" + base.ToString();
```
Sunrunner37: base.toString()?

MarijnS95: Yes, the base class of `VehicleMovement` (which is `Movement`) already stringifies these fields, so it's better to reuse that functionality than to copy it.

PR #16, opened 2017-07-27
Cleanup #15

MarijnS95:
Follows https://github.com/Sunrunner37/Nitrox/pull/15.

Cleanups regarding comments.

I've never before used the `Optional<>` structure in imperative programming, only in functional programming, so I've added one of those constructs (and will add more soon). What is your opinion on this? I personally don't like the clutter (having a variable for the optional, then a condition, then another variable with the value that is held by the `Optional<>`).

Also, another question regarding formatting: spaces between an `if` statement and the open brace? (this applies to other control statements as well).
I have 34 matches of `if(` on master, and 245 of `if (`. My editor is set to insert it.

### DIFF
--- a/NitroxClient/Communication/Packets/Processors/AnimationProcessor.cs
+++ b/NitroxClient/Communication/Packets/Processors/AnimationProcessor.cs
@@ -16,7 +16,11 @@ namespace NitroxClient.Communication.Packets.Processors
 
         public override void Process(AnimationChangeEvent animEvent)
         {
-            remotePlayerManager.ForPlayer(animEvent.PlayerId, p => p.UpdateAnimation((AnimChangeType)animEvent.Type, (AnimChangeState)animEvent.State));
+            var opPlayer = remotePlayerManager.Find(animEvent.PlayerId);
+            if (opPlayer.IsPresent())
+            {
+                opPlayer.Get().UpdateAnimation((AnimChangeType)animEvent.Type, (AnimChangeState)animEvent.State);
+            }
         }
     }
 }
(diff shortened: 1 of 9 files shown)


Sunrunner37:
Hey Marin,

Enflorence and I spoke at length about Optional due to this post.  It does have its roots in functional programming but has made its way outside(this implementation is similar to java's Optional.)  

Some items that we concluded:

1) Although subjective, we agreed that the functional style of optional tends to be jarring.  Compare the two screenshots:

https://media.discordapp.net/attachments/222089961195700225/340323092955922435/unknown.png
https://media.discordapp.net/attachments/222089961195700225/340322712964825088/unknown.png

Saving two lines didn't justify removal of the language's structural elements.  Especially since the reader needs to context switch into a functional mindset to understand the implementation.  Although experienced programmers can context switch easily, the structural elements provide immediate understanding.

2) We believe that adopting the functional style will make people more likely to ignore the "else" clause whether it be logging or exception throwing.  There are many places in the code base that not having a value in the optional is something we should print to the log so we can follow up on it later.

3) The majority of the code base already following the new variable convention.  Since a pattern is already established, we don't think it justifies going back and refactoring code to be functional and think consistency is better.

4) The alternative to optional is to allow null to surface to callers.  We agreed that this will, more often than not, incentivise people to ignore the null check in return values.  There is also no way for programmers to know the fields that are intended to be nullable.  

For your other question... 'if' with a space should be the correct style.

MarijnS95:
Thanks for the in-depth explanation. I've removed the functional style from the commits.

I suggest you to quickly test the changes before merging - I've made these modifications on top of my experimental compatibility branch and rebased them back on master for the PR.

PR #21, opened 2017-08-26
BatchCell sending

nesrak1:


NitroxServer/Communication/Packets/Processors/WorldSending/AskBatchCellProcessor.cs
```diff
@@ -0,0 +1,68 @@
...
+            lock (player.chunkQueues)
+            {
+                Queue<Chunk> batchQueue = player.chunkQueues;
+                if (batchQueue.Count == 0)
+                {
+                    foreach (string file in Directory.GetFiles(folderName + "\\"))
+                    {
+                        string[] chunkName = Path.GetFileNameWithoutExtension(file).Split('-');
+                        Int3 chunkCoords = new Int3(int.Parse(chunkName[2]), int.Parse(chunkName[3]), int.Parse(chunkName[4]));
+                        byte[] data = File.ReadAllBytes(file);
```
Sunrunner37: Maybe these chunks should be kept in a persistent memory store so they don't have to be loaded from a file every time people request them.  If you did this you can store an ID in the chunk queue(instead of duplicating all of this data in memory) and do a lookup before sending the data.

NitroxServer/Communication/Packets/Processors/WorldSending/AskBatchCellProcessor.cs
```diff
@@ -0,0 +1,68 @@
...
+                        batchQueue.Enqueue(new Chunk(chunkCoords, data));
+                    }
+                }
+
+                ReceiveBatchCell chunkPacket = new ReceiveBatchCell(player.Id, batchQueue.Peek().chunkData, batchQueue.Peek().chunkCoord, batchQueue.Count > 1);
+                batchQueue.Dequeue();
```
Sunrunner37: I'm debating on if the queue is sparing the player and headache.  In my mind this is what will end up happening: 

1) Player will request batches 
2) Server will send one 
3) Player will Process
4) Server will send one
5) Player will process

If the players were on LAN then the latency between the player requesting a packet and the server sending it will be close to zero.  That means that the player would be receiving a chunk to process in the next frame.  Effectively equivalent to just sending all of the chunks at once.  

I guess the benefit would be that you at least get a frame in and on higher latencies you can get many frames in.  Probably worth having after thinking out loud.

NitroxModel/Packets/WorldSending/AskWorldEvent.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace NitroxModel.Packets.WorldSending
+{
+    class AskWorldEvent
```
Sunrunner37: Can we remove the unused classes and add them back in when we start using the functionality?

NitroxClient/GameLogic/Logic.cs
```diff
@@ -17,6 +17,7 @@ public class Logic
         public Cyclops Cyclops { get; private set; }
         public Interior Interior { get; private set; }
         public MobileVehicleBay MobileVehicleBay { get; private set; }
+        public ChunkSerializer Serializer { get; private set; }
```
Sunrunner37: Can we rename this Chunks?  Multiplayer.Logic.Chunks is more explicit as to what this is doing.

NitroxServer/Player.cs
```diff
@@ -1,17 +1,22 @@
...
+using System.Collections;
+using System.Collections.Generic;
 
 namespace NitroxServer
 {
     public class Player : IProcessorContext
     {
         public String Id { get; private set; }
         public Vector3 Position { get; set; }
+        public Queue<Chunk> chunkQueues { get; private set; }
```
Sunrunner37: Let's name something more verbose like ChunksPendingTransmission

NitroxClient/GameLogic/ChunkSerializer.cs
```diff
@@ -0,0 +1,63 @@
...
+using NitroxModel.Packets;
+using System.Collections.Generic;
+using System.IO;
+using System.Reflection;
+
+namespace NitroxClient.GameLogic
+{
+    public class ChunkSerializer
+    {
+        private readonly string SERVER_LOCATION = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
```
Sunrunner37: Unused

NitroxServer/Communication/Packets/Processors/WorldSending/AskBatchCellProcessor.cs
nesrak1: There were problems sending chunks to the client at one time. There could be a fix for that but I guess I just went with this as I was too lazy to do otherwise

NitroxModel/NitroxModel.csproj
```diff
@@ -75,6 +79,16 @@
     <Compile Include="Packets\ItemPosition.cs" />
     <Compile Include="Packets\PlaceFurniture.cs" />
     <Compile Include="Packets\VehicleMovement.cs" />
+    <Compile Include="Packets\WorldSending\AskBatchCell.cs" />
+    <Compile Include="Packets\WorldSending\AskGlobalRoot.cs" />
+    <Compile Include="Packets\WorldSending\AskWorldEvent.cs" />
```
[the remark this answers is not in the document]
nesrak1: It was renamed to make it easier to differentiate Request and Receive

NitroxModel/Packets/WorldSending/AskBatchCell.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace NitroxModel.Packets.WorldSending
+{
+    [Serializable]
+    public class AskBatchCell : AuthenticatedPacket
+    {
+        public bool FromChunkQueue { get; private set; }
+        public Int3 ChunkLocation { get; private set; }
+
+        public AskBatchCell(string playerId, bool fromChunkQueue) : this(playerId, fromChunkQueue, new Int3(0, 0, 0)) { }
```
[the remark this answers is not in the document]
nesrak1: This will be used in code later

PR #31, opened 2017-09-04
Cyclops change name

Jannify:


NitroxPatcher/Patches/SubNameInput_OnNameChange_Patch.cs
```diff
@@ -0,0 +1,28 @@
...
+namespace NitroxPatcher.Patches
+{
+    public class SubNameInput_OnNameChange_Patch : NitroxPatch
+    {
+        public static readonly Type TARGET_CLASS = typeof(SubNameInput);
+        public static readonly MethodInfo TARGET_METHOD = TARGET_CLASS.GetMethod("OnNameChange", BindingFlags.Public | BindingFlags.Instance);
+
+        public static void Postfix(SubNameInput __instance)
+        {
+            SubName subname = (SubName)__instance.ReflectionGet("target");
```
Sunrunner37: It looks like OnNameChange() has a null check for target.  We should probably put in a similar check here.

NitroxModel/Packets/CyclopsChangeName.cs
```diff
@@ -0,0 +1,22 @@
...
+
+        public CyclopsChangeName(String playerId, String guid, string name) : base(playerId)
+        {
+            this.Guid = guid;
+            this.Name = name;
+        }
+
+        public override string ToString()
+        {
+            return "[CyclopsActivateShield PlayerId: " + PlayerId + " Guid: " + Guid + " Name: " + Name + "]";
```
Sunrunner37: "CyclopsChangeName"

NitroxClient/Communication/Packets/Processors/CyclopsChangeNameProcessor.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        public CyclopsChangeNameProcessor(PacketSender packetSender)
+        {
+            this.packetSender = packetSender;
+        }
+
+        public override void Process(CyclopsChangeName namePacket)
+        {
+            Optional<GameObject> opCyclops = GuidHelper.GetObjectFrom(namePacket.Guid);
+            SubName subname = (SubName)opCyclops.Get().GetComponentInChildren<CyclopsNameScreenProxy>().subNameInput.ReflectionGet("target"); ;
```
Sunrunner37: You are doing opCyclops.Get() before the IsPresent call.  This can result in a null pointer.

Sunrunner37: We should verify that a CyclopsNameScreenProxy was found before trying to use it.

NitroxPatcher/Patches/SubNameInput_OnNameChange_Patch.cs
```diff
@@ -0,0 +1,28 @@
...
+
+            String guid = GuidHelper.GetGuid(Player.main.GetCurrentSub().gameObject);
```
MarijnS95: You might want to add a null-check here just in case the player is't correctly in the Cyclops.

For instance teleporting causes the player to not be placed in the cyclops every time.

MarijnS95: There's a potential NRE here. If players teleport into the sub (`warpto`, `warpforward`, whatever), they are not correctly attached to it and `GetCurrentSub()` will return null.

It's not your fault though, but a bug in the game itself.

NitroxClient/Communication/Packets/Processors/CyclopsChangeNameProcessor.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        public override void Process(CyclopsChangeName namePacket)
+        {
+            Optional<GameObject> opCyclops = GuidHelper.GetObjectFrom(namePacket.Guid);
+
+            if (opCyclops.IsPresent())
+            {
+                CyclopsNameScreenProxy ScreenProxy = opCyclops.Get().GetComponentInChildren<CyclopsNameScreenProxy>();
+
+                if (ScreenProxy != null)
```
Sunrunner37: Future reference: the current convention is to have an else to log that this was not found. This prevents us from missing a silent failure.

PR #32, opened 2017-09-08
Added player death notification

Enflorence:


Sunrunner37:
Can you do a merge and I'll pull these changes in?

Enflorence:
Done. Please review - this commit was such a long time ago, I actually thought it was already merged.

NitroxClient/Communication/Packets/Processors/PlayerDeathProcessor.cs
```diff
@@ -0,0 +1,15 @@
...
+using NitroxModel.Packets;
+
+namespace NitroxClient.Communication.Packets.Processors
+{
+    class PlayerDeathProcessor : ClientPacketProcessor<PlayerDeathEvent>
+    {
+        public override void Process(PlayerDeathEvent playerDeath)
+        {
+            //TODO: Add any death related triggers (i.e. scoreboard updates, rewards, etc.)
+            ClientLogger.IngameMessage(playerDeath.PlayerId + " died");
```
Sunrunner37: This should be 'Log' - ClientLogger was renamed.

NitroxClient/GameLogic/PlayerDeath.cs
```diff
@@ -0,0 +1,22 @@
+﻿using NitroxClient.Communication;
+using NitroxModel.Packets;
+using UnityEngine;
+
+namespace NitroxClient.GameLogic
+{
+    public class PlayerDeath
```
Sunrunner37: Can we change this to PlayerEvents or something similar?  Just so we can tuck similar logic in with it.

NitroxModel/Packets/PlayerDeathEvent.cs
```diff
@@ -0,0 +1,16 @@
+﻿using NitroxModel.DataStructures;
+using System;
+
+namespace NitroxModel.Packets
+{
+    [Serializable]
+    public class PlayerDeathEvent : AuthenticatedPacket
+    {
+        public Vector3 DeathPosition { get; private set; }
```
Sunrunner37: minor: we removed all of the 'private set' as these are not required for the immutable objects.

NitroxClient/Communication/Packets/Processors/PlayerDeathProcessor.cs
Sunrunner37: Ignore this comment - it was on old diff...

NitroxClient/GameLogic/PlayerDeath.cs
Enflorence: I thought about doing this initially but strayed away because I didn't want a "god" player event class. I felt like there could be future enhancements made to player death alone that may not necessarily transition into other player-driven events.

PR #38, opened 2017-09-15
Cyclops change engine state

Jannify:


NitroxClient/Communication/Packets/Processors/CyclopsToggleEngineStateProcessor.cs
```diff
@@ -0,0 +1,43 @@
...
+
+            if (opCyclops.IsPresent())
+            {
+                CyclopsEngineChangeState engineState = opCyclops.Get().GetComponentInChildren<CyclopsEngineChangeState>();
+
+                if (enginePacket.IsOn == engineState.motorMode.engineOn)
+                {
+                    using (packetSender.Suppress<CyclopsEngineChangeState>())
+                    {
+                        engineState.ReflectionSet("startEngine", true);
```
MarijnS95: Is this necessary? `OnClick` will set it to true anyway when the Cyclops if turned off. And when the Cylcops is turned on, this variable is not even read.

Jannify: Yes it is necessary. Prefix() will be set __state to !motor-state and no Packet will send back...

MarijnS95: You're already encapsulating this method in a suppressor.

Jannify:
Limit is not implemented. I don't know how I do that....

Jannify:
Merge Cyclops-MotorMode/State and -HUD to new branch

PR #39, opened 2017-09-19
Cyclops change engine mode

Jannify:


NitroxPatcher/Patches/CyclopsMotorMode_ChangeCyclopsMotorMode_Patch.cs
```diff
@@ -0,0 +1,35 @@
...
+using System.Reflection;
+
+namespace NitroxPatcher.Patches
+{
+    public class CyclopsMotorMode_ChangeCyclopsMotorMode_Patch : NitroxPatch
+    {
+        public static readonly Type TARGET_CLASS = typeof(CyclopsMotorMode);
+        public static readonly MethodInfo TARGET_METHOD = TARGET_CLASS.GetMethod("ChangeCyclopsMotorMode", BindingFlags.Public | BindingFlags.Instance);
+
+        public static bool Prefix(CyclopsMotorMode __instance, out int __state)
```
MarijnS95: Why is `__state` not of `CyclopsMotorMode.CyclopsMotorModes`?

NitroxClient/Communication/Packets/Processors/CyclopsChangeEngineModeProcessor.cs
```diff
@@ -0,0 +1,56 @@
...
+
+                if (motorMode != null & motorModeButtons != null)
+                {
+                    if ((CyclopsMotorMode.CyclopsMotorModes)motorPacket.Mode != motorMode.cyclopsMotorMode)
+                    {
+                        using (packetSender.Suppress<CyclopsChangeEngineMode>())
+                        {
+                            foreach (CyclopsMotorModeButton motorModeButton in motorModeButtons)
+                            {
+                                if (Player.main.currentSub == (SubRoot) motorModeButton.ReflectionGet("subRoot"))
```
MarijnS95: Doesn't this cause desync issues when setting the MotorMode on a Cyclops that the local player is not in at the moment?

NitroxClient/Communication/Packets/Processors/CyclopsChangeEngineModeProcessor.cs
```diff
@@ -0,0 +1,56 @@
...
+        public override void Process(CyclopsChangeEngineMode motorPacket)
+        {
+            Optional<GameObject> opCyclops = GuidHelper.GetObjectFrom(motorPacket.Guid);
+
+            if (opCyclops.IsPresent())
+            {
+                CyclopsMotorMode motorMode = opCyclops.Get().GetComponentInChildren<CyclopsMotorMode>();
+                CyclopsMotorModeButton[] motorModeButtons = opCyclops.Get().GetComponentsInChildren<CyclopsMotorModeButton>();
+
+                if (motorMode != null & motorModeButtons != null)
```
MarijnS95: Shouldn't this be `&&`?

Sunrunner37:
Hey Jannify - Just to keep you in the loop, I'm not ignoring these pull requests.  I'll merge them in when the headless branch is merged into master.

NitroxModel/Packets/CyclopsChangeEngineMode.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+
+namespace NitroxModel.Packets
+{
+    [Serializable]
+    public class CyclopsChangeEngineMode : AuthenticatedPacket
+    {
+        public String Guid { get; private set; }
```
MarijnS95: We removed all the private set functions; they are not needed if the value is only initialized in the constructor, and we don't (want) to change them anyway.

NitroxClient/Communication/Packets/Processors/CyclopsChangeEngineModeProcessor.cs
```diff
@@ -0,0 +1,56 @@
...
+                {
+                    if (Player.main.currentSub == (SubRoot)motorModeButton.ReflectionGet("subRoot"))
+                    {
+                        motorModeButton.SetCyclopsMotorMode((CyclopsMotorMode.CyclopsMotorModes)motorPacket.Mode);
+                    }
+                    else
+                    {
+                        if (motorPacket.Mode == motorModeButton.motorModeIndex)
+                        {
+                            motorMode.cyclopsMotorMode = motorPacket.Mode;
```
MarijnS95: This code looks very similar to `CyclopsMotorMode.ChangeCyclopsMotorMode`. To be more future proof, it might be a better idea to call that function.

Even better is to reuse the onclick logic. Simply replace the entire foreach loop with:
```C#
motorMode.BroadcastMessage("SetCyclopsMotorMode", this.motorModeIndex, SendMessageOptions.RequireReceiver);
```
All the buttons will now update their own icon/state accordingly, and the selected button will in turn broadcast `ChangeCyclopsMotorMode` to the `CyclopsMotorMode` object again, which updates all the variables like you do here.

As an example of future-proof-ness, you are already missing `.SendMessage("RecalculateNoiseValues", null, SendMessageOptions.RequireReceiver);`.

NitroxClient/Communication/Packets/Processors/CyclopsChangeEngineModeProcessor.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            this.packetSender = packetSender;
+        }
+
+        public override void Process(CyclopsChangeEngineMode motorPacket)
+        {
+            GameObject cyclops = GuidHelper.RequireObjectFrom(motorPacket.Guid);
+            CyclopsHelmHUDManager hud = cyclops.RequireComponentInChildren<CyclopsHelmHUDManager>();
+            CyclopsMotorMode motorMode = cyclops.RequireComponentInChildren<CyclopsMotorMode>();
+            CyclopsMotorModeButton[] motorModeButtons = cyclops.GetComponentsInChildren<CyclopsMotorModeButton>();
```
MarijnS95: This can be made even cleaner, as I explained in my previous review. Instead of manually touching all the buttons, simply call
```C#
motorMode.BroadcastMessage("SetCyclopsMotorMode", this.motorModeIndex, SendMessageOptions.RequireReceiver);
```
And everything should work fine. All the other lines can be removed.

NitroxModel/Packets/CyclopsChangeEngineMode.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+
+namespace NitroxModel.Packets
+{
+    [Serializable]
+    public class CyclopsChangeEngineMode : AuthenticatedPacket
+    {
+        public String Guid { get; set; }
```
MarijnS95: This is dangerous: you should remove the `set` functions entirely (not just `private`), as every other object can modify the data now, which is intended to be immutable.

Jannify:
Merge Cyclops-MotorMode/State and -HUD to new branch

Issue #36, opened 2017-09-12
Consistent coding style


MarijnS95:
Put a styleguide together covering both formatting (eg. settings for the built in formatter) and syntactical matters.
Below is a list of inconsistencies and changes I've noticed in the project.

#### Formatting:
- Spaces between keywords and parentheses `if (` vs `if(`.
- Unnecessary (trailing) white space (VS removes this on format).
- Sorting of fields and methods (dependent on access modifiers mostly).
- Use docstrings `///` instead of regular (`//`) comments.
    - If needed, use the right syntax to refer to variables etc.

### Syntactical / formatting (mixed):
- Unnecessary use of `this.`
- Braces around one liners (and ofc method body on a separate line). Single line method expressions are OK too.
- Parens around assignment expressions `foo = (bar != none);`.
- Unused (and unordered) `using` statements (VS can fix this as well).
- Unnecessary casts.
- Write field initializers at declaration instead of in the constructor (if applicable).
  `public readonly List<string> SomeThings { get; } = new List<string>();` versus assigning it in the contstructor.
- Short notations:
    - Null propagation such as `foo = bar ?? cake;`.
    - Ternary if statements such as `foo = bar < 0 ? bar : foo`;.
    - Safe navigation operator `var foo = bla?.something;`.

### Safe code practices:
- As strict as possible access modifiers (to indicate usage and prevent ambiguity):
    - Use `static` when a method doesn't (need to) touch local variables.
    - `const` and `readonly` where possible.
    - `public` vs `internal`, with multiple libraries going on this might be a good idea.
    - `private` on 'unity functions' that *should not* be called by our code (`Awake`, `Start`, etc)
    - Remove `private set;` when a property is only assigned in the constructor (or directly).
    - `sealed` classes? I'd say that's out of scope, until we start allowing plugins and such (or build a complete modding framework).
- Check for NRE's everywhere;
- **Especially** when retrieving datastructures from game-code. The game is constantly in development, and things **will** change. Defensive programming practices are a must to aid in debugging and correcting these changes later on.

### Naming:
- CTS or aliases (`Int` vs `int` and so on).
- lowerCamelCase for everything private, UpperCamelCase for everything else?

### Other:
- Use of inline functions vs defining a full function with types etc.
- Also applies to lambda notation: `private bool cake() => foo != null;`.
- Sort `Compile` items in csproj files; this reduces faulty decisions in merge conflicts.
[...]

PR #54, opened 2017-10-19
Initial contributing guideline.

MarijnS95:
Initial version of the contributing guide. Needs a lot of work, feel free to mention ideas in a review, or commit them to the branch directly. Check for spelling errors, overall correctness and excessive use of my opinion 😉

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -57,13 +57,13 @@ dotnet_style_predefined_type_for_member_access = true:error
 # Expression-level preferences
 dotnet_style_object_initializer = true:error
 dotnet_style_collection_initializer = true:error
-dotnet_style_coalesce_expression = true:error # Prefer a ?? b
-dotnet_style_null_propagation = true:error #  Prefer foo?.bar
+dotnet_style_coalesce_expression = true:suggestion # Prefer a ?? b
+dotnet_style_null_propagation = false:error # Do not prefer foo?.bar
 
-# Implicit and explicity styles
-csharp_style_var_for_built_in_types = false:none
-csharp_style_var_when_type_is_apparent = false:none
-csharp_style_var_elsewhere = false:none
+# Implicit and explicity styles (disallow use of `var` - types have to be specified explicitly)
+csharp_style_var_for_built_in_types = false:error
+csharp_style_var_when_type_is_apparent = false:error
+csharp_style_var_elsewhere = false:error
 
 # Expression-bodied members
 csharp_style_expression_bodied_methods = false:none
--- a/.editorconfig
+++ b/.editorconfig
@@ -76,3 +76,45 @@ csharp_prefer_braces = true:error # Wrap everything in braces.
 
 # Put `using System.` on top
 dotnet_sort_system_directives_first = true
+
+# Symbols for naming conventions
+dotnet_naming_symbols.public_fields.applicable_kinds           = field
+dotnet_naming_symbols.public_fields.applicable_accessibilities = public
+#, protected, internal, protected_internal
+
+dotnet_naming_symbols.all_const.applicable_kinds           = *
+dotnet_naming_symbols.all_const.applicable_accessibilities = *
+dotnet_naming_symbols.all_const.required_modifiers         = const
+
+dotnet_naming_symbols.private_fields.applicable_kinds           = field
+dotnet_naming_symbols.private_fields.applicable_accessibilities = private
+
+dotnet_naming_symbols.types.applicable_kinds           = class, struct, interface, enum, namespace, type_parameter
+dotnet_naming_symbols.types.applicable_accessibilities = *
+
+dotnet_naming_symbols.parameters.applicable_kinds = parameter
+
+# Styles for naming conventions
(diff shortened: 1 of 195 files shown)


Jannify:
What about a Visual Studio settings file to download?

MarijnS95:
@Jannify Yes, I've added such a file to this PR.

Besides regular formatting settings, there's an entire section for so called "language conventions". These are only checked in VS2017 though. I've added everything that's applicable to C# 6.0 and lower.
A lot of what I've written down [here](https://github.com/SubnauticaNitrox/Nitrox/issues/36) (which is based on https://github.com/SubnauticaNitrox/Nitrox/pull/11) is covered by these settings, though we still have to decide which setting we prefer.

@Sunrunner37 I hope you can give an opinion on the settings. Here's a quick list:

- Use of aliases vs CTS: prever `int` and `string` over
- Usage of `var`  - after a quick code search: I'm the only one using this.
- Use blocks for collection initializers: `new List<Foo> { a, b };` vs `.Add(a);` and `.Add(b);` on the next lines.
- Use blocks for object initializers: `new Foo() { Bar = 1 };`.
- Unnecessary use of `this.`.
- Null checking/coalescing operators `?.` and `??`.

And for each of these, we have to determine a severity, one of `none`, `suggestion`, `warning` or `error`. Error does not yet fail the build in VS2017, but eventually that'll be implemented.

I haven't done anything on the [naming convention](https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-code-style-settings-reference#naming-conventions) settings yet, maybe I'll add some of the default rules there.

Finally, I want to apply this style to all files before merging this PR. There are some tools to do most of it. I/we should, we can't enforce rules when most of the code doesn't adhere to it in the first place.

Sunrunner37:
> Use of aliases vs CTS: prefer int and string over Int32 and String for example.
I think aliases are used most places (except maybe string) - suggestion is probably fine.
 
> Usage of var - after a quick code search: I'm the only one using this.
This is one thing that I'm not favorable on - I don't think saving horizontal space is worth the readability trade off.  - error

> Use blocks for collection initializers: new List<Foo> { a, b }; vs .Add(a); and .Add(b); on the next lines.
Makes sense to me.

> Use blocks for object initializers: new Foo() { Bar = 1 };
eww - c# and it's attributes.... let's not do this please.

> Unnecessary use of this..
Sure, we can nuke these.

Null checking/coalescing operators ?. and ??.
Meh, less readable I believe.

MarijnS95:
> Overall the guidelines look good. Inside of the validation area do we want to call out gameObject.RequireComponent and GuidHelper.RequireObjectFrom?

I'll add that to the guide.

> This is one thing that I'm not favorable on - I don't think saving horizontal space is worth the readability trade off. - error

I think I can agree with that, imho it just makes writing easier to not have to specify the type, especially when working with (nested) generics. If I change this setting, AFAIK VS gives me the option to convert it for me 😀.

> eww - c# and it's attributes.... let's not do this please.

This is not attributes, it's used to set fields and properties after the constructor is called. Basically it changes code like:
```csharp
SomeCrazyObject foo = new SomeCrazyObject();
foo.Bar = "cake";
foo.baz = 1337;
foo.something = anotherThingIDeclaredAbove;
```
into
```csharp
SomeCrazyObject foo = new SomeCrazyObject()
{
    Bar = "cake",
    baz = 1337,
    something = anotherThingIDeclaredAbove
};
```
Which is IMHO much more readable, especially given the indentation. It explicits the fact that these fields/properties are set directly after the constructor, *preventing other lines of code from ending up inbetween*.

> Meh, less readable I believe.

Yup. On the one hand it reduces a lot of ugly `if (foo != null) { foo.Bar(); }` or `foo != null ? foo : bar;`, but on the other hand it makes it more explicit. I think `?.` is too easy to miss, but `foo ?? bar` is fine (they can be configured separately anyway).

Sunrunner37:
Now that it's morning here I'll follow up on these.

>This is not attributes, it's used to set fields and properties after the constructor is called. Basically it changes code like:
You'll have to sometimes read between the lines when I talk as I [come from java](https://stackoverflow.com/questions/8479038/what-is-an-attribute-in-java).  My (subjective) observation is that c# seems to encourage inappropriate intimacy with these types of features.  If you want to create anemic object types (without using the constructor) why not use builder pattern to gain immutability?  

>?. is too easy to miss
That's my opinion as well. (error level maybe?)

>foo ?? bar is fine 
Yup, makes sense to me.  (suggestion level maybe?)

MarijnS95:
> You'll have to sometimes read between the lines when I talk as I come from java. My (subjective) observation is that c# seems to encourage inappropriate intimacy with these types of features. If you want to create anemic object types (without using the constructor) why not use builder pattern to gain immutability?

You are completely right, my apologies. I work in many different languages myself, but only have little experience with Java. When talking about attributes in C#, I think about the `[BlaAttribute]` thingies; it makes no sense for me to think that in this context.

I think you are right though; even if I like the notation over regular `foo.Bar = baz;`, why are we even doing that in the first place? We're not using the builder pattern to begin with, so IMHO the values should be assigned through the constructor.

PR #55, opened 2017-10-20
Add new hierarchy analyzer to DebugUtils

Jannify:
[Example](https://cdn.discordapp.com/attachments/325084661011316737/370613242818461711/unknown.png)

### DIFF
--- a/NitroxClient/Unity/Helper/DebugUtils.cs
+++ b/NitroxClient/Unity/Helper/DebugUtils.cs
@@ -7,6 +7,46 @@ namespace NitroxClient.Unity.Helper
 {
     public class DebugUtils
     {
+        private static string prefix = "";
+
+        public static void DumpHierarchy(GameObject obj)
+        {
+            Log.Debug("Analyze gameobject: ");
+            Log.Debug("------------------------------------");
+            StartHierachy(obj);
+            Log.Debug("------------------------------------");
+        }
+
+        private static void StartHierachy(GameObject obj)
+        {
+            prefix = prefix + "    ";
+            Log.Debug(GetInfoOfGamobject(obj));
+            for (int i = 0; i < obj.transform.childCount; i++)
+            {
+                StartHierachy(obj.transform.GetChild(i).gameObject);
+                prefix = prefix.Substring(4);
+            }
+        }
+
+        private static string GetInfoOfGamobject(GameObject obj)
+        {
+            return prefix + obj.name + "   --<>--   " + GetComponentsOfGamobject(obj);
+        }
+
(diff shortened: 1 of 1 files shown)


NitroxClient/Unity/Helper/DebugUtils.cs
```diff
@@ -7,6 +7,46 @@ namespace NitroxClient.Unity.Helper
...
+            Log.Debug("------------------------------------");
+            StartHierachy(obj);
+            Log.Debug("------------------------------------");
+        }
+
+        private static void StartHierachy(GameObject obj)
+        {
+            prefix = prefix + "    ";
+            Log.Debug(GetInfoOfGamobject(obj));
+            for (int i = 0; i < obj.transform.childCount; i++)
```
MarijnS95: Why the indices? The other `DumpGameObject` example clearly shows that you can do this in tidy `foreach` loop. If `Transform` were using a linked list, `GetChild(i)` would be immensely slow.

NitroxClient/Unity/Helper/DebugUtils.cs
```diff
@@ -7,6 +7,46 @@ namespace NitroxClient.Unity.Helper
...
+
+        public static void DumpHierarchy(GameObject obj)
+        {
+            Log.Debug("Analyze gameobject: ");
+            Log.Debug("------------------------------------");
+            StartHierachy(obj);
+            Log.Debug("------------------------------------");
+        }
+
+        private static void StartHierachy(GameObject obj)
```
MarijnS95: Are you using this typo intentionally to distinguish `StartHierarchy` from `StartHierachy`?

NitroxClient/Unity/Helper/DebugUtils.cs
```diff
@@ -7,6 +7,46 @@ namespace NitroxClient.Unity.Helper
...
+        private static string GetComponentsOfGamobject(GameObject obj)
+        {
+            string list = "";
+            foreach (Component item in obj.GetComponents<Component>())
+            {
+                list += item.GetType() + " | ";
+            }
+            if (list.Length >= 3)
+            {
+                list = list.Substring(0, list.Length-3);
```
MarijnS95: Yuck. Use something like `string.Join` to add the separator.

This entire function can be converted into a oneliner:
```C#
return string.Join(" | ", obj.GetComponents<Component>().Select(item => item.GetType().ToString()).ToArray());
```

Sunrunner37:
I think we will continue using the existing gameObject debug dumper.  Feel free to keep this locally for your own use.

PR #56, opened 2017-10-21
Multiplayer settings options (Name and Color)

Jannify:


NitroxClient/MonoBehaviours/Gui/Chat/PlayerChat.cs
```diff
@@ -16,12 +15,15 @@ class PlayerChat : MonoBehaviour
         private GUIText chatText;
         private Coroutine timer;
         private List<String> messages;
```
Sunrunner37: Let's do a List<ChatMessage> (new nested class) that has member variables for message and color.  You could even tuck away some of the sanitizing and line limiting logic into it.  This allows us to avoid multiple related data structures.

NitroxClient/MonoBehaviours/Gui/Input/KeyBindingManager.cs
```diff
@@ -1,30 +0,0 @@
-﻿using NitroxClient.MonoBehaviours.Gui.Input.KeyBindings;
-using NitroxClient.MonoBehaviours.Gui.Input.KeyBindings.Actions;
-using System.Collections.Generic;
-using System.Linq;
-
-namespace NitroxClient.MonoBehaviours.Gui.Input
-{
-    public class KeyBindingManager
```
Sunrunner37: Let's continue to use the pre-existing key bindings system.

NitroxClient/MonoBehaviours/Gui/Settings/SettingsManager.cs
```diff
@@ -0,0 +1,87 @@
+﻿using NitroxClient.Unity.Helper;
+using NitroxModel.Logger;
+using Steamworks;
+using UnityEngine;
+using UnityEngine.Events;
+using UnityEngine.UI;
+
+namespace NitroxClient.MonoBehaviours.Gui.Settings {
+    public class SettingsManager {
```
Sunrunner37: Can we make this a singleton instead of making everything static?

NitroxClient/MonoBehaviours/Gui/Settings/SettingsManager.cs
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        public static void SetColor_Image(Color color) {
+            Color_Image.color = color;
+        }
+        public static Color GetColor_Image() {
+            if (Color_Image != null) {
+                return Color_Image.color;
+            }
+            return new Color(0, 0, 0);
```
Sunrunner37: How do the colors get assigned at the beginning, is it random?

NitroxClient/MonoBehaviours/Gui/Settings/SettingsOptions.cs
```diff
@@ -0,0 +1,103 @@
+﻿using NitroxClient.Unity.Helper;
+using NitroxModel.Logger;
+using Steamworks;
+using UnityEngine;
+using UnityEngine.Events;
+using UnityEngine.UI;
+
+namespace NitroxClient.MonoBehaviours.Gui.Settings {
```
Sunrunner37: Please audit your entire PR for formatting (in this case { needs to be on a new line)

NitroxClient/MonoBehaviours/Gui/Settings/SettingsManager.cs
```diff
@@ -0,0 +1,87 @@
...
+using UnityEngine.Events;
+using UnityEngine.UI;
+
+namespace NitroxClient.MonoBehaviours.Gui.Settings {
+    public class SettingsManager {
+        private static string Name;
+        private static string Key_Chat;
+        private static float Color_R;
+        private static float Color_G;
+        private static float Color_B;
```
Sunrunner37: Do we need to store these channels separately?  If this color class is mutable we can just call setR()

NitroxPatcher/Patches/Persistent/uGUI_OptionsPanel_AddTabs_Patch.cs
```diff
@@ -0,0 +1,33 @@
...
+        {
+            SettingsOptions settingsOptions = new SettingsOptions();
+
+            int tabIndex = __instance.AddTab("Multiplayer");
+            settingsOptions.AddInputFieldOption(tabIndex, "Name", "Enter Name", SettingsManager.GetName(), __instance, new UnityAction<string>(SettingsManager.SetName));
+            settingsOptions.AddKeybindOption(tabIndex, "Chat", SettingsManager.GetKey_Chat(), __instance, new UnityAction<string>(SettingsManager.SetKey_Chat));
+            __instance.AddHeading(tabIndex, "Player color");
+            __instance.AddSliderOption(tabIndex, "Red", SettingsManager.GetColorR(), new UnityAction<float>(SettingsManager.SetColorR));
+            __instance.AddSliderOption(tabIndex, "Green", SettingsManager.GetColorG(), new UnityAction<float>(SettingsManager.SetColorG));
+            __instance.AddSliderOption(tabIndex, "Blue", SettingsManager.GetColorB(), new UnityAction<float>(SettingsManager.SetColorB));
```
Sunrunner37: We should tuck some of this logic into a class to keep the patchers slime

NitroxClient/MonoBehaviours/Chat.cs
```diff
@@ -20,14 +20,14 @@ public void OnConsoleCommand_chat(NotificationCenter.Notification n)
             {
                 String text = "";
 
-                for(int i = 0; i < n.data.Count; i++)
+                for (int i = 0; i < n.data.Count; i++)
```
MarijnS95: While you're at it, can you change this to a `string.Join` structure?

NitroxClient/MonoBehaviours/Gui/Chat/PlayerChat.cs
```diff
@@ -16,12 +15,15 @@ class PlayerChat : MonoBehaviour
...
+        private Dictionary<String, String> colors;
```
MarijnS95: This can be made readonly.

NitroxClient/MonoBehaviours/Gui/Chat/PlayerChat.cs
MarijnS95: If you're interested, this is called SOA (Structure Of Arrays), whereas we prefer AOS (Array Of Structures), because it groups coupled data together, and allows for neat functions that work on this data.

Sidenote: I recently looked into cInput (a Unity input library), which heavily relies on SOA. The code is extremely hard to follow, everything is crammed into a single structure, and don't even want to think about maintaining it :wink:

NitroxClient/MonoBehaviours/Gui/Chat/PlayerChatEntry.cs
```diff
@@ -14,17 +15,17 @@ class PlayerChatEntry : MonoBehaviour
         private bool chatEnabled = false;
         private string chatMessage = "";
         private PlayerChatManager chatManager;
-        
+
```
MarijnS95: I do like that your editor fixes all these unnecessary (trailing) whitespaces, and spaces between keywords and parentheses :+1:.

NitroxClient/MonoBehaviours/Gui/Settings/SettingsManager.cs
```diff
@@ -0,0 +1,87 @@
...
+        public static Image Color_Image;
+
+        public static void SetName(string name) {
+            Name = name;
+        }
+        public static string GetName() {
```
MarijnS95: All these get and set functions can be converted to properties.

For Name, I prefer an auto property, which is initialized to `SteamFriends.GetPersonaName()` in the constructor.
But you can also add initial validation in a regular property.

NitroxClient/MonoBehaviours/Gui/Settings/SettingsManager.cs
```diff
@@ -0,0 +1,87 @@
...
+                    Name = SteamFriends.GetPersonaName();
+                }
+            }
+            return Name;
+        }
+
+        public static void SetKey_Chat(string key) {
+            Key_Chat = key;
+        }
+        public static string GetKey_Chat() {
```
MarijnS95: For example `public static string ChatKey { get; set; } = "c";`.

NitroxClient/MonoBehaviours/Gui/Settings/SettingsManager.cs
```diff
@@ -0,0 +1,87 @@
...
+using NitroxModel.Logger;
+using Steamworks;
+using UnityEngine;
+using UnityEngine.Events;
+using UnityEngine.UI;
+
+namespace NitroxClient.MonoBehaviours.Gui.Settings {
+    public class SettingsManager {
+        private static string Name;
+        private static string Key_Chat;
```
MarijnS95: Let's change these names to CamelCase to stay consistent with the code, instead of using Camel_Snake_Case.

NitroxClient/MonoBehaviours/Gui/Settings/SettingsOptions.cs
```diff
@@ -0,0 +1,103 @@
...
+    public class SettingsOptions {
+        public void AddInputFieldOption(int tabIndex, string label, string placeholder, string value, uGUI_OptionsPanel uGUI_Options, UnityAction<string> callback = null) {
+            uGUI_MainMenu uGUI_Main = uGUI_MainMenu.main;
+            uGUI_TabbedControlsPanel uGUI_Control = uGUI_Main.GetComponentInChildren<uGUI_TabbedControlsPanel>();
+            GameObject InputOption = new GameObject();
+            GameObject input_field = Object.Instantiate(uGUI_Options.keyRedemptionPrefab.transform.GetChild(1).gameObject);
+            GameObject caption = Object.Instantiate(uGUI_Options.toggleOptionPrefab.transform.GetChild(0).GetChild(2).gameObject);
```
MarijnS95: Will these indices stay the same if UWE changes the layout? Can we change this to use paths (via `.Find`)?

NitroxClient/MonoBehaviours/Gui/Settings/SettingsOptions.cs
```diff
@@ -0,0 +1,103 @@
...
+
+            caption.GetComponent<Text>().text = label;
+            input_field.transform.SetParent(caption.transform);
+            caption.transform.SetParent(InputOption.transform);
+
+            InputOption.name = "uGUI_InputFieldOption";
+            InputOption.AddComponent<LayoutElement>();
```
MarijnS95: You can store the returned value in a field and use in the lines below.

NitroxPatcher/Patches/Persistent/GameSettings_SerializeSettings_Patch.cs
```diff
@@ -0,0 +1,27 @@
...
+    public class GameSettings_SerializeSettings_Patch : NitroxPatch
+    {
+        public static readonly Type TARGET_CLASS = typeof(GameSettings);
+        public static readonly MethodInfo TARGET_METHOD = TARGET_CLASS.GetMethod("SerializeSettings", BindingFlags.NonPublic | BindingFlags.Static);
+
+        public static void Postfix(GameSettings.ISerializer serializer)
+        {
+            SettingsManager.SetName(serializer.Serialize("Multiplayer/Name", SettingsManager.GetName()));
+            SettingsManager.SetKey_Chat(serializer.Serialize("Multiplayer/Chat/Primary", SettingsManager.GetKey_Chat()));
+            SettingsManager.SetColorR(serializer.Serialize("Multiplayer/colorRed", SettingsManager.GetColorR()));
```
MarijnS95: Doesn't the serializer support serializing a color directly?

NitroxPatcher/Patches/Persistent/uGUI_OptionsPanel_AddTabs_Patch.cs
MarijnS95: You could for instance store all this logic in a monobehaviour that encapsulates the multiplayer menu. This also allows to store often-used variables such as uGUI_Control and tabIndex in a field.

NitroxClient/MonoBehaviours/Gui/Settings/SettingsManager.cs
Jannify: No its black. Should it be random at the first start?

NitroxClient/MonoBehaviours/Gui/Settings/SettingsManager.cs
Jannify: That doesn't work because I need a set-function for the UnityAction. Or is there a solution for this?

NitroxPatcher/Patches/Persistent/GameSettings_SerializeSettings_Patch.cs
Jannify: No. :(

MarijnS95: Also, is it really necessary to call the Set functions after serializing. As far as I know, they just return the value that was passed in.

NitroxClient/MonoBehaviours/Gui/Settings/SettingsManager.cs
MarijnS95: I **personally** prefer a lambda expression in such cases, as to not burden the class with this kind of utility 'functions'. For instance `new UnityAction<float>((newVal) => SettingsManager.Color_R = newVal)`.

NitroxPatcher/Patches/Persistent/GameSettings_SerializeSettings_Patch.cs
Jannify: The serializer also read the settings for start.

MarijnS95: Yeah I just noticed by decompiling the code... The `Serialize` method deserializes, based on a `reading` boolean :/.

NitroxClient/MonoBehaviours/Gui/Settings/SettingsManager.cs
MarijnS95: The `Color` class is mutable. It would be cool to make a wrapper property for a `Color` property (which stores it in a private field, and in the image if it exists), but because `Color` is a value-type (`struct`, *not* passed by reference), it's impossible to modify color channels separately. (eg `Color.r = 123;` would fail with a `Cannot modify the return value of 'SettingsManager.Color' because it is not a variable`). In this case, properties delegating the get/set to the fields of the private `Color` field might be desirable.

NitroxClient/MonoBehaviours/Gui/Settings/SettingsOptions.cs
```diff
@@ -0,0 +1,103 @@
...
+            InputOption.GetComponent<LayoutElement>().minHeight = 60;
+            InputOption.GetComponent<LayoutElement>().preferredWidth = -1;
+            InputOption.GetComponent<LayoutElement>().preferredHeight = 60;
+            input_field.transform.localPosition = new Vector3(750, 15, input_field.transform.localPosition.z);
+
+            GameObject gameObject = uGUI_Control.AddItem(tabIndex, InputOption);
+
+            InputField componentInChildren = gameObject.GetComponentInChildren<InputField>();
+            Utils.Assert(componentInChildren != null, "see log", null);
```
MarijnS95: We have our own `Validate.NotNull()`. I wonder though if `"see log"` is applicable to this usecase.

NitroxClient/GameLogic/Chat.cs
```diff
@@ -13,9 +14,9 @@ public Chat(PacketSender packetSender)
             this.packetSender = packetSender;
         }
 
-        public void SendChatMessage(String text)
+        public void SendChatMessage(String text, Color color)
```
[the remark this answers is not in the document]
Jannify: For color change in session: Then we will have a problem with the Chat system because it does not store the author...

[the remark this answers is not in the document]
MarijnS95: But we can change that ;). Simply have a chat object that stores the message and a reference to the player. Which is also what Sunrunner is suggesting: https://github.com/SubnauticaNitrox/Nitrox/pull/56/files/a621aea94c67821300de565d73cd685df685992a#r146141772; making a dictionary to go from chat message to color isn't the right/neat way to do it.

NitroxClient/MonoBehaviours/Chat.cs
Jannify: n is a NotificationCenter.Notification not a string[]. Is there a conversion?

MarijnS95: `.Select(n => n.ToString())`

NitroxClient/MonoBehaviours/MultiplayerSettings.cs
```diff
@@ -0,0 +1,77 @@
...
+
+            RectTransform rect = image.GetComponent<RectTransform>();
+            rect.sizeDelta = new Vector2(200, 25);
+            rect.localScale = new Vector3(2, 2, 1);
+            image.sprite = OptionsPanel.troubleshootingPrefab.GetComponentInChildren<Image>().sprite;
+            image.type = Image.Type.Sliced;
+            image.color = value;
+
+            GameObject gameObject = OptionsPanel.AddItem(TabIndex, InstantiateOptionsRoot(image.gameObject, new Vector3(235, 0, 0), "uGUI_ColorImage", label));
+            SettingsManager.ColorImage = gameObject.RequireComponentInChildren<Image>();
```
MarijnS95: Isn't this the same as the local `image` variable?

NitroxPatcher/Patches/Player_Update_Patch.cs
```diff
@@ -24,12 +24,14 @@ public static void Postfix(Player __instance)
 
             foreach (KeyBinding keyBinding in keyBindingManager.KeyboardKeyBindings)
             {
-                bool isButtonDown = (bool)ReflectionHelper.ReflectionCall<GameInput>(null, "GetButtonDown", new Type[] { typeof(GameInput.Button) }, true, true, keyBinding.Button);
+                //Throws errors in the output_log
```
MarijnS95: Again, if something is broken, **don't just remove it** (locally is fine, just don't include it in a PR). Either try to fix it, or forward it to another Nitrox contributor.

I don't see any exception in my log, can you post them? Though they keybinding option doesn't show up anymore in the keyboard menu either... Maybe some configuration is broken on my system.

NitroxPatcher/NitroxPatcher.csproj
```diff
@@ -75,13 +75,15 @@
...
+    <Compile Include="Patches\Persistent\GameInput_Initialize_Patch.cs" />
+    <Compile Include="Patches\Persistent\GameInput_SetupDefaultKeyboardBindings_Patch.cs" />
+    <Compile Include="Patches\Persistent\uGUI_OptionsPanel_AddTabs_Patch.cs" />
+    <Compile Include="Patches\Persistent\GameSettings_SerializeSettings_Patch.cs" />
     <Compile Include="Patches\ItemsContainer_NotifyAddItem_Patch.cs" />
     <Compile Include="Patches\ItemsContainer_NotifyRemoveItem_Patch.cs" />
     <Compile Include="Patches\MedicalCabinet_OnHandClick_Patch.cs" />
     <Compile Include="Patches\NitroxPatch.cs" />
     <Compile Include="Patches\Openable_PlayOpenAnimation_Patch.cs" />
-    <Compile Include="Patches\Persistent\GameInput_Initialize_Patch.cs" />
```
MarijnS95: Can't these stay in their original position (and move the other `Persistent` patches here)? I just sorted everything in an earlier PR.

NitroxClient/MonoBehaviours/MultiplayerSettings.cs
```diff
@@ -0,0 +1,77 @@
...
+    {
+        private int TabIndex;
+        private uGUI_OptionsPanel OptionsPanel;
+
+        private void Start()
+        {
+            OptionsPanel = gameObject.RequireComponent<uGUI_OptionsPanel>();
+
+            TabIndex = OptionsPanel.AddTab("Multiplayer");
+            AddInputFieldOption("Name", "Enter Name", SettingsManager.GetName(), new UnityAction<string>(newName => SettingsManager.Name= newName));
```
MarijnS95: For some reason the UI has a different scale when in game, compared to the main menu. This causes the textbox to be partially cut off...
image

NitroxClient/MonoBehaviours/Chat.cs
Jannify: n.data is a Hashtable. No definition for Select() :(

MarijnS95: A `Hashtable` implements `IEnumerable`, not `IEnumerable<object>`, so you have to cast it to a generic `IEnumerable` first before methods like `.Select` become available.

```csharp
string text = string.Join(" ", n.data.Cast<object>().Select(d => d.ToString()).ToArray());
```

NitroxClient/MonoBehaviours/MultiplayerSettings.cs
```diff
@@ -0,0 +1,77 @@
...
+            componentInChildren.GetComponentsInChildren<Text>()[0].text = placeholder;
+            if (callback != null)
+            {
+                componentInChildren.onValueChanged.AddListener(callback);
+            }
+        }
+
+        public void AddColorImage(string label, Color value)
+        {
+            Image image = new GameObject().AddComponent<Image>();
```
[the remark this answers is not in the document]
Jannify: Which gameObject did you mean?

NitroxClient/MonoBehaviours/MultiplayerSettings.cs
Jannify: AddItem() instantiate the given component. The link to the Image in game would be lost...

NitroxClient/MonoBehaviours/MultiplayerSettings.cs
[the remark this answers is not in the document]
MarijnS95: Well you're currently extending a `MonoBehaviour`, which is attached to a `gameObject` (accessed by the property with that name).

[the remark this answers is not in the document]
Jannify: Why? Later in code this image will be a child from the option-gameobject.

NitroxClient/GameLogic/Chat.cs
```diff
@@ -13,10 +14,16 @@ public Chat(PacketSender packetSender)
...
         {
-            ChatMessage message = new ChatMessage(packetSender.PlayerId, text);
+            ChatMessage message = new ChatMessage(packetSender.PlayerId, text, color);
             packetSender.Send(message);
         }
+
+        public void SendPlayerJoin(String text, Color color)
```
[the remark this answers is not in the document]
Jannify: Is PlayerAttributes better?

NitroxClient/MonoBehaviours/Gui/MainMenu/JoinServer.cs
```diff
@@ -8,7 +8,6 @@ namespace NitroxClient.MonoBehaviours.Gui.MainMenu
     public class JoinServer : MonoBehaviour
     {
         public string serverIp = "";
-        string username = "username";
         bool showingUsername = false;
```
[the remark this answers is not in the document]
Jannify: Did you mean I should reimplement it? As backup or as option to enter the name before joining?
I am happy to remove this ugly and buggy input field with a nice solution.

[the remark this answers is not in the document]
MarijnS95: You've moved the username input to the menu, so I guess (and by looking at the code) that means that this boolean is doing nothing anymore since no username input field is shown anymore in this 'join' menu. So, I guess you can remove it.

NitroxClient/GameLogic/Chat.cs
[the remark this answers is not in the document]
MarijnS95: Maybe you can merge it with the `Authenticate` packet - let it do the 'authentication' against the server, and send the initial color. Perhaps renamte it to something more appropriate as well.

NitroxClient/MonoBehaviours/Gui/Settings/SettingsManager.cs
```diff
@@ -0,0 +1,56 @@
+﻿using Steamworks;
+using UnityEngine;
+using UnityEngine.UI;
+
+namespace NitroxClient.MonoBehaviours.Gui.Settings
+{
+    public static class SettingsManager
+    {
+        public static string Name { private get; set; }
```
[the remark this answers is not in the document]
Jannify: I can't set the default value with ```SteamFriends.GetPersonaName()``` because I have to check if SteamAPI is there (```SteamAPI.Init()```). I need a function and to be save nobody get a null value because the get is public I set it private.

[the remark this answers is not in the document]
MarijnS95: You can call that in the static constructor, or the property get method. Up to you to see what looks best.

NitroxClient/MonoBehaviours/Chat.cs
Jannify: Dosn't work. Just show the Type...

MarijnS95: Have you tried the "potential fixes" menu (<kbd>ctrl</kbd>+<kbd>.</kbd>)? It should tell you to `using System.Linq;`.

Jannify: Yes I have used ```System.Linq```.

NitroxClient/GameLogic/Chat.cs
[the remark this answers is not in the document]
Jannify: But if we leave it at ```PlayerAttributes``` we can later set the Color, Name, etc if the player use the setting menu in game too. We can also rename it...

NitroxClient/MonoBehaviours/Gui/MainMenu/JoinServer.cs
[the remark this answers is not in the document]
Jannify: Yes, I misunderstood you. I fix this :)

PR #57, opened 2017-10-26
New Status Bar positions.

kylinator25:
Added functionality for putting the health bars next to the players ping icon. Unsure whether Camera.main and Player.main will return correct values.

### DIFF
--- a/NitroxClient/MonoBehaviours/Gui/HUD/RemotePlayerVitals.cs
+++ b/NitroxClient/MonoBehaviours/Gui/HUD/RemotePlayerVitals.cs
@@ -1,5 +1,6 @@
-﻿using NitroxClient.MonoBehaviours.Gui.Helper;
+using NitroxClient.MonoBehaviours.Gui.Helper;
 using NitroxModel.Helper;
+using NitroxModel.Logger;
 using System;
 using UnityEngine;
 using UnityEngine.UI;
@@ -16,7 +17,7 @@ namespace NitroxClient.MonoBehaviours.Gui.HUD
         private static readonly Color FOOD_BAR_BORDER_COLOR = new Color(0.957f, 0.914f, 0.251f, 1.0f);
         private static readonly Color WATER_BAR_COLOR = new Color(0.212f, 0.663f, 0.855f, 1.0f);
         private static readonly Color WATER_BAR_BORDER_COLOR = new Color(0.227f, 0.949f, 0.969f, 1.0f);
-
+        
         private static readonly Vector2 OXYGEN_BAR_POSITION_OFFSET = new Vector2(200, 160);
         private static readonly Vector2 HEALTH_BAR_POSITION_OFFSET = new Vector2(150, 160);
         private static readonly Vector2 FOOD_BAR_POSITION_OFFSET = new Vector2(100, 160);
@@ -43,6 +44,9 @@ namespace NitroxClient.MonoBehaviours.Gui.HUD
         private Bar healthBar;
         private Bar foodBar;
         private Bar waterBar;
+
+        private GameObject playerPing;
+        private int pingId;
 
         private GameObject background;
         private GameObject playerNameText;
@@ -132,6 +136,22 @@ namespace NitroxClient.MonoBehaviours.Gui.HUD
             playerNameText.layer = em.gameObject.layer;
             // em does not have a parent anymore on latest stable, so the stats position is incorrect.
             playerNameText.transform.parent = em.transform;//.parent.transform;
+            GameObject signalBase = Instantiate(Resources.Load("VFX/xSignal")) as GameObject;
+            signalBase.name = "signal " + playerName;
+            signalBase.transform.position = Player.main.transform.position;
+            
+            PingInstance ping = signalBase.GetComponent<PingInstance>();
+            ping.visible = false;
+            ping.name = playerName;
+
+            ping.pingType = PingType.None;
+            ping.displayPingInManager = true;
+            ping.maxDist = 0.01f;
+            ping.minDist = 0f;
+
+            pingId = PingManager.Register(ping);
+
+            playerPing = signalBase;
         }
 
         private Sprite GetPlayerBackgroundSprite()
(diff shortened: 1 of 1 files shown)


NitroxClient/MonoBehaviours/Gui/HUD/RemotePlayerVitals.cs
```diff
@@ -132,6 +136,22 @@ private void CreatePlayerNameText()
             playerNameText.layer = em.gameObject.layer;
             // em does not have a parent anymore on latest stable, so the stats position is incorrect.
             playerNameText.transform.parent = em.transform;//.parent.transform;
+            GameObject signalBase = Instantiate(Resources.Load("VFX/xSignal")) as GameObject;
```
MarijnS95: We already create and assign a signal to a player, can that be reused? (preferably)

NitroxClient/MonoBehaviours/Gui/HUD/RemotePlayerVitals.cs
```diff
@@ -16,7 +17,7 @@ public class RemotePlayerVitals : MonoBehaviour
         private static readonly Color FOOD_BAR_BORDER_COLOR = new Color(0.957f, 0.914f, 0.251f, 1.0f);
         private static readonly Color WATER_BAR_COLOR = new Color(0.212f, 0.663f, 0.855f, 1.0f);
         private static readonly Color WATER_BAR_BORDER_COLOR = new Color(0.227f, 0.949f, 0.969f, 1.0f);
-
+
```
MarijnS95: Unnecessary whitespace.

NitroxClient/MonoBehaviours/Gui/HUD/RemotePlayerVitals.cs
```diff
@@ -211,10 +231,10 @@ public void SetNewPosition(int newPosition)
...
         private void SetBarPostion(GameObject barGameObject, Vector2 positionOffset, Canvas canvas)
         {
-            Vector2 screenPosition = new Vector2(Screen.width - positionOffset.x, Screen.height - (positionOffset.y * position));
-            Vector2 worldPosition;
-            RectTransformUtility.ScreenPointToLocalPointInRectangle(canvas.transform as RectTransform, screenPosition, Camera.main, out worldPosition);
-            barGameObject.transform.position = canvas.transform.TransformPoint(worldPosition);
+            //Vector2 screenPosition = new Vector2(Screen.width - positionOffset.x, Screen.height - (positionOffset.y * position));
+            //Vector2 worldPosition;
+            //RectTransformUtility.ScreenPointToLocalPointInRectangle(canvas.transform as RectTransform, screenPosition, Camera.main, out worldPosition);
+            barGameObject.transform.position = Camera.main.WorldToScreenPoint(PingManager.Get(pingId).transform.position) + new Vector3(positionOffset.x, positionOffset.y);
```
MarijnS95: Just use `+ (Vector3)positionOffset`, `Vector2` has an `implicit operator` defined for it (though you have to explicitly specify it, because the compiler is unsure whether to convert the result of `WorldToScreenPoint` from `Vector3` to `Vector2`, or `positionOffset` from `Vector2` to `Vector3` since both conversions are implicit).

Also, the `PingInstance` returned by `PingManager.Get` is the same as initially registered with `Register`, is there a reason to not store the `PingInstance` in a `private` field in this class?

NitroxClient/MonoBehaviours/Gui/HUD/RemotePlayerVitals.cs
```diff
@@ -132,6 +136,22 @@ private void CreatePlayerNameText()
...
+            PingInstance ping = signalBase.GetComponent<PingInstance>();
+            ping.visible = false;
+            ping.name = playerName;
+
+            ping.pingType = PingType.None;
+            ping.displayPingInManager = true;
+            ping.maxDist = 0.01f;
+            ping.minDist = 0f;
+
+            pingId = PingManager.Register(ping);
```
MarijnS95: I assume this works on stable, but the function returns nothing (`void`) on experimental - this will stop working soon.

NitroxClient/MonoBehaviours/Gui/HUD/RemotePlayerVitals.cs
```diff
@@ -132,6 +136,22 @@ private void CreatePlayerNameText()
             playerNameText.layer = em.gameObject.layer;
             // em does not have a parent anymore on latest stable, so the stats position is incorrect.
             playerNameText.transform.parent = em.transform;//.parent.transform;
+            GameObject signalBase = Instantiate(Resources.Load("VFX/xSignal")) as GameObject;
+            signalBase.name = "signal " + playerName;
+            signalBase.transform.position = Player.main.transform.position;
```
MarijnS95: Is it necessary to temporarily position it at the local player? Can't you leave it empty because it'll be overwritten in `SetBarPosition` soon anyway?

NitroxClient/MonoBehaviours/Gui/HUD/RemotePlayerVitals.cs
kylinator25: I looked into re-using the signal, but I couldn't off the top of my head figure out a way to get the reference from RemotePlayer into RemotePlayerVitals, unless we can use statics?

NitroxClient/MonoBehaviours/Gui/HUD/RemotePlayerVitals.cs
kylinator25: I'm positioning signalBase here, and later I position the health bar to the position of the signal in SetBarPosition. I don't change the signalBase position in SetBarPosition

NitroxClient/MonoBehaviours/Gui/HUD/RemotePlayerVitals.cs
kylinator25: no i changed the function so it returns the int ID

NitroxClient/MonoBehaviours/Gui/HUD/RemotePlayerVitals.cs
kylinator25: ok I see your logic behind this. My initial intention was to Register to the ping manager and then get the position of the UI ping inside the manager, but I couldn't see any plain functions that achieved this, so i am setting it to the signal position.

NitroxClient/MonoBehaviours/Gui/HUD/RemotePlayerVitals.cs
MarijnS95: We have a `PlayerVitalsManager` that returns an existing or new `RemotePlayerVitals` component for a given player. You can use the methods in `RemotePlayer`, but it's up to you to refactor some pieces such that you can access it 😉 

Also, consider where you want to instantiate the label, and how it's gonna be accessed by the other class.

NitroxClient/MonoBehaviours/Gui/HUD/RemotePlayerVitals.cs
MarijnS95: Correct, I meant to edit that question to ask why you are setting it to the local player, beause this will cause the bars to float above the local player (as you're using it in `SetBarPosition`). I'd assume you want these to float above the other players instead.

NitroxClient/MonoBehaviours/Gui/HUD/RemotePlayerVitals.cs
MarijnS95: That's not a good idea, we want to change the least game code possible, and inject everything else at runtime. Currently we're only adding a single function call to the game to start everything up, and we want to keep it at just that.

Consider finding a different way, simpelest and most straightforward would be store the `PingInstance` as a private field in the class.

NitroxClient/MonoBehaviours/Gui/HUD/RemotePlayerVitals.cs
MarijnS95: You'll want to change this a little anyway, since it seems the position of the ping is always equal to that of the local player (when it was created initially), not of the player these bars are for.

Though that probably 'solves itself' as soon as you use the pings from `RemotePlayer` - those are attached to the player and thus have the right position.

PR #60, opened 2017-11-03
Formatting/naming violations and cleanup

MarijnS95:


NitroxClient/Communication/Packets/Processors/CellEntitiesProcessor.cs
```diff
@@ -2,64 +2,42 @@
...
 
 namespace NitroxClient.Communication.Packets.Processors
 {
     class CellEntitiesProcessor : ClientPacketProcessor<CellEntities>
     {
-        private PacketSender packetSender;
-        private HashSet<string> alreadySpawnedGuids = new HashSet<string>();
-        
+        private readonly PacketSender packetSender;
+
```
Sunrunner37: I'm not sure that this refactor will work because entities should unload when they become out of range.  If a chunk unloads and you receive this packet it will spawn another entity.

NitroxClient/Communication/Packets/Processors/CellEntitiesProcessor.cs
```diff
@@ -2,64 +2,42 @@
...
+                if (opGameObject.IsPresent())
                 {
-                    SpawnEntity(entity);
+                    gameObject = opGameObject.Get();
                 }
                 else
                 {
-                    UpdateEntityPosition(entity);
+                    Validate.IsTrue(entity.TechType != TechType.None);
```
Sunrunner37: This isn't really exceptional (as of right now) and shouldn't block other entities from being processed.

PR #71, opened 2018-01-16
multiple bug fixes

urlordjames:
I'm new, some of these might not actually work as intended, not fully tested.


urlordjames:
upon further testing with other computers this breaks submodules

PR #77, opened 2018-01-20
Fixed a "Invalid IL code" error

oliverslott:
Hello, first pull request.
I found that when you join a server the following message would show up in the output_log.txt;
`InvalidProgramException: Invalid IL code in (wrapper dynamic-method) ConstructorInput:OnCraftingBegin_Patch0`
And the client would then fail to connect.
This should fix it, I have also tested it:
image


oliverslott:
I see, I thought it was an unused comment. But good to hear about the future-proof solution!

PR #79, opened 2018-01-21
JoinServer & AddServer menu cleanup

Measuring:

PR #86, opened 2018-01-25
Added scene debugger and a stub-network debugger windows.

Measuring:
Enable debuggers with F7 while the game is running. Network debugger has no functionality but I plan to add that later (or someone else can play with it if they want). The scene debugger shows all the objects in the selected scene. Clicking on an object shows you the monobehaviours.

Editing of scene objects is limited to transform position only. Properly editing other monobehaviours is planned later (something like how Unity Editor does it).

Viewing objects/monobehaviours might be useful for other developers so I created a pull request.


NitroxClient/MonoBehaviours/Debugging/NetworkDebugger.cs
```diff
@@ -0,0 +1,36 @@
+﻿using UnityEngine;
+
+namespace NitroxClient.MonoBehaviours.Debugging
+{
+    public class NetworkDebugger : BaseDebugger
+    {
+        public void Awake()
+        {
+            Hotkey = KeyCode.N;
```
MarijnS95: It looks like these fields are only initialized once, are not supposed to be changed thereafter, and do not need to be set specifically from `Awake`. How about moving their initialisation to a constructor, and specifying the fields as `readonly`? More specifically, since the fields are mandatory, require them to be passed as arguments to a constructor in `BaseDebugger`?

NitroxClient/MonoBehaviours/Debugging/NitroxDebug.cs
```diff
@@ -0,0 +1,140 @@
...
+        private HashSet<BaseDebugger> wereActive;
+        private bool isDebugging;
+        private Rect windowRect;
+
+        public void Awake()
+        {
+            windowRect = new Rect(10, 10, 200, 0);
+
+            EnableDebuggerHotkey = KeyCode.F7;
+            wereActive = new HashSet<BaseDebugger>();
```
MarijnS95: We prefer setting such fields at declaration, and making them `readonly`. Optionally, you can call `.Clear()` from `Awake`.

```csharp
private readonly HashSet<BaseDebugger> wereActive = new HashSet<BaseDebugger>();
```

NitroxClient/MonoBehaviours/Debugging/SceneDebugger.cs
```diff
@@ -0,0 +1,253 @@
...
+            });
+
+            skin.SetCustomStyle("breadcumNav", skin.box, s =>
+            {
+                s.stretchWidth = false;
+                s.fixedWidth = 100;
+            });
+        }
+
+        private void GuiRenderTree(IEnumerable<Transform> objects, int level = 0)
```
MarijnS95: This function doesn't seem to be used anywhere. Rather, it seems to be somewhat implemented when rendering `ActiveTab` `1`.

NitroxClient/MonoBehaviours/Debugging/SceneDebugger.cs
```diff
@@ -0,0 +1,253 @@
...
+            {
+                "Scenes",
+                "Hierarchy",
+                "GameObject"
+            });
+        }
+
+        public override void DoWindow(int windowId)
+        {
+            switch (ActiveTab)
```
MarijnS95: Can `ActiveTab` be changed to an `enum` type to better distinguish the possible pages?

NitroxClient/Unity/Helper/GUISkinUtils.cs
```diff
@@ -0,0 +1,139 @@
...
+
+            // TODO: Create identical copy of font.
+            copy.font = baseSkin.font; // Gives a giberish font: copy.font = UnityEngine.Object.Instantiate(baseSkin.font);
+
+            return copy;
+        }
+
+        public static GUISkin RegisterDerived(string name, GUISkin baseSkin = null)
+        {
+            if (name == null)
```
MarijnS95: We prefere to use the `Validate.NotNull` utility method for null checks (and `Validate.IsTrue/False` a couple lines below).

NitroxClient/Unity/Helper/GUISkinUtils.cs
```diff
@@ -0,0 +1,139 @@
...
+                if (item.name == style.name)
+                {
+                    skin.customStyles[i] = style;
+                    found = true;
+                    break;
+                }
+            }
+            if (!found)
+            {
+                GUIStyle[] newStyles = new GUIStyle[skin.customStyles.Length + 1];
```
MarijnS95: `Array.Resize`?

NitroxClient/Unity/Helper/GUISkinUtils.cs
```diff
@@ -0,0 +1,139 @@
...
+        }
+
+        public static void SetCustomStyle(this GUISkin skin, string name, GUIStyle baseStyle, Action<GUIStyle> modify)
+        {
+            GUIStyle style = new GUIStyle(baseStyle);
+            style.name = name;
+            modify(style);
+
+            bool found = false;
+            for (int i = 0; i < skin.customStyles.Length; i++)
```
MarijnS95: Personally not a fan of loops accessing something in an iterable in C#. At least this can be mitigated with `Array.FindIndex(skin.customStyles, item => item.name == style.name);`.

NitroxClient/Unity/Helper/GUIStyleUtils.cs
```diff
@@ -0,0 +1,18 @@
+﻿using UnityEngine;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace NitroxClient.Unity.Helper
+{
+    public static class GUIStyleUtils
```
MarijnS95: This class is not referenced anywhere.

MarijnS95:
Besides, with regards to your [previous PR](https://github.com/SubnauticaNitrox/Nitrox/pull/79), you seem to have a lot of Unity UI knowledge. I'm more than happy with the cleanup you did on the awful join menu, ad I've wanted to do that for a while now. However, using UWE's GUI so that it blends completely into the game UI. Is that something you're planning to do?

And now that I randomly have the `DebugUtils` class on my screen, would you mind pulling `ByteArrayToHexString` into `StringUtils`.

And regarding `StringUtils`: We've done a bunch of refactors to clean and sort `using` statements. If you have Visual Studio 2017, that can be done automatically, and it'd be nice to do so, again for consistency.

Measuring:
@MarijnS95 Thanks for the detailed review. I'll take some time to take everything in and do a polishing pass.

I plan on keeping the UI in this PR with using IMGUI. It's old but easy and pretty enough for tech-users ;). 

The join/create server could do with using the style of Subnautica UI but I think we really need a new API for that to easy recreate the style OR load a custom unity scene holding our UI. IMO the latter is easier to do and to maintain but it does require people to install Unity if they want to modify it properly.

MarijnS95:
I actually have to thank you! Glad to see some contributions again, since previously active people (including myself and @Sunrunner37) [haven't been able to work on it](https://github.com/SubnauticaNitrox/Nitrox/graphs/contributors) as much as we'd like. Or better said: do anything at all in the past few months.

I totally agree on staying with IMGUI for *development tools*, the game UI is just extra fluff taking up screenspace and wasting development time.

I've never worked with Unity besides touching the script side within this mod, so I can't provide any judgement there. Though building it from code is ugly and annoying; tried to refactor the current system myself, and there's a PR lying around to implement a settings menu. Both of which are far from ideal.

NitroxClient/MonoBehaviours/Debugging/SceneDebugger.cs
Measuring: The initial plan was as you say but because I wanted to support multiple debuggers/windows you would need to define an enum per debugger. Still it's better than having to remember numbers.

MarijnS95: Exactly, I think the extra burden of defining them somewhere outweighs numbers that are easy to forget (and may cause confusion). Enums are basically numbers with a name anyway.

NitroxClient/MonoBehaviours/Debugging/SceneDebugger.cs
Measuring: Good catch. I first hard Resharper warn me about many things like this but I don't have a license for it anymore. You got recommendations for Visual Studio 2017 extensions and tools?

MarijnS95: I don't sadly. You should be able to generate some offline reports, though in this case I just noticed because I wanted to know where this was used.

NitroxClient/MonoBehaviours/Debugging/NetworkDebugger.cs
Measuring: public NetworkDebugger() : base(400, KeyCode.N, true, false, false, GUISkinCreationOptions.DERIVEDCOPY)

Doesn't work because Unity won't call the `base()` constructor because classes implement `MonoBehaviour` and act like 'scripts'. Has to do with how Unity works. I can't use `readonly` either because I can't set it from the derived constructor (in SceneDebugger/NetworkDebugger).

I could refactor as much code as possible into a normal C# class for both debuggers but it won't be pretty in the projects hierarchy imo. I can break convention or work around Unity. Which would you say is best?

MarijnS95: Of course, I forgot Unity instantiates objects similarly to deserializing; without calling the constructor.

Maybe use get-properties?

MadMaxOfYore: I've had success in the past by building a "core" library containing models that encapsulated all of my behaviors. The public API's were structured with Unity in mind, but each and every one was a SOLID object.

My MonoBehavior objects then take on an Adapter pattern as they are really only responsible for communicating user interactions between Unity and my game's real engine. I initialize an instance of a model that implements a specific behavior during the Awake of a MonoBehavior. Then, it is just a matter of calling the appropriate engine API method whenever an action is invoked on the MonoBehavior.

I don't know what the rules are for linking to other repos, but I do have a very simple example that demonstrates the design pattern I am referring to. If you're interested, then I think that could lend itself towards solving the problem of encapsulation while also dealing with the procedural nature Unity itself.

Measuring:
@MarijnS95 I've made the changes, didn't have the time to comment yesterday. Do you approve this PR as is?

NitroxClient/Unity/Helper/GUISkinUtils.cs
```diff
@@ -0,0 +1,139 @@
...
+        }
+
+        /// <summary>
+        /// Switches the active skin until after the <paramref name="render"/> action.
+        /// </summary>
+        /// <param name="skin">Skin to switch to.</param>
+        /// <param name="render">Render function to run.</param>
+        public static void RenderWithSkin(GUISkin skin, Action render)
+        {
+            if (render == null)
```
MarijnS95: `Validate.NotNull`

NitroxClient/Unity/Helper/StringUtils.cs
```diff
@@ -0,0 +1,50 @@
...
+using System.Linq;
+using System.Text;
+
+namespace NitroxClient.Unity.Helper
+{
+    public static class StringUtils
+    {
+        public static string TruncateRight(this string value, int maxChars, string appendix = "...")
+        {
+            if (value == null)
```
MarijnS95: `Validate.NotNull`.

Measuring:
@MarijnS95 

Got those changes in as well.

Measuring:
@Sunrunner37 PR got reviewed. Can you merge it for me?

MarijnS95:
@Measuring I just asked him, he's (probably) at work and not able to do so right now.

NitroxClient/Debuggers/SceneDebugger.cs
```diff
@@ -0,0 +1,234 @@
...
+                s.fixedWidth = 100;
+            });
+        }
+
+        protected override void Render(int windowId)
+        {
+            switch ((TabTypes)ActiveTab)
+            {
+                default:
+                case TabTypes.SCENES:
```
Sunrunner37: Do you think this would be more readable if the tab render resolution was included in the base class?  I was thinking that you could have a base class method called AddTab that takes a name and a function pointer to the render function.   This could simply add it to a map in the base class that can be looked up in the render function.  Some benefits of doing this:

1) No need to override render in all of the children
2) No need to declare a new enum in all of the children
3) Allows us to break up some of these really huge switch statements
4) Allows better information hiding by not bubbling up more fields to children classes 

Feel free to take or leave this suggestion.  Either way, I would recommend breaking the cases into methods for readability.

NitroxClient/MonoBehaviours/NitroxDebugManager.cs
```diff
@@ -0,0 +1,129 @@
...
+        }
+
+        public void Update()
+        {
+            if (Input.GetKeyDown(EnableDebuggerHotkey))
+            {
+                ToggleDebugging();
+            }
+
+            if (Input.GetKeyDown(KeyCode.C) && Input.GetKey(KeyCode.LeftControl))
```
Sunrunner37: Do we want to check the rest of this stuff if we are not debugging?

NitroxClient/Debuggers/SceneDebugger.cs
Measuring: This is a very good suggestion. I've changed it and it's even more clear than before. Thanks.

NitroxClient/MonoBehaviours/NitroxDebugManager.cs
Measuring: Only the ToggleCursor part worked when you're not debugging but it probably clutters the game more than it is useful (you can always just enable debugging with F7 then CTRL+C to toggle the cursor). I've changed it.

PR #88, opened 2018-01-26
Assets tools reading

nesrak1:
Can read data directly from assets files. Either reads from the steam subnautica dir or the resources.assets file in local dir (possibly dedicated server maybe?)


NitroxServer/Serialization/BatchCellsParser.cs
```diff
@@ -80,8 +80,7 @@ public List<EntitySpawnPoint> ParseBatchData(Int3 batchId)
 
         public void ParseFile(Int3 batchId, string pathPrefix, string suffix, List<EntitySpawnPoint> spawnPoints)
         {
-            // This isn't always gonna work.
-            string path = @"C:\Program Files (x86)\Steam\steamapps\common\Subnautica\SNUnmanagedData\Build18\";
+            string path = Path.Combine(SteamFinder.FindSteamGamePath(264710, "Subnautica"), "SNUnmanagedData/Build18") + "/";
```
MarijnS95: We should detect the path only once in the code, allowing us to overwrite it later based on a config file, if desired.

NitroxServer/GameLogic/Spawning/EntitySpawner.cs
```diff
@@ -106,5 +110,74 @@ private void SpawnEntities()
...
         }
+
+        private bool GetDataFiles(out string lootDistributions, out Dictionary<string, WorldEntityInfo> worldEntityData)
+        {
+            lootDistributions = "";
+            worldEntityData = new Dictionary<string, WorldEntityInfo>();
+            string resourcesPath = "";
+            string gameResourcesPath = Path.Combine(SteamFinder.FindSteamGamePath(264710, "Subnautica"), "Subnautica_Data/resources.assets");
+            AssetsFile resourcesFile;
+            if (File.Exists(gameResourcesPath))
```
MarijnS95: Not a fan of `File.Exists` before opening a file. It's safer to just open the file (in `new FileStream`) and deal with the exceptions appropriately. That way the code is safe against race conditions with someone/something else accessing the file.

NitroxServer/GameLogic/Spawning/EntitySpawner.cs
```diff
@@ -106,5 +110,74 @@ private void SpawnEntities()
                 }
             }
         }
+
+        private bool GetDataFiles(out string lootDistributions, out Dictionary<string, WorldEntityInfo> worldEntityData)
```
MarijnS95: Shouldn't this be `static`, because you're using `out` to write into a member field and are not accessing anything else?

NitroxServer/GameLogic/Spawning/EntitySpawner.cs
```diff
@@ -106,5 +110,74 @@ private void SpawnEntities()
...
+            {
+                resourcesPath = Path.GetFullPath("../resources.assets");
+            }
+            else if (File.Exists("resources.assets"))
+            {
+                resourcesPath = Path.GetFullPath("resources.assets");
+            }
+            else
+            {
+                Log.Error("Could not find resources.assets in parent or current directory.", new FileNotFoundException());
```
MarijnS95: `throw new FileNotFoundException("Could not find resources.assets in parent or current directory.")`? Then deal with the exception at a higher level rather than passing a bool?

NitroxModel/Helper/SteamFinder.cs
```diff
@@ -0,0 +1,76 @@
...
+using NitroxModel.Logger;
+using System;
+using System.IO;
+using System.Text.RegularExpressions;
+
+namespace NitroxModel.Helper
+{
+    public class SteamFinder
+    {
+        public static string FindSteamGamePath(int appid, string gameName)
```
MarijnS95: Can you return an `Optional<string>` or throw an exception when the path cannot be resolved from the registry? This to properly indicate that it isn't available.

NitroxServer/Program.cs
```diff
@@ -9,15 +9,15 @@ static void Main(string[] args)
         {
             Log.SetLevel(Log.LogLevel.ConsoleInfo | Log.LogLevel.ConsoleDebug);
 
-            try
-            {
+            //try
```
MarijnS95: Shouldn't this be uncommented again?

You could wrap it in `#if DEBUG`.

NitroxModel/Helper/SteamFinder.cs
```diff
@@ -0,0 +1,76 @@
...
+        {
+            if (ReadRegistrySafe("Software\\Valve\\Steam", "SteamPath") == null)
+            {
+                Log.Info("You either don't have steam installed or your registry variable isn't set.");
+                return "";
+            }
+
+            string appsPath = (string)ReadRegistrySafe("Software\\Valve\\Steam", "SteamPath") + "/steamapps/";
```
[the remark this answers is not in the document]
nesrak1: Sure thing, looks like this is an old version from years ago before I knew about path.combine.

NitroxServer/GameLogic/Spawning/EntitySpawner.cs
```diff
@@ -106,5 +110,74 @@ private void SpawnEntities()
...
+                return false;
+            }
+
+            using (FileStream resStream = new FileStream(resourcesPath, FileMode.Open))
+            {
+                resourcesFile = new AssetsFile(new AssetsFileReader(resStream));
+                AssetsFileTable resourcesFileTable = new AssetsFileTable(resourcesFile);
+                foreach (AssetFileInfoEx afi in resourcesFileTable.pAssetFileInfo)
+                {
+                    if (afi.curFileType == 0x31) //TextAsset
```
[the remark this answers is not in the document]
nesrak1: I can compile up a classdatabase to use so that all the serialization data comes from file rather than manually coding the names and the readxxx methods. It's another file though, and requires a little more setup.

NitroxServer/Serialization/BatchCellsParser.cs
nesrak1: Sure, I just don't know where to put it where both classes to reference it. Static string seems to hacky, however it is true the path will never change.

NitroxServer/GameLogic/Spawning/EntitySpawner.cs
nesrak1: Good point. The template was copied from the library itself, and it uses some weird practices since it was ported from c++.

NitroxServer/GameLogic/Spawning/EntitySpawner.cs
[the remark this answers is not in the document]
MarijnS95: As you might know, I'm not a fan of precompiled files, especially if it's not integrated with the VS project and generated/updated when the project is rebuilt. That way we rely on a file that some contributor to the project once made and added. Of course, providing the tools and documentation helps, but it's still an extra dependency we have to watch out for.

Anyway, that wasn't really the point. If you can migrate these values to an enum, I'm happy.
```csharp
enum AssetFileType : int {
    TextAsset = 0x31,
    MonoBehaviour = 0x72
}
```
Besides, are these values fixed for all (Unity) asset files, or is this specific to Subnautica?

Finally, as you know this has sparked some discussion in Discord regarding rolling our own code that effectively does the same as the game. I still want to see if we can look further into reusing the game and library code, as this saves us a lot of time and dependency maintenance.

NitroxServer/GameLogic/Spawning/EntitySpawner.cs
```diff
@@ -111,6 +115,22 @@ private void SpawnEntities()
             }
         }
 
+        private bool IsValidSpawnType(string id, bool creatureSpawn)
+        {
+            if (worldEntitiesByClassId.ContainsKey(id))
```
MarijnS95: How about `TryGetValue`? Saves an extra lookup in case the key exists.

NitroxServer/UnityStubs/Transform.cs
```diff
@@ -7,12 +7,12 @@ public class Transform
     {
         [ProtoMember(1)]
         public Vector3 Position { get; }
-
+
```
MarijnS95: Whitespace.

NitroxServer/Serialization/BatchCellsParser.cs
MarijnS95: That's solved once we have an IoC container as mentioned by @MadMaxOfYore in #101. This allows us to create a singleton class responsible for providing the settings from a configuration file (and sensible default values), which is then accessible through the container from everywhere else in the program without explicitly passing it around.

MarijnS95:
I'm not sure what to do with this PR (but I'm not the maintainer so it's not up to me to decide 😉). In it's current state it's a mix of about 3 different features which have little to nothing to do with eachother, of which 2 are blocking on other code or further research (unless they are split differently).

- Steam path detection: This could probably live on it's own, but it's better to implement some kind of settings manager as mentioned in the review.
- Asset bundle reading: Also mentioned in the review: It works, but we might want to look into some other solutions that don't require our own implementation of something the game engine can do for us. Research takes time, so it's probably not worth blocking the PR for that, though it's on my todo/tech-debt list for sure.
- Fixes for spawn position and rotation: If you kick this commit off the current branch, they can be filed separately without blocking on the other two.

NitroxServer/GameLogic/Spawning/EntitySpawner.cs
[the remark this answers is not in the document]
nesrak1: I understood the whole enum thing, I probably was going off the subject. I usually only use things like enums when they are used multilple times and having an enum just there and only being used once seems kind of weird to me. The values are halfway fixed. The TextAsset will always read that way and the MonoBehaviour as you know can change if they decide to change any fields in that class. Since they are using a custom build of unity, it is very very unlikely that they will use a different unity version, and the IDs will remain the same.
The code that reads these two values comes from the assets files, and these are read natively by the player exe, not by the Assembly-CSharp, otherwise I would've just used the method in unity.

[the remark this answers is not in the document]
MarijnS95: I agree, but I also prefer to use them to hide such magic values, even if they are only used once. Makes the intent clear without having to specify it in a comment, or looking up what value you need elsewhere.

Thanks for the info on Unity, I think it's okay to accept the ID's for now, assuming they won't change anytime soon.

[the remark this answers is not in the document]
MadMaxOfYore: The general rule of thumb is to use an enum whenever you need to define a static range of values that a specific value *must* be within. They also have the added benefit of being (mostly) human readable. However; it is an implementation detail that clarifies a specific intent.

The catch here is that we cannot enforce what the value on the provided instance *should* be. We can only write some control statements to react to possible values that it *could* be. As such, I do not think that an enum would be the best choice in this context.

There still remains the issue of a hard coded value within the control statement itself - which is generally considered a bast practice for readability concerns. To address that, consider creating some class-level constants that describes the value that you are storing within it and reference those in your code instead. 

Ex:
```c#
class Demo {
private const int TEXT_ASSET_TYPE = 0x31;

   public void doThis(instance) {
      if(instance.type == TEXT_ASSET_TYPE)...
   }
}
```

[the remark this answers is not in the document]
nesrak1: I think I agree with this more. It's only two lines and I don't have to cast the enum to a uint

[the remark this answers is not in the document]
MarijnS95: Indeed, `const` is the alternative I was looking for, since there is currently no function or variable that can only accept that enum type to specify exactly what range *or set* of values is allowed. `Const` does exactly what I was looking for in the first comment: a human-readable 'accessor' to a magic number.

NitroxServer/Serialization/BatchCellsParser.cs
MadMaxOfYore: Yeah - something like this would be a great candidate to roll up within a ConfigurationProvider. However; given the game world that we live in, any sort of centralized ConfigurationProvider may need to take too many references and could end up being a pretty nasty God class. 

In this minimalist environment, it may be prudent to simply roll an ISteamPathProvider that carries the responsibility of providing this information on demand. The ISteamPathProvider abstraction need not be implemented in isolation. Rather, we could write a facade called SteamProvider that implements multiple interfaces. Everyone would reference ISteamPathProvider but would be provided with an instance of SteamProvider after resolving the dependency from an IoC container.

This isn't something that I'd expect nes to address right now though. I'd say let it roll as-is until we can get an IoC implementation prioritized. For now - I think commenting a TODO in the code that addresses the tech-debt and intended solution would be "good enough" for this PR.

NitroxServer/GameLogic/Spawning/EntitySpawner.cs
MadMaxOfYore: Agreed - throwing the exception and reacting to it accordingly in consumer code is a clearer intent that this paradigm.

NitroxServer/GameLogic/Spawning/EntitySpawner.cs
MadMaxOfYore: That depends. If this is an area where the exception could be raised many times, then it is better to make sure that you're not going to run into that edge case repeatedly. Generating and reacting to an exception is a heavy-handed approach that actually lead to performance problems in some edge cases.

Generating exceptions in .NET can be a pretty hefty task - especially if you have a very deep call stack. Avoiding exceptions when possible (and appropriate) is generally the safer approach.

MarijnS95: Sure, if this code is in a critical path, then it's better to check `.Exists` first. Even then, the actual opening of the file should also be wrapped in an exception handler and dealt with appropriate on errors. As far as I'm aware, it's possible to lock a file for exclusive access, meaning that no other application can even read it.

NitroxServer/GameLogic/Spawning/EntitySpawner.cs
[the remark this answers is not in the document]
MadMaxOfYore: Bingo! Using an enum in this case would be like using a hammer as a paper weight. No one can argue that itserves the function perfectly, but everyone else that walks into the room is really confused about why there is a hammer sitting on a stack of papers on your desk...

MadMaxOfYore:
I *think* that nes had already spoken with @Sunrunner37 about lumping them all together in Discord - but I may be wrong on that. 

The Steam path detection can probably be taken now as-is and refactored once we address the dependency management concerns. 

Are the spawn fixes broken still? If so, then I agree that we will at least want to create one other branch that cherry-picks the other two acceptable features into it. This branch can continue to live on while we accept the PR on the others.

NitroxServer/GameLogic/Spawning/EntitySpawner.cs
MadMaxOfYore: You're absolutely right that the file could be locked. If we controlled the environment, then we could implement an OS level semaphore that we use to synchronize access to that file.

Now that it is being mentioned - don't we have complete access control to this resource? If so, then I'd say roll a semaphore to address that concern if it is a reality.

That being said - "file exists" is still a different scenario from "file is locked" and should be handled in isolation. In the absence of access control, then handling the exception is the only recourse. Even then, I would expect to see some retry logic in such cases rather than a fail-fast - particularly in this context. 

However; I really don't like the idea of using exceptions in place of control statements. An exception flow should log the exception at a minimum and try to rectify the current exception at most. It would give me heartburn to see an exception being used as an indication to "try something else" when there are alternatives that cleanly express the intent.

MarijnS95: Agreed, this piece of code is the worst place to use this technique, because `.Exists` is genuinely used for providing alternatives.

However, what I don't like is the unchecked file access. I've been involved in tons of debugging/discussion sessions with other studens, wondering why `file.Exists` sometimes returned `true` while the opening of the file crashed with a `FileNotFound` exception. They knew the file was being touched by another program, but didn't "understand" the programs were executing in parallel. Since then I always try to advice to wrap file IO in a `try-catch` block; rather deal with an edgecase appropriately than to crash the program. And if the file is really expected to be there (for instance, if it's part of the download/installation; not something optional), I usually omit the `.Exists` check entirely.

Again, I should've simply mentioned to wrap the `new FileStream` part in a `try-catch` block rather than comment on this block of alternative-checking. This is really an edgecase since no other program, or thread in this mod should touch the file.

nesrak1: Imo, best and cleanest idea is to make a method that checks for file existence, then checks if it can read it. Too much to repeat and would rather not code something knowing it's going to throw exceptions

nesrak1:
Spawns just need to be offset, Sunrunner and I were already discussing this problem in discord. Seems the world center isn't exactly the center of the 12,18,12 batch

MadMaxOfYore:
Do you have an ETA on when the spawn issues will be addressed in this branch?

nesrak1:
Today for sure. It's a 3 line fix, so relatively easy to do. I just need to make sure the numbers are perfect before I commit again.

MadMaxOfYore:
Understood - then I would be amenable to accepting this PR as-is.

PR #94, opened 2018-01-28
Fix bug that prevented fishes to load

ioncodes:
This removes the absolute path in BatchCellsParser and introduces a settings variable. You now have to right-click on NitroxServer and click Properties. Then go to settings and create a new file. Then you have to add a property called "BuildDir" with the value to the Build18 folder of Subnautica.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -274,3 +274,7 @@ paket-files/
 # JetBrains Rider
 .idea/
 *.sln.iml
+
+NitroxServer/Properties/Settings\.Designer\.cs
+
+NitroxServer/Properties/Settings\.settings
--- a/NitroxServer/NitroxServer.csproj
+++ b/NitroxServer/NitroxServer.csproj
@@ -82,6 +82,11 @@
     <Compile Include="Player.cs" />
     <Compile Include="Program.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
+    <Compile Include="Properties\Settings.Designer.cs">
+      <AutoGen>True</AutoGen>
+      <DesignTimeSharedInput>True</DesignTimeSharedInput>
+      <DependentUpon>Settings.settings</DependentUpon>
+    </Compile>
     <Compile Include="Serialization\BatchCellsParser.cs" />
     <Compile Include="Serialization\LootDistributionsParser.cs" />
     <Compile Include="Serialization\ServerProtobufSerializer.cs" />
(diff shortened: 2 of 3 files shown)


MarijnS95:
That is correct, but like I said: We plan to have a local settings file, next to the `.exe` (rather than the builtin way that places this file somewhere in the `AppData` folder), but haven't decided on the format yet. If we now temporarily allow to use `Properties` that only causes confusion later on.

As far as we are aware, all distributions of the game require Steam in order to run (eg. no DRM-free version). We do not support (or rather: intentionally try to break) illegal copies of the game.

ioncodes:
@MarijnS95 Yes but I'm legally allowed to play a copy of a game if I'm already the rightful owner of it (eg bought on Steam). I'd recommend a simple json for the properties. And also don't use Steam detection, in case there are custom folders or it's installed on a removable hard drive. I'd rather scan the OpenGL hooks and check for instance paths. The ID should be smth like Subnautica_<smth here>. Then in the GL_init you can backtrace the folder. Like that you also have runtime access to the wrapper.

MarijnS95:
There's also `yaml`, `xml`, regular "`conf`" files, and the list goes on. Tons of formats, and we probably choose one based on builtin support.

Steam detection is only for local usage, and easy development setup. Similarly, detection through OpenGL hooks is not possible if the server is running externally, on a PC/server that does not have the game installed and running. Hence the necessity for a config file.

ioncodes:
@MarijnS95 There's only 1 standard and that's xml, but JSON is becoming the new standard in C#. Also, the client could just send the directory to the server?

MarijnS95:
The server maintainer can also copy it manually, implementing file transfer just for these few files is out of scope of the project *IMHO*.

ioncodes:
Yeah ofcourse but then he would have to manually edit config files and you even said that this project tries to keep it simple for the end user.

MarijnS95:
Disregard what I've previously said, here's a rewrite:

1. Another PR has been filed a few days ago, that fixes the exact same issue, albeit in a different manner.
2. **In my opinion**, both methods in these two PR's should be combined (as I've already mentioned in a  comment on the other PR).
  - Detecting the files through Steam makes it easier for developers and users running the server on their local gaming machine without prior configuration.
  - But providing the option to override this path in a config file is invaluable, especially if this server application runs on a separate PC/server that does not have the game installed. Or if someone wants to override the default detected Steam path.
  - This config file will also contain other server properties, such as the initial gamemode, port, maximum number of players and so on.
3. The `Properties` method used in this PR stores the config file in an obnoxious folder somewhere hidden far away in the `AppData` folder. Ideally this file is placed in the same folder as the server executable, making it easy to find and access.
4. Letting a random client upload the files needed by the server is unsafe:
  - First off, this requires us to write and maintain code to do so.
  - Everyone can modify the game files; if they're the first to connect to a server they can send anything they'd like. Malware doesn't do much as long as it isn't executed, but changing the spawn file to spawn tons and tons of reapers everywhere ruins the game experience (but probably makes the server grind to a halt long before that).
  - A server owner doesn't like that approach - if they go as far as setting up a separate machine to run the server they probably know how to copy some required files to it.

PR #103, opened 2018-01-31
Refactored code-patch-restore and main-menu-mp-btn to monobehaviours

Measurity:
Also fixed the chat button binding ambiguous method call exception. Chat button setting is visible again while in-game playing mp.

Want to refactor more patches but don't want to change too much at a time.


NitroxClient/MonoBehaviours/NitroxDebugManager.cs
```diff
@@ -125,5 +126,34 @@ private void ShowDebuggers()
...
+        }
+
+        private void OnDisable()
+        {
+            SceneManager.sceneLoaded -= SceneManager_sceneLoaded;
+            SceneManager.sceneUnloaded -= SceneManager_sceneUnloaded;
+            SceneManager.activeSceneChanged -= SceneManager_activeSceneChanged;
+        }
+
+        private void SceneManager_sceneLoaded(Scene scene, LoadSceneMode loadMode)
```
MarijnS95: Was this for your own debugging purposes, or do you intend to leave it in here for future adventures?

Measurity: It is purely for getting more technical information in the output-log.txt. It's easier to see where an error occurred in the game if you know what got loaded etc.

MarijnS95: Right, sounds like a helpful thing to keep in there.

PR #148, opened 2018-02-09
Cyclops Syncing

RabidCrab:
Syncs overall Cyclops health status, including:

- Spawning
- Destruction
- External hull damage points
- Fires
- Collision damage
- Depth damage
- Creature damage


NitroxClient/Communication/Packets/Processors/CyclopsDamageProcessor.cs~HEAD
```diff
@@ -0,0 +1,58 @@
...
+    /// Cyclops fire damage management is done in <see cref="SubFire"/> located in <see cref="CyclopsExternalDamageManager.subFire"/>. They do not have any dependencies on each other,
+    /// but I may be wrong. The <see cref="LiveMixin"/> is at <see cref="SubFire.liveMixin"/>, and looks to be a reference of <see cref="SubRoot.live"/>. If the health is below 80%,
+    /// there's a chance a fire will start.
+    /// </para><para>
+    /// Cyclops fire extinguishing is handled at <see cref="Fire.Douse(float)"/>. A large douse amount triggers the private <see cref="Fire.Extinguish()"/> method.
+    /// </para><para>
+    /// Currently this does a full re-sync of damage points and fires due to having no set owner. Any client can call for an change of the Cyclops' health state.
+    /// </para>
+    /// </summary>
+    public class CyclopsDamageProcessor : ClientPacketProcessor<CyclopsDamage>
```
Sunrunner37: This file was commited twice (CyclopsDamageProcessor.cs~HEAD)

NitroxClient/Communication/Packets/Processors/CyclopsDamageProcessor.cs
```diff
@@ -0,0 +1,58 @@
...
+        public CyclopsDamageProcessor(IPacketSender packetSender)
+        {
+            this.packetSender = packetSender;
+        }
+
+        public override void Process(CyclopsDamage packet)
+        {
+            SubRoot cyclops = GuidHelper.RequireObjectFrom(packet.Guid).GetComponent<SubRoot>();
+
+            Multiplayer.Logic.Cyclops.SetActiveDamagePoints(cyclops, packet.DamagePointIndexes,
```
Sunrunner37: Are all of the files in this PR?  I'm not seeing the implementation for SetActiveRoomFires or SetActiveDamagePoints

NitroxClient/Communication/Packets/Processors/CyclopsDamageProcessor.cs
```diff
@@ -0,0 +1,58 @@
...
+                packet.SubHealth,
+                packet.DamageManagerHealth,
+                packet.SubFireHealth);
+
+            Multiplayer.Logic.Cyclops.SetActiveRoomFires(cyclops, packet.RoomFires,
```
Sunrunner37: This looks slightly non-standard as we normally only call the static Logic path from patching logic.  Normally the processor will directly call the relevant in-game code.  However, I'll have to see your implementation(above comment) first to decide.

NitroxClient/GameLogic/ServerTime.cs
```diff
@@ -0,0 +1,68 @@
+﻿using System;
+using NitroxClient.Communication;
+using NitroxModel.Logger;
+
+namespace NitroxClient.GameLogic
+{
+    /// <summary>
+    /// The synchronized UTC server time.
+    /// </summary>
+    public class ServerTime
```
Sunrunner37: Seems like a lot of effort to sync random numbers.  Was there not an option to make the random number deterministic instead?

NitroxPatcher/Patches/Fire_Douse_Patch.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    /// <summary>
+    /// Both owners and non-owners will be able to douse fires on a ship.
+    /// </summary>
+    public class Fire_Douse_Patch : NitroxPatch
+    {
+        public static readonly Type TARGET_CLASS = typeof(Fire);
+        public static readonly MethodInfo TARGET_METHOD = TARGET_CLASS.GetMethod("Douse", BindingFlags.Public | BindingFlags.Instance, null, new Type[] { typeof(float) }, null);
+
+        public static void Postfix(Fire __instance, float amount)
```
Sunrunner37: It looks like the Fire.Douse() code is called via unity update (every frame).  This appears that it will cause a massive amount of packets to go out.  In the past, we have stored a buffer and broadcasted every 10% or so.  You can see an example in construction packet sending code.

NitroxClient/Communication/Packets/Processors/CyclopsDamageProcessor.cs
MarijnS95: @Sunrunner37 I bet you were using <kbd>ctrl</kbd>+<kbd>f</kbd> - GitHub hides that file because it's considered a large diff.

NitroxClient/GameLogic/ServerTime.cs
MarijnS95: I had a discussion with @RabidCrab about this in discord; turns out this was only used for prediction, something which is completely irrelevant in this case. It was also mentioned that we need the simulation ownership framework to decide which client is authorative on the fires and damage numbers; as far as I understand this is to prevent clients spawning fires at different positions, only to be removed again once the server validated the damage/fire state array of one of the clients.

It wouldn't make sense to synchronize the random number generator anyway, since the entire codebase uses it. Once the clients call it a different amount of times it's already out of sync. A much better solution would be to rng the numbers on the simulating/owning client, and distribute that to the others (where the random generation of fires/damages is blocked, obviously).

NitroxPatcher/Patches/Fire_Douse_Patch.cs
MarijnS95: Color selection (in the cyclops) faces the same issue. I *tried* to make a generic approach to ratelimiting such packets, but haven't finished it yet. See https://github.com/MarijnS95/Nitrox/tree/ratelimiter.

NitroxClient/GameLogic/ServerTime.cs
Sunrunner37: Ok, so we will forego this logic in favor of updates to simulation framework?

RabidCrab: I dumped this idea after talking with @MarijnS95 about it and coming to the conclusion that it's not necessary to synchronize the damage and fire if there's a designated owner. At the moment there's no ownership designation, but @MarijnS95 is working on a system that I'll be using when it's ready.

I'll remove all of the time syncing code, unless we want to keep it for a day/night cycle accurate down to the milliseconds. Considering the game code rounds time to the nearest second, I doubt it'll be needed.

MarijnS95: IIrc the time is currently only synced once with the server, when the client joins. Over time this may get skewed, or command might be used to alter the time. In any case, we should reuse or improve the existing system (if I read correctly, this PR adds a separate method of syncing time).

NitroxPatcher/Patches/Fire_Douse_Patch.cs
RabidCrab: I completely forgot to look into rate limiting this, good thing you caught it. 

I found the relevant buffer code, and it looks like I can use most of it. I think I'm going to modify it to send a packet once the player douses the fire a set amount or extinguishes it since the health of the fire can grow or shrink rapidly.

NitroxClient/Communication/Packets/Processors/CyclopsDamageProcessor.cs
RabidCrab: If I move over the logic to the processor, I will either need to duplicate DouseFire and RepairDamagePoint, or call them statically from the Processor.

After thinking about it a bit, I think it'd be best to move the two called methods to the processor, and keep the DouseFire and RepairDamagePoint in the Cyclops class. A complete reset of the damage and fire points should be something only the Processor does, whereas getters for the damage points and fires, and specific point damage, repair, and fire dousing should be something possible for external classes to use.

NitroxClient/GameLogic/ServerTime.cs
RabidCrab: @MarijnS95 , it uses the existing time system, but extends it to include the server's UTC time. When the client receives a TimeChange packet without a ClientSentAtMilliseconds (happens after authentication), it sets that value to the current time in milliseconds and sends the packet back to the server. The server will take the returned packet, set CurrentServerUtcTimeMilliseconds, and send the packet back to the client. The round-trip time is divided by 2 and added to the server time passed with the packet to account for latency.

Time is then determined by adding the current time delta to that received time. As long as their clock doesn't stop working, the time will be accurate with the server, generally within a few milliseconds.

Although the solution works, I can't think of any reason why we would need to know the server's time down to the milliseconds beyond what I originally intended for it to do.

MadMaxOfYore:
What is the status of this? It seems to be getting stale at this point.

MarijnS95:
@MadMaxOfYore it's blocking on EntityOwnership, which as you know is still in discussion, design and development phase.

MadMaxOfYore:
@MarijnS95 

Interesting. What about this feature is blocked on entity spawning? It seems like we should be able to accept it if it is synchronizing the cyclops state and events.

MarijnS95:
@MadMaxOfYore entity **ownership**; in this case the cyclops is an entity that needs to have some form of simulation right management. Last I remember this PR tried to do it manually by deciding on the server which of the update packets were distributed. We've already decided to need such a framework, which replaces the server management side entirely by only allowing a single simulating client to decide what happens.

Unfortunately I haven't yet, and will not have the time to update the framework like how we decided in discord. Not sure how to proceed with this PR: merge as-is, or wait for *someone* to finalize the ownership system.

Sunrunner37:
Per the newest batch of commits, this PR is now unblocked.  https://github.com/SubnauticaNitrox/Nitrox/commit/a869dd87c6359783c11da41e520a6f8ea4aeb73b

RabidCrab:
I should be able to take a look at this over the weekend, although I expect it won't be until the middle of next week I will be able to put some work into implementing ownership changes. 

I haven't taken a close look into how ownership is determined, but from a cursory look it looks like I'll need to make a few changes on packet sending. Fortunately, it looks like the actual damage logic can be converted without any large changes.

Sunrunner37:
Let me know when this is ready for review 👍

RabidCrab:
Will do. It looks like some of the functions no longer work correctly.

I have a couple of questions about registering dev console commands and simulation ownership.

Previously, I used a hack to spawn a cyclops that no longer works. I noticed we have two different places to execute commands. One is DevConsole with the commands warpto, mplayer, and disconnect. The other is ConsoleCommands with say, list, kick.

Spawning a sub can be a client or server executed command, or both. For the client, it would do the equivalent of the "sub" command in the game, where it spawns a sub in front of the player. For the server, it would be more like a spawn command that requires a spawn coordinate. Which would be the preferable option? Or should I do both?

The other question has to do with ownership. Is SimulationOwnership.HasExclusiveLock() what I'm looking for to determine simulation ownership? Would it be preferable to have the client check its ownership status each damage update, or should I have every client broadcast a sub's damage status and have the server manage the sub status?

Sunrunner37:
1) For now patching just 'sub' will probably be fine.  Later we can make a server command that just executes the same code path.
2) You should check the exclusive lock on the client before broadcasting damage. (exactly how you described doing it)

RabidCrab:
Checking for an exclusive lock does not seem to work for crafted or command created Cyclops. Do I need to request an exclusive lock for whenever the Cyclops takes damage?

Sunrunner37:
Sorry, you should actually use the `hasAnyLock` as someone simulating, but not piloting, would have a transient lock.

RabidCrab:
Damage syncing is ready to go again.

There's a couple of things I want to do to make it a more polished experience, but the actual functionality of Cyclops damage syncing is ready.

NitroxClient/GameLogic/Cyclops.cs
```diff
@@ -53,5 +75,452 @@ public void ActivateShield(string guid)
...
+
+        /// <summary>
+        /// This is run during every major state change. If this player is the current owner, a <see cref="CyclopsDamage"/> packet will be sent out
+        /// </summary>
+        private void BroadcastDamageChange(SubRoot subRoot, Optional<DamageInfo> info)
+        {
+            string subGuid = GuidHelper.GetGuid(subRoot.gameObject);
+
+            // If it isn't the owner, we don't want to broadcast
+            if (!simulationOwnershipManager.HasAnyLockType(subGuid))
```
Sunrunner37: Do we want to prevent this damage from happening in the first place?  Seems like at this point it is out of sync.

NitroxClient/GameLogic/Cyclops.cs
```diff
@@ -53,5 +75,452 @@ public void ActivateShield(string guid)
...
+            {
+                return;
+            }
+
+            LiveMixin subHealth = subRoot.gameObject.RequireComponent<LiveMixin>();
+
+            SerializableDamageInfo damageInfo = null;
+
+            if (info.IsPresent())
```
Sunrunner37: Should this block be inside the below `if (subHealth.healh > 0)` block? (we don't want to compile this data if the cyclops ist going to be destroyed.)  If you were to do that, might make sense to have a new function to contain all of the logic like: `buildCyclopsDamagePacket()`

NitroxClient/GameLogic/Cyclops.cs
```diff
@@ -53,5 +75,452 @@ public void ActivateShield(string guid)
...
+        {
+            SubFire subFire = subRoot.gameObject.RequireComponent<SubFire>();
+            Dictionary<CyclopsRooms, SubFire.RoomFire> roomFiresDict = (Dictionary<CyclopsRooms, SubFire.RoomFire>)subFire.ReflectionGet("roomFires");
+            string fireGuid = GuidHelper.GetGuid(fire.gameObject);
+
+            if (!fireDouseAmount.ContainsKey(fireGuid))
+            {
+                // 500 fires in a single session? Are they trying to cook Peepers on open fires? Yes, I've tried to do it, and to my disappointment it
+                // doesn't cook them.
+                if (fireDouseAmount.Count > 500)
```
Sunrunner37: Why is this check needed?

NitroxClient/GameLogic/Cyclops.cs
```diff
@@ -1,15 +1,37 @@
...
         private readonly IPacketSender packetSender;
+        private readonly SimulationOwnership simulationOwnershipManager;
 
-        public Cyclops(IPacketSender packetSender)
+        /// <summary>
+        /// KeyValuePair<Guid, douseAmount>. Used in <see cref="OnFireDoused(Fire, SubRoot, float)"/> to reduce
+        /// the <see cref="CyclopsFireHealthChanged"/> packet spam as fires are being doused. A packet is only sent after
+        /// the douse amount surpasses <see cref="FIRE_DOUSE_AMOUNT_TRIGGER"/>
+        /// </summary>
+        private readonly Dictionary<string, float> fireDouseAmount = new Dictionary<string, float>();
```
Sunrunner37: Should we have a helper class to separate out all of the first related logic?  There is a ton of it and that would allow for some smaller method sizes when refactoring.

NitroxClient/MonoBehaviours/Gui/MainMenu/ConsoleJoinServer.cs
```diff
@@ -92,6 +97,25 @@ public void OnConsoleCommand_warpto(NotificationCenter.Notification n)
             }
         }
 
+        public void OnConsoleCommand_sub(NotificationCenter.Notification n)
```
Sunrunner37: That is cool, I didn't know you could have two handlers to the dev console events.

NitroxClient/MonoBehaviours/Gui/MainMenu/ConsoleJoinServer.cs
```diff
@@ -92,6 +97,25 @@ public void OnConsoleCommand_warpto(NotificationCenter.Notification n)
...
+        {
+            if (n?.data?.Count > 0)
+            {
+                string subName = (string)n.data[0];
+
+                switch (subName.ToLowerInvariant())
+                {
+                    case "cyclops":
+                        NitroxServiceLocator.LocateService<Vehicles>().CreateVehicle(TechType.Cyclops, Optional<string>.Empty(), MainCamera.camera.transform.position + 20f * MainCamera.camera.transform.forward, Quaternion.LookRotation(MainCamera.camera.transform.right), Optional<List<InteractiveChildObjectIdentifier>>.Empty());
```
Sunrunner37: Shouldn't we generate a GUID here to assign to the cyclops?  Also, I think this might be the place that we want to `BroadcastCreatedVehicle` and not in the create method above.

NitroxClient/GameLogic/Cyclops.cs
```diff
@@ -1,15 +1,37 @@
...
 {
     public class Cyclops
     {
         private readonly IPacketSender packetSender;
+        private readonly SimulationOwnership simulationOwnershipManager;
 
-        public Cyclops(IPacketSender packetSender)
+        /// <summary>
+        /// KeyValuePair<Guid, douseAmount>. Used in <see cref="OnFireDoused(Fire, SubRoot, float)"/> to reduce
+        /// the <see cref="CyclopsFireHealthChanged"/> packet spam as fires are being doused. A packet is only sent after
```
Sunrunner37: In a future PR (not this one) we should look into doing the generic rate limiter.  I don't think Marin finished it.

NitroxClient/GameLogic/Cyclops.cs
RabidCrab: This was one of the things I realized I missed just after committing. Currently, the fires are blocked from being created client side by an ownership check at SubFire_CreateFire_Patch.

Damage points don't get blocked like fires do, meaning there can be a de-sync between damage points until the owner broadcasts a new CyclopsDamage packet. I'll add a patch to CyclopsExternalDamageManager.CreatePoint() to intercept damage point creation.

I was hesitant to hook onto and block SubRoot.OnTakeDamage only because it's a sizeable chunk of code that doesn't do much beyond shake the player's screen. It's possible for one client to get a strike warning while the others do not, but the damage status of the Cyclops should remain unchanged if I hook onto and block the damage point creation event.

I think what I'll do is block the function if they aren't the owner, but have it run the client side sound and screen shake for non-owners. That way it'll feel responsive but also leave the Cyclops unaffected until the owner verifies it was an actual hit. Phantom hits should be a rare occurrence, but if they are too common, it'll be easy enough to remove.

NitroxClient/GameLogic/Cyclops.cs
RabidCrab: Debugging code I forgot to remove. Can't remember the exact reason, but it's no longer possible for the fire dousing amount to exceed 15.

NitroxClient/GameLogic/Cyclops.cs
RabidCrab: At least some of the fire logic will have to be refactored out of Cyclops at some point. I haven't taken a close look, but it seems like the Fire class is pretty much the only class that the Cyclops code shares with the rest of the code base.

Despite that, it is still wrapped in it's own special damage system via SubFire that manages fire creation and storage. It seems like the only thing it shares is extinguishing.

I can't seem to find any dynamic fire creation beyond the Cyclops. I suspect there's no other dynamic fires, and the ones in the Aurora are all static and unable to be respawned. Either way, I need to modify Fire_Douse_Patch to allow for fires that are not in a Cyclops, and rename CyclopsFireHealthChangedProcessor to FireHealthChangedProcessor and allow for non-Cyclops fires as well.

NitroxPatcher/Patches/SubRoot_OnTakeDamage_Patch.cs
```diff
@@ -0,0 +1,28 @@
...
+    /// Hook onto <see cref="SubRoot.OnTakeDamage(DamageInfo)"/>. If the function made it to the end, that means it created a new fire.
+    /// </summary>
+    class SubRoot_OnTakeDamage_Patch : NitroxPatch
+    {
+        public static readonly Type TARGET_CLASS = typeof(SubRoot);
+        public static readonly MethodInfo TARGET_METHOD = TARGET_CLASS.GetMethod("OnTakeDamage", BindingFlags.Instance | BindingFlags.Public, null, new Type[] { typeof(DamageInfo) }, null);
+
+        public static void Postfix(SubRoot __instance, DamageInfo info)
+        {
+            // NitroxServiceLocator.LocateService<Cyclops>().OnTakeDamage(__instance, Optional<DamageInfo>.OfNullable(info));
```
Sunrunner37: Commented out?

NitroxServer/Communication/Packets/Processors/CyclopsDamageProcessor.cs
```diff
@@ -0,0 +1,32 @@
...
+        private readonly PlayerManager playerManager;
+
+        public CyclopsDamageProcessor(PlayerManager playerManager)
+        {
+            this.playerManager = playerManager;
+        }
+
+        public override void Process(CyclopsDamage packet, Player simulatingPlayer)
+        {
+            Log.Debug("[CyclopsDamageProcessor PlayerId: " + simulatingPlayer.Id
```
Sunrunner37: Can you add this as toString() to Cyclops damage.  Would be nicer to do 

```
Log.Debug("New cyclops damage from " + simulatingPlayer.Id + " " + packet);
```

NitroxModel/DataStructures/GameLogic/SerializableRoomFire.cs
```diff
@@ -0,0 +1,35 @@
...
+        {
+            return "[SerializableRoomFire" 
+                + " Room: " + Room.ToString()
+                + " SerializableFireNodes: " + string.Join(", ", ActiveRoomFireNodes.Select(x => x.ToString()).ToArray())
+                + "]";
+        }
+
+        public override bool Equals(object obj)
+        {
+            if (obj is SerializableRoomFire)
```
Sunrunner37: do we also have to check for null?

RabidCrab: Definitely should, I'll add a check.

Sunrunner37:
Let me know when this is ready for review again.

killzoms:
btw about the ConsoleCommands they are server-client commands while DevConsole commands are mostly client only commands

micaww:
I see you're patching `Fire.Douse` to send packets when fires are doused by an extinguisher, but you're also executing this method in the packet processor on other clients. Won't this cause recursive packet sending back and forth?

NitroxPower:
Rabidcrab i found out that when somebody is driving the cyclops for the other people its really easy that they will glitch out from it when its moving there is any way you can improve it?

Sunrunner37:
Yeah it looks like we should add in packet suppression using the packet suppressor disposable.

RabidCrab:
> Rabidcrab i found out that when somebody is driving the cyclops for the other people its really easy that they will glitch out from it when its moving there is any way you can improve it?

It's a pretty big issue that happened to me a lot while I was testing. I'll be taking a look at it. Simply setting the player as a child of the Cyclops root fixes a lot of the problems, but also creates several more. I suspect it has to do with kinematics. If you go full speed on a cyclops, then hop of the wheel and stand still, you'll slide around, even in singleplayer.

The developers haven't fixed this bug themselves, and I suspect it's because the bug is a difficult one to fix. I'm not sure I can come up with a complete fix without dumping a lot of time into the problem, but I can at least fix the issue of players sliding all the way out of a Cyclops, or getting launched into orbit from glitching into one of the walls.

NitroxClient/Communication/Packets/Processors/CyclopsDamageProcessor.cs
```diff
@@ -0,0 +1,202 @@
...
+                            if (keyValuePair.Value.spawnNodes[nodeIndex].childCount > 0)
+                            {
+                                keyValuePair.Value.spawnNodes[nodeIndex].GetComponentInChildren<Fire>().Douse(10000);
+                            }
+                        }
+                        else
+                        {
+                            if (keyValuePair.Value.spawnNodes[nodeIndex].childCount < 1)
+                            {
+                                // fires.Create(fireNode.FireGuid, subRootGuid, Optional<CyclopsRooms>.Of(fireNode.Room), Optional<int>.Of(fireNode.NodeIndex));
```
Sunrunner37: minor: remove

NitroxClient/GameLogic/Fires.cs
```diff
@@ -0,0 +1,149 @@
...
+        /// Create a new <see cref="Fire"/>. Majority of code copied from <see cref="SubFire.CreateFire(SubFire.RoomFire)"/>. Currently does not support Fires created outside of a Cyclops
+        /// </summary>
+        /// <param name="fireGuid">Guid of the Fire. Used for identification when dousing the fire</param>
+        /// <param name="subRootGuid">Guid of the Cyclops <see cref="SubRoot"/></param>
+        /// <param name="room">The room the Fire will be spawned in</param>
+        /// <param name="spawnNodeIndex">Each <see cref="CyclopsRooms"/> has multiple static Fire spawn points called spawnNodes. If the wrong index is provided,
+        ///     the clients will see fires in different places from the owner</param>
+        public void Create(string fireGuid, Optional<string> subRootGuid, CyclopsRooms room, int spawnNodeIndex)
+        {
+            // A Fire is either on a Cyclops or not. Currently, creating a Fire is only possible if it's in a Cyclops.
```
Sunrunner37: Is this function even reusable for other fires?  It takes in a CyclopsRoom

NitroxClient/GameLogic/Fires.cs
```diff
@@ -0,0 +1,149 @@
...
+                    componentInChildren.fireSubRoot = subFire.subRoot;
+                    GuidHelper.SetNewGuid(componentInChildren.gameObject, fireGuid);
+                }
+
+                subFire.ReflectionSet("roomFires", roomFiresDict);
+                subFire.ReflectionSet("availableNodes", availableNodes);
+            }
+            else
+            {
+                // This is where non-Cyclops Fire creation logic should go
```
Sunrunner37: Maybe split out the cyclops stuff into its own method for readability.

NitroxClient/GameLogic/Fires.cs
```diff
@@ -0,0 +1,149 @@
...
+                    if (transform.childCount == 0)
+                    {
+                        availableNodes.Add(transform);
+                    }
+                }
+                roomFiresDict[room].fireValue++;
+                PrefabSpawn component = transform2.GetComponent<PrefabSpawn>();
+                if (component == null)
+                {
+                    return;
```
Sunrunner37: Should we log an error?

NitroxModel/DataStructures/GameLogic/FireData.cs
```diff
@@ -0,0 +1,69 @@
+﻿using System;
+using NitroxModel.DataStructures.Util;
+using ProtoBuf;
+
+namespace NitroxModel.DataStructures.GameLogic
+{
+    [Serializable]
+    [ProtoContract]
+    public class FireData
```
Sunrunner37: Maybe a `CyclopsFireData` class instead of having the cyclops members optional?

NitroxModel/DataStructures/GameLogic/FireData.cs
```diff
@@ -0,0 +1,69 @@
...
+            get { return (CyclopsGuid.IsPresent()) ? CyclopsGuid.Get() : null; }
+            set { CyclopsGuid = Optional<string>.OfNullable(value); }
+        }
+        [ProtoIgnore]
+        public Optional<string> CyclopsGuid { get; set; }
+
+
+        [ProtoMember(3)]
+        public CyclopsRooms Room { get; set; }
+        /*public CyclopsRooms? SerializableRoom
```
Sunrunner37: Lots of dead code in here.

NitroxClient/Communication/Packets/Processors/FireDousedProcessor.cs
```diff
@@ -0,0 +1,42 @@
...
+            this.packetSender = packetSender;
+        }
+
+        /// <summary>
+        /// Finds and executes <see cref="Fire.Douse(float)"/>. If the fire is extinguished, it will pass a large float to trigger the private
+        /// <see cref="Fire.Extinguish()"/> method.
+        /// </summary>
+        public override void Process(FireDoused packet)
+        {
+            Optional<GameObject> fireGameObject = GuidHelper.GetObjectFrom(packet.Guid);
```
Sunrunner37: Let's make a `RequireObjectFrom(guid)` for these cases.

Sunrunner37:
It also says that you updated `Assets/chatlog.unity` as part of this review (at the very end).  Can you remove this?  I think it was a merge conflict or something.

NitroxModel/DataStructures/GameLogic/FireData.cs
RabidCrab: I can create CyclopsFireData, and have FireData as its base.

NitroxModel/DataStructures/GameLogic/FireData.cs
RabidCrab: Legacy from testing nullable and non-nullable data. Adding a CyclopsFireData that extends FireData would resolve the issues I've had with passing Protobuf nulls.

NitroxClient/GameLogic/Fires.cs
RabidCrab: Added an error to log if it didn't get a PrefabSpawn

NitroxModel/DataStructures/GameLogic/FireData.cs
RabidCrab: I changed FireData to CyclopsFireData. I looked all throughout the code and couldn't find a single instance of a fire being spawned outside of the Cyclops.

If they so happen to add fires in the future, or we want to, it would be easy to add a Fires.CreateFire to handle it. I narrowed the scope of the current Fires.CreateFire to make it specific to Cyclops. I should probably rename it to Fires.CreateCyclopsFire.

RabidCrab:
I think I got rid of the Assets/chatlog.unity, but I'm not entirely sure. I wasn't able to find any changes I made to the file beyond merging.

NitroxClient/GameLogic/Fires.cs
RabidCrab: I removed all of the non-Cyclops checks. All of the fire creation code is specific to the Cyclops.

NitroxPower:
RabidCrab also is it possible to make so we can build inside the Cyclops currently iwhat ever we build inside it when we move with the cyclops it moves fine and its perfect the only issue is that when you rejoin what ever you building in cyclops return to the original place where they first got builded and flying over there

PR #152, opened 2018-02-10
Disabled pausing in multiplayer

Duckuk:


NitroxPatcher/Patches/IngameMenu_OnSelect_Patch.cs
```diff
@@ -0,0 +1,23 @@
...
+namespace NitroxPatcher.Patches
+{
+    public class IngameMenu_OnSelect_Patch : NitroxPatch
+    {
+        public static readonly Type TARGET_CLASS = typeof(IngameMenu);
+        public static readonly MethodInfo TARGET_METHOD = TARGET_CLASS.GetMethod("OnSelect", BindingFlags.Public | BindingFlags.Instance);
+
+        public static void Postfix(IngameMenu __instance)
+        {
+            FreezeTime.End("IngameMenu");
```
Measurity: This change will also disable pausing in singleplayer. Can you add a check if multiplayer is running?

MarijnS95: This patch is not `Persistent`, meaning it'll be automatically injected when MP is started, and reverted when leaving a session.

MarijnS95:
@MadMaxOfYore The code will correctly handle "unknown" `userId`s passed to `.End`. 

```cs
public static void End(string userId)
{
	if (FreezeTime.freezers.Contains(userId))
	{
		FreezeTime.freezers.Remove(userId);
		if (FreezeTime.freezers.Count == 0)
		{
			Time.timeScale = FreezeTime.oldTimeScale;
			FreezeTime.PauseSound(false);
		}
	}
}
```

In any case, what is your view on rapidly calling `End` after a `Begin` as opposed to eliminating the call to `Begin` when multiplayer is loaded? I liked the patch in the `Begin` method itself (https://github.com/SubnauticaNitrox/Nitrox/pull/152/commits/cd840036033f06ec736a9b6d1900d7deb58c0294), but then we have to check all 5 usages of this method; most likely disable the `F8` feedback menu (we definitely don't want people spamming the devs with issues caused by our mod).

MadMaxOfYore:
I would also kill the pause completely rather than start/stop. I think even a momentary freeze in a multiplayer game has the potential to create some instability for the ecosystem of a long-running multiplayer session. 

I am not sure why we would need to check for 5 different usages, but I also don't have a ton of context on this particular problem.

Killing the F8 menu feels like it is out of scope for this story, but I definitely agree that we need a card for that. I'd hate to get on the bad side of UWE's support group o_O

MarijnS95:
Pardon the wording, I should've elaborated more.

Therea are currently five usages of `FreezeTime`. Patching `FreezeTime.Begin` means all usecases are affected, requiring us to check each of them for potential isses. One of them is the `F8` menu. My gut feeling tells me not pausing the game while generating/using statistics causes even more trouble than issues reported with just the mod loaded.

MadMaxOfYore:
Gotcha. I still feel like we should try and kill the feedback menu entirely. Anything that happens while the mod is loaded is definitely our problem and not UWE's. I don't trust our users to be discerning with were they vent any frustrations they have in the heat of the moment either. 

If they can make a boneheaded decision, then they will -_- (no offense intended, of course =P )

I am indifferent about whether or not we should omit the feedback menu from this patch. It needs to be addressed, but I think killing it altogether is the desired end-game, but that is beyond the scope of this PR.

I don't know enough about our patches to really offer any other insight on them this at this time. As the code stands, I am fine with this PR landing onto master. Pausing the game already screws things up -
 messing things up by preventing a pause doesn't really feel like a huge risk to me.

So long as @Duckuk has done some due diligence in testing out the different pause scenarios, of course.

PR #156, opened 2018-02-10
Nitroxconsole

Measurity:
## The main changes:
- Refactored Multiplayer init code into the Multiplayer.OnEnable/OnDisable methods.
- Added a NitroxConsole.
- Fixed that mplayer command didn't work.

## NitroxConsole
Press \` key (just like with DevConsole) to open the NitroxConsole.

### Command syntax
The NitroxConsole uses a git style command system where parameters can either be just names like `-f` or have values like `-name this_is_my_name`. A string can have spaces as long as it is surrounded by the `'` character like `find -player 'That strange player's name'`. 

Parameter values are matched as either `string` or `number` allowing for ints or floats.

mplayer command became: `mplayer -u NAMEHERE -i 127.0.0.1` (-i is optional)
`disconnect` is still `disconnect`.
`warpto` became `warpto -o ID`. -o for -other.

### The command categories
NitroxConsole defaults to 'nitrox' category of commands. To run subnautica commands enter it like so: `subnautica fps`. Giving no category (like `fps`) will run the command as `nitrox fps`. I plan to later let users change the category from the console UI as a dropdown.

### Adding commands
To add new commands define a method similar to the following. Note the attributes and the required parameter:
image

**A command doesn't get registered automatically** and needs to be added with the `AddCommand` function. I opted for this because I want to later on disable commands on the fly based the game's state:
`NitroxConsole.Main.AddCommand(OnConsoleCommand);`

### How does it work?
After something is entered into the console: NitroxConsoleUI -> NitroxConsole.Submit() -> new CommandCandidateEventArgs(input) -> new CommandEventArgs(candidate) -> CommandReceived applicable handler invoked.

## Final thoughts
Even though this change is relatively transparent I hope the console will get used more often now that it is more powerful.

Future plans for improvement: adding auto-complete/command search.

### DIFF
--- a/ClientTester/Commands/DefaultCommands/AniCommand.cs
+++ b/ClientTester/Commands/DefaultCommands/AniCommand.cs
@@ -13,7 +13,7 @@ namespace ClientTester.Commands.DefaultCommands
 
         public override void Execute(MultiplayerClient client, string[] args)
         {
-            assertMinimumArgs(args, 2);
+            AssertMinimumArgs(args, 2);
 
             AnimChangeState state;
 
--- a/ClientTester/Commands/DefaultCommands/ChatCommand.cs
+++ b/ClientTester/Commands/DefaultCommands/ChatCommand.cs
@@ -13,7 +13,7 @@ namespace ClientTester.Commands.DefaultCommands
 
         public override void Execute(MultiplayerClient client, string[] args)
         {
-            assertMinimumArgs(args, 1);
+            AssertMinimumArgs(args, 1);
 
             if (args.Length >= 2)
             {
--- a/ClientTester/Commands/DefaultCommands/DropCommand.cs
+++ b/ClientTester/Commands/DefaultCommands/DropCommand.cs
@@ -13,7 +13,7 @@ namespace ClientTester.Commands.DefaultCommands
 
         public override void Execute(MultiplayerClient client, string[] args)
         {
-            assertMinimumArgs(args, 4);
+            AssertMinimumArgs(args, 4);
 
             client.Logic.Item.Dropped(new GameObject(), UWE.Utils.ParseEnum<TechType>(args[0]), CommandManager.GetVectorFromArgs(args, 1));
         }
--- a/ClientTester/Commands/DefaultCommands/MoveCommand.cs
+++ b/ClientTester/Commands/DefaultCommands/MoveCommand.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Timers;
+using NitroxModel.DataStructures;
 using NitroxModel.DataStructures.ServerModel;
 using NitroxModel.DataStructures.Util;
 using UnityEngine;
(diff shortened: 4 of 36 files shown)


Measurity:
I'll put this one on hold until after #159 has been approved and merged so I can incorporate the changes in this PR.

Sunrunner37:
Hello Measurity,

First I want to thank you again for contributing - I think it is good that we have some talented engineers like yourself work on the project now.  In regards to this PR, I think this may be too far into feature creep for what we would like to do with the mod.  Here are a few observations that I would like you to consider:

1) UWE has already implemented a fully working console system that, although sometimes difficult to use, has fully fulfilled our use-cases.

2) It would be extremely confusing to players to have another console in the game.  It's even confusing for new developers who are not familiar with our system.

3) We have very few console commands - certainly not enough to warrant an entire custom system.  This is a large amount of complexity that we will need to maintain for very little (if any) benefit.

I understand that working on features is not glamorous (I've done tons of that leg work and hate it).  But we should focus on things that get us closer to an alpha release.  We can table this PR and revisit if we have a strong need in the future.

Measurity:
Hi Sunrunner37,

Thank you for taking a look at my PR.

You've given a few counter arguments but they don't give the whole picture of what this PR entails.

Short and sweet, the improvements that this PR brings are:

1. Full control of how the console works without having to understand what the SN console can and can't do.
2. Future SN patches have an almost zero chance to break this code because it uses the Unity API.
3. Provides a robust system for all tools that need or want a command-line: NitroxClient, ClientTester and likely NitroxServer in the future to kick/ban/configure for example.
4. The NitroxConsoleUI provides an in-game log. A future improvement can make all InGame logs go to this console as well.
5. It provides a way to run predefined code in-game to test features without having to recompile. Same as SN console but with full and easy control over the execution.

With this in mind I'll respond directly to your feedback:

1. The current console in Subnautica is not fully working the way we want. There is an error in the current master branch that makes the `mplayer` command broken. The SN console unloads commands from it's "notify list" when the game loads, making its usage non-straightforward. Refitting the DevConsole to our needs is more involving (injecting patches or wrapper class) than having our own system.
2. The SN DevConsole will be disabled with this PR as it is when you bought the game. It wouldn't be more or less confusing because the new console is a full replacement of the SN console and supports the same workflow and more. All the code is our code, fully explorable without a decompiler. This should make it _less_ confusing to understand, use and extend.
3. When looking at the game commands now we have only 3. However when you include the client tester there are 12. Admittedly only Nes seems to use the ClientTester and it needs a new face-lift. That does not mean that there isn't a need for a centralized command-line. In future we are likely to need it for the server as well. The backend of this command system is fully reusable (everything excluding the NitroxConsoleUI class, aka the view.). It can replace the current client tester implementation of the command system, making the code base less complex and more centralized.

This solution might seem like feature-creep because it contains many lines of code/classes for 'one feature' (I don't know why you call it feature-creep). Though there is only 1 feature that is not directly used by the current 3 commands: command descriptions. Which is just one property in an event class. All other features are used by the 3 commands alone. The NitroxConsoleUI isn't more complex than the DevConsole UI apart from the log and design.

I am willing to further improve and change this PR. If you want it to be less complex or different in any way I'll gladly accept feedback. Closing this PR (and moving away from a centralized console) is just wasting an improvement to the code-base.

Issue #119, opened 2018-02-04
[UI] Remote Player Colors


Sunrunner37:
- [ ] Equipment coloring 
- [x] Chat coloring 
- [x] Ping Coloring
- [x] Color picker in settings
- [x] Persist picked color
- [x] Begin with a randomly selected color

MadMaxOfYore:
I'm working on this - but can't claim at this time.

Jannify:
#56 Covers all exclusive equipment and random color.
Have to talk to @MadMaxOfYore for cleaning and implementing...

PR #159, opened 2018-02-12
F/multiplayer settings

MadMaxOfYore:
First - thanks mightily to @MarijnS95 for coming to my rescue at the end of this to help drive it home. He jumped in and fixed my Git tree after I tried to rebase a long running branch with a ton a changes at 2:30 AM. He hopped in and fixed things up very nicely - and also added the final touches to the post-build process for the new Autofac dependency. It is much appreciated!

As to the meat of this feature:

From a visibility perspective, the biggest change here is that players will now spawn into the world with a randomly assigned color. In this current iteration, the coloring effect is only applied to chat names, ping icons and associated player names, as well as the initial dive suit that the player spawns into the world with. 

Under the hood, I've completely refactored the way that we manage dependencies in the client project. The reasoning can effectively be summed up as slowing down to speed up. I felt like the way we were kind of "stitching" dependency management together in the client was making it hard to move quickly because even little changes implied a large regression footprint. These changes allow us to isolate dependencies into a single container providing us with consistent dependency access, dependency lifetime configuration, and centralized management. All of this should help us move faster because we don't need to worry about re-configuring dependency construction because it is already handled for us. Additionally, it increases our ability to unit test objects which take a global dependency by allowing us to install mocks into the new service locator.

Aside from that, I also provided a more sophisticated refactor of the ClientBridge - now known as the MultiplayerSessionManager. The new model should allow for easy pivots to approach and configuration as we progress into deep feature development. Provides a state change event so we aren't just polling for a specific state change to occur. This all became necessary due to the need to be able to support a more complex comm channel as we move towards an initial release. 

Finally, the color change feature brings with it a framework for handling remote player appearance in the game. Currently, the PlayerModelDirector and associated build context only support attaching a ping instance and applying a color filter to the starter dive suit. We need help from other devs, specifically in determining the best color filter approach required for each piece of equipment. This framework allows us to focus time solely to that effort rather than worry about "how" we will get that filter applied to the player. 

#119


NitroxClient/Communication/MultiplayerSession/MultiplayerSessionManager.cs
```diff
@@ -0,0 +1,110 @@
...
+        {
+            this.client = client;
+            CurrentState = initialState;
+        } 
+
+        public event MultiplayerSessionConnectionStateChangedEventHandler ConnectionStateChanged;
+        public IMultiplayerSessionConnectionState CurrentState { get; private set; }
+
+        //Eww...
+        public string PlayerId { get; private set; }
```
MadMaxOfYore: This actually needs to be deprecated. I split player ID and player Name. The player ID is now provided on the session reservation object.

NitroxServer/GameLogic/Spawning/EntitySpawner.cs
```diff
@@ -127,7 +127,10 @@ private bool IsValidSpawnType(string id, bool creatureSpawn)
                 if (creatureSpawn && worldEntityInfo.slotType == EntitySlot.Type.Creature)
                 {
                     return true;
-                } else if (!creatureSpawn && worldEntityInfo.slotType != EntitySlot.Type.Creature)
```
MadMaxOfYore: Merge gone bad...

ClientTester/MultiplayerClient.cs
```diff
@@ -1,83 +1,93 @@
...
-                if (iterations >= 20)
-                {
-                    break;
-                }
-            } while (ClientBridge.CurrentState == ClientBridgeState.WaitingForRerservation);
+        private Dictionary<Type, PacketProcessor> GeneratePacketProcessorMap()
+        {
+            Dictionary<Type, PacketProcessor> packetProcessorMap = new Dictionary<Type, PacketProcessor>
+            {
+                { typeof(MultiplayerSessionPolicy), new MultiplayerSessionPolicyProcessor(multiplayerSession) },
```
MarijnS95: Can we infer the packet type here? It's only two items now, but this procedure is error prone nevertheless. Especially if someone changes the type of packet handled by the processor later on (very unlikely though).

```cs
Dictionary<Type, PacketProcessor> packetProcessorMap = new List<PacketProcessor>
{
    new MultiplayerSessionPolicyProcessor(multiplayerSession),
    new MultiplayerSessionReservationProcessor(multiplayerSession)
}.ToDictionary(k => k.GetType().BaseType.GetGenericArguments()[0], v => v);
```

NitroxClient/Communication/DeferringPacketReceiver.cs
```diff
@@ -1,24 +1,27 @@
...
+    // TODO: Spinlocks don't seem to be necessary here, but I don't know for certain.
     public class DeferringPacketReceiver
     {
         private const int EXPIDITED_PACKET_PRIORITY = 999;
         private const int DEFAULT_PACKET_PRIORITY = 1;
         private const int DESIRED_CELL_MIN_LOD_FOR_ACTIONS = 1;
 
         private readonly Dictionary<AbsoluteEntityCell, Queue<Packet>> deferredPacketsByAbsoluteCell = new Dictionary<AbsoluteEntityCell, Queue<Packet>>();
-        private readonly NitroxModel.DataStructures.PriorityQueue<Packet> receivedPackets = new NitroxModel.DataStructures.PriorityQueue<Packet>();
+        private NitroxModel.DataStructures.PriorityQueue<Packet> receivedPackets;
```
MarijnS95: Has the removal of `readonly` (and initialisation moved to the constructor) only been done to reinitialize the PQ (rather than defining a `.Clear` method on it?).

Leave it for now - the PQ is something on our list anyway.

NitroxClient/Communication/DeferringPacketReceiver.cs
```diff
@@ -109,14 +112,26 @@ public void CellLoaded(VisibleCell visibleCell)
...
                     {
-                        Log.Debug("Found deferred packet... adding it back with high priority.");
                         Packet packet = deferredPackets.Dequeue();
                         receivedPackets.Enqueue(EXPIDITED_PACKET_PRIORITY, packet);
                     }
                 }
             }
         }
+
+        public void Flush()
```
MarijnS95: Isn't `Flush` supposed to send all outstanding packets to the server, and deal with incoming packets immediately? How about some different terminology (such as `Clear`), or make this function do what people assume `Flush` means?

NitroxClient/Communication/MultiplayerSession/ConnectionState/AwaitingReservationCredentials.cs
```diff
@@ -0,0 +1,85 @@
...
+        {
+            if (!sessionConnectionContext.Client.IsConnected)
+            {
+                throw new InvalidOperationException("The client is not connected.");
+            }
+        }
+
+        private static void PlayerSettingsIsNotNull(IMultiplayerSessionConnectionContext sessionConnectionContext)
+        {
+            try
```
MarijnS95: This is an interesting way of using `Validate.NotNull`...

NitroxClient/GameLogic/PlayerModelBuilder/ColorValueRange.cs
```diff
@@ -0,0 +1,19 @@
+﻿namespace NitroxClient.GameLogic.PlayerModelBuilder
+{
+    public class ColorValueRange
```
MarijnS95: Just `ValueRange`?

NitroxClient/GameLogic/PlayerModelBuilder/HsvColorFilter.cs
```diff
@@ -0,0 +1,53 @@
...
+using UnityEngine;
+
+namespace NitroxClient.GameLogic.PlayerModelBuilder
+{
+    public class HsvColorFilter
+    {
+        private float targetHue;
+        private float targetSaturation;
+        private float targetVibrance;
+        public List<ColorValueRange> HueValueRanges => new List<ColorValueRange>();
```
MarijnS95: These getters return a **new** List every time they are evaluated/called - I don't think that's intended?

NitroxClient/MonoBehaviours/Multiplayer.cs
```diff
@@ -85,112 +48,57 @@ public void ProcessPackets()
...
-                    try
-                    {
-                        PacketProcessor processor = PacketProcessorsByType[packet.GetType()];
-                        processor.ProcessPacket(packet, null);
-                    }
-                    catch (Exception ex)
-                    {
-                        Log.Error("Error processing packet: " + packet, ex);
-                    }
+                    Type clientPacketProcessorType = typeof(ClientPacketProcessor<>);
```
MarijnS95: Don't we want to preprocess this once like in `MultiplayerClient.GeneratePacketProcessorMap`? I have not profiled it but I'd not be surprised if this is slow.

NitroxClient/MonoBehaviours/PowerMonitor.cs
```diff
@@ -54,8 +57,13 @@ public void ChargeChanged(float amount, GameObject gameObject)
...
         }
 
+        public void Awake()
+        {
+            powerBoardcaster = NitroxServiceLocator.LocateService<Power>();
+        }
+
         /**
-         * We allow small bursts of power to build up so we don't flood with packets.
+         * We allow small bursts of power to Build up so we don't flood with packets.
```
MarijnS95: ?

NitroxClient/MonoBehaviours/NitroxBootstrapper.cs
```diff
@@ -12,6 +12,11 @@ public void Awake()
 
             gameObject.AddComponent<MainMenuMods>();
             gameObject.AddComponent<CodePatchManager>();
+
+#if DEBUG
+            gameObject.AddComponent<ConsoleJoinServer>();
```
MarijnS95: The console resets registered commands when the game loads; as such the commands registered in `ConsoleJoinServer` are only available in the menu (where they shouldn't be), not in the game itself (where they should be).

ClientTester/MultiplayerClient.cs
MadMaxOfYore: Fair enough. I know killing the client tester has been discussed so did not want to spend a lot of time here, but a quick fix to make it less brittle is cool with me.

NitroxClient/Communication/DeferringPacketReceiver.cs
MadMaxOfYore: Yeah - we disconnect from the service now. The "flush" may need to do more, but presently it just needs to allow the player to join a new session without issue. The fact that they still can't reconnect without closing the client is another matter -_-

NitroxClient/Communication/MultiplayerSession/ConnectionState/AwaitingReservationCredentials.cs
MadMaxOfYore: I was torn - but feel that an InvalidOperationException was more appropriate given the context. Figured this was the cleanest way to represent the different concerns.

NitroxClient/GameLogic/PlayerModelBuilder/HsvColorFilter.cs
MadMaxOfYore: Yeah - dumb last second move. I already fixed this on a different branch and planned to land the fix with that branch.

NitroxClient/MonoBehaviours/PowerMonitor.cs
MadMaxOfYore: Caught in a refactor most likely.

NitroxClient/MonoBehaviours/Multiplayer.cs
MadMaxOfYore: Not slow as far as I can tell. I've done this exact same thing when routing messages received on a REST endpoint into the appropriate DDD service.

NitroxClient/Communication/DeferringPacketReceiver.cs
MadMaxOfYore: I'd started with `Purge` but do not recall why I ended up back here. Maybe because I intended to flush at the point it is called? I am only clearing it out because we don't have reconnects ironed at all. I am indifferent about the name at this point personally.

ClientTester/MultiplayerClient.cs
MarijnS95: I've been going through the PR on GitHub - missed/misread the fact that this is `ClientTester`. Nice to fix for consistency, but if we kill it I'm happy as well.

NitroxClient/Communication/DeferringPacketReceiver.cs
MarijnS95: I've seen the discussion on Discord regarding removing outstanding packets in the queue and deferred structure. We can leave this for now, and add the desired functionality later on.

NitroxClient/Communication/DeferringPacketReceiver.cs
MarijnS95: I've seen the discussion on Discord regarding removing outstanding packets in the queue and deferred structure. We shall be adding reconnect functionality soon enough, so this function can stay the way it is for now.

NitroxClient/Communication/MultiplayerSession/ConnectionState/AwaitingReservationCredentials.cs
MarijnS95: Well, that way you get parameter resolving as well. Looks okay to me, though defining a `NotNull` override throwing an exception of your choosing seems nicer to me; especially since these similar `try-catch` blocks are used in many places.

NitroxClient/MonoBehaviours/Multiplayer.cs
MarijnS95: As discussed yesterday, it's probably fine, though I do like to profile the difference between direct lookup from a map and reconstructing the generic type for every packet. Every CPU cycle we can spare is one; the game itself is heavy as is already.

NitroxClient/GameLogic/PlayerModelBuilder/PlayerModelBuildContext.cs
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        public PlayerModelBuildContext WithRebreather()
+        {
+            PlayerRebreatherModelBuilder modelBuilder = new PlayerRebreatherModelBuilder(modelGeometry);
+            LinkBuilder(modelBuilder);
+            return this;
+        }
+
+        private void LinkBuilder(BasePlayerModelBuildHandler buildHandler)
```
MarijnS95: I assume all (future) `With` methods are supposed to be safe from any ordering? This is prepending `BuildHandler`s backwards.

NitroxClient/GameLogic/PlayerModelBuilder/PlayerModelDirector.cs
```diff
@@ -0,0 +1,35 @@
+﻿using NitroxClient.GameLogic.PlayerModelBuilder.Abstract;
+using UnityEngine;
+
+namespace NitroxClient.GameLogic.PlayerModelBuilder
+{
+    // The vision here is to use this director as our central tool for managing remote model appearance alterations like changing equipment or in-game changes to appearance.
+    // This could probably very easily be extended to work with the local Player later down the line.
+    public class PlayerModelDirector
```
MarijnS95: Is this extra level of indirection necessary? I don't see any added value in `StagePlayer()` - this could be put in the constructor, and then (similar to Android semantics):

```cs
new PlayerModel.Builder(player)
    .WithPing()
    .WithRegularDiveSuit()
    .Build();
```
Or even more advanced:
```cs
RemotePlayer player = new RemotePlayer.Builder(playerId, playerName, playerSettings)
    .WithPing()
    .WithRegularDiveSuit()
    .Build();
```

NitroxModel/Packets/Exceptions/UncorrelatedPacketException.cs
```diff
@@ -0,0 +1,36 @@
...
+            ExpectedCorrelationId = expectedCorrelationId;
+        }
+
+        public UncorrelatedPacketException(string message, Exception innerException, CorrelatedPacket invalidPacket, string expectedCorrelationId) : base(message, innerException)
+        {
+            InvalidPacket = invalidPacket;
+            ExpectedCorrelationId = expectedCorrelationId;
+        }
+
+        protected UncorrelatedPacketException([NotNull] SerializationInfo info, StreamingContext context, CorrelatedPacket invalidPacket, string expectedCorrelationId) : base(info, context)
```
MarijnS95: We are not using annotations elsewhere, do we want to use them?

Interesting, these are bundled with the game.

MadMaxOfYore: That was auto-generated by the IDE.

NitroxClient/GameLogic/PlayerModelBuilder/PlayerModelBuildContext.cs
MadMaxOfYore: Yes - ordering should be irrelevant by implementation.

NitroxClient/GameLogic/PlayerModelBuilder/PlayerModelDirector.cs
MadMaxOfYore: The Director is an intrinsic artifact of the Builder pattern.

NitroxClient/MonoBehaviours/Multiplayer.cs
MadMaxOfYore: Been doing some research. The use of `typeof` incurs far less of a performance cost at run-time than `GetType` because it is a static type lookup that is resolved once at compile-time.

Calling `MakeGenericType` does take longer than grabbing the resolved type from a dictionary, but it only takes about 10 - 20 ticks in total from what I can gather from research. It feels like excessive optimization to be overly concerned about it.

At any rate - simply caching the generic type reference after it is initially resolved would be the ideal solution in my mind. Scanning the entire assembly can get expensive and I'd hate for the mod to put too much of a burden on the initial load than is necessary.

Seems like something that can be easily addressed if/when it becomes a problem.

NitroxClient/GameLogic/PlayerModelBuilder/ColorValueRange.cs
MadMaxOfYore: Hrmm, I only *intended* for this to be a value object that is used only for the color filter - hence the name.

NitroxClient/MonoBehaviours/NitroxBootstrapper.cs
MadMaxOfYore: Interesting. I assumed that this persisted into the game itself. No matter - just need to create a new game object and set it to not be destroyed on load. It will simply be alive and well for the life of the client - but will stop/start multiplayer when it is appropriate to do so.

Objections?

PR #171, opened 2018-02-24
Entity cell math fixes

MarijnS95:
Depends on #170.

Adds `Level` property to `AbsoluteEntityCell` and deprecates `VisibleCell`. This parameter is required to calculate the cell, as level 0 has `10x10x10` cells per batch whereas all other levels (`1-3`) have `5x5x5` cells per batch.

Default levels have been assigned to certain packets, these may need to be revisited.


NitroxClient/Communication/DeferringPacketReceiver.cs
```diff
@@ -49,42 +49,24 @@ public Queue<Packet> GetReceivedPackets()
...
+                Log.Debug($"Action {packet} was deferred, cell not loaded (with required lod): {playerAction.AbsoluteEntityCell}");
+                AddPacketToDeferredMap(playerAction, playerAction.AbsoluteEntityCell);
+                return true;
             }
 
             return false;
         }
 
-        private void AddPacketToDeferredMap(PlayerActionPacket playerAction, AbsoluteEntityCell cell)
+        private void AddPacketToDeferredMap(RangedPacket playerAction, AbsoluteEntityCell cell)
```
Sunrunner37: minor: change the variable naming

NitroxClient/Communication/DeferringPacketReceiver.cs
```diff
@@ -97,21 +79,20 @@ private void AddPacketToDeferredMap(PlayerActionPacket playerAction, AbsoluteEnt
             }
         }
 
-        public void CellLoaded(VisibleCell visibleCell)
+        public void CellLoaded(AbsoluteEntityCell visibleCell)
```
Sunrunner37: same here

NitroxClient/GameLogic/Terrain.cs
```diff
@@ -33,31 +33,33 @@ public void CellLoaded(Int3 batchId, Int3 cellId, int level)
...
         }
-        
+
         private IEnumerator WaitAndAddCell(Int3 batchId, Int3 cellId, int level)
         {
             yield return new WaitForSeconds(0.5f);
 
-            VisibleCell cell = new VisibleCell(batchId, cellId, level);
+            AbsoluteEntityCell cell = new AbsoluteEntityCell(batchId, cellId, level);
+            Log.Debug("Cell {0} loaded", cell);
```
Sunrunner37: Should we leave these in?  They might be a little noisy.

NitroxServer/UnityStubs/Vector3.cs
```diff
@@ -5,6 +5,7 @@ namespace NitroxServer.UnityStubs
     [ProtoContract]
     public class Vector3
     {
+#pragma warning disable IDE1006 // Naming Styles
```
Sunrunner37: I wish IDE supported built in ignore list mastered in a new file that didn't muddy up the code with pragma warning disables.

NitroxClient/Communication/DeferringPacketReceiver.cs
MarijnS95: I wasn't entirely sure about the `Ranged` naming in the first place, there should be a better synonym. Completely forgot to add that question to the PR, so if you have a better name please let me know; I'll add it to #172.

NitroxServer/UnityStubs/Vector3.cs
MarijnS95: Interesting, I cannot remember adding these manually on purpose. I can alter the `.editorconfig` to stop complaining about naming, because that's where this is coming from.

NitroxClient/GameLogic/Terrain.cs
MarijnS95: This one should definitely be removed, as it is extremely noisy.

NitroxClient/Communication/DeferringPacketReceiver.cs
Sunrunner37: Yeah ranged doesn't really tell you its use - unless we put a boat load of class documentation.  We could say DeferrablePacket.

MarijnS95: That sounds like a proper alternative.

PR #172, opened 2018-02-24
Batch lazy loading

MarijnS95:
Depends on #171 

Lazyloads batches on the server, depending on the locations visited by clients. The original system took between 40 seconds to 2 minutes at full CPU load, on a relatively powerful quadcore machine.

Please review the locking carefully - it has been written to be more granular and allow more concurrency, though this may not be optimal in all cases.

### DIFF
--- a/NitroxServer/GameLogic/Spawning/EntitySpawner.cs
+++ b/NitroxServer/GameLogic/Spawning/EntitySpawner.cs
@@ -15,9 +15,6 @@ namespace NitroxServer.GameLogic.Spawning
 {
     public class EntitySpawner
     {
-        private readonly Dictionary<AbsoluteEntityCell, List<Entity>> entitiesByAbsoluteCell = new Dictionary<AbsoluteEntityCell, List<Entity>>();
-        private readonly Dictionary<string, Entity> entitiesByGuid = new Dictionary<string, Entity>();
-
         private readonly HashSet<Int3> parsedBatches = new HashSet<Int3>();
         private readonly Dictionary<string, WorldEntityInfo> worldEntitiesByClassId;
         private readonly LootDistributionData lootDistributionData;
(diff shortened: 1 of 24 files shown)


Sunrunner37:
Can you merge so the diff can be updated?

MarijnS95:
@Sunrunner37 Done. Please review the comments in #171 so I can add the fixes to this PR.

NitroxServer/GameLogic/Spawning/EntitySpawner.cs
```diff
@@ -1,118 +1,135 @@
...
+                        selectedPrefab = prefab;
+                        break;
                     }
                 }
+            }
 
-                if (!ReferenceEquals(selectedPrefab, null) && worldEntitiesByClassId.ContainsKey(selectedPrefab.classId))
-                {
-                    WorldEntityInfo worldEntityInfo = worldEntitiesByClassId[selectedPrefab.classId];
+            if (!ReferenceEquals(selectedPrefab, null) && worldEntitiesByClassId.ContainsKey(selectedPrefab.classId))
```
MadMaxOfYore: I know it sounds trivial, but we should stick with TryGetValue here given the number of spawn points we are iterating over. It is twice as fast as ContainsKey.

NitroxServer/GameLogic/Spawning/EntitySpawner.cs
```diff
@@ -1,118 +1,135 @@
...
+            Log.Debug("Batch {0} not parsed yet; parsing...", batchId);
+
+            List<EntitySpawnPoint> batchSpawnPoints = batchCellsParser.ParseBatchData(batchId);
+
+            parsedBatches.Add(batchId);
+
+            Dictionary<AbsoluteEntityCell, List<Entity>> newSpawnPoints = new Dictionary<AbsoluteEntityCell, List<Entity>>();
 
-            foreach (EntitySpawnPoint entitySpawnPoint in entitySpawnPoints)
+            Parallel.ForEach(batchSpawnPoints, spawnPoint =>
```
MadMaxOfYore: This method of multi threading has a lot of overhead associated with it in .Net. Have you bench-marked this to see if there is actually a performance benefit to concurrency/parallelism here? If we're locking on the same resources under the hood when we load game objects, then it may be faster to just serialize this process.

MarijnS95: You are right - I've blindly copied it over from the previous version without any consensus. A little profiling session indicates the deserialisation process (in `ParseBatchData`) takes the most CPU *by far*. The overhead of `Parallel.ForAll` and the `lock` in it fluctuates from 0 to 1 second on a non-mt test that takes 9 seconds (the test loads all batches in  `[8, 14)` for `x`, `y` and `z`). 

Currently batches are loaded in serial, when they are "touched" by some other code (as part of the lazy-loading process). To take optimal use of multithreaded processors, we may load all batches around it in parallel, high chance they'll be required soon anyway.
I don't think it's that relevant though (more code, more maintenance); when traversing the world very few new batches are needed every second, it seems fast enough to keep up for now.

NitroxServer/GameLogic/Spawning/EntitySpawner.cs
```diff
@@ -6,123 +6,132 @@
...
 
-                double randomNumber = random.NextDouble();
-                double rollingProbability = 0;
+                entities.AddRange(SpawnEntities(esp));
+            }
+
+            return Optional<IDictionary<AbsoluteEntityCell, List<Entity>>>.Of(entitiesByAbsoluteCell);
+        }
+
+        private IEnumerable<Entity> SpawnEntities(EntitySpawnPoint entitySpawnPoint)
```
Sunrunner37: minor: For my understanding (As I am not a c# programmer), why do we want to use IEnumerable with yield operator over returning a collection?  From my understanding, yield operator implies to the reader that we want the code to have deferred execution.  However, in this usage we are simply doing AddRange which takes all of the elements anyways.

MarijnS95: It's not something C# specific, and `yield` has many more usecases than deferred execution. I'm not fond of creating and returning lists because it's too verbose, and wasting time/memory because it's only used for a very short amount of time; once the function returns the items are copied to the other collection after which the list will be GC'd. An enumerable solves that nicely, as the items are evaluated and and directly stored in the right collection. Yield is a perfect, concise way to get such an enumerable. And what would you do in case of the `yield break` at the beginning of the function? Return `null`? Return an empty list?

In any case, I wasn't expecting a merge already, and still wanted to convert this to a oneliner:

```cs
batchCellsParser
    .ParseBatchData(batchId)
    ‎.GroupBy(esp => esp.AbsoluteEntityCell)
    ‎.ToDictionary(grp => grp.Key, grp.SelectMany(SpawnEntities).ToList());
```
(From the top of my head)

PR #177, opened 2018-02-28
F/multiplayer settings local player coloring

MadMaxOfYore:
Other than showing that I can submit a PR that doesn't make everyone's eyes bleed ;)

This PR introduces coloring for the local player so that they themselves are able to perceive the recoloring effect applied to their avatar. 

Why? Because who wants to constantly ask: "Dude, what's my color?"

image

I've also created an abstraction around the player objects in preparation for equipment sync - which needs to be completed before applying a color effect to all wearable equipment really makes sense.


MadMaxOfYore:
#119

NitroxClient/GameLogic/RemotePlayer.cs
```diff
@@ -123,7 +111,7 @@ public void SetPilotingChair(PilotingChair newPilotingChair)
                     mpCyclops.Exit();
                 }
 
-                RigidBody.isKinematic = AnimationController["cyclops_steering"] = (newPilotingChair != null);
+                RigidBody.isKinematic = AnimationController["cyclops_steering"] = newPilotingChair != null;
```
Sunrunner37: minor: can we leave parenthesis around statements as-is?  This statement is already hard to read.

PR #178, opened 2018-02-28
f/clean disconnect

MadMaxOfYore:
A small set of changes aimed at isolating the scope of our singleton dependencies to a specific multiplayer session using Autofac as the dependency tree container. We will now initialize a set of dependencies whenever a user selects a multiplayer session to join, and dispose of them entirely whenever the user leaves the session.

This free's us up in a lot of ways. One of which is that we are free to maintain state in some services where it makes sense to do so without needing to worry about how to clear that state out later on down the line. 

Additionally, we no longer need to refactor any existing code that currently maintains state inappropriately as we simply no longer care unless it negatively impacts the operation of a single multiplayer session. Since the client is not responsible for persisting any state from the session itself, we are free to dispose of the entire graph whenever we are done with it.

Finally - my hope is that this leads to a pretty clear path for cleanly rooting the client into the current session state as dictated by the server they are joining.


NitroxClient/MonoBehaviours/Multiplayer.cs
```diff
@@ -89,6 +90,11 @@ public void StopCurrentSession()
             remotePlayerManager.RemoveAllPlayers();
```
Sunrunner37: Do we still need to do these things (removeAllPlayers and flush)?  If not, can we remove those code paths?

MadMaxOfYore: Sure can.

MadMaxOfYore: Sure can.

NitroxModel/Core/NitroxServiceLocator.cs
```diff
@@ -7,35 +7,58 @@ namespace NitroxModel.Core
...
+            dependencyRegistrar.RegisterDependencies(builder);
+
+            //IgnoreStartableComponents - we don't want to cause "phantom" executions of the Start() method 
+            //on a Monobehaviour because someone accidentally did something funky with a DI registration.
+            DependencyContainer = builder.Build(ContainerBuildOptions.IgnoreStartableComponents);
+        }
+
+        public static void BeginNewLifetimeScope()
         {
             if (DependencyContainer == null)
```
Sunrunner37: minor: Validate.NotNull(obj, msg) here and other places.

PR #181, opened 2018-03-01
Post merge cleanups

MarijnS95:
- Deleted unreferenced files (I have a python script to check this periodically, could be set up on the CI if we ever want to check this).
- Shortened some code: It's more concise and readable IMO.
- Removed double `Dictionary` lookups (`dict.ContainsKey(key)` followed by `dict[key]`, which can be shortened into `dict.TryGetValue(key, out value)`).
- Fixed 13 failing tests (10 remaining).

Are we still aiming for C# 6.0? I'd like to use `dict.TryGetValue(key, out ValueType value)` from C# 7.0 ;)


NitroxServer/GameLogic/Spawning/EntitySpawner.cs
```diff
@@ -76,53 +70,40 @@ private IEnumerable<Entity> SpawnEntities(EntitySpawnPoint entitySpawnPoint)
...
-                float probabilityDensity = prefab.probability / entitySpawnPoint.Density;
-                rollingProbabilityDensity += probabilityDensity;
+                yield break;
             }
 
             double randomNumber = random.NextDouble();
-            double rollingProbability = 0;
-
-            if (rollingProbabilityDensity > 0)
+            if (rollingProbabilityDensity > 1f)
```
MarijnS95: @Sunrunner37 Shouldn't the random number always scale to the sum of probabilities?

NitroxModel/DataStructures/AbsoluteEntityCell.cs
```diff
@@ -1,69 +0,0 @@
-﻿using System;
-using NitroxModel.Helper;
-
-namespace NitroxModel.DataStructures
-{
-    [Serializable]
-    public class AbsoluteEntityCell
```
Sunrunner37: Did you mean to remove this?

NitroxServer/GameLogic/Spawning/EntitySpawner.cs
Sunrunner37: Dunno - i think the algorithm was copied from UWE.

NitroxModel/DataStructures/AbsoluteEntityCell.cs
MarijnS95: Yes, see the first commit. The file has been modified and moved; this original version was reintroduced by a pull request.

NitroxServer/GameLogic/Spawning/EntitySpawner.cs
MarijnS95: I'll check the source code. It doesn't seem correct, but the lower probability for all items might be intended.

MarijnS95: It's still working like that in the game code - if the sum of prefab probabilties isn't at least `1f` this allows a SpawnPoint to not spawn anything, which is a valid usecase.

NitroxModel/DataStructures/AbsoluteEntityCell.cs
MadMaxOfYore: My bad...

NitroxModel/MultiplayerSession/RandomColorGenerator.cs
```diff
@@ -4,9 +4,11 @@ namespace NitroxModel.MultiplayerSession
 {
     public static class RandomColorGenerator
     {
+        private static readonly System.Random random = new System.Random();
+
         public static Color GenerateColor()
         {
-            return new Color32((byte)Random.Range(0, 255), (byte)Random.Range(0, 255), (byte)Random.Range(0, 255), 255);
+            return new Color32(random.NextByte(), random.NextByte(), random.NextByte(), 255);
```
MadMaxOfYore: This needs to take the appropriate range into account. I think it is 0, 256

MarijnS95: @MadMaxOfYore That is incorrect. A byte consists of 8 bits, with which it can represent a total of `256` distinct values; from `0` (inclusive) to `255` (inclusive). `NextByte` is an extension method implemented as:
```cs
public static byte NextByte(this System.Random rng) => return Convert.ToByte(rng.Next(256));
```
Where `256` is the **exclusive** `maxValue`.

Any value above `255` would simply wrap around because the upper bytes are discarded.

MarijnS95:
@Sunrunner37 I'm not rebasing this on top of https://github.com/SubnauticaNitrox/Nitrox/commit/ea3756dd99ad47c1bea05af04667aa781a050a34 yet:

- The `Add` and `Remove` operations in `EntitySwitchedCells` are a no-op because they work on a **copy** of the original collection, as returned by `GetEntities`.
- I'd like to relax the collections in certain places:
  - In some cases a collection is only iterated: I have not profiled the overhead (probably negligible anyway), but if we aim for as much performance as possible, allocating, filling and iterating intermediate lists sounds like too much work IMO.
  - As [discussed earlier](https://github.com/SubnauticaNitrox/Nitrox/pull/172#discussion_r171455609), I'm not very fond of empty intermediate lists to indicate no elements had to be added, as is the case with `LoadUnspawnedEntities`. With an `Optional` we can prevent superfluous locking in `AddEntities`, it shows the intent properly, and does not require an extra allocated list.
- I still like to profile `LoadUnspawnedEntities` in conjunction with `AddEntities`; the implementation in this PR groups entities by `AbsoluteEntityCell` such that it only has to look up the list for that cell once per group of entities, whereas your implementation does that lookup once for every entity. I assume `GroupBy` has more overhead though: Dictionary lookups are `O(1)`, and `.Add` on a list is `O(1)` as well, for a total of `n` items, so `O(n)`. I cannot find information on `GroupBy` but assume it's higher than that.

NitroxModel/MultiplayerSession/RandomColorGenerator.cs
Measurity: [Random.NextBytes(byte[] buffer)](https://msdn.microsoft.com/en-us/library/system.random.nextbytes(v=vs.110).aspx) should also work so you don't waste data with `Convert.ToByte`.

MadMaxOfYore: .NET Random implements an exclusive upper bound - so a `Next` call would absolutely be 0, 256

However; I did not notice that the method was `NextBytes` and not `Next` - my mistake ;)

PR #184, opened 2018-03-02
Cleaned deployment step, fixed incompatibility.

### LINKED ISSUE #182: Compilation error
After the last update while compiling the code it returns the following error:
"Parameter "SkipNonexistentTargets" is not supported by the activity "MSBuild". Check that the parameter is present for the activity and that it is a public instance for the proprerty".
It's translated so it can be not exactly the same, but that's the meaning.
Can someone check/confirm the issue? It does not happens with the version just before this one.

MarijnS95:
Fixes #182.

- Now uses GetTargetPath which does not trigger a build to resolve target file.
- Deploys target library and dependencies directly from a built project, rather than relying on NitroxPatcher to do so after it finishes building.
- Removed SkipNonexistentTargets which is not available in older versions of Visual Studio.


MadMaxOfYore:
@Sunrunner37 Can you pull this down and see if it fixes your build issues?

PR #189, opened 2018-03-07
F/multiplayer settings persistence

MadMaxOfYore:
This PR implements persistent player preferences at the server level. 

Included is a revamped join menu that allows players to set a name and a color for a specific server:

(Appears after clicking on the appropriate server button)

image

Player preferences operate as follows:

- Persisted as JSON within Unity's player preferences framework.
- Stored by IP Address, so players can customize their in-game persona for each server they join.
- Settings persisted whenever player clicks "Join"
- If the player is re-joining a server, then the join menu will initialize with the player name and color that was last used to access that server.
- Join New Server Behavior:
        - Joining a multiplayer server for the first time ever: 
                   Player Name: unset
                   Color: Random
        - Re-Joining a multiplayer server:
                   Player Name: Last name used for server
                   Color: Last color used for server
        - Joining a new multiplayer server after having joined at least one prior:
                   Player Name: Last preference set*
                   Color: Last preference set*

Last preference set: The last "new" preference that the player saved. Ergo: Preferences are that-point-forward for new servers, but previously joined servers maintain their configurations without changing the "last set" preference. 

Special note: The ability to select your own color has more or less forced us to do more than a simple hue adjustment to the equipment textures. Given that we are allowing for some alterations to vibrancy of the pixels, we are experiencing a degree of fidelity loss along the boundaries of the different regions of the textures. This can be further mitigated with more specialized filtering techniques that can come with additional iterations.


NitroxClient/GameLogic/PlayerPreferences/UnityPreferenceStateStateProvider.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System;
+using System.Collections.Generic;
+using LitJson;
+using UnityEngine;
+
+namespace NitroxClient.GameLogic.PlayerPreferences
+{
+    //This abstraction allows us to write tests against the preferences manager. Otherwise - we are unduly tied to Unity.
```
Sunrunner37: Do you know of any options for avoiding with a non-paid mocking framework?  From my research, not being able to just mock everything seems to be a very unfortunate side-affect of  c#'s 'every method is final' decision.  However, I wonder if there are weaver mocking frameworks that do this out-of-the-box.

NitroxClient/GameLogic/PlayerPreferences/PlayerPreference.cs
```diff
@@ -0,0 +1,100 @@
+﻿using System;
+using UnityEngine;
+
+namespace NitroxClient.GameLogic.PlayerPreferences
+{
+    [Serializable]
+    public class PlayerPreference : IEquatable<PlayerPreference>
```
Sunrunner37: Why not serialize and persist the Player object? We'll need to store that data anyways.

MadMaxOfYore: Persistence of the player object seems like it should be maintained by the server rather than the client. That also feels like it is out-of-scope for this feature.

NitroxClient/GameLogic/PlayerPreferences/PlayerPreference.cs
```diff
@@ -0,0 +1,100 @@
...
+                int hashCode = PlayerName != null ? PlayerName.GetHashCode() : 0;
+                hashCode = (hashCode * 397) ^ RedAdditive.GetHashCode();
+                hashCode = (hashCode * 397) ^ GreenAdditive.GetHashCode();
+                hashCode = (hashCode * 397) ^ BlueAdditive.GetHashCode();
+                return hashCode;
+            }
+        }
+    }
+
+    //LitJson does not seem to be capable of ignoring certain properties.
```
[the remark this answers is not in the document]
MadMaxOfYore: Should be. That didn't feel any *less* hacky TBH, but I am indifferent.

NitroxClient/GameLogic/PlayerPreferences/UnityPreferenceStateStateProvider.cs
MadMaxOfYore: The short answer is that what you want is simply not possible without leaking testing related code into all of our concrete implementations - definitely an anti-pattern. You'd effectively need to implement your entire public API with virtual keyword declarations which is an incredibly heavy-handed approach towards testing.

The long answer is that testing dependencies with a mock of a concrete implementation is quickly becoming viewed as an anti-pattern throughout the industry as well. The bottom line is that a reusable public API should *always* implement an interface. The notion that "there is only one implementation" is a logical fallacy. There should always be two implementations at the very least - the production implementation and the test implementation. If you are not living by that, then I already know that you have coupling problems in your code base without even looking at it.

Before anyone freaks out - know that value objects do not fall into this category. It is perfectly acceptable to use concrete value objects in tests because they are; by definition, nothing more than a data structure without complicated business logic.

Boundary interfaces should never be seen as a form indirection. They are; in fact, the embodiment of every aspect of SOLID. Love them!

FWIW: Java has been moving in this direction for a long time: "An interface for every bean, and a bean for every interface!"

NitroxClient/MonoBehaviours/Multiplayer.cs
```diff
@@ -95,7 +94,12 @@ public void InitMonoBehaviours()
         public void StopCurrentSession()
         {
             SceneManager.sceneLoaded -= SceneManager_sceneLoaded;
-            multiplayerSession.Disconnect();
+			
+            if (multiplayerSession.CurrentState.CurrentStage != MultiplayerSessionConnectionStage.Disconnected)
```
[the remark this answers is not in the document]
MadMaxOfYore: If we encounter an exception while trying to join a server, then the underlying state objects will automatically disconnect the underlying client.

NitroxTest/Client/PlayerPreferences/PlayerPreferenceManagerTests.cs
```diff
@@ -0,0 +1,186 @@
...
+namespace NitroxTest.Client.PlayerPreferences
+{
+    [TestClass]
+    public class PlayerPreferenceManagerTests
+    {
+        [TestMethod]
+        public void ShouldBeAbleToRetrieveANewPreferenceEntry()
+        {
+            //Given
+            PlayerPreferenceState playerPreferenceState = new PlayerPreferenceState();
```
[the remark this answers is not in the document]
MadMaxOfYore: I'd started down the BDD path, but all I would be doing is encapsulating about 10 lines of logic. The rest would all need to reside within that specific test case. The domain just didn't seem complicated enough to warrant that level of sophistication in the tests. However; I am happy to clean this up.

NitroxClient/GameLogic/PlayerPreferences/UnityPreferenceStateStateProvider.cs
Measurity: `Boundary interfaces should never be seen as a form of indirection.` while I see the value in interfaces and especially so for testing, I do hate them just as much. They hide implementation so well that it's hard to figure out at a glance what is happening in a code base. I don't like how Enterpise java has a IFactoryBeanServiceProvider etc.. where every function hides behind another couple interfaces with every function just a few lines of trivial code.

I think there's a YAGNI in here somewhere as well but in this case it's just 1 interface for testing so that's an improvement imo.

MadMaxOfYore: When you truly come to know the ways of TDD and DDD, you will change the way that you trying to look at a code base in the first place.

At the end of the day, it is far easier to understand an application in an abstract sense than it is to have a concrete understanding of what it is doing across the board.

Looking at the interface, you can see that *something* out there gives me a state object and also saves that state object for me when I am ready. That is far easier to understand than knowing that everything is actually being stored by Unity's player preference mechanism under the hood. 

It simplifies your world and forces you to isolate responsibilities. As developers, we have a pervasive need to know exactly how something is accomplishing a task when we don't *need* that information in any context that does not directly involve changing preference persistence. 

It is an imagined necessity, not a real one.

MadMaxOfYore: I guess my core disagreement with that boils down to dependency inversion. Modules should always communicate with one another through abstractions and not implementation details. I simply don't agree that the logical number of production implementations should dictate when an interface should be used. 

I am doing a great deal of Java development myself these days, and the practice of rolling interfaces is alive and well here as far as I can tell. Almost all literature I've gotten my hands on advocates for implementing interfaces for all services as a best practice - even in Spring Boot. 

YAGNI would only apply if numerous implementations where the litmus test for whether or not an interface is warranted - which isn't the intent behind interfaces in Java or .NET. If multiple concrete implementations is really your overriding concern, then what you want is an abstract class with abstract methods - which clearly indicates a polymorphic ecosystem where many concrete implementations should be expected. 

Interfaces - those are just inverted dependencies.

Sunrunner37: If we are practicing good information hiding techniques we would not be leaking any of the concrete class's implementation details in the first place.  If we were to take the example here and implement it as a single concrete class it would be PreferenceStateProvider and NOT UnityPreferenceStateProvider.  It would also have the exact same method signatures that you defined in the interface.  That would give us the following:

* Any dependent classes that used this class wouldn't know that it used Unity preference states.
* Anyone who does a list on the public APIs wouldn't know that it used Unity specific stuff.
* A curious developer can easily click into the class to gain immediate implementation details without having to go through a layer of indirection.

This is literally the best of both worlds.  If you were to actually need a new type of PreferenceStateProvider, you can rename the current implementation, toss in an interface, and do the build out of the new one.  If you never need a new one, you don't pollute your package with unneeded layers of indirection.  

My comments about Java were not implying that we do not use interfaces.  I'm simply making a point that in Java we don't need to care if the item we are mocking is an interface or concrete class.  It simply does not matter - it is a class dependency that has methods and returns data.  To me this interface debate is just a side-effect of the main question: Is there a c# framework that allows us to not care what the type is?  Because after my thinking on information hiding principals above, the only thing that the interface seems to provide is a shim to account for c#'s limitation.

MadMaxOfYore: The only thing you do with that mindset is destroy ubiquitous language within a specific domain. If I am outside of the persistence domain, then I only need to know that I am working with some persistence provider. If I am working within the domain, then the name adds value because it tells anyone looking at a high level what is going on without needing to muck through code to figure it out. 

Your attitude on interfaces is one that is loosing popularity I am afraid - even in the Java world. All modern IDE's provide mechanisms for navigating directly to interface implementations with simple hotkeys because it is incredibly common to have one and only one class that implements a specific interface. 

The decision to default methods as sealed was made specifically to keep developers from doing what you are suggesting. The original design team looked at Java, and that was one of the things that they almost unanimously did not like about the language. It certainly would not be hard or even breaking for Microsoft to change the default behavior of .NET to virtual implementations, but they are not going to do that because their opinion (and I do not mean to offend you) is that it leads to teams taking a nonchalant mindset towards dependency inversion.

Sunrunner37: It is not accurate to say that "all this does is destroy ubiquitous language" when in reality it solves all of the major problems you stated above (about interacting with dependencies in an abstract way) with the added benefit of one less level of indirection.  As a counter point, do people really gain significant context by seeing this as UnityPreferenceStateProvider?  What about areas that have the interface and implementation named exactly the same like LocalPlayer/ILocalPlayer?  Clearly there is no 'ubiquitous language' to be destroyed there.  Domain based naming makes sense in a multi-implementation environment but this seems to be a rather trivial detail in the single implementation use-case that we are discussing.  Certainly more minor than an additional layer of indirection for a small project like Nitrox.

Microsoft's decision had nothing to do with dependency inversion nor had anything to do with what I am suggesting (https://www.artima.com/intv/nonvirtual.html).  They specifically chose this approach to maintain a class's implicit contract across versions.  In fact, this choice was mostly selfish in nature.  They decided to use a language construct to blanketly seal everything instead of designing for inheritance and using the sealed keyword.  These details are totally outside of this discussion; however, it has significant negative implications for anything that leverages proxy classes: mocking frameworks, dependency injection frameworks, and ORMs.  It would be absurd to say that Microsoft is keeping methods non-virtual to prevent concrete mocks.

NitroxClient/GameLogic/PlayerPreferences/PlayerPreference.cs
Sunrunner37: This was my bad, I thought this was a server abstraction.

NitroxClient/GameLogic/PlayerPreferences/UnityPreferenceStateStateProvider.cs
MadMaxOfYore: I never stated it was for the purposes of preventing concrete mocking - I said it was for the purposes of encouraging the practice of dependency inversion: which speaks directly towards responsible API versioning. The point was that the decision to make a method virtual should be made deliberately with a mindset towards allowing for behavioral overrides in a way that a consumer will not inadvertently invalidate the internal state of the base classes - not just so you can test it later. Anders has gone on record (just not in that article you found) by stating that he believes the default virtual behavior in Java has been abused in a number of ways - doing it just so you can mock it is one of the many abuses that it encourages. 

Making a method (or leaving one) virtual just so it can be tested is a hack  - plain and simple. If it wasn't a hack, then you wouldn't need a framework just to suppress the constructor and jump through a number of other hoops just so someone can press Ctrl+B instead of Ctrl+Alt+B....

I cannot disagree more that a mock should be thought of as a derived sub class. Inheriting a base abstraction simply to completely override every method implementation knowing that you will never take advantage of the base behavior is a clear violation of Liskov's substitution principle. You are quite literally taking a duck, ripping all of its innards out, and wrapping the remaining flesh around something fake and artificial. Gruesome, ugly, and absolutely not polymorphic.

A BMW is an extension of a Car all day.
A plastic (pretend) duck is in no way an extension of a real duck, but a Mallard is....

On top of all of that; what you are advocating would effectively mean that you can't use the `final` keyword for no other reason than it prevents testing - and now we are right back to leaking testing details into production code rather than structuring our code to be testable. They are two completely different worlds entirely.

Sunrunner37: I would love to debate the finer details of which language made the correct design decision.  However, I can't help but notice that you have all but ignored my first paragraph which is talking about our specific project.  In regards to a single-implementation interface, I have provided you with a clear-cut example in which information hiding provides the same benefits as an interface (abstract view for dependencies with no indirection) and directly replied to your concerns surrounding destruction of ubiquitous language.  You have only provided me with a response on why using virtual is bad, which leads me into my main point....

You are under the false pretense that I am advocating the use of virtual within our code base - this is completely false. My original question was asking if there was a weaving mocking framework that can mock a concrete class (which I then asked two more times in subsequent posts).  I even said that using virtual was a hack in my first response.  I have simply pointed out that java does not have mocking limitations because everything is virtual (whether or not this is the correct choice for c#, we can debate at a later time). 

Mocks are only a tool for unit testing - we shouldn't design our software differently because we know we'll be using a mock.  I consider the 'pepper everything with virtual' as well as 'testing is my other implementation so I'll make an interface' to both be instances of testing leaking into production code.  When creating a mock, what does it matter if it is mocking an interface or a concrete class?  In the context of a unit test, the mock is simply a class dependency that gets invoked - why should we care if it has its own implementation that we are overriding?

Above, I see that you have some confusion over how I consider mocks.  I distinctly said that *concrete* mocks are derived subclasses.  This is always true because it is the only way you can implement them.  You are extending a class so you must subclass.  I did NOT say that all mocks should be considered derived subclasses.  As for the duck example, a mock is simply a proxy class, used for testing, that has pieces of its implementation overridden and is injected as a class dependency - it seems weird and unnecessary to introspectively consider its polymorphism.  Yes, in the real world, a mock can be that plastic bird because who really cares?

PR #200, opened 2018-03-24
Added server side change

killzoms:
I added some customization options for server side port selection and just finished code for client side port selection!
A quick install script was also made for this build using the windows batch file.


NitroxServer/Communication/TcpServer.cs
```diff
@@ -1,84 +1,84 @@
...
+        private readonly PacketHandler packetHandler;
+        private readonly PlayerManager playerManager;
+
+        public TcpServer(PacketHandler packetHandler, PlayerManager playerManager)
+        {
+            this.packetHandler = packetHandler;
+            this.playerManager = playerManager;
+        }
+        
+        public void Start(int port = 11000, int pEnd = 4000)
```
killzoms: Set default values if port and pEnd are left blank

NitroxClient/Communication/TcpClient.cs
```diff
@@ -1,83 +1,96 @@
...
+
+        public void Start(string ipAddress)
+        {
+            try
+            {
+                //If Ip address includes a port use the included port
+                port = 11000;
+                string[] splitIp = ipAddress.SplitByChar(':');
+
+                for (int i = 0; i < splitIp.Length; i++)
```
MarijnS95: Such a complicated loop is not necessary here, and it'll set `ipAddress = splitIp[0];` on every iteration. Consider doing something like this instead:

```cs
Validate.IsTrue(!string.IsNullOrEmpty(ipAddress));
string[] splitIp = ipAddress.Split(':');
ipAddress = splitIp[0];
if (splitIp.Length > 1)
{
    port = int.Parse(splitIp[1]);
}
```

... and eventually throw an error if more than one colon is found.

We might want to check if .NET includes such an `address:port` parser already (returning an `IPEndPoint` combining the two).

NitroxServer/Server.cs
```diff
@@ -1,29 +1,34 @@
...
+            PacketHandler packetHandler = new PacketHandler(playerManager, timeKeeper, simulationOwnership);
+            EventTriggerer eventTriggerer = new EventTriggerer(playerManager);
+
+
+            tcpServer = new TcpServer(packetHandler, playerManager);
+        }
+
+        public void Start()
+        {
+            int port = int.Parse(ConfigurationManager.AppSettings.Get("port"));
```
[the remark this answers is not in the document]
killzoms: Has been rectified.

NitroxClient/Communication/TcpClient.cs
killzoms:
```C#
if (splitIp.Length > 1)
{
    port = int.Parse(splitIp[1]);
    if (splitIp.Length > 2)
    {
        throw new Exception("Multiple ports detected!");
    }
}
```

Something like this should help prevent errors with multiple colons

Sunrunner37:
Hello,

Thank you for your first PR.  Unfortunately, we will need to close this one out for a few reasons:

- As Marin specified above, we have decided not to go with this configuration management system.  Maybe hop on discord and we can settle on the correct implementation.
- There are too many style changes in this commit which makes it very hard to see the functional changes.   Global stylistic refactors are OK as long as they are included in their own PR.  Lastly, we tend to leave code untouched for the easy git blame history.
- The new line handling makes it look like the entire file has been modified.  Please fix this upon future submissions.

Issue #122, opened 2018-02-04
[Misc] Disable lifepod exit cinematics / creature spawns


Sunrunner37:

MadMaxOfYore:
Note: you can disable the intro screen with a call to Utils.SetContinueMode(true)

Can probably do this in Multiplayer.StartSession()

Leaving this open as I have bigger fish to fry and this seems like a good story to leave for any potential new-comers looking to get their feet wet. Will pick it up if it is still lingering around near our first alpha release.

Coding-Hen:
I assume that this be closed as its been marked as complete?

PR #212, opened 2018-04-30
Disable intro sequences in multiplayer

Kossetsoup:
Implements #122 

Introductory sequence is disabled in multiplayer, and first time escape pod exit cutscenes don't play in multiplayer either (and the creatures aren't spawned).

Intro and pod cutscenes play as normal in singleplayer.

### DIFF
--- a/NitroxClient/MonoBehaviours/Multiplayer.cs
+++ b/NitroxClient/MonoBehaviours/Multiplayer.cs
@@ -42,6 +42,11 @@ namespace NitroxClient.MonoBehaviours
             }
         }
 
+        public bool IsMultiplayer()
+        {
+            return multiplayerSession.Client.IsConnected;
+        }
+
         public void ProcessPackets()
         {
             Queue<Packet> packets = packetReceiver.GetReceivedPackets();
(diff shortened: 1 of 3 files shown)


MarijnS95:
You don't have to worry about singleplayer/`IsMultiplayer` when making this a **non** Persistent patch (take it out of the `Persistent` namespace). It'll only be injected during multiplayer sessions.

Kossetsoup:
I made it a `Persistent` patch because I needed it to be applied sooner than when the non-persistent patches are applied. The `EscapePodFirstUseCinematicsController.Initialize()` method is called before the other patches are applied so I was unable to approach it that way.

Kossetsoup:
I tried for a few hours to try to rectify the problem with a non-`Persistent` patch, but by the time the patches were executed it is already decided that the sequences are going to execute. 

If my solution is unacceptable, I'm happy to take another look at it. However, this is the cleanest solution I could come up with. A solution using a non-`Persistent` patch would involve more than a single change. Would need to make changes in several places as the camera, spawning entities and sound are handled in different places from what I saw.

Looking forward to your reply!

PR #222, opened 2018-06-22
Fixed movement not syncing properly

ahk1221:
Nitrox syncs movement based on velocity, this causes lots of issues with player positions not being the same on all clients. With this, instead of relying on the velocity, the player positions are extrapolated from what the server sends us, making sure that all clients' players' have the same position.


Sunrunner37:
Was this meant to be deleted?

PR #230, opened 2018-06-28
Fix Player container wipe during the normal shutdown routine

garsia0:


NitroxPatcher/Patches/ItemsContainer_NotifyRemoveItem_Patch.cs
```diff
@@ -13,9 +13,13 @@ public class ItemsContainer_NotifyRemoveItem_Patch : NitroxPatch
 
         public static void Postfix(ItemsContainer __instance, InventoryItem item)
         {
+            LocalPlayer localPlayer = NitroxServiceLocator.LocateService<LocalPlayer>();
             if (item != null && __instance.tr.parent.name != "EscapePod" && __instance.tr.parent.name != "Player")
             {
-                NitroxServiceLocator.LocateService<ItemContainers>().BroadcastItemRemoval(item.item, __instance.tr.parent.gameObject);
+                if (!localPlayer.Shutdown)
```
Sunrunner37: Do we know why this code path is being called when the client disconnects?  Would it be better to patch the code path that triggers this to not trigger this in multiplayer?

garsia0: if we path the function make that bug on user go to main screen the client dont release memory resources from current session i   find a better file for catch the event work went user press Alt+F4 to

 
namespace NitroxPatcher.Patches
{
    public class Player_OnDestroy_Patch : NitroxPatch
    {
        public static readonly Type TARGET_CLASS = typeof(Player);
        public static readonly MethodInfo TARGET_METHOD = TARGET_CLASS.GetMethod("OnDestroy", BindingFlags.NonPublic | BindingFlags.Instance);

        public static void Prefix()
        {
            LocalPlayer localPlayer = NitroxServiceLocator.LocateService<LocalPlayer>();
            if (localPlayer != null)
            {
                localPlayer.Shutdown = true;
            }
            Log.Info("Player OnDestroyEvent");
        }

        public override void Patch(HarmonyInstance harmony)
        {
            PatchPrefix(harmony, TARGET_METHOD);
        }
    }
}

NitroxServer/Communication/Packets/PacketHandler.cs
```diff
@@ -74,6 +74,10 @@ private void ProcessAuthenticated(Packet packet, Player player)
             {
                 defaultPacketProcessor.ProcessPacket(packet, player);
             }
+            if (packet.GetType().ToString() != "NitroxModel.Packets.Movement")
```
Sunrunner37: Can you remove this?

NitroxPatcher/Patches/Equipment_AddItem_Patch.cs
```diff
@@ -13,8 +14,10 @@ public class Equipment_AddItem_Patch : NitroxPatch
 
         public static void Postfix(Equipment __instance, bool __result, string slot, InventoryItem newItem)
         {
+            Log.Info("Equipment_AddItem_Patch #########################################################################################################");
```
Sunrunner37: Can you remove this?

NitroxServer/Communication/Packets/PacketHandler.cs
garsia0: dont im firt time working witch GitHub need lean how make PR only files i want send

garsia0:
sorry for the disaster i make clean up

PR #237, opened 2018-07-01
add Sign for World Saving

garsia0:


Sunrunner37:
I think we need to solve this problem a little more generically.  Having to create a new list for every stateful base item will become too much.  This is what I propose:

* Add a new interface called BasePieceMetaData and extend with an initial implementation of SignMetaData.  (You'll need to use ProtoInclude(SignMetaData) in BasePieceMetaData for serialization)
* Add a BasePieceMetaData as a new field to BasePiece.
* Add a new interface for BasePieceMetaDataProcessor<T> and extend with an initial implementation of SignMetaDataProcessor<SignMetaData>
* In the initial sync packet processor, register all BasePieceMetaDataProcessors
* When an initial sync packed is processed, looped through all meta data processors and call them accordingly. (You may have to wait until all base pieces are placed.  To do this, look at how InventoryItemAdder listens to ThrottledBuilder.main.QueueDrained event before adding inventory items to base pieces)

garsia0:
any example in the code of structure you want ?

Sunrunner37:
An example can be found by looking at entity bootstrapping (code that is specific to an tech type and runs when first creating an entity).  It follows almost the same pattern:

1) We declare an interface for entity spawn bootstrappers: https://github.com/SubnauticaNitrox/Nitrox/blob/master/NitroxServer/GameLogic/Entities/Spawning/IEntitySpawner.cs

2) We have a specific implementation (in this example crashhome needs to spawn crashfish and sulfer): https://github.com/SubnauticaNitrox/Nitrox/blob/master/NitroxServer/GameLogic/Entities/Spawning/EntityBootstrappers/CrashFishBootstrapper.cs

3) We add custom bootstrappers by tech type: https://github.com/SubnauticaNitrox/Nitrox/blob/master/NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs#L44

4) We call a custom bootstrapper if one exists for the given tech type: https://github.com/SubnauticaNitrox/Nitrox/blob/master/NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs#L156

Sunrunner37:
This was completed using the new base piece metadata system https://github.com/SubnauticaNitrox/Nitrox/commit/41b530bdb616db858dcaa7ca3dadedb9bfcdd3b2

PR #238, opened 2018-07-01
* Fixed bug with flora spawn duplication

### LINKED ISSUE #236: Double spawn of resources
image

It might be related to fish overspawning. I think everything is spawned two times. It's not related to the number of players. I'll try to find more information debugging.

killzoms:
Fixes #236

Reorganized method keeping end result whilst filtering and removing duplicated materials.

### DIFF
--- a/NitroxClient/NitroxClient.csproj
+++ b/NitroxClient/NitroxClient.csproj
@@ -123,6 +123,7 @@
     <Compile Include="Communication\Packets\Processors\PDAEncyclopediaEntryAddProcessor.cs" />
     <Compile Include="Communication\Packets\Processors\PDALogEntryAddProcessor.cs" />
     <Compile Include="Communication\Packets\Processors\SeamothModuleActionProcessor.cs" />
+    <Compile Include="Communication\Packets\Processors\SimulationOwnershipResponseProcessor.cs" />
     <Compile Include="Communication\Packets\Processors\VehicleOnPilotModeChangedProcessor.cs" />
     <Compile Include="Communication\Packets\Processors\VehicleDestroyedProcessor.cs" />
     <Compile Include="Communication\Packets\Processors\VehicleCreatedProcessor.cs" />
--- a/NitroxModel/NitroxModel.csproj
+++ b/NitroxModel/NitroxModel.csproj
@@ -155,6 +155,7 @@
     <Compile Include="Packets\SignChanged.cs" />
     <Compile Include="Packets\SimulationOwnershipChange.cs" />
     <Compile Include="Packets\SimulationOwnershipRequest.cs" />
+    <Compile Include="Packets\SimulationOwnershipResponse.cs" />
     <Compile Include="Packets\StoryEventSend.cs" />
     <Compile Include="Packets\TimeChange.cs" />
     <Compile Include="Packets\ToggleLights.cs" />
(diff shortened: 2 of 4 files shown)


killzoms:
just occurred to me that i forgot to add a filter for nickel sulfur and kyanite... diamonds that spawn naturally on the seabed will also need to be fixed... will do when i wake up.

quimnuss:
Then I guess Magnetite as well?

Sunrunner37:
I'm not sure that this is the correct fix for the problem.  This code path is invoked when the server tells the client to spawn entities.  Filtering out entities, and allowing the clients original entities to remain, will cause the server and client guids to become out of sync.  Generally this might be OK for flora (as guids are mostly static); however, we want the server to always be authoritative on the guids.  I think we may want to dig deeper and figure out why the client is spawning flora, disable that code path, and allow the server entities to spawn as they would normally.

killzoms:
From what I can tell 
```c#
ParseFile(batchId, "", "other", spawnPoints);
```
is the origin of the spawning for flora however i am uncertain as to how many other things are spawned by it.

Sunrunner37:
That is the origin of the spawning on the server side which should remain as-is because it is authoritative.  We need to find the source of the client side spawning (somewhere in assembly-csharp) and patch it to disable.

killzoms:
I believe this will work.

MarijnS95:
There was a recent PR (https://github.com/SubnauticaNitrox/Nitrox/pull/212) that used Persistent patches for a similar reason; as you can see it requires an extra check to not change behaviour in the single player environment.

I didn't get to reply to it, but I'd much rather prefer fixing **Non**Persistent patches to run before **anything** we possibly want to do, instead of having a few of these inbetween _semi-persistent_ patches.

killzoms:
Thank you for referencing that PR it was very informative! I wish to dig deeper however I have hit a bit of a snag. One of the methods I am trying to patch makes use of ``this`` key words and I do not know how to access ``this`` key words from the patch. I have attempted to patch manually via dnSpy to no avail...
The class and method that I think is the origin is ``LargeWorldEntity`` which is also a type and ``Start`` I have been trying to patch in a simple ``UnityEngine.MonoBehaviour.print()`` which logs information to output_log

MarijnS95:
@killzoms It is possible to use an argument named `__instance` in the Prefix/Postfix method, Harmony will fill that with a reference to `this`. There are more such substitutions: check the Harmony wiki/documentation and/or other patches.

Good luck finding the source :wink:

AlexejheroYTB:
Fixes #236

MarijnS95:
@AlexejheroYTB It has to be in the PR description for that to work; see how it has an underline and informative tooltip?

AlexejheroYTB:
Yep. Thanks for the info!

Sunrunner37:
Let us know when this is ready for review again.

killzoms:
Running tests on PaxTerrainController.LoadAsync() method.

Sunrunner37:
Do you need to change that exact line or can you patch the code inside that function?

killzoms:
either or should do however i have no skill whatsoever with dealing with Transpilers...

killzoms:
Noticed that calling Multiplayer.Main.IsMultiplayer() before loading multiplayer returns null

Sunrunner37:
Thanks for compiling the full list!

PR #251, opened 2018-07-13
Save Oxigen & Spawn Location

garsia0:


NitroxModel/DataStructures/GameLogic/PlayerSpawnData.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System;
+using ProtoBuf;
+
+namespace NitroxModel.DataStructures.GameLogic
+{
+    [Serializable]
+    [ProtoContract]
+    public class PlayerSpawnData
```
Sunrunner37: Can we just use a Vector3?  That way we don't have to create a new class.

NitroxServer/Communication/Packets/Processors/PlayerStatsProcessor.cs
```diff
@@ -0,0 +1,23 @@
...
+        private readonly World world;
+
+        public PlayerStatsProcessor(World world)
+        {
+            this.world = world;
+        }
+
+        public override void Process(PlayerStats packet, Player player)
+        {
+            world.PlayerData.PlayerStats(player.Name, new PlayerStatsData(packet.Oxygen, packet.MaxOxygen, packet.Health, packet.Food, packet.Water));
```
Sunrunner37: We can put PlayerStatsData as a member variable inside PlayerStats so you can simply pass in packet.data.

PR #257, opened 2018-07-14
 Add Remove Seamoth Sync & Save

garsia0:


garsia0:
found problem reopen later

PR #260, opened 2018-07-15
Vehicle Create Destroy Event

garsia0:


NitroxPatcher/Patches/Vehicle_Awake_Patch.cs
```diff
@@ -0,0 +1,25 @@
...
+namespace NitroxPatcher.Patches
+{
+    public class Vehicle_Awake_Patch : NitroxPatch
+    {
+        public static readonly Type TARGET_CLASS = typeof(Vehicle);
+        public static readonly MethodInfo TARGET_METHOD = TARGET_CLASS.GetMethod("Awake", BindingFlags.Public | BindingFlags.Instance);
+
+        public static void Postfix(Vehicle __instance)
+        {
+            NitroxServiceLocator.LocateService<Vehicles>().Add(__instance);
```
Sunrunner37: Is this for the console command?  I believe we are already syncing when the vehicle constructor creates a vehicle.  That code path might execute at the same time as this and cause a conflict. We may have to delete ConstructorInput_OnCraftingBegin_Patch.

NitroxClient/GameLogic/Vehicles.cs
```diff
@@ -106,5 +120,170 @@ private void OnVehiclePrefabLoaded(GameObject prefab, string guid, Vector3 spawn
...
+                        gameObject.transform.rotation = vehicle.transform.rotation;
+                    }
+                    UnityEngine.Object.Destroy(vehicle.gameObject);
+                }
+            }
+        }
+
+        public void Add(Vehicle vehicle) //Add Vehicle Fron Network Packet
+        {
+            string guid;
```
Sunrunner37: Can we declare and initialize these local variables on a single line?

NitroxClient/GameLogic/Vehicles.cs
```diff
@@ -106,5 +120,170 @@ private void OnVehiclePrefabLoaded(GameObject prefab, string guid, Vector3 spawn
...
+                        GameObject gameObject = UnityEngine.Object.Instantiate<GameObject>(vehicle.destructionEffect);
+                        gameObject.transform.position = vehicle.transform.position;
+                        gameObject.transform.rotation = vehicle.transform.rotation;
+                    }
+                    UnityEngine.Object.Destroy(vehicle.gameObject);
+                }
+            }
+        }
+
+        public void Add(Vehicle vehicle) //Add Vehicle Fron Network Packet
```
Sunrunner37: Let's rename this to CreateNewVehicle and make a new method called BroadcastNewVehicle.  The new method, BroadcastNewVehicle can call CreateNewVehicle then send the packet.  This would allow you to call only CreateNewVehicle without having to do packet surpression in the PacketProcessor. 

We should do the same thing for RemoveVehicle as well.

NitroxClient/Communication/Packets/Processors/VehicleOnPilotModeProcessor.cs
```diff
@@ -0,0 +1,31 @@
...
+
+        public VehicleOnPilotModeProcessor(IPacketSender packetSender, Vehicles vehicles)
+        {
+            this.packetSender = packetSender;
+            this.vehicles = vehicles;
+        }
+
+        public override void Process(VehicleOnPilotMode packet)
+        {
+            using (packetSender.Suppress<VehicleRemoveEntry>())
```
Sunrunner37: It doesn't look like OnPilotModeSet needs to suppress VehicleRemoveEntry.

NitroxClient/GameLogic/Vehicles.cs
```diff
@@ -106,5 +120,170 @@ private void OnVehiclePrefabLoaded(GameObject prefab, string guid, Vector3 spawn
...
+            Rigidbody rigidbody = vehicle.gameObject.GetComponent<Rigidbody>();
+
+            velocity = rigidbody.velocity;
+            angularVelocity = rigidbody.angularVelocity;
+
+            // Required because vehicle is either a SeaMoth or an Exosuit, both types which can't see the fields either.
+            steeringWheelYaw = (float)vehicle.ReflectionGet<Vehicle, Vehicle>("steeringWheelYaw");
+            steeringWheelPitch = (float)vehicle.ReflectionGet<Vehicle, Vehicle>("steeringWheelPitch");
+
+            // Vehicles (or the SeaMoth at least) do not have special throttle animations. Instead, these animations are always playing because the player can't even see them (unlike the cyclops which has cameras).
```
Sunrunner37: It looks like this code was copied from PlayerMovement.  Can we move that function here and have player movement call into vehicle instead? (we might have to do some work to make it generic) That way we don't have redundant code.

NitroxClient/GameLogic/Vehicles.cs
```diff
@@ -106,5 +120,170 @@ private void OnVehiclePrefabLoaded(GameObject prefab, string guid, Vector3 spawn
...
+
+        public void OnPilotModeSet(string vehicleGuid, string playerGuid, byte type) // Set & Unset Remote Player Value Inside of Vehicle Receiver
+        {
+            Log.Info("Receiver: VehicleGuid: " + vehicleGuid + " PlayerGuid: " + playerGuid + " Type: " + type);
+
+            Optional<GameObject> Vehicle = GuidHelper.GetObjectFrom(vehicleGuid);
+
+            if (Vehicle.IsPresent())
+            {
+                Log.Info("Object Found");
```
Sunrunner37: Remove

NitroxClient/GameLogic/Vehicles.cs
```diff
@@ -106,5 +120,170 @@ private void OnVehiclePrefabLoaded(GameObject prefab, string guid, Vector3 spawn
...
+                {
+                    if (type == 0)
+                    {
+                        V.pilotId = playerGuid;
+                    }
+                    else
+                    {
+                        V.pilotId = string.Empty;
+                    }
+                    Log.Info("Component Found");
```
Sunrunner37: Remove these two log lines

NitroxClient/GameLogic/Vehicles.cs
```diff
@@ -106,5 +120,170 @@ private void OnVehiclePrefabLoaded(GameObject prefab, string guid, Vector3 spawn
...
+
+            if (Vehicle.IsPresent())
+            {
+                Log.Info("Object Found");
+                GameObject Go = Vehicle.Get();
+                Vehicle V = Go.GetComponent<Vehicle>();
+
+                if (V != null)
+                {
+                    if (type == 0)
```
Sunrunner37: what does type of 0 mean?  Can you add a comment?

NitroxModel/DataStructures/GameLogic/VehiclePlayer.cs
```diff
@@ -0,0 +1,24 @@
+﻿using ProtoBuf;
+using System;
+using UnityEngine;
+
+namespace NitroxModel.DataStructures.GameLogic
+{
+    [Serializable]
+    [ProtoContract]
+    public class VehiclePlayer
```
Sunrunner37: This looks to be unused.  If it is not, why do we need it when it only has one string member?

NitroxModel/Packets/VehicleAddEntry.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+using NitroxModel.DataStructures.Util;
+using UnityEngine;
+using NitroxModel.DataStructures.GameLogic;
+
+namespace NitroxModel.Packets
+{
+    [Serializable]
+    public class VehicleAddEntry : VehicleMovement
```
Sunrunner37: Can we change to VehicleCreated?

NitroxModel/Packets/VehicleRemoveEntry.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+using NitroxModel.DataStructures.Util;
+using UnityEngine;
+using NitroxModel.DataStructures.GameLogic;
+
+namespace NitroxModel.Packets
+{
+    [Serializable]
+    public class VehicleRemoveEntry : Packet
```
Sunrunner37: Can we change to VehicleDestroyed?

NitroxModel/Packets/VehicleOnPilotMode.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+using NitroxModel.DataStructures.Util;
+using UnityEngine;
+using NitroxModel.DataStructures.GameLogic;
+
+namespace NitroxModel.Packets
+{
+    [Serializable]
+    public class VehicleOnPilotMode : Packet
```
Sunrunner37: Can we change to VehicleOnPilotModeChanged?

NitroxClient/GameLogic/Vehicles.cs
MarijnS95: I'm against comments if usage can be made clear with the correct type and name (enum, integer constant, or even a bool should be sufficient in this case).

NitroxClient/GameLogic/Vehicles.cs
MarijnS95: And then, most if not all of the movement/animation code is of no importance when creating the vehicle. This can all be kept in the movement update codepath.

NitroxPatcher/Patches/Vehicle_Awake_Patch.cs
garsia0: for the moment (Mobile Vehicle Bay) no sync the  ConstructorInput_OnCraftingBegin_Patch no work because a missing owner went repair that can delete Vehicle_Awake_Patch

NitroxClient/GameLogic/Vehicles.cs
garsia0: made clean entire function

NitroxModel/DataStructures/GameLogic/VehiclePlayer.cs
garsia0: hahaha i forget create that xD

garsia0:
Changes Done

PR #267, opened 2018-07-26
Dialog Conection Lost

garsia0:


NitroxClient/Communication/UdpClient.cs
```diff
@@ -59,6 +60,11 @@ public void ReceivedMessage(object peer)
                             connectedEvent.Set();
                         }
 
+                        if (LargeWorldStreamer.main != null && LargeWorldStreamer.main.IsReady() && !IsConnected)
```
Sunrunner37: Can we do:

`if (NetConnectionStatus.Disconnected)`

This will allow us to have a pop-up during the loading screen and even the color selection screen.

NitroxClient/MonoBehaviours/Gui/InGame/LostConnections.cs
```diff
@@ -0,0 +1,43 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using UnityEngine;
+using UnityEngine.UI;
+
+namespace NitroxClient.MonoBehaviours.Gui.InGame
+{
+    public class LostConnections
```
Sunrunner37: Can we rename this to LostConnectionModal?

PR #276, opened 2018-08-07
Cyclops Locker Fix

garsia0:


NitroxClient/GameLogic/ItemContainers.cs
```diff
@@ -89,5 +108,28 @@ public void RemoveItem(string ownerGuid, string itemGuid)
...
+                        StorageContainer storageContainer = tr2.GetComponent<StorageContainer>();
+                        if (storageContainer != null)
+                        {
+                            return GuidHelper.GetGuid(storageContainer.gameObject);
+                        }
+                    }
+                }
+            }
+
+            return string.Empty;
```
Sunrunner37: We should throw an exception so that this doesn't become a silent failure if UWE changes the tree structure.

NitroxClient/GameLogic/ItemContainers.cs
```diff
@@ -89,5 +108,28 @@ public void RemoveItem(string ownerGuid, string itemGuid)
...
+            string LockerId = ownerTransform.gameObject.name.Substring(7, 1);
+
+            GameObject locker = ownerTransform.parent.gameObject.FindChild("submarine_locker_01_0" + LockerId);
+            if (locker != null)
+            {
+                foreach (Transform tr in locker.transform)
+                {
+                    foreach (Transform tr2 in tr)
+                    {
+                        StorageContainer storageContainer = tr2.GetComponent<StorageContainer>();
```
Sunrunner37: Can we do GetComponentInChildren instead of these nested loops?

NitroxClient/GameLogic/ItemContainers.cs
```diff
@@ -89,5 +108,28 @@ public void RemoveItem(string ownerGuid, string itemGuid)
                 Log.Error("Could not find container field on object " + owner.name);
             }
         }
+
+        public string GetLockerGuid(Transform ownerTransform)
```
Sunrunner37: Let's name GetCyclopsLockerGuid

NitroxClient/GameLogic/ItemContainers.cs
```diff
@@ -31,10 +41,19 @@ public void BroadcastItemAdd(Pickupable pickupable, GameObject owner)
             packetSender.Send(add);
         }
 
-        public void BroadcastItemRemoval(Pickupable pickupable, GameObject owner)
+        public void BroadcastItemRemoval(Pickupable pickupable, Transform ownerTransform)
         {
+            string ownerGuid = null;
+
+            if (Regex.IsMatch(ownerTransform.gameObject.name, @"Locker0([0-9])StorageRoot$", RegexOptions.IgnoreCase))
```
Sunrunner37: let's assign to a local bool for readability:

> bool isCyclopsLocker = Regex.IsMatch(ownerTransform.gameObject.name, @"Locker0([0-9])StorageRoot$", RegexOptions.IgnoreCase);

Same in the function above.

NitroxClient/GameLogic/ItemContainers.cs
garsia0: cant, there have lots same  component in Children and for object name  the children name is different on every locker

NitroxClient/GameLogic/Vehicles.cs
```diff
@@ -186,7 +176,21 @@ private void OnVehiclePrefabLoaded(GameObject prefab, string guid, Optional<stri
                 }
             }
         }
-        
+
+        public void UpdateVehicleChildren(string guid, Optional<List<InteractiveChildObjectIdentifier>> interactiveChildrenGuids) //Destroy Vehicle From network
```
Sunrunner37: > //Destroy Vehicle From network

What does this comment mean?

PR #277, opened 2018-08-08
Nitrox Installer

dm430:
A WIX 3.x based installer for Nitrox.

Dependent on:
- WIX Toolset
- dnlib


Nitrox.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 14
-VisualStudioVersion = 14.0.25420.1
+# Visual Studio 15
```
Sunrunner37: Is it possible to leave this file untouched?

NitroxInstallerActions/packages.config
```diff
@@ -0,0 +1,4 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="dnlib" version="1.6.1" targetFramework="net461" />
```
Sunrunner37: Should we add a git submodule for dnlib? (https://github.com/0xd4d/dnlib)

AlexejheroYTB:
What would this installer compile into? (.exe?). And would it need to have the dll files in the same folder as it? (So an archive is required?)

Nitrox.sln
dm430: Yes, I just figured we would want the projects to be part of the solution for easy management. If not I can revert this file.

NitroxInstallerActions/packages.config
dm430: That is up to you. If you want I will add it as a sub module.

dm430:
@AlexjheroYTB it is a self contained MSI built with the windows installer xml (WIX) toolset. This allows for full control and customization of the installer.

AlexejheroYTB:
Does it need to have the other dll files in the same folder/archive to run, or are they archived/compiled in it?

PR #286, opened 2018-08-21
Console to WPF

garsia0:


Measurity:
I don't think that converting the server console to WPF is a good idea. WPF isn't cross-platform and the console can be if it was ported to .NET Core.

If there's a need for WPF I think the best approach is to keep the console backbone and create a new WPF project that listens in to the console by redirecting the console in/output stream and/or a server API that the WPF project can use.

garsia0:
i add Conditional build you can build console and WPF independent no have reference from another

NitroxServer/Log.cs
```diff
@@ -0,0 +1,106 @@
+﻿using System;
+using System.Diagnostics;
+
+namespace NitroxServer
+{
+    public class LogServer
```
Sunrunner37: Instead of creating a new LogServer class can we keep the old one and have a function to pass in an optional textbox?  That way we can keep the same class and don't need to maintain two of the same loggers.  It would also allow us to get rid of the static Instance in MainWindow.

NitroxServer/Communication/Packets/Processors/ServerCommandProcessor.cs
```diff
@@ -7,7 +7,13 @@ public class ServerCommandProcessor : AuthenticatedPacketProcessor<ServerCommand
     {
         public override void Process(ServerCommand packet, Player player)
         {
+#if Debug_WPF
+           MainWindow._Server.Save(); 
+#endif
+#if Debug
```
Sunrunner37: What about release?

Is there a way to pass in an instance of server instead of accessing these statically?

NitroxServer/NitroxServer.csproj
```diff
@@ -15,27 +14,53 @@
...
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
+    <OutputType>Exe</OutputType>
     <PlatformTarget>AnyCPU</PlatformTarget>
     <DebugSymbols>true</DebugSymbols>
     <DebugType>full</DebugType>
     <Optimize>false</Optimize>
     <OutputPath>bin\Debug\</OutputPath>
-    <DefineConstants>DEBUG;TRACE</DefineConstants>
+    <DefineConstants>TRACE;DEBUG;Debug</DefineConstants>
```
Sunrunner37: Do we need to define a new constant?  Why Can't we use DEBUG?

PR #295, opened 2018-08-29
Graceful server shutdown

Cha0s2nd:
I added a closing event on the main window to call the shutdown of the NetworkPeer and stop the listen thread. This solves the issue of the server still running when the window is closed.

### DIFF
--- a/NitroxServer/Communication/UdpServer.cs
+++ b/NitroxServer/Communication/UdpServer.cs
@@ -14,6 +14,8 @@ namespace NitroxServer.Communication
 {
     public class UdpServer
     {
+        private bool isStopped = true;
+
         private readonly PacketHandler packetHandler;
         private readonly PlayerManager playerManager;
         private readonly EntitySimulation entitySimulation;
@@ -36,6 +38,19 @@ namespace NitroxServer.Communication
 
             Thread thread = new Thread(Listen);
             thread.Start();
+
+            isStopped = false;
+        }
+
+        public void Stop()
+        {
+            isStopped = false;
+
+            thread.Stop();
+
+            //Close all connections here or let them timeout?
+
+            server.Stop();
         }
         
         private void Listen()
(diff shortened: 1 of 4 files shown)


NitroxServer/Communication/UdpServer.cs
```diff
@@ -36,11 +38,20 @@ public void Start()
 
             Thread thread = new Thread(Listen);
```
Measurity: Can you move this thread init to the constructor? That way you can do `thread.Join()` at the `Stop()` method to wait for the listen thread to exit.

NitroxServer/Communication/UdpServer.cs
```diff
@@ -36,11 +38,20 @@ public void Start()
...
+            server.Shutdown("Shutting down server...");
         }
         
         private void Listen()
         {
-            while (true)
+            while (!isStopped)
             {
                 // Pause reading thread and wait for messages.
                 server.MessageReceivedEvent.WaitOne();
```
Measurity: `WaitOne` blocked indefinitely right now so just add a 500 to it like so: `server.MessageReceivedEvent.WaitOne(500);`

NitroxServer/Server.cs
```diff
@@ -35,6 +35,13 @@ public void Start()
             Log.Info("Nitrox Server Started");
             EnablePeriodicSaving();
         }
+
+        public void Stop()
+        {
+            worldPersistence.Save(world);
```
Measurity: It would be proper to also stop the timer here that saves the world.

Can use `Save()` here too since it already exists. Same for the `timer.Elapsed+=`.

NitroxServer/Communication/UdpServer.cs
Cha0s2nd: Adding a timeout here might unexpected results, the NetServer.Shutdown resets the WaitOne.

Measurity: I'm not sure if the thread can stop gracefully without a timeout. If it looks fine to you (like, no hanging processes) then we'll just keep an eye on it.

PR #309, opened 2018-10-17
Changed from Any to AnyCPU to fix issue

killzoms:
Changed Any to AnyCPU to match AnyCPU Condition
and added SteamID support for Reservations


killzoms:
Added a Log file to NitroxServer and fixed bugs with SteamID

Measurity:
I think `LPlayerId` means "this Player Id is of type long? C# doesn't use that convention because it has really good tooling. You mind changing it back?

killzoms:
will change all second i can.

killzoms:
This fork is currently undergoing testing Via Custume server with several people connected

NitroxClient/Communication/Packets/Processors/VehicleDockingProcessor.cs
```diff
@@ -36,7 +36,7 @@ public override void Process(VehicleDocking packet)
             vehicle.StartCoroutine(DisablePilotingAfterAnimation(packet.VehicleGuid, packet.PlayerId));
         }
 
-        IEnumerator DisablePilotingAfterAnimation(string vehicleGuid, ushort playerId)
+        IEnumerator DisablePilotingAfterAnimation(string vehicleGuid, ulong playerId)
```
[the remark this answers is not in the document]
Measurity: The SteamAPI uses a ulong id so it's easier to store this way.

Code that works with SteamID is in JoinServer class:

`AuthenticationContext authenticationContext = new AuthenticationContext(playerName, Steamworks.SteamUser.GetSteamID().m_SteamID);`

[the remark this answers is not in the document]
Sunrunner37: Hmm, does this break multiple copies running for development purposes?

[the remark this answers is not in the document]
killzoms: havent tested yet however it shouldnt break it will likely treat all the characters inventories as the same inventory with the duplicate character models going to 0,0,0

killzoms:
I am wondering if its fine to replace the GUID here 
``public override void NegotiateReservation(IMultiplayerSessionConnectionContext sessionConnectionContext)
        {
            try
            {
                ValidateState(sessionConnectionContext);

                string reservationCorrelationId = Guid.NewGuid().ToString();
                RequestSessionReservation(sessionConnectionContext, reservationCorrelationId);
                AwaitSessionReservation(sessionConnectionContext, reservationCorrelationId);
            }
            catch (Exception)
            {
                Disconnect(sessionConnectionContext);
                throw;
            }
        }``
with the steamId implemented in a later class

NitroxClient/Communication/MultiplayerSession/ConnectionState/AwaitingReservationCredentials.cs
```diff
@@ -15,8 +15,8 @@ public override void NegotiateReservation(IMultiplayerSessionConnectionContext s
             try
             {
                 ValidateState(sessionConnectionContext);
-
-                string reservationCorrelationId = Guid.NewGuid().ToString();
+                ulong steamId = Steamworks.SteamUser.GetSteamID().m_SteamID;
+                string reservationCorrelationId = steamId.ToString();
```
Sunrunner37: Are there any edge cases that occur when the playerId and the reservation id are the same? For example, what if I am authenticating, cancel, then try to authenticate again.  

What was the original motivation for having these two ids be the same?

NitroxPatcher/Patches/Persistent/ProtobufSerializer_DeserializeIntoGameObject_Patch.cs
```diff
@@ -49,15 +50,27 @@ class ProtobufSerializer_DeserializeIntoGameObject_Patch : NitroxPatch
             "eb6634e5-3a58-4a0d-ae4e-b673e1fa51ea", // Coral_reef_jeweled_disk_purple_01_02
             "c197a6ca-f910-43db-92ab-2e35e423a6f1", // Coral_reef_jeweled_disk_purple_01_04
             "9a643563-9278-4c77-8bd2-f9b4b1a1053a", // Coral_reef_jeweled_disk_red_01_03
-
+            "37ea521a-6be4-437c-8ed7-6b453d9218a8", // FloaterLarge
```
Sunrunner37: Thanks for adding more of these.

NitroxServer/GameLogic/PlayerManager.cs
```diff
@@ -55,12 +56,24 @@ public List<Player> GetPlayers()
                     MultiplayerSessionReservationState rejectedState = MultiplayerSessionReservationState.Rejected | MultiplayerSessionReservationState.UniquePlayerNameConstraintViolated;
                     return new MultiplayerSessionReservation(correlationId, rejectedState);
                 }
+                
+                ulong playId = playerData.PlayerId(playerName, Convert.ToUInt64(correlationId));
+                if (allowMultipleIds) // Duplicate steam ids will be emulated as seperate players
+                {
+                    foreach (PlayerContext Context in reservations.Values)
```
Sunrunner37: Wouldn't these PlayerContexts need to have their ids sorted for this to work?  If there are three local instances they may be out of order example:

Trying to Connect with: 1 
Already Connected Ids: 2, 1
Output Id: 2 (incorrect)

Or are these always in correct sorted order?

Additionally, it would be ideal to have this whole block in a separate method for readability. Something like, resolveOverlappingSteamIds or something similar.

NitroxServer/GameLogic/PlayerManager.cs
```diff
@@ -55,12 +56,24 @@ public List<Player> GetPlayers()
                     MultiplayerSessionReservationState rejectedState = MultiplayerSessionReservationState.Rejected | MultiplayerSessionReservationState.UniquePlayerNameConstraintViolated;
                     return new MultiplayerSessionReservation(correlationId, rejectedState);
                 }
+                
+                ulong playId = playerData.PlayerId(playerName, Convert.ToUInt64(correlationId));
+                if (allowMultipleIds) // Duplicate steam ids will be emulated as seperate players
```
Sunrunner37: If we don't allow, shouldn't we have a mechanism to reject the authentication?

NitroxServer/GameLogic/PlayerManager.cs
```diff
@@ -17,6 +17,7 @@ public class PlayerManager
         private readonly Dictionary<string, PlayerContext> reservations = new Dictionary<string, PlayerContext>();
         private readonly Dictionary<Connection, ConnectionAssets> assetsByConnection = new Dictionary<Connection, ConnectionAssets>();
         private readonly PlayerData playerData;
+        private readonly bool allowMultipleIds = true;
```
Sunrunner37: minor: to be clearer, allowAuthenticationOfDuplicateSteamIds

NitroxClient/Communication/MultiplayerSession/ConnectionState/Disconnected.cs
```diff
@@ -66,7 +66,8 @@ private static void StartClient(string ipAddress, IClient client)
 
         private static void EstablishSessionPolicy(IMultiplayerSessionConnectionContext sessionConnectionContext, IClient client)
         {
-            string policyRequestCorrelationId = Guid.NewGuid().ToString();
+            ulong steamId = Steamworks.SteamUser.GetSteamID().m_SteamID;
```
Sunrunner37: Would this throw an exception if steam was not authenticated? (duplicate instances for testing)

NitroxServer/ServerLogger.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.IO;
+
+namespace NitroxServer
+{
+    class ServerLogger : TextWriter
```
Sunrunner37: Does the Log class not handle writing to console?  I forget

NitroxServer/GameLogic/Players/PlayerData.cs
```diff
@@ -168,11 +165,11 @@ public PersistedPlayerData()
             {
                 // Constructor for serialization purposes
             }
-
-            public PersistedPlayerData(string playerName, ushort playerId)
+            
+            public PersistedPlayerData(string playerName, ulong lPlayerId)
```
Sunrunner37: minor: lowercase playerId.  No Hungarian notation.

NitroxServer/ServerLogger.cs
Measurity: Yup it does. `ServerLogger` should be removed and use `Log` instead.

killzoms: The log class writes to console however it doesnt write console to a file hence the addition of ServerLogger which writes to both a file and console

NitroxModel/MultiplayerSession/MultiplayerSessionReservationState.cs
```diff
@@ -18,7 +18,10 @@ public enum MultiplayerSessionReservationState
         ServerPlayerCapacityReached = 1 << 2,
 
         [Description("The password that you provided for the server is incorrect.")]
-        AuthenticationFailed = 1 << 3
+        AuthenticationFailed = 1 << 3,
+
+        [Description("The SteamID is already in use. Are you sure your not logged in?")]
```
Measurity: _your_ -> _you're_ ;)

NitroxServer/GameLogic/PlayerManager.cs
Measurity: Instead of adding 1 to the steam id maybe use a cpu ticks instead?

`playId += System.Diagnostics.Stopwatch.GetTimestamp();`

NitroxServer/ServerLogger.cs
Measurity: Maybe enhance or inherit from `Log` and add an additional file writing option instead? Or should we opt for a third-party logger that does it for us?

killzoms:
Recent developments have made this an incorrect thing to implement at this time

Issue #306, opened 2018-10-10
A way to redirect port/s or to change IP


custume:
Hello to all,

Got your server running on a VM inside my server but I had a hard time redirecting port because the server don't latch on the external IP, just the internal, so inside my network all good (inside my Vlan) but outside I get port close all the time if I try to connect directly to the server.

It is online, you can connect by domain : custume-server.com on port 11000

But I had to redirect the port 11000 to 5000 inside my DMZ and them from port 5000 to 11000 inside my network, and it works now but was hard to find out the error

can you guys add some way ( if you have time ) to accept other port or IP so that we can config the server better and not running port redirects all over the place 

Thank you all for the mod and best regards

MarijnS95:
It is open source, so anyone is free to contribute a properly working configuration mechanism (https://github.com/orgs/SubnauticaNitrox/projects/1#card-7411616).

As a temporary solution, you could change the ports ([here](https://github.com/SubnauticaNitrox/Nitrox/blob/master/NitroxClient/Communication/UdpClient.cs#L14) and [here](https://github.com/SubnauticaNitrox/Nitrox/blob/master/NitroxServer/Communication/UdpServer.cs#L159)) and recompile. Of course every player needs to have that change locally as well but that shouldn't be an issue given that this software is still pre-alpha (without any release).

custume:
Great reply, thank you for the directions.

I can change the port but the string ipAddress where is the variable located ?? that is what I like to change, that way I can add a static IP and move on.

For the moment is already running with 2 redirects but with the info you give me I can custom tailed the server

I already offer my help to contribute with testing and debug but got no answer, even try on the subnautica forum but nothing, is not like I have to much time but I'm happy to help in any way.

Thx and best regards

quimnuss:
I didn't even know there was a forum. You can put your issues here on the repository or visit the subnautica modding discord for a more interactive experience.

Althought the mp devs are very busy lately, you're welcome. 

El 10 d’octubre de 2018 17:35:40 CEST, Nuno Ribeiro <notifications@github.com> ha escrit:

custume:
there is a discord ?!? , got it

Thx for the reply

MarijnS95:
> I already offer my help to contribute with testing and debug but got no answer, even try on the subnautica forum but nothing, is not like I have to much time but I'm happy to help in any way.

[...]

PR #310, opened 2018-10-23
Added support for custom ports and no. of max connections

FahadAhmad2001:
- Added support for both clients and servers to use ports other than 11000.
- For servers, a config.ini file is created where the server exe is located. From here you can change the port number and number of max connections.
- For clients, in the add server, if you want to use a custom port, eg 6500, just add it to the host textbox, so for example 127.0.0.1:6500. If no port is added, it will automatically use port 11000

### DIFF
--- a/NitroxClient/Communication/Abstract/IMultiplayerSessionState.cs
+++ b/NitroxClient/Communication/Abstract/IMultiplayerSessionState.cs
@@ -7,7 +7,7 @@ namespace NitroxClient.Communication.Abstract
     {
         IClient Client { get; }
         string IpAddress { get; }
-        int serverPort { get; }
+        int ServerPort { get; }
         MultiplayerSessionPolicy SessionPolicy { get; }
         PlayerSettings PlayerSettings { get; }
         AuthenticationContext AuthenticationContext { get; }
--- a/NitroxClient/Communication/MultiplayerSession/MultiplayerSessionManager.cs
+++ b/NitroxClient/Communication/MultiplayerSession/MultiplayerSessionManager.cs
@@ -28,7 +28,7 @@ namespace NitroxClient.Communication.MultiplayerSession
 
         public IClient Client { get; }
         public string IpAddress { get; private set; }
-        public int serverPort{ get; private set; }
+        public int ServerPort{ get; private set; }
         public MultiplayerSessionPolicy SessionPolicy { get; private set; }
         public PlayerSettings PlayerSettings { get; private set; }
         public AuthenticationContext AuthenticationContext { get; private set; }
@@ -39,7 +39,7 @@ namespace NitroxClient.Communication.MultiplayerSession
         public void Connect(string ipAddress,int port)
         {
             IpAddress = ipAddress;
-            serverPort = port;
+            ServerPort = port;
             CurrentState.NegotiateReservation(this);
         }
 
@@ -94,7 +94,7 @@ namespace NitroxClient.Communication.MultiplayerSession
         public void ClearSessionState()
         {
             IpAddress = null;
-            serverPort = 11000;
+            ServerPort = 11000;
             SessionPolicy = null;
             PlayerSettings = null;
             AuthenticationContext = null;
--- a/NitroxClient/MonoBehaviours/Gui/MainMenu/JoinServer.cs
+++ b/NitroxClient/MonoBehaviours/Gui/MainMenu/JoinServer.cs
@@ -157,7 +157,7 @@ namespace NitroxClient.MonoBehaviours.Gui.MainMenu
             }
             catch (ClientConnectionFailedException)
             {
-                Log.InGame($"Unable to contact the remote server at: {ServerIp}:{serverPort.ToString()}");
+                Log.InGame($"Unable to contact the remote server at: {ServerIp}:{serverPort}");
                 OnCancelClick();
             }
         }
--- a/NitroxServer/Server.cs
+++ b/NitroxServer/Server.cs
@@ -23,9 +23,9 @@ namespace NitroxServer
             worldPersistence.Save(world);
         }
 
-        public Server(ServerConfigReader ConfigReader)
+        public Server(ServerConfigReader configReader)
         {
-            ServerOptions = ConfigReader;
+            ServerOptions = configReader;
             Instance = this;
             worldPersistence = new WorldPersistence();
             world = worldPersistence.Load();
--- a/NitroxClient/Communication/MultiplayerSession/ConnectionState/Disconnected.cs
+++ b/NitroxClient/Communication/MultiplayerSession/ConnectionState/Disconnected.cs
@@ -19,7 +19,7 @@ namespace NitroxClient.Communication.MultiplayerSession.ConnectionState
 
                 IClient client = sessionConnectionContext.Client;
                 string ipAddress = sessionConnectionContext.IpAddress;
-                int port = sessionConnectionContext.serverPort;
+                int port = sessionConnectionContext.ServerPort;
                 StartClient(ipAddress, client,port);
                 EstablishSessionPolicy(sessionConnectionContext, client);
             }
--- a/NitroxTest/Client/Communication/MultiplayerSessionTests/MultiplayerSessionMangerTests.cs
+++ b/NitroxTest/Client/Communication/MultiplayerSessionTests/MultiplayerSessionMangerTests.cs
@@ -48,7 +48,7 @@ namespace NitroxTest.Client.Communication.MultiplayerSessionTests
 
             // Assert
             multiplayerSession.IpAddress.ShouldBeEquivalentTo(TestConstants.TEST_IP_ADDRESS);
-            multiplayerSession.serverPort.ShouldBeEquivalentTo(TestConstants.TEST_SERVER_PORT);
+            multiplayerSession.ServerPort.ShouldBeEquivalentTo(TestConstants.TEST_SERVER_PORT);
         }
 
         [TestMethod]


FahadAhmad2001:
#306

FahadAhmad2001:
#306

NitroxClient/MonoBehaviours/Gui/MainMenu/JoinServer.cs
```diff
@@ -157,7 +157,17 @@ private void StartMultiplayerClient()
             }
             catch (ClientConnectionFailedException)
             {
-                Log.InGame($"Unable to contact the remote server at: {ServerIp}:11000");
+                string Splitter = ":";
```
Measurity: Instead of using an if check on the ipaddress string I would just always include the port number given. You can use `multiplayerSession.Port` if you added it.

`Log.InGame($"Unable to contact the remote server at: {multiplayerSession.IpAddress}:{multiplayerSession.Port}");`

NitroxServer/Communication/UdpServer.cs
```diff
@@ -156,8 +156,34 @@ private Connection GetConnection(long remoteIdentifier)
         private NetPeerConfiguration BuildNetworkConfig()
         {
             NetPeerConfiguration config = new NetPeerConfiguration("Nitrox");
-            config.Port = 11000;
-            config.MaximumConnections = 100;
+            string INIContents;
```
Measurity: You can use `File.ReadAllText(@".\config.ini")` instead of `StreamReader.ReadToEnd`. Saves 2 lines.

The UdpServer is getting messy if it also contains INI reading and parsing logic. It would be better to put an INI file reading/writing/parsing class in the `NitroxModel` project and then use it here. Though I'm ok with it to keep it like this for now because of the work required.

NitroxServer/Communication/UdpServer.cs
```diff
@@ -156,8 +156,34 @@ private Connection GetConnection(long remoteIdentifier)
...
+                ParseINIContents=SplitINIContents[0].Split(equals);
+                PortNumber = int.Parse(ParseINIContents[1]);
+                ParseINIContents = SplitINIContents[1].Split(equals);
+                MaxConn = int.Parse(ParseINIContents[1]);
+            }
+            else
+            {
+                string NewINIContents;
+                NewINIContents = "[NetworkSettings]\nDefaultPortNumber=11000;\nMaxConnections=100;";
+                StreamWriter WriteINI = new StreamWriter(@".\config.ini");
```
Measurity: Similar to code above you can use `File.WriteAllText(@".\config.ini", NewINIContents);` instead of StreamWriter.

NitroxClient/Communication/UdpClient.cs
```diff
@@ -26,13 +26,28 @@ public UdpClient(DeferringPacketReceiver packetManager)
         public void Start(string ipAddress)
```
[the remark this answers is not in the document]
FahadAhmad2001: As this causes errors with NitroxTest and ClientTester as port number is not specified in these projects, so in these 2 projects, do I set the port number as 11000, like this:

NitroxServer/Communication/UdpServer.cs
FahadAhmad2001: Fixed this in the new commit

FahadAhmad2001:
Added the requested changes in the new commit

NitroxClient/Communication/Abstract/IClient.cs
```diff
@@ -8,7 +8,7 @@ namespace NitroxClient.Communication.Abstract
     public interface IClient
     {
         bool IsConnected { get; }
-        void Start(string ipAddress);
+        void Start(string ipAddress, int ServerPort);
```
Measurity: Method parameters should be camcelCase. `serverPort`

FahadAhmad2001: Should I change it to serverPort

Measurity: Yes, that's the naming convention for method parameters.

NitroxServer/Communication/UdpServer.cs
```diff
@@ -156,8 +156,30 @@ private Connection GetConnection(long remoteIdentifier)
         private NetPeerConfiguration BuildNetworkConfig()
         {
             NetPeerConfiguration config = new NetPeerConfiguration("Nitrox");
-            config.Port = 11000;
-            config.MaximumConnections = 100;
+            string INIContents;
+            int PortNumber, MaxConn;
+            if (File.Exists(@".\config.ini"))
```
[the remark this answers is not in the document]
FahadAhmad2001: I created a new class to read the config, but do I call it before server.Start(), or call it when config.Port and config.MaximumConnections are being set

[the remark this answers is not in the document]
Measurity: Ideally, we should pass a custom config class into the UdpServer constructor so that file reading happens before the `UdpServer` is created.

Something like: `new UdpServer(..., UdpServerConfig.FromFile("config.ini"));` and then in the `FromFile` function you use the IniReader/Writer class you created and store the relevant INI fields in the custom config class.

Lastly, map the custom config class to `NetPeerConfiguration` inside UdpServer constructor so that Lidgren classes (like `NetPeerConfiguration`) don't leave the `UdpServer` class.

To answer your question: [`Server.Start()`](https://github.com/SubnauticaNitrox/Nitrox/blob/4cececc388920cef22b0f83313b25ff01ee0a939/NitroxServer/Server.cs#L44) sounds like the best location because reading files is slow. Move the `new UdpServer()` from the `Server` class constructor into the `Server.Start()` to do this.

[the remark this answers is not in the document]
Sunrunner37: I'm not sure if the config class should be specific to UdpServer.  I think that we will have a wide range of server configurations that we'll want to access in the future. (An example being how often we auto save).  When the server first starts we can parse the config file and pass around a data structure that encapsulates all server configs.  A subset of these configs could be NetworkConfigs.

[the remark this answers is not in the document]
Measurity: You're right. A more abstract server config file would be better. But I'm not sure how much time  @FahadAhmad2001 has to implement all this. Please let us know what you think.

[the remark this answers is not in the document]
FahadAhmad2001: I tried moving the new UdpServer() to server.Start(), but then I have to change it from private readonly UdpServer to private UdpServer. Also, I cant make Server.Start() static because then EnablePeriodicSaving() does not work (unless I make that static as well).

So, should I read from the config file before UdpServer is created, ie in Server.cs

This way, values such as time before auto save can also be read and set here
The code in UdpServer.cs:

and

Also, here is the ServerConfigReader class:

Finally, should I place the serverConfigreader class in NitroxModel or in NitroxServer?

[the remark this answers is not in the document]
Measurity: Looks good so far. Yes, the config should be read before the `UdpServer` and `Server` are started so that both can be configurated by the same INI file. It would be better to read the config file and pass the config class instance to the `Server` constructor instead of the file path itself. This way you don't have to bother making things static because the IO reading happens before the server constructor. I'll let you decide whether to keep `ServerConfigReader` static or pass it as an instance to the `Server` and `UdpServer` constructors. Both work.

So in short:
- Move the ServerConfigReader.ReadServerConfig(configPath) to `Program` class where the `Server` instance is created.
- Then read out the `ServerConfigReader` class in `Server` (if you have any server options in the INI file that is) and `UdpServer` (get the port from INI file).

Discard what I said about making `UdpServer.Start()` static in the comments above.

FahadAhmad2001:
Should I add/change anything?

NitroxClient/Communication/Abstract/IMultiplayerSessionState.cs
```diff
@@ -7,6 +7,7 @@ public interface IMultiplayerSessionState
     {
         IClient Client { get; }
         string IpAddress { get; }
+        int serverPort { get; }
```
Measurity: Could you make `serverPort` pascal case as `ServerPort`? Only for this property. We use pascal case for properties.

NitroxClient/Communication/MultiplayerSession/MultiplayerSessionManager.cs
```diff
@@ -28,16 +28,18 @@ internal MultiplayerSessionManager(IClient client, IMultiplayerSessionConnection
 
         public IClient Client { get; }
         public string IpAddress { get; private set; }
+        public int serverPort{ get; private set; }
```
Measurity: Same case here, also pascal case for `ServerPort`

NitroxClient/MonoBehaviours/Gui/MainMenu/JoinServer.cs
```diff
@@ -153,11 +153,11 @@ private void StartMultiplayerClient()
 
             try
             {
-                multiplayerSession.Connect(ServerIp);
+                multiplayerSession.Connect(ServerIp,serverPort);
             }
             catch (ClientConnectionFailedException)
             {
-                Log.InGame($"Unable to contact the remote server at: {ServerIp}:11000");
+                Log.InGame($"Unable to contact the remote server at: {ServerIp}:{serverPort.ToString()}");
```
Measurity: `ToString()` is not needed when you use interpolated strings (or string.Format).

NitroxServer/Server.cs
```diff
@@ -13,21 +15,22 @@ public class Server
...
+        private ServerConfigReader ServerOptions;
         public static Server Instance;
 
         public void Save()
         {
             worldPersistence.Save(world);
         }
 
-        public Server()
+        public Server(ServerConfigReader ConfigReader)
```
Measurity: parameters for methods/functions are camelCase. `ConfigReader` to `configReader`.

PR #312, opened 2018-10-28
Changed logging to use DI instead of being static everywhere

Measurity:
This change improves maintainability of logging and allows logging to be more easily extended. Log.InGame is now an extension method.

In the future, I want to move game specific `NitroxModel` code into a `NitroxModel.Game` project and remove the UnityEngine/Game references in NitroxModel.


Sunrunner37:
Hmm, dependency injection for logging seems non-standard to me.  Although it makes sense that in different contexts you may want different loggers, it seems like the majority of the time this can be more cleanly accomplished via static instances.  The problem with the approach in the PR is that it requires you to have a logger instance in every class and pass that in through the constructor.  This seems to be very large overhead when the static logger can be easily accessed everywhere.  It also doesn't add much value as logging is not really a conceptual component of a class (ex: a 'Wheel' is conceptually tied to a 'Car' but a 'Logger' is not).  You would also never want to pass in a mock for the logger for unit tests.

Measurity:
Both have their pros and cons. Static logger is easier to code and has better performance but it violates open-closed principle (no extension methods or switching out loggers when Nitrox runs somewhere else). Meaning it's harder to separate concerns and split library and game code. 

Conceptually a logger could be tied to a car in the form of a sensor that reports diagnostics. It's not directly useful for the car owner but it can decrease repair costs. Right now the `Log` class is a global service that any class can access which isn't good design. With DI all important services are (or should be) in de `ClientAutoFaqRegistrar` and every service should be appropriately injected into the constructor. This PR is, to me, a good direction towards that design.

I agree that it isn't very consistent. The only reason I didn't use DI in other projects is that they don't have it (like ClientTester). I think this is just a side effect of projects being very different which is a good thing (or they wouldn't be different projects). There's something to be said about NitroxServer project though. It already uses some kind of DI for `PacketProcessors`. I could try to implement DI there but it's gonna be a bit of work for one PR.

Sunrunner37:
I understand what you are saying from a conceptual standpoint - a logger is a class dependency and class dependencies are injected.  However, from a practical standpoint this change has added a massive amount of boilerplate code for only marginal benefit.  I suggest a balance between idealistic view of software development and simplified implementation.  These things always come down to trade-offs and in this instance I'm not seeing significant benefit for the amount of overhead.  Even if the code could become more elegant in one location, developers now need to figure out how to get an instance of the logger, classes become messier with new member variables, and every constructor now needs another field injected. 

On a different note, I don't necessarily agree that extension methods are the correct implementation for game logging.  It has two problems: 1) A developer now needs to remember to import yet another thing to log in-game. 2) This functionality is still in scope and available everywhere - even if the current context didn't support in-game logging.... so the problem it is solving still doesn't really get solved.  I think the better solution is a GameLogger that extends the normal logger.  However, I don't think the added complexity of any of these options warrants the change.  Our current logging implementation is easy, functional, and has not had issues - even if it has some slight static clunkiness at initialization time.

Lastly, I feel like we are going full circle in regards to splitting up the Model and Game dependencies.  This used to be the way the application was developed (when the server took absolutely no dependency on unity).  However, the application became massively simplified by removing this layer of indirection.  The model package is very simple in its current form; it is really just a collection of packets and some common models.  Adding another layer of abstraction may not make sense as there is nothing that needs those models outside of the context of unity.  I think the changes are well intended but we should decide on an exact plan first.

PR #327, opened 2018-12-14
* Switching to LiteNetLib due to lidgren having issues with reliable …

issc20:
…ordering and packet loss


Aaronstar-McClure:
bump

PR #346, opened 2018-12-16
Updated for NitroxServer installation

killzoms:
Included NitroxServer in the installer


NitroxInstallerActions/Patches/NitroxServerSetup.cs
```diff
@@ -0,0 +1,76 @@
...
+using System.Threading.Tasks;
+using System.IO;
+
+namespace InstallerActions.Patches
+{
+    public static class NitroxServerSetup
+    {
+        public static void Init(string dir)
+        {
+            if (File.Exists(Path.Combine(Path.GetFullPath(dir), "..\\..\\SubServer\\Assembly-CSharp.dll")))
```
dm430: For both readability, and maintainability I would suggest creating variables for the Path.Combine results, and constants for the relative paths.

killzoms:
Got some bug fixes chat logging in the server chat log showing when you press Y

NitroxInstallerActions/Patches/NitroxServerSetup.cs
```diff
@@ -0,0 +1,126 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using System.IO;
+
+namespace InstallerActions.Patches
+{
+    public static class NitroxServerSetup
+    {
+        private static string assemblyCSharpDLL = "Assembly-CSharp.dll";
```
Sunrunner37: Would recommend refactoring:

`
private static List<string> assemblies = new List<String> {"Assembly-CSharp.dll", "Newtonsoft.Json.dll", "Assembly-CSharp-firstpass.dll"}
`

You could even do a list directory and copy all DLLs so this doesn't have to be hard coded.

NitroxInstallerActions/Patches/NitroxServerSetup.cs
```diff
@@ -0,0 +1,126 @@
...
+
+        // dir = (Subnautica Directory)\Subnautica_Data\Managed
+        public static void Init(string dir)
+        {
+            string file = Path.Combine(Path.GetFullPath(dir), serverAssemblyCSharpDLL); // (Subnautica Directory)\SubServer\Assembly-CSharp.dll
+            if (File.Exists(file))
+            {
+                File.Delete(file);
+            }
+            File.Copy(dir + assemblyCSharpDLL, file);
```
Sunrunner37: You can refactor all of these out into a single method:

```
private void copyAssembly(String assemblyName, String directory) {
             string file = Path.Combine(Path.GetFullPath(dir), assemblyName); 
             if (File.Exists(file))
             {
                 File.Delete(file);
             }
             File.Copy(dir + assemblyName, file);
}
````

then do 

```
foreach(String assembly in assemblies) {
    String assemblyPath = "..\\..\\SubServer\\" + assembly;
    copyAssembly(path, dir);
}
```

File: NitroxModel/Logger/Log.cs [hunk 4]
```diff
@@ -23,6 +23,12 @@ public static void SetLevel(LogLevel level)
             Write("Log level set to " + Log.level);
         }
         
+        // For chat logging
+        public static void ChatMsg(string name, string msg)
+        {
+            Console.WriteLine("{0}: {1}", name, msg);
```
Sunrunner37: I don't know if we need this.  We already have Log.Write().  You can pass in an already formatted string.

NitroxServer/Serialization/BatchCellsParser.cs
```diff
@@ -44,7 +44,14 @@ public List<EntitySpawnPoint> ParseBatchData(Int3 batchId)
             ParseFile(batchId, "", "creatures", spawnPoints);
             ParseFile(batchId, "", "other", spawnPoints);
 
-            Log.Debug($"Loaded {spawnPoints.Count} entity-spawn-points for batch {batchId}");
+            if (spawnPoints.Count > 0)
```
Sunrunner37: I think the old way was fine.  This was logged as a debug message.  I think we should toggle off debug messages for a release build.

NitroxInstallerActions/Patches/NitroxServerSetup.cs
killzoms: Since only specific files are used for the server I figured it would be better to not clog the server with unneeded files

NitroxInstallerActions/Patches/NitroxServerSetup.cs
killzoms: the full path is already passed through the dir variable btw

NitroxClient/MonoBehaviours/Gui/Chat/PlayerChatEntry.cs
```diff
@@ -42,7 +43,7 @@ public void OnGUI()
 
                 if (Event.current.isKey && Event.current.keyCode == KeyCode.Escape)
                 {
-                    Hide();
+                    Hide(true);
```
Sunrunner37: I'm not too keen on optional flags for methods like this.  I think it would be cleaner to do

```
chatManager.HideLog();
Hide();
```

instead of passing in a parameter.  Mainly because this is the only place that passes in the parameter.

NitroxModel/Logger/Log.cs
```diff
@@ -34,7 +34,7 @@ public static void InGame(string msg)
             Info(msg);
         }
 
-        private static void Write(string fmt, params object[] arg)
+        public static void Write(string fmt, params object[] arg)
```
Sunrunner37: Sorry this was my fault... we don't want to use write.  We want to use Log.Info().  Info means information log level which is just general info for someone reading the logs.

Let's set this back to private.

dexter74:
Not work to compile.

dexter74:
Compile rework

PR #374, opened 2018-12-17
Adding functions to silently log missing files.

Hacksawfred3232:
The commit message i written for the commit on my fork of the git repo is slightly off. So i will try again here:

- Added 2 new functions. CheckForIndexAndSilenceWarnings() which checks for index.txt using path.txt and LogMissingFileName() which is a helper function for logging to missing.txt
- Added calls for these functions into ParseFile(). These calls only trigger when "WarnSilence" is set to true and a file that was attempted to be parsed does not exist. 

Main reason why i did this: On my run through of compiling the mod and testing if it works, i noticed that the console fills up with errors, but most or all of the mobs spawned correctly. This is just for cleaning up for that case. But this also should narrow down if someone has a corrupted path for the mod, as if the path was not set correctly when compiling the mod yet it still compiled correctly, the check should error out when checking for index.txt
By the way, i know that using a file to "set a variable" is a shortcut, but i have not delve into the source code more thoroughly. Planning to do that at a later time. This patch's main purpose is to silence unnecessary warnings while also still leaving some evidence of what went wrong.

Side note: Don't be alarmed if it shows that i replaced everything in the new version of the file, i uploaded the file manually instead of doing the whole "git commit" and "git push" routine.

### DIFF
--- a/NitroxServer/Serialization/BatchCellsParser.cs
+++ b/NitroxServer/Serialization/BatchCellsParser.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
@@ -31,7 +31,6 @@ namespace NitroxServer.Serialization
             /**
            * Implemented check to see if index.txt in our Build18 folder exists.
            * If it does, warnings for this check are then silenced on the console, and instead, missing files are dumped to a missing.txt file
-           * For this purpose and as a precautionary measure, path.txt is used.
            * Warnings can be turned back on by the presence of a nosilence.txt file. Could be later implemented as an option. */
             if (WarnSilence)
             {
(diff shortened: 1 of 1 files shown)


Hacksawfred3232:
Never mind, i just realised that "path.txt" is used for the Epic Games launcher as said here: #329, fixing the "CheckForIndexAndSilenceWarnings" function to use the "steampath finder" as well.

NitroxServer/Serialization/BatchCellsParser.cs
```diff
@@ -1,128 +1,207 @@
...
+                WarnSilence = true;
+            }
+            else
+            {
+                Log.Info("Not silencing warnings about missing files!");
+                DoNotAttempt = true;
+            }
+            return;
+        }
+        public void LogMissingFileName(string fileName)
```
Sunrunner37: Missing some of these files is perfectly normal as not every batch has corresponding files.  To take care of the over logging we need to bundle a release version that has debug logging disabled.  Then no one would be confused by the error messages.

Sunrunner37: Yeah I think we should close without merge as a recent commit has reduced this logging.  Thanks for submitting though - we look forward to future PRs :)

PR #401, opened 2018-12-19
Server displays its IP (for convenience)

FahadAhmad2001:
- The server now shows its IP (LAN IP/ Hamachi/ Public IP) for convenience.


NitroxServer/Server.cs
```diff
@@ -44,6 +46,51 @@ public Server(ServerConfigReader configReader)
...
         public void Start()
         {
+            NetworkInterface[] allInterfaces = NetworkInterface.GetAllNetworkInterfaces();
+            foreach(NetworkInterface eachInterface in allInterfaces)
+            {
+                if (eachInterface.Name == "Hamachi")
+                {
+                    string hamachiAddresses = "";
+                    int count = 0; //Multiple Hamachi IP's , IPv4 and IPv6
+                    foreach(IPAddressInformation hamachiIP in eachInterface.GetIPProperties().UnicastAddresses)
```
Measurity: You could make this foreach almost a one liner with LINQ to make it easier to read this method. Something like:

```
var ips = eachInterface.GetIPProperties().UnicastAddresses
    .SelectMany(address => address)
    .Where(address => address.Address.ToString().Contains("fe80::"));
Log.Info("If using Hamachi, use one of the following IPs: " + string.Join(" or ", ips))
```

It's of the top of my head so it might not compile directly or work.

NitroxServer/Server.cs
```diff
@@ -44,6 +46,51 @@ public Server(ServerConfigReader configReader)
...
+                        count += 1;
+                    }
+                    Log.Info("If using Hamachi, use the following IP: " + hamachiAddresses);
+                }
+                if (!(eachInterface.GetIPProperties().GatewayAddresses.Count==0)) // To avoid VMWare / other virtual interfaces
+                {
+                    foreach (IPAddressInformation eachIP in eachInterface.GetIPProperties().UnicastAddresses)
+                    {
+                        char splitter = '.';
+                        string[] splitIpParts = eachIP.Address.ToString().Split(splitter);
```
Measurity: You're only using splitter here so you can also directly put '.' in here.

NitroxServer/Server.cs
```diff
@@ -44,6 +46,51 @@ public Server(ServerConfigReader configReader)
...
+                        int secondPart = 0;
+                        if (splitIpParts.Length>1)
+                        {
+                            int.TryParse(splitIpParts[1], out secondPart);
+                        }
+                        if (splitIpParts[0] == "10" || (splitIpParts[0] == "192" && splitIpParts[1] == "168") || (splitIpParts[0] == "172" && (secondPart > 15 && secondPart < 32))) //To get if IP is private
```
Measurity: Since this is the crux of this foreach maybe put it higher in this foreach and use `continue;` to skip if it doesn't meet these checks. Early exit > nested ifs.

FahadAhmad2001: As in the if statement, it requires an integer (secondPart), which is parsed above the if statement. As this integer is only used once, do I just remove the parsing code above the if statement and parse the integer in the actual if statement instead?

Measurity: Hmm.. I tried cleaning it up but this "determine private ip address" doesn't seem to be much smaller than this. You can ignore this conversation.

NitroxServer/Server.cs
```diff
@@ -44,11 +47,53 @@ public Server(ServerConfigReader configReader)
 
         public void Start()
         {
+            ListServerIPs();
```
Sunrunner37: I think we should put this into a new class: IpResolver (return List<String>) or IpLogger (print internally).

NitroxServer/Server.cs
```diff
@@ -44,11 +47,53 @@ public Server(ServerConfigReader configReader)
...
             udpServer.Start();
             Log.Info("Nitrox Server Started");
             EnablePeriodicSaving();
         }
+        public void ListServerIPs()
+        {
+            NetworkInterface[] allInterfaces = NetworkInterface.GetAllNetworkInterfaces();
+            foreach (NetworkInterface eachInterface in allInterfaces)
+            {
+                if (eachInterface.Name == "Hamachi")
```
Sunrunner37: Let's break these out into a couple new methods:

```
private printIfHamachi(interface);
private printIfLan(interface);
private printIfExternal();
```

PR #415, opened 2018-12-19
Feature/refactor server configuration

luis-paulo-silva:
This is the standard way in .Net to use configuration and it will be a valuable asset to your project.
Also, this way you always get a default configuration on the build


NitroxServer/Communication/UdpServer.cs
```diff
@@ -26,13 +26,13 @@ public class UdpServer
...
 
-        public UdpServer(PacketHandler packetHandler, PlayerManager playerManager, EntitySimulation entitySimulation, ServerConfigReader ConfigReader)
+        public UdpServer(PacketHandler packetHandler, PlayerManager playerManager, EntitySimulation entitySimulation, ServerConfig ConfigReader)
         {
             this.packetHandler = packetHandler;
             this.playerManager = playerManager;
             this.entitySimulation = entitySimulation;
-            PortNumber = ConfigReader.serverPort;
-            MaxConn = ConfigReader.maxConn;
+            PortNumber = ConfigReader.ServerPort;
```
Measurity: Could you make the `ConfigReader` parameter to `configReader`. The code style was wrong before your commits but since you're touching it.. :)

NitroxServer/ConfigParser/ServerConfig.cs
```diff
@@ -0,0 +1,45 @@
...
+            }
+        }
+
+        private int? _maxConnections = null;
+        public int MaxConnections
+        {
+            get
+            {
+                int configValue;
+                if(_maxConnections == null && Int32.TryParse(ConfigurationManager.AppSettings[MAX_CONNECTIONS_SETTING], out configValue)) //Using csharp 7 this could be done like "out int configValue"
```
Measurity: Can you remove the comment about C# 7? Also for the one in `ServerPort` getter.

PR #417, opened 2018-12-20
Update IpLogger.cs to better handle IPv6 detection

Machuga14:
Update IPLogger.cs to better handle IPv6 and use redundant IP checking service

Some ISP's (like Comcast) allocate IPv4 and IPv6 IP Addresses, which can cause some IP Checking services to inconsistently post different IP address. The original ip address service being used, bot.whatismyipaddress.com was retrieving IPv6 Addresses for Comcast, but checkip.dyndns.org was correctly retrieving the IPv4 IP Address.

As can be seen on the following image, the IP Address retrieved was in IPV6 form, which is unhelpful (and I confirmed manually, Nitrox currently doesn't support IPv6 Routing):
ipv6 on comcast isp with bot whatismyipaddress com

With the correction I made, the correct IPv4 address is being retrieved, by detecting that the first attempt at pulling a correctly formatted IPv4 address actually retrieved an IPv6 address, as seen in action here:
ipv4 with fix

Here's a third IP Checking utility which displays both my public and private IP addresses:
ip address lookup with both ipv4 and ipv6

It may not be a bad idea to replace the entire first IP check and just use the 2nd check, or maybe use an alternate URL, with better HTML-scraping logic to differentiate between IPv4 and IPv6, but I'm just posting this initial modification as a starting point to consider making further changes in the future.

### DIFF
--- a/NitroxServer/IpLogger.cs
+++ b/NitroxServer/IpLogger.cs
@@ -59,6 +59,22 @@ namespace NitroxServer
                     using (WebClient client = new WebClient())
                     {
                         string externalIP = client.DownloadString("http://bot.whatismyipaddress.com"); // from https://stackoverflow.com/questions/3253701/get-public-external-ip-address answer by user_v
+                        if (externalIP.Contains(":"))
+                        {
+                          // If the externalIP we retrieved from bot.whatismyipaddress.com has colons in it, it's an IPV6 address.
+                          // If so, use backup source which more reliably retrieves the IPV4 address. This behavior can be
+                          // observed when connecting via Comcast, which supports allocation of both IPV4 and IPV6 addresses
+                          // simultaneously.
+                          // Basic idea came from: https://stackoverflow.com/a/7838551/1136388
+                          string ipPage = client.DownloadString("http://checkip.dyndns.org");
+                          int startIP = ipPage.IndexOf(": ") + 1;
+                          int endIP = ipPage.IndexOf("<", startIP);
+
+                          if (endIP - startIP > 0 && endIP > 0)
+                          {
+                            externalIP = ipPage.Substring(startIP, endIP - startIP);
+                          }
+                        }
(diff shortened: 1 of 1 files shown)


Machuga14:
That's what I'm on the fence about; I don't know if http://checkip.dyndns.org always gives the IPv4 instead of the IPv6 address; If I knew that it did, I would argue that we replace the first check entirely and just default to the 2nd IP checking service... For my particular network setup, it reliably grants the IPv4 address, but for other people, the 1st check was reliably granting the IPv4 address - granted, probably for users who don't have both IPv4 and IPv6 addresses.

The thing with the redundant check is it only executes if the first check is determined to be IPv6 (because of the colons formatted in the address), so it would only slow down the instantiation of the server a bit if the first check grabbed an IPv6.

I see value in the two checks if http://checkip.dyndns.org will sometimes grab the IPv6 address and the first one could grab an IPv4 address, which is why I ended my comment on the initial PR with a statement that it may be better if we used another IP Checking service, which provides both the IPv4 and IPv6 address in 1 go (if both are available), and then better HTML-scraping logic could be written to intelligently select only the IPv4 address.

Ultimately, I'm happy either way - replace the old, or keep it as it currently is with a redundant check, and I leave it to you to make a judgment call on how it's handled.

Machuga14:
@Fahad, I think that is exactly what we want to do. Later today after I get home from work I will make that change on my local repo and include in in this PR. I know there are many IP checking utilities out there, and most aren't well documented on which kind of IP addresses they will display, but the url you provided is pretty good at self-documenting.

Machuga14:
@Sunrunner37, I figured out how to sync my Fork from upstream, I resolved the conflict on my end, and then I modified the External IP Address discovery to use the other IP Service mentioned by @FahadAhmad2001. I think at this point it's ready to go; I've tested the old and new versions on my end, and this is correcting the IPv4 versus IPv6 issues on my ISP.

PR #453, opened 2018-12-22
Current bugs list

DukeVenator:
A list to keep what bugs are currently known to keep the public informed.

This list will also contain fixes as they become known to prevent pointless issues from being generated on the issues page.

### DIFF
--- a/Current Bugs list
+++ b/Current Bugs list
@@ -1,0 +1,23 @@
+A proposed bug list for people to know what bugs are in and to prevent duplicate issues from being posted over and over!
+
+
+    * Exploding fish have a 80%ish percent chance of being stuck in their pods while the other 20% go after you and kill you just fine (terrifying)
+    * Sometimes Habitat storage will duplicate and have two in the same space
+    * Deconstruction appears not to save properly or sync between restarts (Tinnnnnny bit annoying)
+    * Seamoths Damage values need tobe sent to other clients along with if a client has done repairs on the seamoth
+    * Wearables sometimes disappear on relog (Currently using debug to spawn shit back in)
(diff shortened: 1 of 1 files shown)


Current Bugs list
```diff
@@ -0,0 +1,23 @@
...
+    * Infinite o2 bug (Fixed by reequiping tanks)
+    * Nuclear reactors break everything (Really annoying but thats why we backup)
+    * Bio reactors ^ (major problem)
+    * Scanner rooms are not synced and upgrades reset on restart of server (To be expected)
+    * Using grav gun on players can cause issues.... (funny issues)
+    * Prawn suit has same issue as sea moth with repairs not being sent to other players (you might need a sync event for this)
+    * Player animations do what you do (minor as all hell fix this last)
+    * Fish go inside habitats and become stuck on the floor ( free fish to eat really don't mind)
+    * Fish spawn numbers are over the top and appear to have no limit or are dupping (Might need to have a limit in place)
+    * stalker teeth do not drop (For now you will just have to enable the console and type "item stalkertooth")
```
ambid17: I have a few to add:

-acid shrooms and coral tubes randomly despawn. They respawn after a restart and then disappear after a bit. The also cannot be put in lockers, they will disappear
-equipped items will seemingly randomly disappear. Mostly the normal oxygen tanks. Definitely on relog as stated, but not always.
-My buddy has a server on a separate pc. When he connects on a pc on the same network, it has faster water decrease

ambid17:
There are tons missing, but i think this is a really good effort

Current Bugs list
lewis2470: Cyclops will launch players standing inside of them out of the Cyclops and into the air unless they sit in a seat

PR #464, opened 2018-12-22
Fixes domainsupport

Jannify:


NitroxClient/MonoBehaviours/Gui/MainMenu/MainMenuMultiplayerPanel.cs
```diff
@@ -112,20 +117,36 @@ public void OpenJoinServerMenu(string serverIp)
 
             joinServerGameObject = new GameObject();
             JoinServer joinServerComponent = joinServerGameObject.AddComponent<JoinServer>();
-            string seperator = ":";
-            if (serverIp.Contains(seperator))
+
+            if (Regex.IsMatch(serverIp, "^[0-9.:]+$"))
```
Sunrunner37: Can we break these into methods?  Like resolveIpv4 and resolveHostName.

Agent77326: ipv6 is not covered with the regex

Jannify: The Lidgren methods we use doesn't support IPv6 too.

PR #517, opened 2018-12-25
QModManager Integration

Fabianflol:
### This is a Headline
I thought it would be useful to integrate the QModManager into Nitrox, because it is used by most Subnautica mods. In the current state it schould only be used for client sided mods, due to the lack of Server-Synchronisation. Modded inventories for example can't be synchronized.

### This is another Headline
QModManager uses a different version of harmony. To fix this problem I had to slightly modify the harmony used by Nitrox.
nitroxharmony
I changed the namespace and everything else to NitroxHarmony (The modified version is in the ZIP-file below).

To finish it up the following dll's are added to Subnautica\Subnautica_Data\Managed\:
[NitroxQModManagerDLLs.zip](https://github.com/SubnauticaNitrox/Nitrox/files/2709106/NitroxQModManagerDLLs.zip)

### Here are some screenshots with my patch:
Main Menu
sub0
MoreQuickSlots loaded fine
sub1
SubnauticaMap
sub2
Options Menu
sub3


Measurity:
Does this mean that Nitrox now hosts QMods? I don't think that's the right way to do it because QMods isn't by us and Nitrox instead should be made compatible with QMods.

I know it's more work and that's why it hasn't happened yet.

killzoms:
Wouldn't it be better to run a check for QMods existence rather than having QMods included with Nitrox? Always including QMods in the build even when QMods is not wanted by the host/player would be problematic

Fabianflol:
That's the way QModManager works, it looks for mods in a folder and if there are some it is calling their patch method. Actually quite simple. If the user don't place mods in the QMods folder nothing happens after Nitrox is loaded.

Sunrunner37:
I think I like the intent of the PR but have some reservations about the implementation.  Would it be better to have our own built-in loader that can optionally parse open source QMOD files?  That way we can support nitrox-specific enhancements on top of the loading (such as custom packet processors) while still supporting qmod format.

Fabianflol:
OK I will try to implement the specific loader in the next few days.

killzoms:
if we do go down the QMods route we would likely want to make sure there are no harmony patching issues...

Measurity:
@Sunrunner37 You might have your own reservations to this.

I have some news that will probably be of major influence to this PR. QModManager will soon have the ability to make mods dependent on one another. With it, we can allow others to make mods for Nitrox. Personally, I think this is the way to go. As soon as 'dev 1.4' branch @ QModManager is merged into master we should look into it.

Afaik the new version of QModManager will allow mods to load their own versions of specific DLLs so we might be able to still use our custom Harmony version.

Right now I recommend holding of additional work on this PR or try making Nitrox compatible as a QMod with the 'dev 1.4' branch.

AlexejheroYTB:
Please dont merge unti i take a look at this - as the maintainer of qmods i believe i have a saying in this

Fabianflol:
> You don't need to change the namespace name of the custom Nitrox Harmony dll, just the filename is enough. Also, I don't think reimplementing QModManager's loading system is the best idea for various reasons.

Just changing the filename didn't work for me. I'm new to C# so I don't know how to do it in another way.

I thought QMods is a finished piece of code, but it isn't. So I'll try the other way around, let QMods load Nitrox.

ahk1221:
That is weird, it has worked for us while we were updating QMods' Harmony to the latest version.

Sunrunner37:
We had a discussion about this in the discord.  We think the correct long term solution can be implemented on QMM side as a MP compatibility flag (tri-state: sp, mp, sp-mp).  Nitrox will remain a non-qmod but will expose interfaces that other mods can interact with.  Nitrox will assume that someone else loaded the mods (generally qmod).  Nitrox developers will need to work closely with QMM to remove outstanding bugs with patching and uninstalling.  Nitrox will investigate migrating their hook into another location.

Issue #498, opened 2018-12-24
Bulkhead disappearing


RylenWolf:
Hello, this one is relatively simple. Server is not saving bulkhead. When you build one it works until the server restarts. Then it's completely missing.

PR #560, opened 2019-01-01
Bulkhead UpdatePlacement patch

quimnuss:
Mimicked the BaseAddModuleGhost Nitrox patch to use the MutiplayerBuilder instead of the Player main subroot for the UpdatePlacement.


quimnuss:
Fixes #498

Issue #545, opened 2018-12-29
Acid Mushroom GUID Lookup Exception


Foltik:
### Description
When a single client is connected to the server, acid mushrooms can be added and removed from lockers properly. When a second client joins, however, all acid mushrooms are removed from all storage, and adding an acid mushroom to a storage results in it disappearing.

### Steps to Reproduce
1. Start a server
2. Have Player 1 join the server and create a habitat with a locker inside
3. Have Player 1 insert an acid mushroom into the locker
4. Have Player 2 join

The acid mushroom will disappear from the locker, and any mushrooms that are added after will also disappear.

### Reason for Removal from Storage
I added some logging in `NitroxServer/ItemContainer{Add,Remove}PacketProcessor.cs`, and get the following when acid mushrooms disappear:
```
[Nitrox] I: REMOVE PACKET: 0dcc1705-04ea-4f01-b8f3-9718bf6778e5 FROM: Gray
[Nitrox] I: ADD PACKET: [ItemData ContainerGuid: 5ddfbec7-4fa0-4dd5-9fd0-f6ce86d8f3ea Guid: 0dcc1705-04ea-4f01-b8f3-9718bf6778e5] FROM: Gray
[Nitrox] I: REMOVE PACKET: 0dcc1705-04ea-4f01-b8f3-9718bf6778e5 FROM: Jack
```
The first player, Gray, removes the acid mushroom from his inventory and adds it to a locker. Immediately after, the second player, Jack, sends the server a remove packet. This happens without input from the player.

### Cause
When the second client receives the add packet, the following exception is triggered:
```
Error: GetTransformById could not find '8ddb83df-7684-446c-b612-27d7ecf2a719' among 1836 ids

UnityEngine.DebugLogHandler:Internal_LogException(Exception, Object)
UnityEngine.DebugLogHandler:LogException(Exception, Object)
UnityEngine.Logger:LogException(Exception, Object)
UnityEngine.Debug:LogException(Exception, Object)
UWE.Utils:LogReport(String, Object)
UWE.Utils:LogReportFormat(Object, String, Object[])
UWE.Utils:LogReportFormat(String, Object[])
ProtobufSerializer:GetTransformById(String)
ProtobufSerializer:DeserializeIntoGameObject_Patch1(Object, Stream, GameObjectData, UniqueIdentifier, Boolean, Boolean, Transform, Int32)
<DeserializeObjectsAsync>c__Iterator3:MoveNext()
[...]

PR #561, opened 2019-01-02
Got working the storage in the Escape Pod.

CtrlAltDavid01:
I could only test it by relogging and restarting the server, I could not try it with 2 players.

Edit: The storage box is now starting empty and I didn't managed to figure out how to put the starting items into it on world gen.

PS: I'm new to this code and I don't fully understand it, sorry if I broke something!

### DIFF
--- a/NitroxClient/Communication/Packets/Processors/BroadcastEscapePodsProcessor.cs
+++ b/NitroxClient/Communication/Packets/Processors/BroadcastEscapePodsProcessor.cs
@@ -33,7 +33,7 @@ namespace NitroxClient.Communication.Packets.Processors
                 AssignPlayerToEscapePod(packet);
             }
 
-            Player.main.StartCoroutine(SyncEscapePodGuids(packet));
+            SyncEscapePodGuids(packet);
         }
 
         private void AssignPlayerToEscapePod(BroadcastEscapePods packet)
@@ -68,10 +68,8 @@ namespace NitroxClient.Communication.Packets.Processors
 
         // Done in a coroutine because we want to wait until the escape pod has fully loaded
         // at the beginning of the game.
-        private IEnumerator SyncEscapePodGuids(BroadcastEscapePods packet)
+        private void SyncEscapePodGuids(BroadcastEscapePods packet)
         {
-            yield return new WaitUntil(() => EscapePod.main.gameObject);
-
             foreach (EscapePodModel model in packet.EscapePods)
             {
                 if (!escapePodsByGuid.ContainsKey(model.Guid))
@@ -99,6 +97,7 @@ namespace NitroxClient.Communication.Packets.Processors
             escapePod.transform.position = model.Location;
 
             StorageContainer storageContainer = escapePod.RequireComponentInChildren<StorageContainer>();
+            storageContainer.container.Clear();
             GuidHelper.SetNewGuid(storageContainer.gameObject, model.StorageContainerGuid);
 
             MedicalCabinet medicalCabinet = escapePod.RequireComponentInChildren<MedicalCabinet>();
--- a/NitroxModel/DataStructures/GameLogic/EscapePodModel.cs
+++ b/NitroxModel/DataStructures/GameLogic/EscapePodModel.cs
@@ -10,33 +10,18 @@ namespace NitroxModel.DataStructures.GameLogic
     public class EscapePodModel
     {
         [ProtoMember(1)]
-        public List<ushort> SerializedAssignedPlayers
-        {
-            get
-            {
-                lock (AssignedPlayers)
-                {
-                    return AssignedPlayers;
-                }
-            }
-            set
-            {
-                AssignedPlayers = value;
-            }
-        }
-        [ProtoMember(2)]
         public string Guid { get; set; }
-        [ProtoMember(3)]
+        [ProtoMember(2)]
         public Vector3 Location { get; set; }
-        [ProtoMember(4)]
+        [ProtoMember(3)]
         public string FabricatorGuid { get; set; }
-        [ProtoMember(5)]
+        [ProtoMember(4)]
         public string MedicalFabricatorGuid { get; set; }
-        [ProtoMember(6)]
+        [ProtoMember(5)]
         public string StorageContainerGuid { get; set; }
-        [ProtoMember(7)]
+        [ProtoMember(6)]
         public string RadioGuid { get; set; }
-        [ProtoIgnore]
+        [ProtoMember(7)]
         public List<ushort> AssignedPlayers { get; set; } = new List<ushort>();
 
         public void InitEscapePodModel(string guid, Vector3 location, string fabricatorGuid, string medicalFabricatorGuid, string storageContainerGuid, string radioGuid)
--- a/NitroxModel/DataStructures/GameLogic/EscapePodModel.cs
+++ b/NitroxModel/DataStructures/GameLogic/EscapePodModel.cs
@@ -1,21 +1,45 @@
 ﻿using System;
 using System.Collections.Generic;
+using ProtoBuf;
 using UnityEngine;
 
 namespace NitroxModel.DataStructures.GameLogic
 {
     [Serializable]
+    [ProtoContract]
     public class EscapePodModel
     {
-        public string Guid { get; }
-        public Vector3 Location { get; }
-        public string FabricatorGuid { get; }
-        public string MedicalFabricatorGuid { get; }
-        public string StorageContainerGuid { get; }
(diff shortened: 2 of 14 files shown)


NitroxClient/Communication/Packets/Processors/BroadcastEscapePodsProcessor.cs
```diff
@@ -68,10 +68,8 @@ private void AssignPlayerToEscapePod(BroadcastEscapePods packet)
 
         // Done in a coroutine because we want to wait until the escape pod has fully loaded
         // at the beginning of the game.
-        private IEnumerator SyncEscapePodGuids(BroadcastEscapePods packet)
+        private void SyncEscapePodGuids(BroadcastEscapePods packet)
         {
-            yield return new WaitUntil(() => EscapePod.main.gameObject);
```
Sunrunner37: Why was this removed?  Do we not want to wait until the gameobject is loaded?

NitroxModel/DataStructures/GameLogic/EscapePodModel.cs
```diff
@@ -1,21 +1,45 @@
...
+                }
+            }
+            set
+            {
+                AssignedPlayers = value;
+            }
+        }
+        [ProtoMember(2)]
+        public string Guid { get; set; }
+        [ProtoMember(3)]
```
Sunrunner37: Let's space these out with returns so that they don't all run together.

NitroxModel/DataStructures/GameLogic/EscapePodModel.cs
```diff
@@ -1,21 +1,45 @@
...
+        public Vector3 Location { get; set; }
+        [ProtoMember(4)]
+        public string FabricatorGuid { get; set; }
+        [ProtoMember(5)]
+        public string MedicalFabricatorGuid { get; set; }
+        [ProtoMember(6)]
+        public string StorageContainerGuid { get; set; }
+        [ProtoMember(7)]
+        public string RadioGuid { get; set; }
+        [ProtoIgnore]
```
Sunrunner37: We don't want to persist the assigned players? We have the ability to have multiple life pods so may need this.

NitroxServer/GameLogic/EscapePodManager.cs
```diff
@@ -1,22 +1,72 @@
...
+using System.Collections.Generic;
+using ProtoBufNet;
 using NitroxModel.DataStructures.GameLogic;
+using NitroxServer.GameLogic.Items;
 using UnityEngine;
+using System.Linq;
 
 namespace NitroxServer.GameLogic
 {
+    [ProtoContract]
```
Sunrunner37: I think we'll want to put the serialized data into a new model.  This will allow us to decouple the persistence layer and not litter the Manager with serialization attributes.

NitroxServer/GameLogic/EscapePodManager.cs
```diff
@@ -47,18 +97,33 @@ public EscapePodModel[] GetEscapePods()
...
+            const string FILTERED_WATER_DATA = "AggBhgEIABAAGAAiCFVudGFnZ2VkMiRkNDdmM2FiZi1kNzMxLTQ4MjktYWY0Yy1kNmNhMzhhODFiZjU6JDIyYjBjZTA4LTYxYzktNDQ0Mi1hODNkLWJhN2ZiOTlmMjZiMEIkM2Q0NjdiMDEtODExNC00OTExLWJmODgtYWVlMmI3M2JlMGE4SAFQAAIIBBkKFVVuaXR5RW5naW5lLlRyYW5zZm9ybRABOAoPDQAAAAAVAAAAAB0AAAAAEhQNAAAAABUAAAAAHQAAAAAlAACAPxoPDQAAgD8VAACAPx0AAIA/DgoKUGlja3VwYWJsZRABE8I+EAgAEAAYACAAKAEwATgAQAMUChBMYXJnZVdvcmxkRW50aXR5EAECEAALCgdFYXRhYmxlEAEFDQAAAAA=";
+            const string FLARE_DATA = "AggBhgEIABAAGAgiCFVudGFnZ2VkMiQ1YzE4ODExNS03NWZhLTQ2NTMtYWYxMi02ZGYzMjFhM2E0NDk6JGY1YjZlYmVkLTlhZDktNGU0Ni1iMGI0LTBmOGQ2NDY3ZTA5MUIkM2Q0NjdiMDEtODExNC00OTExLWJmODgtYWVlMmI3M2JlMGE4SAFQAAIIBBkKFVVuaXR5RW5naW5lLlRyYW5zZm9ybRABOAoPDQAAAAAVAAAAAB0AAAAAEhQNAAAAABUAAAAAHQAAAAAlAACAPxoPDQAAgD8VAACAPx0AAIA/DgoKUGlja3VwYWJsZRABE8I+EAgAEAAYACAAKAEwATgAQAMJCgVGbGFyZRABDg0AAOFEEAAYACUAAAAAFAoQTGFyZ2VXb3JsZEVudGl0eRABAhAA";
```
Sunrunner37: Instead of putting items in the escape pod, why don't we put items into the players inventory when they first join the server.  This would allow everyone to get the starting items.  If you do it that way, I think you might be able to run the routine in client side code (so you don't have crazy stuff like this here)

NitroxServer/Serialization/World/PersistedWorldData.cs
```diff
@@ -43,6 +43,9 @@ public class PersistedWorldData
         [ProtoMember(9)]
         public GameData GameData { get; set; }
 
+        [ProtoMember(10)]
+        public EscapePodManager EscapePodManager { get; set; }
```
Sunrunner37: As stated above, let's create a new model.  EscapePodData.

NitroxModel/DataStructures/GameLogic/EscapePodModel.cs
CtrlAltDavid01: Sorry, never have worked with this.

NitroxServer/GameLogic/EscapePodManager.cs
CtrlAltDavid01: Great idea, didn't think of that :D

NitroxClient/Communication/Packets/Processors/BroadcastEscapePodsProcessor.cs
CtrlAltDavid01: It didn't seem to be needed and I removed it because it only let the code continue after the InitialPlayerSync so the items could not be placed in it. I'll try to restore it and see if it works that way.

CtrlAltDavid01:
I made the changes, but "yield return new WaitUntil(() => EscapePod.main.gameObject);" can not be put back, because it delays the time too much and the items can't be placed inside the container and I have never noticed anything it would break.

Sunrunner37:
For the future, we may want to put some of the escape pod stuff inside the initial sync.  (Though we still need to tell other players if another escape pod spawns because the code does support multiples).

CtrlAltDavid01:
I'll work on it tomarrow, I'll put some of the escape pod stuff inside the initial sync as you mentioned and make it if a new pod is created every player get a packet about it.

Sunrunner37:
Btw, you've done a great job.  Looks like you had a smooth ramp up!

bannmann:
@PurpleTentacle81 and I tested commit edc0ed02e84ecbf16f176cc86f9256e1665ced25. For each test run, we started the server from scratch (no `save.nitrox`).

We can confirm that putting an item into escape pod storage is synchronized between players. It also works when the other player is offline and joins after the item was stored. The player that stored the item can even go offline before the other joins.

However, we tested four kinds of items, and they behave differently: fishes (peeper, boomerang, airsack, garryfish), metal salvage and copper work fine, but acid mushrooms never appear for the other player. Maybe there are more items that don't work.

Another problem is that removing an item from storage is not synchronized at all. For the other player, the item is still in storage and can be removed again. It seems like the item was duplicated.

CtrlAltDavid01:
Acid mushrooms are a different bug, if you pick up one of the and relog it disappears from your inventory, I think it's not synced at all, so the problem is not in this code.

quimnuss:
#545 to be precise

CtrlAltDavid01:
I just fixed the item removal so anyone can test it if they want to (It was a dumb mistake), still working on the other stuff.

bannmann:
We tested 9a9c0afb4b2bcede3d50539ab143d5c6197d5e64 and can confirm that item removal is now working fine.

Thanks, @CtrlAltDavid01! Looking forward to see this on `master` 😃

PR #564, opened 2019-01-03
Fixed the spawning of a new EscapePod when the server is restarted  and a new player joins.

CtrlAltDavid01:


NitroxPower:
Thank you for all the fixes Dave keep going :+1:

PR #566, opened 2019-01-04
Fix for exception when throwing item

ASimpson87:
fix catch exception when throwing an item ( key already exists in dict) no need to re add.

Needs to be tested across clients.


ASimpson87:
Are there any tips for testing across clients ? Currently i have no way of doing this. I'm not entirely sure if it is replicated without the add.

PR #567, opened 2019-01-04
Added Save Interval to app.config for later use with session settings

ASimpson87:
Added Save Interval to app.config for later use with session settings also removed non required add on existing key in dict when throwing item.

### DIFF
--- a/NitroxServer/GameLogic/Entities/EntityData.cs
+++ b/NitroxServer/GameLogic/Entities/EntityData.cs
@@ -37,7 +37,7 @@ namespace NitroxServer.GameLogic.Entities
         private Dictionary<string, Entity> entitiesByGuid = new Dictionary<string, Entity>();
 
         [ProtoIgnore]
-        private Dictionary<string, Entity> globalRootEntitiesByGuid = new Dictionary<string, Entity>();
+        public Dictionary<string, Entity> GlobalRootEntitiesByGuid = new Dictionary<string, Entity>();
 
         public void AddEntities(IEnumerable<Entity> entities)
         {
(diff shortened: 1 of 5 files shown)


NitroxServer/GameLogic/Entities/EntityData.cs
```diff
@@ -67,7 +67,10 @@ public void AddEntity(Entity entity)
             {
                 lock (globalRootEntitiesByGuid)
                 {
-                    globalRootEntitiesByGuid.Add(entity.Guid, entity);
+                    if (!globalRootEntitiesByGuid.ContainsKey(entity.Guid))
```
Sunrunner37: Although this is technically correct (and nice to have safety) I think it is masking a problem.  We shouldn't be trying to add an item to the global root if it is already there.  This probably points to another problem (maybe we aren't correctly removing the item from this dict when we pick up items).

ASimpson87: Ok no worries , I'll take a quick look at how its being handled and see if i can change it to be removed on pickup.

NitroxServer/GameLogic/Entities/EntityData.cs
```diff
@@ -37,7 +37,7 @@ public class EntityData
         private Dictionary<string, Entity> entitiesByGuid = new Dictionary<string, Entity>();
 
         [ProtoIgnore]
-        private Dictionary<string, Entity> globalRootEntitiesByGuid = new Dictionary<string, Entity>();
+        public Dictionary<string, Entity> GlobalRootEntitiesByGuid = new Dictionary<string, Entity>();
```
Measurity: Instead of exposing the Dictionary I recommend creating a remove method `public bool Remove(GUID guid)` on EntityData which will lock the dictionary and try to remove said item returning true if successful.

NitroxServer/GameLogic/Entities/EntityManager.cs
```diff
@@ -71,6 +71,10 @@ public void RegisterNewEntity(Entity entity)
         public void PickUpEntity(string guid)
         {
             entityData.RemoveEntity(guid);
+            if (entityData.GlobalRootEntitiesByGuid.ContainsKey(guid))
```
Measurity: Change this to `entityData.Remove(guid)`.

ASimpson87: Much better thanks @Measurity  , i shall add this in.

ASimpson87:
Cheers guys.🤙Onto the next!

Issue #570, opened 2019-01-04
Outdated mod description.


GuzioMG:
I know, that I'm probably taking you out of job and you are angry at me, that you need to repair it, while you've got more important things to do, but it's a small issue which I noticed and I just wanted to tell about it, because it seems, that you didn't notice that:

-outdated milestones and in README.MD file:
I think, that you should add 2 milestones because they are important changes:
>publishing on NexusMods

and
>installer is available

-Mod descriptions in NexusMods and GitHub are out of sync:
On https://www.nexusmods.com/subnautica/mods/159 it says
>"This mod is still in its early stages of development, and a public, easy to install version has not been released yet. Those who are keen to try can follow the guide on the wiki. This requires compilation of the mod, and injecting it into Subnautica; not suited for the faint-hearted."

, while the installer is already available. Or on GitHub, there is nothing about donating, while there is something about donations on NexusMods.

-The missing paragraph in NexusMods description:
I found a mistake in a description on NexusMods. It says 
>"..for the game.How can I contribute?Simply submit a pull request. Feel free to join us in our discord channel."

I think, that
>"How can I contribute?"

should be in a separated paragraph, shouldn't it? (I know, that this is a mistake, but this post is about small mistakes, that's why I'm posting it).

GuzioMG:
OK, I decided to help with this project a bit, so I forked it (fork is available here: https://github.com/GuzioMG/Nitrox ) and created a pull request (here: https://github.com/SubnauticaNitrox/Nitrox/pull/571 ).

I know, that I probably could do it without forking repository, but I'm new to that more "advanced" elements of Git and GitHub. I was usually only git-cloning repositories and opening issues. This was my first pull request in my life, so please, don't be angry.

Anyway, I'm not closing this issue, because there are still issues in https://www.nexusmods.com/subnautica/mods/159

PR #571, opened 2019-01-04
Small changes in README.MD file.

GuzioMG:
On https://github.com/SubnauticaNitrox/Nitrox/issues/570 I told about some things missing in README.MD file and in NexusMods description. And while I'm not able to edit NexusMods description, I can edit README.MD. And I did it. And you can add it.

I know, that I probably could do it without forking repository (fork is available here: https://github.com/GuzioMG/Nitrox ), but I'm new to that more "advanced" elements of Git and GitHub. I was usually only git-cloning repositories and opening issues. This is my first pull request in my life, so please, don't be angry.


GuzioMG:
I also removed two unnecessary files.
I just removed two useless files, because I can't figure out, why do they actually exist. They are just a Windows shortcut. It is nonsense. Can someone tell me, what the hell are that files doing here? By the way, if they are doing something, please, don't reject this pull request. I'll put them back there.

_config.yml
```diff
@@ -0,0 +1 @@
+theme: jekyll-theme-hacker
```
GuzioMG: This is an accidental edit, but I don't have an idea, how to remove it.

PR #573, opened 2019-01-05
Patched CheckLoseTooth with hard-coded hardness values

quimnuss:
As-is, titanium, scrap, seamoth and prawn suit drop at 15%, live things drop at 5%.

### DIFF
--- a/NitroxPatcher/Patches/Stalker_CheckLoseTooth_Patch.cs
+++ b/NitroxPatcher/Patches/Stalker_CheckLoseTooth_Patch.cs
@@ -1,0 +1,47 @@
+﻿using Harmony;
+using NitroxClient.GameLogic;
+using NitroxClient.GameLogic.Helper;
+using NitroxModel.Core;
+using NitroxModel.DataStructures;
+using NitroxModel.Helper;
+using NitroxModel.Logger;
+using System;
+using System.Reflection;
+using UnityEngine;
+
+namespace NitroxPatcher.Patches
+{
+    public class Stalker_CheckLoseTooth_Patch : NitroxPatch
+    {
+        public static readonly Type TARGET_CLASS = typeof(Stalker);
+        public static readonly MethodInfo TARGET_METHOD = TARGET_CLASS.GetMethod("CheckLoseTooth", BindingFlags.NonPublic | BindingFlags.Instance);
+
+        public static bool Prefix(Stalker __instance, GameObject target)
+        {
+            float hardness = 0f;
+            TechType techType = CraftData.GetTechType(target);
+            LiveMixin liveMixin = target.GetComponent<LiveMixin>();
+            
+            if (techType == TechType.ScrapMetal || techType == TechType.Titanium
+             || techType == TechType.Seamoth || techType == TechType.Exosuit)
+            {
+                hardness = 0.3f; //15% probability
+            }
+            else if(liveMixin != null && liveMixin.IsAlive())
+            {
+                hardness = 0.1f; //5% probability
+            }
+
+            if (UnityEngine.Random.value < hardness && UnityEngine.Random.value < 0.5f)
+            {
+                __instance.ReflectionCall("LoseTooth");
+            }
+            return false;
+        }
+
+        public override void Patch(HarmonyInstance harmony)
+        {
+            PatchPrefix(harmony, TARGET_METHOD);
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


NitroxPatcher/Patches/Stalker_CheckLoseTooth_Patch.cs
```diff
@@ -0,0 +1,42 @@
...
+             || techType == TechType.Seamoth || techType == TechType.Exosuit)
+            {
+                hardness = 0.3f; //15% probability
+            }
+            else if(liveMixin != null && liveMixin.IsAlive())
+            {
+                hardness = 0.1f; //5% probability
+            }
+
+            if (UnityEngine.Random.value < hardness && UnityEngine.Random.value < 0.5f)
```
Sunrunner37: Why do we have `UnityEngine.Random.value < 0.5f`?  For the tooth to drop it would have to be less than the hardnesses above (0.3 & 0.1) So it seems that doesn't do anything.

NitroxPatcher/Patches/Stalker_CheckLoseTooth_Patch.cs
```diff
@@ -0,0 +1,42 @@
...
+            LiveMixin liveMixin = target.GetComponent<LiveMixin>();
+            
+            if (techType == TechType.ScrapMetal || techType == TechType.Titanium
+             || techType == TechType.Seamoth || techType == TechType.Exosuit)
+            {
+                hardness = 0.3f; //15% probability
+            }
+            else if(liveMixin != null && liveMixin.IsAlive())
+            {
+                hardness = 0.1f; //5% probability
```
Sunrunner37: Doesn't unity return a value between 0.0 and 1.0 making anything 0.1 or less a 10% drop chance?  or am I missing something?

NitroxPatcher/Patches/Stalker_CheckLoseTooth_Patch.cs
```diff
@@ -0,0 +1,42 @@
...
+using UnityEngine;
+
+namespace NitroxPatcher.Patches
+{
+    public class Stalker_CheckLoseTooth_Patch : NitroxPatch
+    {
+        public static readonly Type TARGET_CLASS = typeof(Stalker);
+        public static readonly MethodInfo TARGET_METHOD = TARGET_CLASS.GetMethod("CheckLoseTooth", BindingFlags.NonPublic | BindingFlags.Instance);
+
+        public static bool Prefix(Stalker __instance, GameObject target)
```
Sunrunner37: Let's add a block comment of why we are making the change (you can simply include your comment from github)

NitroxPatcher/Patches/Stalker_CheckLoseTooth_Patch.cs
```diff
@@ -0,0 +1,42 @@
...
+        {
+            float hardness = 0f;
```
Sunrunner37: Maybe we skip the naming `hardness` and just say probability or something similar.

NitroxPatcher/Patches/Stalker_CheckLoseTooth_Patch.cs
quimnuss: It does, although it could be expressed in one random, those are two different Randoms. It means that it always drops half the time it would drop. So, statistically it's the same as setting 15% and 5%, but stating clearly that it will never drop for sure. E.g. hardness of 1f will drop 50% of the time.

NitroxPatcher/Patches/Stalker_CheckLoseTooth_Patch.cs
quimnuss: Same as above, since the probability of dropping is always 50% you have to multiply it by 50% to get the final probability. (5% in the case of alive things)

NitroxPatcher/Patches/Stalker_CheckLoseTooth_Patch.cs
quimnuss: Should I merge the two probabilities (50% and h%) or given my explanations above it makes semantic sense to keep them separate? I like to also keep the reference to the properties of the material semantically for future reference. If something is hard, it should make it easier for teeth to drop. If we call it "probability" we remove this semantic reference. Do my arguments convince you? I'm fine either way.

NitroxPatcher/Patches/Stalker_CheckLoseTooth_Patch.cs
Sunrunner37: Ahh, I get it.  Let's express in one random.

PR #574, opened 2019-01-05
Fixes issues running in background, fixes death message, cleanup

Dilicor:
**1. Fixing running server in the background using nohup:**
When executing `nohup mono NitroxServer.exe &`,
```
Unhandled Exception:
System.UnauthorizedAccessException: Access to the path "/home/users/dy1zan/SubServer/[Unknown]" is denied.
  at System.IO.FileStream.ReadData (System.Runtime.InteropServices.SafeHandle safeHandle, System.Byte[] buf, System.Int32 offset, System.Int32 count) [0x00036] in <8f2c484307284b51944a1a13a14c0266>:0
  at System.IO.FileStream.ReadInternal (System.Byte[] dest, System.Int32 offset, System.Int32 count) [0x0002c] in <8f2c484307284b51944a1a13a14c0266>:0
  at System.IO.FileStream.Read (System.Byte[] array, System.Int32 offset, System.Int32 count) [0x000b9] in <8f2c484307284b51944a1a13a14c0266>:0
  at System.IO.StreamReader.ReadBuffer () [0x000bf] in <8f2c484307284b51944a1a13a14c0266>:0
  at System.IO.StreamReader.ReadLine () [0x00027] in <8f2c484307284b51944a1a13a14c0266>:0
  at System.IO.UnexceptionalStreamReader.ReadLine () [0x00000] in <8f2c484307284b51944a1a13a14c0266>:0
  at System.IO.TextReader+SyncTextReader.ReadLine () [0x00000] in <8f2c484307284b51944a1a13a14c0266>:0
  at (wrapper synchronized) System.IO.TextReader+SyncTextReader:ReadLine ()
  at System.Console.ReadLine () [0x0001f] in <8f2c484307284b51944a1a13a14c0266>:0
  at NitroxServer.Program.Main (System.String[] args) [0x00050] in <1ef184afea584e109dc210ace98a6956>:0
```
Fix: only ReadLine when the console is in Interactive mode.

**2. Fixing death message**
It used playerid rather than username.

**3. Making LocalPlayer consistent with RemotePlayer**
I have converted it to use PlayerContext, which can then be passed to player-triggered events. This also cleans up passing player IDs to events.


NitroxServer/Program.cs
```diff
@@ -31,11 +31,15 @@ static void Main(string[] args)
...
-            while (server.IsRunning)
-            {
-                CmdProcessor.ProcessCommand(Console.ReadLine());
+            if (Environment.UserInteractive)
+            {
+                ConsoleCommandProcessor CmdProcessor = NitroxServiceLocator.LocateService<ConsoleCommandProcessor>();
+                while (server.IsRunning)
+                {
+                    CmdProcessor.ProcessCommand(Console.ReadLine());
+                }
```
Sunrunner37: Won't the application immediately close if it isn't stopped by the while loop?

NitroxServer/Program.cs
```diff
@@ -31,11 +31,15 @@ static void Main(string[] args)
                 return;
             }
 
-            ConsoleCommandProcessor CmdProcessor = NitroxServiceLocator.LocateService<ConsoleCommandProcessor>();
-            while (server.IsRunning)
-            {
-                CmdProcessor.ProcessCommand(Console.ReadLine());
+            if (Environment.UserInteractive)
```
Sunrunner37: Maybe we should add a comment about what this is fixing.

NitroxServer/Program.cs
Dilicor: Apparently not, I thought that too. But since it's running in the background with other threads, it works. My guess is the main program must wait for other threads to terminate?

Dilicor: I can now confirm that the thread which executes Listen in NitroxServer/Communication/UdpServer.cs (thread), runs in the foreground and hence ensures the application does not terminate.
https://docs.microsoft.com/en-us/dotnet/api/system.threading.thread?view=netframework-4.7.2#foreground-and-background-threads

I have also added a few more things to this pull request.

NitroxServer/Program.cs
Dilicor: Comment added, also added a comment to UdpServer.cs to avoid confusion that the `Listen` thread is running in the foreground.

Dilicor:
Okay, tested running NitroxServer in the background without the patch using [PM2](http://pm2.keymetrics.io/), and works okay.
The issue only occurs with nohup. Nohup could be commonly used, some could complain about it. I would recommend the patch OR having instructions in the Linux installation guide (e.g using PM2).
I'll leave it to the reviewer to determine what happens from here.

NitroxClient/GameLogic/LocalPlayer.cs
```diff
@@ -35,7 +42,7 @@ public LocalPlayer(IMultiplayerSession multiplayerSession, IPacketSender packetS
 
         public void BroadcastStats(float oxygen, float maxOxygen, float health, float food, float water)
         {
-            PlayerStats playerStats = new PlayerStats(multiplayerSession.Reservation.PlayerId, oxygen, maxOxygen, health, food, water);
+            PlayerStats playerStats = new PlayerStats(PlayerContext, oxygen, maxOxygen, health, food, water);
```
Sunrunner37: I'm not sure that the correct solution is to add player context to all of these packets.  That is going to add a lot of unneeded data transfer.  I believe when a player connects (or we receive the initial sync packet) we can locally store the context for later use.

PR #579, opened 2019-01-05
Fixed cleanup and escape pod.

CtrlAltDavid01:
-When a player quits items no longer disappear.
-Escape pod is now part of the initial sync.
-When a new escape pod spawns players get a packet about it.


NitroxPatcher/Patches/Inventory_OnApplicationQuit_Patch.cs
```diff
@@ -0,0 +1,30 @@
...
+namespace NitroxPatcher.Patches
+{
+    public class Inventory_OnApplicationQuit_Patch : NitroxPatch
+    {
+        public static readonly Type TARGET_CLASS = typeof(Inventory);
+        public static readonly MethodInfo TARGET_METHOD = TARGET_CLASS.GetMethod("OnApplicationQuit", BindingFlags.NonPublic | BindingFlags.Instance);
+
+        public static void Prefix()
+        {
+            //EscapePodManager.STOP_SENDING_PACKETS = true;
```
Sunrunner37: remove

NitroxClient/Communication/Packets/Processors/InitialPlayerSyncProcessor.cs
```diff
@@ -66,6 +69,16 @@ private void SpawnGlobalRootEntities(List<Entity> globalRootEntities)
             entities.Spawn(globalRootEntities);
         }
 
+        private void SetEscapePodInfo(List<EscapePodModel> escapePodsData)
+        {
+            if (escapePodManager.MyEscapePodGuid == null)
```
Sunrunner37: Shouldn't this always be null in the initial sync?

NitroxClient/GameLogic/EscapePodManager.cs
```diff
@@ -1,45 +1,34 @@
...
-            if (myEscapePodGuid == null)
-            {
-                AssignPlayerToEscapePod(packet);
-            }
-
-            SyncEscapePodGuids(packet);
-        }
-
-        private void AssignPlayerToEscapePod(BroadcastEscapePods packet)
+        public void AssignPlayerToEscapePod(List<EscapePodModel> escapePods)
```
Sunrunner37: This method signature is a little odd to me.  `escapePodManager.AssignPlayerToEscapePod(escapePods). ` ideally this would simply take a guid.  Maybe we can have the assigned escape pod on the initial sync playerData so that we can simply pass it in.  This would require you to probably make a change in the server to do:
```String escapePodGuid = escapePodManager.assignPlayerToEscapePod(player, out newlyCreatedPod);```

File: NitroxClient/GameLogic/ItemContainers.cs [hunk 4]
```diff
@@ -68,6 +68,8 @@ public void BroadcastItemRemoval(Pickupable pickupable, Transform ownerTransform
             {
                 ownerGuid = GuidHelper.GetGuid(ownerTransform.transform.parent.gameObject);
             }
+
+            Log.Debug("!! " + ownerTransform.parent.name + " ! " + ownerGuid + " !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!");
```
Sunrunner37: Remove

NitroxClient/GameLogic/EscapePodManager.cs
```diff
@@ -1,77 +1,67 @@
...
 
-        public BroadcastEscapePodsProcessor(IPacketSender packetSender, IMultiplayerSession multiplayerSession)
+        public EscapePodManager(IPacketSender packetSender, IMultiplayerSession multiplayerSession)
         {
             this.packetSender = packetSender;
             this.multiplayerSession = multiplayerSession;
         }
 
-        public override void Process(BroadcastEscapePods packet)
+        public void AssignPlayerToEscapePod(List<EscapePodModel> escapePods, string assignedEscapePodGuid)
```
Sunrunner37: Let's change to `AssignPlayerToEscapePod(EscapePodModel escapePod)` and do the `EscapePodModel escapePod = escapePods.Find(x => x.Guid == assignedEscapePodGuid);` right before calling.

NitroxClient/GameLogic/ItemContainers.cs
```diff
@@ -27,9 +27,8 @@ public ItemContainers(IPacketSender packetSender, LocalPlayer localPlayer)
         public void BroadcastItemAdd(Pickupable pickupable, Transform ownerTransform)
         {
             string ownerGuid = null;
-
             bool isCyclopsLocker = Regex.IsMatch(ownerTransform.gameObject.name, @"Locker0([0-9])StorageRoot$", RegexOptions.IgnoreCase);
-            bool isEscapePodStorage = ownerTransform.parent.name == "EscapePod";
+            bool isEscapePodStorage = ownerTransform.parent.name == "EscapePod" || ownerTransform.parent.name == "EscapePod(Clone)";
```
Sunrunner37: You can do .StartsWith("EscapePod")

CtrlAltDavid01: I tried it, but then the escapePodStorage starts with that as well, so it would break things, but I'll try it again

PR #581, opened 2019-01-05
Fix compatibility for NitroxServer on Linux

DKingCN:
proof:
[https://imgur.com/a/QTvA4DE](https://imgur.com/a/QTvA4DE)

wiki will be added by shalix later.

### DIFF
--- a/NitroxInstallerActions/Patches/NitroxServerSetup.cs
+++ b/NitroxInstallerActions/Patches/NitroxServerSetup.cs
@@ -31,12 +31,12 @@ namespace InstallerActions.Patches
                 CopyFile(tmpFile, dir);
             }
 
-            string file = Path.Combine(Path.GetFullPath(dir), "..\\..\\SubServer\\path.txt");
+            string file = Path.Combine(Path.GetFullPath(dir), "..", "..", "SubServer", "path.txt");
             if (File.Exists(file))
             {
                 File.Delete(file);
             }
-            string tmpDir = dir.Replace("\\Subnautica_Data\\Managed", "");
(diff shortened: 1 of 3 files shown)


DKingCN:
1. Fix the bug in MainMenuMPPanel.cs that did not cover to resolve domain:port, so it would fail to connect a server in that format.

2. Refactor the code that interacts with serverStr. not using string any more. Instead, use ServerInfo with lazyload now.

DKingCN:
I have been in touch with Dilicor about that topic.

I would prefer using `screen` rather than nohup, screen worries about no background compatibility, but I will help that part.

DKingCN:
Hi, ServerInfo looks better now.

DKingCN:
complete Online Server features finished.

Final

Client side and Server side is in this PR

Server Manager code is at [DKingCN/NitroxServerManager](https://github.com/DKingCN/NitroxServerManager) as simple as a single script

Sunrunner37:
A few comments:

- Please break up PRs by using different GIT branches.  That will limit the amount of churn on a PR and act as an isolated unit to review/merge.
- Ensure to action previous PR comments so we can merge.
- Removing the entity log line is just masking the problem.  We should diagnose why this issue is occuring.
- As mentioned on discord, I already have a listing server in the works.  Although I definitely appreciate the contribution.

DKingCN:
1. The mess of PR is caused more like a historical reason. Feature didnt came as a whole part and Fixes are too small to split to another PR. Force push didnt made the PR cleaner, so I will close this one later and sort the update and create a new one.
2. I thought I finished the problem you mentioned in comments. Did I missed some?
3. You are right, masking the log is just making it worse.
4. I have done most of the feature before you could reply on discord, so I had to finish and share it as an option.

PR #585, opened 2019-01-05
Basic Permissions system and Server Game Mode!

killzoms:
Finally got around to adding a basic permissions system and server game mode option!

### DIFF
--- a/NitroxServer/App.config
+++ b/NitroxServer/App.config
@@ -8,5 +8,6 @@
         <add key="DefaultPortNumber" value="11000"/>
         <add key="SaveInterval" value="60000"/>
         <add key="GameMode" value="Survival"/>
+        <add key="DisableConsole" value="true"/>
     </appSettings>
 </configuration>
--- a/NitroxServer/App.config
+++ b/NitroxServer/App.config
@@ -7,5 +7,6 @@
         <add key="MaxConnections" value="100"/>
         <add key="DefaultPortNumber" value="11000"/>
         <add key="SaveInterval" value="60000"/>
+        <add key="GameMode" value="Survival"/>
     </appSettings>
 </configuration>
--- a/NitroxServer/App.config
+++ b/NitroxServer/App.config
@@ -8,6 +8,6 @@
         <add key="DefaultPortNumber" value="11000"/>
         <add key="SaveInterval" value="60000"/>
         <add key="GameMode" value="Survival"/>
-        <add key="DisableConsole" value="true"/>
+        <add key="DisableConsole" value="false"/>
     </appSettings>
 </configuration>
--- a/NitroxServer/Communication/Packets/Processors/ServerCommandProcessor.cs
+++ b/NitroxServer/Communication/Packets/Processors/ServerCommandProcessor.cs
@@ -1,23 +1,34 @@
 ﻿using NitroxModel.Packets;
 using NitroxServer.Communication.Packets.Processors.Abstract;
 using NitroxServer.ConsoleCommands.Processor;
+using NitroxServer.GameLogic.Players;
+using NitroxModel.DataStructures.GameLogic;
 
 namespace NitroxServer.Communication.Packets.Processors
 {
     public class ServerCommandProcessor : AuthenticatedPacketProcessor<ServerCommand>
     {
         private readonly ConsoleCommandProcessor cmdProcessor;
+        private readonly PlayerData playerData;
 
-        public ServerCommandProcessor(ConsoleCommandProcessor cmdProcessor)
+        public ServerCommandProcessor(ConsoleCommandProcessor cmdProcessor, PlayerData playerData)
         {
             this.cmdProcessor = cmdProcessor;
+            this.playerData = playerData;
         }
 
         public override void Process(ServerCommand packet, Player player)
         {
             string msg = string.Join(" ", packet.CmdArgs);
 
-            cmdProcessor.ProcessCommand(msg);
+            if (playerData.Permissions(player.Name) >= Perms.Admin)
+            {
+                cmdProcessor.ProcessCommand(msg);
+            }
+            else
+            {
+                player.SendPacket(new ChatMessage(ChatMessage.SERVER_ID, "You do not have permission for this command!"));
+            }
         }
     }
 }
--- a/NitroxServer/Communication/Packets/Processors/ServerCommandProcessor.cs
+++ b/NitroxServer/Communication/Packets/Processors/ServerCommandProcessor.cs
@@ -20,15 +20,8 @@ namespace NitroxServer.Communication.Packets.Processors
         public override void Process(ServerCommand packet, Player player)
         {
             string msg = string.Join(" ", packet.CmdArgs);
-
-            if (playerData.GetPermissions(player.Name) >= Perms.Admin)
-            {
-                cmdProcessor.ProcessCommand(msg);
-            }
-            else
-            {
-                player.SendPacket(new ChatMessage(ChatMessage.SERVER_ID, "You do not have permission for this command!"));
-            }
+            
+            cmdProcessor.ProcessPlayerCommand(msg, player, playerData.GetPermissions(player.Name));
         }
     }
 }
--- a/NitroxServer/Communication/Packets/Processors/ServerCommandProcessor.cs
+++ b/NitroxServer/Communication/Packets/Processors/ServerCommandProcessor.cs
@@ -21,7 +21,7 @@ namespace NitroxServer.Communication.Packets.Processors
         {
             string msg = string.Join(" ", packet.CmdArgs);
             
-            cmdProcessor.ProcessPlayerCommand(msg, player, playerData.GetPermissions(player.Name));
+            cmdProcessor.ProcessCommand(msg, player, playerData.GetPermissions(player.Name));
         }
     }
 }
--- a/NitroxServer/Communication/Packets/Processors/ServerCommandProcessor.cs
+++ b/NitroxServer/Communication/Packets/Processors/ServerCommandProcessor.cs
@@ -21,7 +21,7 @@ namespace NitroxServer.Communication.Packets.Processors
         {
             string msg = string.Join(" ", packet.CmdArgs);
 
-            if (playerData.Permissions(player.Name) >= Perms.Admin)
+            if (playerData.GetPermissions(player.Name) >= Perms.Admin)
             {
                 cmdProcessor.ProcessCommand(msg);
             }
--- a/NitroxServer/ConsoleCommands/OpCommand.cs
+++ b/NitroxServer/ConsoleCommands/OpCommand.cs
@@ -1,0 +1,43 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using NitroxServer.ConsoleCommands.Abstract;
+using NitroxServer.GameLogic.Players;
+using NitroxModel.DataStructures.GameLogic;
+using NitroxServer.GameLogic;
+
+namespace NitroxServer.ConsoleCommands
+{
+    class OpCommand : Command
+    {
+        private readonly PlayerData playerData;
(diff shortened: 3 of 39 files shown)


NitroxServer/Serialization/ServerConfig.cs
```diff
@@ -41,6 +43,50 @@ public int MaxConnections
...
+                }
+                return _gameMode ?? DEFAULT_GAMEMODE;
+            }
+        }
+
+        GameModeOption ParseGameMode(string stringGameMode)
+        {
+            GameModeOption gameMode = GameModeOption.Survival;
+            stringGameMode = stringGameMode.ToLower(); // Lets be frank people have habits
+            switch (stringGameMode)
```
Sunrunner37: Might be cleaner as a dictionary above:

```
Dictionary<string, GameModeOption> gameModeByConfigKey = new Dictionary<string, GameModeOption> {
  {"survival", GameModeOption.Survival},
  {"creative", GameModeOption.Creative},
  {"hardcore", GameModeOption.Hardcore},
  {"permadeath, GameModeOption.Survival | GameModeOption.Permadeath}
};
```

Then you can remove this method and simply use the map in the `GameMode` getter

NitroxServer/Communication/Packets/Processors/ServerCommandProcessor.cs
```diff
@@ -1,23 +1,34 @@
...
             this.cmdProcessor = cmdProcessor;
+            this.playerData = playerData;
         }
 
         public override void Process(ServerCommand packet, Player player)
         {
             string msg = string.Join(" ", packet.CmdArgs);
 
-            cmdProcessor.ProcessCommand(msg);
+            if (playerData.GetPermissions(player.Name) >= Perms.Admin)
```
Sunrunner37: Maybe this should be handled outside of the actual command?  Commands can have a field like `MinPermissionRequired` and we don't even invoke if they don't meet the requirement.  That way you don't have to copy/paste this packet sending into every command.

NitroxServer/GameLogic/Players/PlayerData.cs
```diff
@@ -194,6 +212,9 @@ public class PersistedPlayerData
             [ProtoMember(6)]
             public string SubRootGuid { get; set; }
 
+            [ProtoMember(7)]
+            public Perms Permissions { get; set; }
```
Sunrunner37: Is this nullable?  Should we assign a default value if the serializer doesn't have one from the file (backwards compatibility)

NitroxServer/ConsoleCommands/OpCommand.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        public override void RunCommand(string[] args)
+        {
+            playerData.UpdatePlayerPermissions(args[0], Perms.Admin);
+        }
+
+        public override bool VerifyArgs(string[] args)
+        {
+            bool playerFound = false;
+            foreach(Player player in playerManager.GetPlayers())
```
Sunrunner37: Maybe add an `Optional<Player> GetPlayerByName()` in player manager to make this cleaner.

NitroxServer/Serialization/ServerConfig.cs
killzoms: used TryGetValue after creating Dictionary

NitroxServer/Communication/Packets/Processors/ServerCommandProcessor.cs
killzoms: checked off the list

NitroxServer/GameLogic/PlayerManager.cs
```diff
@@ -123,6 +123,30 @@ public void PlayerDisconnected(Connection connection)
...
+        public bool TryGetPlayerByName(string playerName, out Player foundPlayer)
+        {
+            lock (assetsByConnection)
+            {
+                foundPlayer = null;
+                foreach (Player player in ConnectedPlayers())
+                {
+                    if (player.Name == playerName)
+                    {
+                        foundPlayer = player;
```
Sunrunner37: minor: return true after this line as an optimization.

NitroxServer/App.config
```diff
@@ -7,5 +7,7 @@
         <add key="MaxConnections" value="100"/>
         <add key="DefaultPortNumber" value="11000"/>
         <add key="SaveInterval" value="60000"/>
+        <add key="GameMode" value="Survival"/>
+        <add key="DisableConsole" value="true"/>
```
Sunrunner37: Maybe default to false while we are still in development.

NitroxServer/ConsoleCommands/Processor/ConsoleCommandProcessor.cs
```diff
@@ -54,6 +59,39 @@ public void ProcessCommand(string msg)
...
+
+            if (cmd.RequiredPermLevel > perms)
+            {
+                player.SendPacket(new ChatMessage(ChatMessage.SERVER_ID, "You do not have permission for this command!"));
+                return;
+            }
+
+            if (cmd.SupportsClientSide)
+            {
+                RunPlayerCommand(cmd, parts, player);
```
Sunrunner37: Do we have the use-cases to support an entirely separate method signature?  I think a command like `/give <item> <player>` would always have player as parameter.  And if it always assumed the executing player you would have to have two different commands in this model. Thoughts?

NitroxServer/ConsoleCommands/SayCommand.cs
```diff
@@ -3,22 +3,28 @@
...
+using NitroxModel.DataStructures.GameLogic;
 
 namespace NitroxServer.ConsoleCommands
 {
     internal class SayCommand : Command
     {
         private readonly PlayerManager playerManager;
 
-        public SayCommand(PlayerManager playerManager) : base("say", Optional<string>.Of("<message>"), "say Even the lowliest of cogs needs to say something SO SAY SOMETHING!", new[] {"broadcast"})
+        public SayCommand(PlayerManager playerManager) : base("say", NitroxModel.DataStructures.GameLogic.Perms.Admin, Optional<string>.Of("<message>"), "say Even the lowliest of cogs needs to say something SO SAY SOMETHING!", new[] {"broadcast"})
```
Sunrunner37: minor: why fully qualified name?

NitroxServer/GameLogic/PlayerManager.cs
```diff
@@ -123,6 +123,30 @@ public void PlayerDisconnected(Connection connection)
             }
         }
 
+
+        /// <summary>
+        /// Trys to return a player by name
+        /// </summary>
+        /// <param name="playerName">The player to find</param>
+        /// <param name="foundPlayer">Will be null if return value is false</param>
+        /// <returns>true if the player was found false otherwise</returns>
```
Sunrunner37: minor: commenting doesn't add value.  Summary is the name of the method, parameters are self explainable, and the usage aligns with c#'s Try* that is used everywhere in the code.

NitroxServer/Serialization/ServerConfig.cs
```diff
@@ -41,6 +55,42 @@ public int MaxConnections
...
+            {
+                if (_gameMode == null && ConfigurationManager.AppSettings[GAMEMODE_SETTING] != null)
+                {
+                    _gameMode = ParseGameMode(ConfigurationManager.AppSettings[GAMEMODE_SETTING]);
+                }
+                return _gameMode ?? DEFAULT_GAMEMODE;
+            }
+        }
+
+        GameModeOption ParseGameMode(string stringGameMode)
```
Sunrunner37: minor: explicit accessor `private` per style.

NitroxServer/ConsoleCommands/SayCommand.cs
```diff
@@ -3,22 +3,28 @@
...
         {
             this.playerManager = playerManager;
+            SupportsClientSide = true;
```
Sunrunner37: Are some not supported client side?

NitroxServer/ConsoleCommands/Processor/ConsoleCommandProcessor.cs
```diff
@@ -66,5 +104,18 @@ private void RunCommand(Command command, string[] parts)
...
         }
+
+        private void RunPlayerCommand(Command command, string[] parts, Player player)
+        {
+            string[] args = parts.Skip(1).ToArray();
+
+            if (command.VerifyArgs(args))
+            {
+                command.RunCommand(args, player);
+                return;
```
Sunrunner37: minor: in this circumstance, i think `else` is more intuitive than `return`

NitroxServer/ConsoleCommands/Processor/ConsoleCommandProcessor.cs
```diff
@@ -54,6 +59,39 @@ public void ProcessCommand(string msg)
...
+            }
+
+            string[] parts = msg.Split()
+                .Where(arg => !string.IsNullOrEmpty(arg))
+                .ToArray();
+
+            Command cmd;
+            if (!commands.TryGetValue(parts[0], out cmd))
+            {
+                return;
```
Sunrunner37: Should we send a packet saying the command is invalid?

NitroxServer/ConsoleCommands/Processor/ConsoleCommandProcessor.cs
killzoms: Though i used the different methods due to sending info back to the player without creating a new method in PlayerManager to try getting the player based on the player name

NitroxServer/ConsoleCommands/SayCommand.cs
killzoms: used the SupportsClientSide to differentiate between commands like exit and say, there isnt a need to tell the client about the exit command so we dont have to differentiate between the Player execution method and the server execution method

killzoms:
should be a new command added with the latest batch of commits "/w <name> <msg>" Everyones favorite whisper

NitroxServer/Player.cs
```diff
@@ -71,7 +72,21 @@ public bool CanSee(Entity entity)
 
         public void SendPacket(Packet packet)
         {
-            connection.SendPacket(packet);            
+            if (Id == ChatMessage.SERVER_ID) // this is probably just a small hack
```
Sunrunner37: Can we put this hack in the caller instead?

NitroxServer/ConsoleCommands/WhisperCommand.cs
```diff
@@ -0,0 +1,50 @@
...
+            }
+            else
+            {
+                if (player.Id != ChatMessage.SERVER_ID)
+                {
+                    player.SendPacket(new ChatMessage(ChatMessage.SERVER_ID, "Player not found!"));
+                }
+                else
+                {
+                    Log.Info("Player not found!");
```
Sunrunner37: We don't want to log an error as this is normal operation (someone running the command on the server)

NitroxServer/ConsoleCommands/WhisperCommand.cs
```diff
@@ -0,0 +1,50 @@
...
+    class WhisperCommand : Command
+    {
+        private readonly PlayerManager playerManager;
+
+        public WhisperCommand(PlayerManager playerManager) : base("w", Perms.Player, Optional<string>.Of("w <PlayerName> <msg>"))
+        {
+            this.playerManager = playerManager;
+        }
+
+        public override void RunCommand(string[] args, Player player)
```
Sunrunner37: Let's pass an Optional<Player> into these and we can do Optional<Player>.empty() if it is the server.  Then we don't have to check for the server id.

NitroxServer/ConsoleCommands/WhisperCommand.cs
killzoms: the reason for the player not found being logged is in the event someone mistypes a players name

PR #586, opened 2019-01-06
Add Discord rich presence for server info and joining

Jannify:


NitroxClient/MonoBehaviours/DiscordRP/DiscordRpc.cs
```diff
@@ -0,0 +1,205 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Runtime.InteropServices;
+using System.Text;
+using AOT;
+
+public class DiscordRpc
```
Measurity: This class isn't in a namespace. I think it's best to move all the Discord specific classes (including AcceptResponse class) into a `NitroxClient.Discord` namespace.

NitroxClient/GameLogic/PlayerManager.cs
```diff
@@ -75,12 +77,22 @@ public void RemovePlayer(ushort playerId)
             {
                 opPlayer.Get().Destroy();
                 playersById.Remove(playerId);
+                DiscordController.Main.UpdateDRPDiving(GetPlayerCount() + 1);
```
Measurity: Shouldn't this be `GetPlayerCount() -1`?

NitroxClient/MonoBehaviours/DiscordRP/AcceptRequest.cs
```diff
@@ -0,0 +1,104 @@
+﻿using System.Collections;
+using NitroxClient.MonoBehaviours.DiscordRP;
+using UnityEngine;
+
+namespace NitroxClient.MonoBehaviours.Gui.MainMenu
+{
+    public class AcceptRequest : MonoBehaviour
```
Measurity: Rename to `DiscordJoinRequestGui`

NitroxClient/MonoBehaviours/DiscordRP/AcceptRequest.cs
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        public void OnGUI()
+        {
+            discordWindowRect = GUILayout.Window(0, discordWindowRect, DoDiscordWindow, "Server join request from Discord");
+        }
+
+        private void SetGUIStyle()
+        {
+            GUI.skin.textArea.fontSize = 18;
```
Measurity: This will overwrite the default IMGUI skin. You should make a copy of this skin and then change it.

```CS
private GUISkin SetGUIStyle()
{
    return GUISkinUtils.RegisterDerivedOnce("menus.discord", s =>
    {
        s.textArea.fontSize = 18;
        s.textArea.richText = false;
        s.textArea.alignment = TextAnchor.MiddleLeft;
        s.textArea.wordWrap = true;
        s.textArea.stretchHeight = true;
        s.textArea.padding = new RectOffset(10, 10, 5, 5);

        s.label.fontSize = 14;
        s.label.alignment = TextAnchor.MiddleRight;
        s.label.stretchHeight = true;
        s.label.fixedWidth = 140;
        s.label.fixedHeight = 140;

        s.button.fontSize = 14;
        s.button.stretchHeight = true;
    }
}
```

NitroxClient/MonoBehaviours/DiscordRP/AcceptRequest.cs
```diff
@@ -0,0 +1,104 @@
...
+            GUI.skin.label.fixedWidth = 140;
+            GUI.skin.label.fixedHeight = 140;
+
+            GUI.skin.button.fontSize = 14;
+            GUI.skin.button.stretchHeight = true;
+        }
+
+        private void DoDiscordWindow(int windowId)
+        {
+            SetGUIStyle();
```
Measurity: Change to:

```CS
GUISkinUtils.RenderWithSkin(SetGUIStyle(), () =>
{
            using (GUILayout.VerticalScope v = new GUILayout.VerticalScope("Box"))
            {
                using (GUILayout.HorizontalScope h = new GUILayout.HorizontalScope())
                {
                    GUILayout.Label(avatar);
                    GUILayout.Label("Player " + Request.username + " wants to join your server.");
                }
                using (GUILayout.HorizontalScope b = new GUILayout.HorizontalScope())
                {
                    if (GUILayout.Button("Accept"))
                    {
                        CloseWindow(1);
                    }

                    if (GUILayout.Button("Deny"))
                    {
                        CloseWindow(0);
                    }
                }
            }
}
```

NitroxClient/MonoBehaviours/DiscordRP/AcceptRequest.cs
```diff
@@ -0,0 +1,104 @@
...
+
+                    if (GUILayout.Button("Deny"))
+                    {
+                        CloseWindow(0);
+                    }
+                }
+            }
+        }
+
+        private void CloseWindow(int respond)
```
Measurity: Right now the request window will stay open forever as far as I've seen. Can you add a close button to the UI and make it auto close when the request token is expired?

You can start a coroutine and just `yield return new WaitForSeconds(x)` where x is the time until the token expires. Or just enter a suffient but short amount.

NitroxClient/MonoBehaviours/DiscordRP/AcceptRequest.cs
```diff
@@ -0,0 +1,104 @@
...
+using UnityEngine;
+
+namespace NitroxClient.MonoBehaviours.Gui.MainMenu
+{
+    public class AcceptRequest : MonoBehaviour
+    {
+        public DiscordController DiscordRp;
+        public DiscordRpc.DiscordUser Request;
+
+        Rect discordWindowRect = new Rect(Screen.width / 2 - 250, 100, 500, 250);
```
Measurity: This window is in the center of the screen but this means that anyone can annoy you very easily by spamming invites. I think it's better to move the request window to a corner of the screen so that it is less distracting.

NitroxClient/MonoBehaviours/DiscordRP/DiscordController.cs
```diff
@@ -0,0 +1,196 @@
...
+            handlers.requestCallback += RequestCallback;
+            DiscordRpc.Initialize(applicationId, ref handlers, true, optionalSteamId);
+
+            ShowingWindow = false;
+        }
+
+        void OnDisable()
+        {
+            Log.Info("Discord: shutdown");
+            DiscordRpc.Shutdown();
```
Measurity: Unregister the `handlers` callbacks here.

NitroxInstaller/Product.wxs
```diff
@@ -100,6 +103,14 @@
       </Component>
     </ComponentGroup>
   </Fragment>
+  
+  <Fragment>
+    <ComponentGroup Id="SUBNAUTICAMONO_files" Directory="SUBNAUTICAMONO">
+      <Component Id="discordRPC.dll" Guid="*">
+        <File Id="discordRPC.dll" Name="discord-rpc.dll" Source="$(var.NitroxClient_TargetDir)MonoBehaviours/DiscordRP/discord-rpc.dll" />
```
Measurity: This doesn't seem like a good location for the DLL (path is along side .cs files). Also I prefer not to put .dll files on a git repo directly but if it has to happen then put the DLL in the resources. 

We should discuss the best solution on Discord.

NitroxClient/MonoBehaviours/DiscordRP/DiscordController.cs
```diff
@@ -0,0 +1,196 @@
...
+                if (t.name == name)
+                {
+                    return t.gameObject;
+                }
+            }
+
+            return null;
+        }
+
+        private string CheckIP(string ipPort)
```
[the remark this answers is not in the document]
Sunrunner37: Also, these web request calls tend to fail with stricter fire wall rules (like on windows server).  We should make to to appropriately handle them not being available. (same with fetching avatar pictures above.)

NitroxClient/GameLogic/PlayerManager.cs
```diff
@@ -75,12 +77,22 @@ public void RemovePlayer(ushort playerId)
...
 
         private GameObject CloneLocalPlayerBodyPrototype()
         {
             return Object.Instantiate(localPlayer.BodyPrototype);
-        }        
+        }
+
+        public int GetPlayerCount()
+        {
+            if (playersById == null)
```
Sunrunner37: Do we have to check for null here?  It looks like the object is always initialized above.

NitroxClient/GameLogic/PlayerManager.cs
Jannify: No, the "+1" is for the player himself because "GetPlayerCount()" just return the "multiplayer-player".

NitroxClient/MonoBehaviours/DiscordRP/DiscordController.cs
Jannify: I have this code from the official discord example, so I don't think I have to do this because it completely shuts down.

NitroxInstaller/Product.wxs
Jannify: No this path is in the output dir. VS copies the file on compiling there. So the path is"..\Nitrox\bin\Debug\MonoBehaviours\DiscordRP".
But yeah I will move it to the resources dir.

NitroxClient/MonoBehaviours/DiscordRP/DiscordController.cs
```diff
@@ -0,0 +1,196 @@
...
+        {
+            string ip = ipPort.Split(':')[0];
+            string port = ipPort.Split(':')[1];
+
+            if (ip == "127.0.0.1")
+            {
+                WebRequest req = WebRequest.Create("http://checkip.dyndns.org");
+                WebResponse resp = req.GetResponse();
+                StreamReader sr = new StreamReader(resp.GetResponseStream());
```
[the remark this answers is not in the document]
Jannify: WebClient fails (on my pc) because of certificate issues with mono. We have to figure out something we can rely on.
For now I will use my way in the IpHelper and we should rework it later with a system what not ask a website.

NitroxClient/MonoBehaviours/DiscordRP/DiscordRpc.cs
Jannify: Yeah was there. Gets lost after update because I replaced it.

Sunrunner37:
Let us know when this is ready for re-review

NitroxClient/MonoBehaviours/DiscordRP/DiscordController.cs
cher123: I suggest that every fucking one of you leave or just continue this account before the FBI contacts you, because it's only a matter of time before they will. You fucked with the wrong one this time!

Jannify:
unknown

PR #587, opened 2019-01-06
Adding players starting items when they first join.

CtrlAltDavid01:


NitroxServer/GameLogic/Players/PlayerData.cs
```diff
@@ -39,6 +39,9 @@ public class PlayerData
         [ProtoMember(3)]
         public ushort currentPlayerId = 0;
 
+        [ProtoIgnore]
+        public bool currentPlayerFirstConnecting = false;
```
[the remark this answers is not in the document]
CtrlAltDavid01: You are right, didn't think of players joining at once

NitroxServer/GameLogic/PlayerManager.cs
```diff
@@ -70,14 +71,16 @@ public List<Player> GetPlayers()
...
-        public Player CreatePlayer(Connection connection, string reservationKey)
+        public Player CreatePlayer(Connection connection, string reservationKey, out bool wasBrandNewPlayer)
         {
             lock (assetsByConnection)
             {
                 ConnectionAssets assetPackage = assetsByConnection[connection];
                 PlayerContext playerContext = reservations[reservationKey];
                 Validate.NotNull(playerContext);
 
+                wasBrandNewPlayer = playerContext.WasBrandNewPlayer;
```
Sunrunner37: do you need to pass it through the playerContext? can you simply do` wasBrandNewPlayer = playerData.hasSeenPlayerBefore(playerName);?`

CtrlAltDavid01: I have to, because by that time the player is created in playersByPlayerName and when I call it it's the last chance and I have to somehow get it to CreatePlayer.

Issue #290, opened 2018-08-23
No oxygen loss with tank.


fleeting-aardvark:
Updated and compiled a few of hours ago {Wed, 22 Aug 18 @ 5pm CST} on a client version of 60072.

When wearing an oxygen tank or high capacity oxygen tank, there is no oxygen loss for the player.
This occurs for the host {myself} and another client computer {a friend}.

----
With the recent changes to player positioning, I do not believe this is an intended side effect.

crashfly:
Further play on the new save seems to limit this bug on initial startup of the server.  Disconnecting and reconnecting later has corrected the problem.

Still, it was kind of nice to not have to go up for air.

tprussak:
I've also experienced this issue. Another fix for it is to manually remove and reequip the tank. It seems to be related to the tank being put into the player's equipment automatically when crafted.

TroelsL:
We've also experienced this bug. Exiting and restarting the clients did not resolve it. As soon as any tank is equipped, unlimited O2.

Kurgol:
It's only when the oxygen tank (and advanced oxygen tank) are auto-equipped by the crafting station. If you immediate unequip and requip either tank it starts working correctly. Hope that helps!

GW-GhostWolf:
We have found that the bug is related to equipping any tank at or above the surface. We have a fix that is being committed and PR'd shortly.

PR #592, opened 2019-01-06
Resolve inifinite oxygen bug when tank equipped at surface (#290)

GW-GhostWolf:
Only true change is line 28. Everything else is sort and remove usings and VS 2017 code formatting

### DIFF
--- a/NitroxClient/GameLogic/EquipmentSlots.cs
+++ b/NitroxClient/GameLogic/EquipmentSlots.cs
@@ -25,7 +25,10 @@ namespace NitroxClient.GameLogic
         {
             string ownerGuid = GuidHelper.GetGuid(owner);
             string itemGuid = GuidHelper.GetGuid(pickupable.gameObject);
-            pickupable.gameObject.transform.SetParent(owner.transform); // On Deserialized Function Try to find non-existent Parent set null to prevent that bug
+            // save current parent to prevent infinite oxygen when tank equipped above water #290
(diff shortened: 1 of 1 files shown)


NitroxClient/GameLogic/EquipmentSlots.cs
```diff
@@ -25,17 +25,17 @@ public void BroadcastEquip(Pickupable pickupable, GameObject owner, string slot)
         {
             string ownerGuid = GuidHelper.GetGuid(owner);
             string itemGuid = GuidHelper.GetGuid(pickupable.gameObject);
-            pickupable.gameObject.transform.SetParent(null); // On Deserialized Function Try to find non-existent Parent set null to prevent that bug
+            pickupable.gameObject.transform.SetParent(owner.transform); // On Deserialized Function Try to find non-existent Parent set null to prevent that bug
```
Sunrunner37: Not sure if this is the correct solution.  We intentionally remove the parent so that it deserializes with no parent (other clients that receive an equipment event would not be able to deserialize as the parent doesn't exist).  I think the correct solution would be to keep the parent set to null then at the end of this function reset the parent.  Also add commenting on why we are doing it (the current comment is not clear)

GW-GhostWolf: 1 - I didn't put the comment there. it was there when I started. 
2 - I'll check and see if we can move / change the set parent functions so that it is still null for the packet send

PR #596, opened 2019-01-08
Exit Event Handling for gracefully exit

DKingAlpha:
The patch catches exit event on different platforms, so we can exit server gracefully.

### DIFF
--- a/NitroxServer/Program.cs
+++ b/NitroxServer/Program.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Globalization;
+using System.Runtime.InteropServices;
 using System.Threading;
 using NitroxModel.Core;
 using NitroxModel.DataStructures.GameLogic;
@@ -31,6 +32,8 @@ namespace NitroxServer
                 return;
             }
 
+            CatchExitEvent();
+
             ConsoleCommandProcessor cmdProcessor = NitroxServiceLocator.LocateService<ConsoleCommandProcessor>();
             while (server.IsRunning)
             {
@@ -58,5 +61,49 @@ namespace NitroxServer
             Thread.CurrentThread.CurrentCulture = cultureInfo;
             Thread.CurrentThread.CurrentUICulture = cultureInfo;
         }
+
+        private static void CatchExitEvent()
+        {
+            // Catch Exit Event
+            PlatformID platid = Environment.OSVersion.Platform;
+
+            // using *nix signal system to catch Ctrl+C
+            if (platid == PlatformID.Unix || platid == PlatformID.MacOSX || platid == PlatformID.Win32NT || (int)platid == 128) // mono = 128
+            {
+                Console.CancelKeyPress += OnCtrlCPressed;
+            }
+
+            // better catch using WinAPI. This will handled process kill
+            if (platid == PlatformID.Win32NT)
+            {
+                SetConsoleCtrlHandler(ConsoleEventCallback, true);
+            }
+        }
+
+        // See: https://docs.microsoft.com/en-us/windows/console/setconsolectrlhandler
+        private delegate bool ConsoleEventDelegate(int eventType);
+        [DllImport("kernel32.dll", SetLastError = true)]
+        private static extern bool SetConsoleCtrlHandler(ConsoleEventDelegate callback, bool add);
+
+        private static bool ConsoleEventCallback(int eventType)
+        {
+            if (eventType == 2) // close
+            {
+                StopAndExitServer();
+            }
+            return false;
+        }
+        private static void OnCtrlCPressed(object sender, ConsoleCancelEventArgs e)
+        {
+            StopAndExitServer();
+        }
+
+        private static void StopAndExitServer()
+        {
+            Log.Info("Exiting ...");
+            Server.Instance.Stop();
+            Environment.Exit(0);
+        }
+
     }
 }


NitroxServer/Program.cs
```diff
@@ -31,6 +30,30 @@ static void Main(string[] args)
                 return;
             }
 
+            // Catch Exit Event
+            PlatformID platid = Environment.OSVersion.Platform;
+            switch(platid)
```
Measurity: Isn't doing `Console.CancelKeyPress += (sender, args) => { Server.Instance.Stop(); Environment.Exit(0); }` enough for all platforms? Seems to work fine without the additional Win32 DLLImport and code.

NitroxServer/Program.cs
```diff
@@ -58,5 +81,26 @@ static void configureCultureInfo()
...
+            if (eventType == 2)
+            {
+                Log.Info("Exiting ...");
+                Server.Instance.Stop();
+            }
+            return false;
+        }
+        private static void OnCtrlCPressed(object sender, ConsoleCancelEventArgs e)
+        {
+            Log.Info("Exiting ...");
```
Measurity: Logging this isn't needed because the server already does it.

NitroxServer/Program.cs
```diff
@@ -58,5 +81,26 @@ static void configureCultureInfo()
...
+            Server.Instance.Stop();
```
Measurity: Can you add `Environment.Exit(0)` below this line? Otherwise it'll hang for me. I noticed it is because `Console.ReadLine()` is now waiting for text on a stream which is not receiving any text anymore because of Ctrl+C.

NitroxServer/Program.cs
DKingCN: CancelKeyPress only handle Ctrl+C request, it will not be invoked when User **kill** it (signal 9, or closing console window). You can try closing window on windows to see if cancelkeypress delegate invoked.

SetConsleCtrlHandler will catch that closing.

I tried signal handling in the beginning but its just hard for me to implement on C#.

NitroxServer/Program.cs
DKingCN: Saving takes a few seconds for me running it on GCP, this is to notify users that their cancel request is received.

NitroxServer/Program.cs
DKingCN: I didnt notice that. will test it.

DKingCN: You are right, this happens on windows. On linux it does not hang tho. Anyway I am adding an exit on it.

DKingCN:
Fixed & pushed again. Please check it. Thx.

btw I am in a different timezone so you might have trouble contacting me on discord, feel free to leave a msg if im not online.

NitroxServer/Program.cs
```diff
@@ -57,6 +58,51 @@ static void configureCultureInfo()
...
+            if (platid == PlatformID.Unix || platid == PlatformID.MacOSX || platid == PlatformID.Win32NT || (int)platid == 128) // mono = 128
+            {
+                Console.CancelKeyPress += new ConsoleCancelEventHandler(OnCtrlCPressed);
+            }
+
+            // better catch using WinAPI. This will handled process kill
+            if (platid == PlatformID.Win32NT)
+            {
+                consoleHandler = new ConsoleEventDelegate(ConsoleEventCallback);
+                SetConsoleCtrlHandler(consoleHandler, true);
```
Measurity: Can you remove the `consoleHandler` var here and from the class itself? Just set the SetConsoleCtrlHandler directly.

NitroxServer/Program.cs
```diff
@@ -57,6 +58,51 @@ static void configureCultureInfo()
...
+
+        private static void CatchExitEvent()
+        {
+            // Catch Exit Event
+            PlatformID platid = Environment.OSVersion.Platform;
+
+            // using *nix signal system to catch Ctrl+C
+            if (platid == PlatformID.Unix || platid == PlatformID.MacOSX || platid == PlatformID.Win32NT || (int)platid == 128) // mono = 128
+            {
+                Console.CancelKeyPress += new ConsoleCancelEventHandler(OnCtrlCPressed);
```
Measurity: Don't need to do `new ConsoleCancelEventHandler` because it will be inferred automatically.

`Console.CancelKeyPress += OnCtrlCPressed`

DKingCN:
1. One same binary should work on both platform. This is  what a .net project is expected.
2. You can not decide what platform  the program will run on AT build time.
3. Preprocessor only work at build time.

These are reasons why I did not add a preprocessing condition there.

The best practice I can figure out is to try to import them in code. As I tested on linux, the program will not throw exception on WinAPI import UNLESS we actually invoked them. So this is the best i can do.

If you have a better idea, please let me know.


Thanks for the other issues you mentioned. I am noob in C#.

Measurity:
Thank you for investigating it. If the error only happens on invoke like you say it's fine to do without preprocessor.

NitroxServer/Program.cs
cher123: I have screenshotted every single thing that everyone has put on my computer. it is being taken to the FBI first thing Monday morning and will be investigated by their TI department. So, I suggest that if any of you are messing with my computer, get out now before it's too late and you end up in federal prison.

Sunrunner37: Wut

DKingCN:
Hi, fixed the problems you mentioned.

1. removed redundant type conversion
2. formatted code.
3. force-pushed for a cleaner commit history.

PR #597, opened 2019-01-08
Online Server Menu - NitroxClient

DKingCN:
This commit is a part of the feature **Online Server Menu**

It fetchs server list from a [Online Server Manager](https://github.com/DKingCN/NitroxServerManager), which provides (Name, Address:Port,  \#players, Latency) in Nitrox Client

Preview

The code is compatibile with IPv6 network.


Sunrunner37:
Hi DKing, we appreciate the contribution; however, as said in the previous PR we are not interested in server menu implementations at the time.  I have an alternate branch that I've been working on that solves for this.  I would recommend next time consulting us before developing something that is completely new like this.  Especially using a completely different language and protocol.  We should communicate and decide on an implementation before investing effort.

Issue #557, opened 2019-01-01
Subnautica singleplayer sound glitch


DagrMine:
Earlier today I had to uninstall nitrox because when I started a singleplayer game it kept the music from the loading screen going when I loaded in and all of the sound effects (thermoblade, knife, hatch, etc...) didn't work. For some reason the dialogue still worked. Hope someone finds this useful

jayjar291:
same here

jimmyvbuck01:
Same but only my music is screwed up, when I turn it off all the other sound effects work

eliw00d:
This happens to me as well, but after some period of time it starts working again.

RylenWolf:
If mods are curious, this bug fixes itself for me when a different track plays. Specifically, entering the Observatory plays a specific audio, and that fixes the problem for me. I suspect it's related to the music for the menu just not being told to stop by anything else.

jimmyvbuck01:
> This bug fixes itself for me when a different track plays. Specifically, entering the Observatory plays a specific audio, and that fixes the problem for me. I suspect it's related to the music for the menu just not being told to stop by anything else.

It must have something to do with how the mod edits the main page by adding the multiplayer tab to it. I uninstalled the mod and now the sound works fine, I don't plan to play again for a few weeks so I will just play singleplayer till then.

LaikaIwanowa:
Here's how to fix it: Enter another biom than the one you were when you loaded the save. That way the music and sound go back to normal.

RylenWolf:
@caroga my idea is to open a ticket on the issue, wait for coders to see it, and help players with workarounds until the issue is fixed.

eliw00d:
@Caroga I believe that is the goal: https://github.com/SubnauticaNitrox/Nitrox/pull/599#issuecomment-452379504

Coding-Hen:
@Measurity this can probably be closed as single player is disabled

PR #599, opened 2019-01-08
Add ability to enable/disable nitrox

FahadAhmad2001:
This allows users to disable nitrox without having to uninstall it. A file is created in the Subnautica directory called nitrox.enabled, and NitroxPatcher.Main.Execute() is only called if this file contains "true". Useful for occasions where nitrox might affect singleplayer experience, such as in #557


Sunrunner37:
There are only 5 persistent patches - I think we should aim to make them less destructive.  So far we've done a pretty good job but sometimes things leak over.  Although this may be used once or twice, I think this will probably just be a feature no one knows about.  Let's skip on it for now and fix the outstanding SP bug.

PR #601, opened 2019-01-09
Installer now creates firewall exceptions

FahadAhmad2001:
The installer now creates firewall exceptions for TCP and UDP port 11000, NitroxServer.exe and Subnautica.exe


NitroxInstallerActions/InstallPatchAction.cs
```diff
@@ -16,7 +16,7 @@ public static ActionResult InstallPatch(Session session)
             try
             {
                 string managedDirectory = session.CustomActionData["MANAGEDDIR"];
-                                
+                string gameDir = Path.GetDirectoryName(Path.Combine(Path.GetFullPath(managedDirectory), "..\\..\\Subnautica.exe"));
```
JustinHoeky: Best way is to use the "linux path fix" like in #595 so it will be the same like the other code

NitroxInstallerActions/Patches/FirewallRules.cs
```diff
@@ -0,0 +1,206 @@
...
+            setFirewallRulesStartInfo.RedirectStandardOutput = true;
+            Char quotation = '"';
+            string status = "success";
+            foreach (string item in RulesArray)
+            {
+                if (item == "PortsTCP" || item == "PortsUDP" || item == "ProgramServer" || item == "ProgramClient")
+                {
+                    if (item == "PortsTCP")
+                    {
+                        setFirewallRulesStartInfo.Arguments = "advfirewall firewall add rule name=" + quotation + "Nitrox Multiplayer Mod TCP" + quotation + " dir=in action=allow protocol=TCP localport=11000";
```
Sunrunner37: Would recommend a data structure to encapsulate some of this, example:

```
class FirewallRule {
    string protocol;
    string port;
    string name;

    string AddArgument { get { return "advfirewall firewall add rule ........."} }
}
```

NitroxInstallerActions/Patches/FirewallRules.cs
```diff
@@ -0,0 +1,158 @@
...
+                    string output = RunNetshCommand(command);
+                    if (!output.Contains("Ok."))
+                    {
+                        status = false;
+                    }
+                }
+            }
+            return status;
+        }
+        bool RemoveFirewallRules(List<string> RulesArray, int PortNumber)
```
Sunrunner37: I still think there is room for a FirewallRule data structure to encapsulate the name, addComand, and removeCommand.

NitroxInstallerActions/Patches/FirewallRules.cs
```diff
@@ -0,0 +1,158 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using System.Diagnostics;
+using System.IO;
+
+namespace InstallerActions.Patches
+{
+    public class FirewallRules
```
Sunrunner37: Please conform to our general style guidelines.  There should be a return between each method and between logical groupings of code.  In addition to that, we place spaces before and after each equal sign.  Please update the PR to conform to these.

killzoms: https://github.com/SubnauticaNitrox/Nitrox/blob/master/CONTRIBUTING.md guidelines in case you couldnt find them

NitroxInstallerActions/Patches/FirewallRules.cs
```diff
@@ -46,112 +27,146 @@ string RunNetshCommand(string Argument)
...
+            }
+            return commands;
         }
-        bool RemoveFirewallRules(List<string> RulesArray, int PortNumber)
+        bool GetRuleInstalled(string RuleName)
         {
-            bool status=true;
-            foreach(string item in RulesArray)
+            string command = "", output = "";
+            if (RuleName == "PortsUDP")
```
alexanderheuts: Use a switch or else if statements. There's only one if-statement that gets triggered.

NitroxInstallerActions/Patches/FirewallRules.cs
```diff
@@ -46,112 +27,146 @@ string RunNetshCommand(string Argument)
...
+        public string addCommand;
+        public string removeCommand;
+        public string name;
+        public bool isInstalled;
+        string[] GetAddRemoveCommand(string RuleName, string GamePath, int Port = 11000)
         {
-            bool status = true;
-            foreach(string item in RulesArray)
+            string[] commands = new string[2] { "", "" };
+            if (RuleName == "PortsUDP")
```
alexanderheuts: Use an enum for RuleName instead of string. Less sensitive to errors and easier to check.

NitroxInstallerActions/Patches/FirewallRules.cs
```diff
@@ -46,112 +27,146 @@ string RunNetshCommand(string Argument)
...
-            command= "advfirewall firewall show rule name=\x22Nitrox Multiplayer Mod UDP\x22";
-            output = RunNetshCommand(command);
-            if(output.Contains("No rules match the specified criteria"))
+            List<FirewallRule> rules = new List<FirewallRule>();
+            string[] ruleNames = new string[3] { "PortsUDP", "ProgramServer", "ProgramClient" };
+            foreach(string eachName in ruleNames)
             {
-                requiredNitroxRules.Add("PortsUDP");
+                FirewallRule rule = new FirewallRule();
+                rule.CreateRule(eachName, subPath, portNumber);
```
alexanderheuts: CreateRule can throw an exception, but it's not being handled.

FahadAhmad2001:
Please let me know if I need to add or change anything

FahadAhmad2001:
Closing PR as I just realized this won't work on Linux devices ( and maybe even crash the installer)  as netsh.exe is only on windows

FahadAhmad2001:
Reopening PR, as the Nitrox installer is not used on Linux devices

FahadAhmad2001:
@Sunrunner37 Please let me know if I need to change anything else

PR #603, opened 2019-01-10
AutoUpgrade saveFile Version 4 to Version 5

BattleHH:
Hi, 
a small adjustment so that version 4 savefiles are saved as version 5

Kind regards 
Battle

- if SaveFile Version 4
- CreateWorld with new EscapepodData
- Set Version to 5

I´ve tested with two Buddy 4h, works fine :-)

### DIFF
--- a/NitroxServer/Serialization/World/PersistedWorldData.cs
+++ b/NitroxServer/Serialization/World/PersistedWorldData.cs
@@ -50,6 +50,11 @@ namespace NitroxServer.Serialization.World
         {
             if(version < CURRENT_VERSION)
             {
+                if (version == 4)
+                {
+                    return true;
+
+                }
                 Log.Error("Version " + version + " save file is no longer supported.  Creating world under version " + CURRENT_VERSION);
                 return false;
             }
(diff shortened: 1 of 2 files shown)


BattleHH:
CatSZekelyheute um 15:35 Uhr
mmh might be a bad idea because we're still in alpha, we won't support every save version
it'd be nice to have a separate upgrader, so to speak
it's a short-term vs long-term conflict of interest :stuck_out_tongue:
@Battle
Battleheute um 15:42 Uhr
@CatSZekely  This is just for this case, here only the EscapePodData is missing and this will be added. I measure the effort and here the effort is low. :smiley:
CatSZekelyheute um 15:45 Uhr
I figured you would say that, it's just cumbersome to add this, and then remove it when everybody has updated. Hence my suggestion to keep it separate on your repository. I imagine the PR will be rejected because of this...
Battleheute um 15:46 Uhr
it´s ok i will close the Pr :smiley:

Sunrunner37:
I think it is valuable if this could be made into a generic version upgrading solution.  However, at this stage in the development a `if (persistedData.version == 4)` right in the middle of the initial sync is definitely a no-go.

PR #605, opened 2019-01-10
Scrolling chat log

ASimpson87:
Scrolling chat log , mouse activated on chat entry mouse wheel to scroll chat up and down. Asset bundle  chatlog will need to be shipped with installer.


ASimpson87:
Yes we do...

Measurity:
When this is merged we should probably have a tut on how to maintain the assetbundle.

- Install Unity 5.6
- Add assetbundle to a Unity project
- Open it

I assume we might also need the .meta files that Unity exports alongside assets. Maybe it's a good idea to create a new repo for the Unity assets as a Unity 5.6 project @Sunrunner37

Sunrunner37:
Ok I can create the project and can include this stuff in the installer.  Should Simpson configure unity to export metadata for the asset (so it can be included in this PR) before I merge?  Also, can you take a crack at creating that documentation Simpson?

ASimpson87:
Ok i have the meta files so i can add them to the new repo when its up. I'll do the readme for the asset bundles.

PR #608, opened 2019-01-11
Updated Harmony

killzoms:
This is a paired PR with https://github.com/SubnauticaNitrox/Harmony/pull/2

### DIFF
--- a/NitroxPatcher/Patches/Persistent/PAXTerrainController_LoadAsync_Patch.cs
+++ b/NitroxPatcher/Patches/Persistent/PAXTerrainController_LoadAsync_Patch.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Reflection;
 using System.Reflection.Emit;
+using NitroxModel.Logger;
 using Harmony;
 using NitroxModel.Helper;
 using System.Linq;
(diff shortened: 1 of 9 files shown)


NitroxPatcher/Patches/Persistent/PAXTerrainController_LoadAsync_Patch.cs
```diff
@@ -36,8 +37,8 @@ public static IEnumerable<CodeInstruction> Transpiler(MethodBase original, IEnum
                    {
                         //Call our injection code and set paxTerratinController.setIsWorking(true); (this keeps up the loading screen)
                         yield return new ValidatedCodeInstruction(OpCodes.Call, typeof(Multiplayer).GetMethod(nameof(Multiplayer.SubnauticaLoadingCompleted), BindingFlags.Public | BindingFlags.Static));
-                        yield return new ValidatedCodeInstruction(OpCodes.Ldc_I4_1, instruction.labels);
-                   }
+                        yield return new CodeInstruction(OpCodes.Ldc_I4_1) { labels = instruction.labels };
```
Sunrunner37: Why can we no longer use this in conjunction with ValidatedCodeInstruction?

Sunrunner37:
Please ensure to do some end-to-end testing on most features to make sure that injections aren't silently breaking due to the upgrade.

NitroxPatcher/Patches/Persistent/PAXTerrainController_LoadAsync_Patch.cs
killzoms: The ValidatedCodeInstruction when attempting to pass a null Operand with the set label fails to run properly, and passing the labels as an operand does not agree well with Harmony

PR #611, opened 2019-01-12
updated debugUtils

alexanderheuts:


alexanderheuts:
The old one can be removed as well. The new one is a little more detailed and has more options. Left the old one in as it was being used here and there. I'll remove the old references and update the PR.

alexanderheuts:
@Sunrunner37 I've rebased with the latest master. Should all be fine now.

Issue #621, opened 2019-01-13
Server-spawned models at wrong size


quimnuss:
Just documenting for future reference.
Some of the hand-placed models are of the wrong size.

I've launched the singleplayer vs the multiplayer with client-spawning de-authorized (= blacklist everything) and done some screenshots. (Also very useful for the seeing the duplication bug)

![image](https://user-images.githubusercontent.com/4179721/51090910-7f92ea00-1783-11e9-8dd4-9d95a9c86cd4.png)

This is also very patent in the CargoBoxes that are supposed to block your way like:

![image](https://user-images.githubusercontent.com/4179721/51091035-157b4480-1785-11e9-88ed-9dbff133a962.png)

The best one to see the difference is the piece of fuselage is in the way at 988.1 10.0 -38.0

anoviel:
same with plants on floating island, and giant floaters under this island

anoviel:
same with plants on floating island, and giant floaters under this island

PR #622, opened 2019-01-14
Bug with double parsing of ingame files and object scaling

killzoms:
Edit:
GameObject scaling has been included now

### DIFF
--- a/NitroxModel/DataStructures/GameLogic/Entity.cs
+++ b/NitroxModel/DataStructures/GameLogic/Entity.cs
@@ -18,31 +18,34 @@ namespace NitroxModel.DataStructures.GameLogic
         public Quaternion Rotation { get; set; }
 
         [ProtoMember(3)]
-        public TechType TechType { get; set; }
+        public Vector3 Scale { get; set; }
 
         [ProtoMember(4)]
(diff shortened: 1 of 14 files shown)


killzoms:
Fixes https://github.com/SubnauticaNitrox/Nitrox/issues/621

NitroxModel/DataStructures/GameLogic/Entity.cs
```diff
@@ -18,42 +18,46 @@ public class Entity
         public Quaternion Rotation { get; set; }
 
         [ProtoMember(3)]
-        public TechType TechType { get; set; }
+        public Vector3 Scale { get; set; }
```
Sunrunner37: You'll need to increment the world save version since this is a breaking contract change.

NitroxServer/Serialization/ServerProtobufSerializer.cs
```diff
@@ -84,6 +84,7 @@ private void RegisterMembers(MemberInfo[] info, Type type)
 
                         try
                         {
+                            //Log.Info(type + ": " + tag + ": " + property.Name); // ignore this
```
Sunrunner37: remove

PR #623, opened 2019-01-15
Bug with server spawned objects not being movable

ASimpson87:
This allows movable objects spawned by the server to be moved so players can progress further through the story ie boxes blocking the aurora entries.

Only movable objects can be moved... I'm not sure if this affects anything else game related i did some limited testing and seems to be fine.

### DIFF
--- a/NitroxClient/GameLogic/Spawning/DefaultEntitySpawner.cs
+++ b/NitroxClient/GameLogic/Spawning/DefaultEntitySpawner.cs
@@ -28,7 +28,12 @@ namespace NitroxClient.GameLogic.Spawning
             gameObject.transform.localRotation = entity.Rotation;
             GuidHelper.SetNewGuid(gameObject, entity.Guid);
             gameObject.SetActive(true);
-            gameObject.AddComponent<Rigidbody>();
+            // Makes movable objects movable... we can probably do this before the server sends the spawner packet?
+            if(gameObject.GetComponent<Rigidbody>() ==  null)
+            {
+                gameObject.AddComponent<Rigidbody>();
+            }
+
             LargeWorldEntity.Register(gameObject);
             CrafterLogic.NotifyCraftEnd(gameObject, entity.TechType);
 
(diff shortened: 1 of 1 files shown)


NitroxClient/GameLogic/Spawning/DefaultEntitySpawner.cs
```diff
@@ -28,6 +28,12 @@ public Optional<GameObject> Spawn(Entity entity, Optional<GameObject> parent)
             gameObject.transform.localRotation = entity.Rotation;
             GuidHelper.SetNewGuid(gameObject, entity.Guid);
             gameObject.SetActive(true);
+            // Makes movable objects movable... we can probably do this before the server sends the spawner packet?
+            if(gameObject.GetComponent<Rigidbody>() ==  null)
+            {
+                gameObject.AddComponent<Rigidbody>();
```
Sunrunner37: I'm not sure the implications of having all entities spawning with a rigid body.  I think we should dive deeper into why the prefabs are not being initialized with the correct components.  Maybe do a diff from the result of `CraftData` and `PrefabDatabase` above to see if there are any differences for affected tech types.

ASimpson87: Yep I wasn't sure either I tested a fair bit after and it seems everything is fine. This can be used as a quick fix for story line progression safely if people want that. 

I beileve we are diving into proper initialization now along with sever only spawning. Can we create a master branch for this as it's quite a major overhaul and I think it would be nice to have some clarity on master that this is being worked on. 

Just my thoughts feel free to close PR anyway and I will carry on with looking for better initialization methods.

Sunrunner37: I'm fine merging this in as a temporary fix.  Let's create an issue so we remember to remove it.  

Also, let's wait to see what the actual fix for spawning is.  I think there are two issues: 1) incorrect prefab used for loading or incorrect bootstrapping 2) A missing source of parsed files.  If that is the case, these are probably minor changes.  We'll see when we get there though.

PR #630, opened 2019-01-17
Adds Admin login & fixes to Chat

ASimpson87:
Adds an admin login through chat rather then having someone to do it through console. Also fixes to the layout.

### DIFF
--- a/NitroxServer/ConsoleCommands/LoginCommand.cs
+++ b/NitroxServer/ConsoleCommands/LoginCommand.cs
@@ -1,0 +1,54 @@
+﻿using System.Text.RegularExpressions;
+using NitroxServer.ConsoleCommands.Abstract;
+using NitroxServer.GameLogic.Players;
+using NitroxModel.DataStructures.GameLogic;
+using NitroxServer.GameLogic;
+using NitroxModel.DataStructures.Util;
+using NitroxModel.Logger;
+using NitroxServer.ConfigParser;
+
+namespace NitroxServer.ConsoleCommands
+{
+    class LoginCommand : Command
+    {
+        private readonly PlayerData playerData;
+        private readonly PlayerManager playerManager;
+        private readonly ServerConfig serverConfig;
+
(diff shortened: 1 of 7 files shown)


NitroxServer/ConsoleCommands/LoginCommand.cs
```diff
@@ -0,0 +1,54 @@
...
+            {
+                if (playerData.UpdatePlayerPermissions(playerName, Perms.ADMIN))
+                {
+                    message = "Updated permissions to admin";
+                }
+                else
+                {
+                    message = "Could not update permissions " + player.ToString();
+                }
+            }else
```
Measurity: Format the document to conform to code style.

NitroxServer/ConsoleCommands/LoginCommand.cs
```diff
@@ -0,0 +1,54 @@
...
+        {
+            string pass = args[0];
+            string message;
+            string playerName = player.Get().Name;
+
+            if (pass == serverConfig.ServerAdminPassword)
+            {
+                if (playerData.UpdatePlayerPermissions(playerName, Perms.ADMIN))
+                {
+                    message = "Updated permissions to admin";
```
Measurity: Would be good to add the playername that got updated so server owner knows what's happening.

NitroxServer/ConsoleCommands/LoginCommand.cs
```diff
@@ -0,0 +1,54 @@
...
+                }
+                else
+                {
+                    message = "Could not update permissions " + player.ToString();
```
Measurity: Add playername here as well.

Measurity:
The default password can be a security vulnerability if the server owner does not know this login command exists. Can you add a warning to the server startup so that server owners can do something about it? If you do this I also recommend adding a command to easily set the password (or any other server setting) at runtime and save it.

Maybe the latter is a bit too much work for 1 PR. Just adding the message to server console is good enough too.

Sunrunner37:
Maybe randomly generate a password (and spit to the console) if a default one is not set

ASimpson87:
Password is now randomly generated if there is nothing set in the config. You can also update the password at run-time with /changepassword [args]

PR #632, opened 2019-01-17
Replaced logging with log4net

DerTyan:
I replaced the custom logger with an implementation of log4net.

It creates on every launch a new log file with the format "nitrox-{yyyy-MM-DD}-{index}.log". The latest file has no index. Only 10 old files + the latest one will stay, the rest will be deleted.


DerTyan:
Forgot to check the client, was focused on the server: 😨

PR #633, opened 2019-01-17
Disable Client side entity spawning

killzoms:
@Sunrunner37 here ya go


NitroxPatcher/Patches/Persistent/EntityCell_AwakeAsync_Patch.cs
```diff
@@ -0,0 +1,68 @@
...
+                yield return instruction;
+            }
+        }
+
+        public override void Patch(HarmonyInstance harmony)
+        {
+            PatchTranspiler(harmony, getMethod());
+        }
+
+        private static Type getLoadAsyncEnumerableMethod()
```
Measurity: Make it a pascal case method name

NitroxPatcher/Patches/Persistent/EntityCell_AwakeAsync_Patch.cs
```diff
@@ -0,0 +1,68 @@
...
+                    targetEnumeratorClass = type;
+                }
+            }
+
+            Validate.NotNull(targetEnumeratorClass);
+
+            return targetEnumeratorClass;
+        }
+
+        private static MethodInfo getMethod()
```
Measurity: This one too

NitroxPatcher/Patches/Persistent/EntityCell_AwakeAsync_Patch.cs
```diff
@@ -0,0 +1,68 @@
...
+        private static Type getLoadAsyncEnumerableMethod()
+        {
+            Type[] nestedTypes = TARGET_CLASS.GetNestedTypes(BindingFlags.NonPublic | BindingFlags.Static);
+            Type targetEnumeratorClass = null;
+
+            foreach (Type type in nestedTypes)
+            {
+                if (type.FullName.Contains("AwakeAsync"))
+                {
+                    targetEnumeratorClass = type;
```
Measurity:
```CS
Validate.NotNull(type)
return type;
```

NitroxPatcher/Patches/Persistent/EntityCell_AwakeAsync_Patch.cs
```diff
@@ -0,0 +1,68 @@
...
+    public class EntityCell_AwakeAsync_Patch : NitroxPatch
+    {
+        public static readonly Type TARGET_CLASS = typeof(EntityCell);
+
+        private static readonly FieldInfo current = getLoadAsyncEnumerableMethod().GetField("$current", BindingFlags.Instance | BindingFlags.NonPublic);
+
+        public static IEnumerable<CodeInstruction> Transpiler(MethodBase original, IEnumerable<CodeInstruction> instructions, ILGenerator generator)
+        {
+            List<CodeInstruction> codeInstructions = instructions.ToList();
+            yield return new CodeInstruction(OpCodes.Ldarg_0);
```
Measurity: Can you add some comments on what the IL is supposed to do? Ex: "Injecting field to set X so that Y happens."

NitroxPatcher/Patches/Persistent/EntityCell_AwakeAsync_Patch.cs
killzoms: followed the convention in https://github.com/SubnauticaNitrox/Nitrox/blob/master/NitroxPatcher/Patches/Persistent/PAXTerrainController_LoadAsync_Patch.cs#L45

NitroxPatcher/Patches/Persistent/EntityCell_AwakeAsync_Patch.cs
killzoms: I believe it was set up in this manner to throw an error when the type could not be found

killzoms:
Going to close this as a better method of doing this has surfaced

PR #637, opened 2019-01-18
SeaMoth/Prawn Suit/Cyclops  Colours & Name persistence

ASimpson87:
Cyclops needs more work.


NitroxPower:
Good job :+1: does this also fix the bug that upgrade modules dont get saved  when they are on a vehicle?

ASimpson87:
> Good job 👍 does this also fix the bug that upgrade modules dont get saved when they are on a vehicle?

This doesn't look at upgrade modules , however it's on my to do list.

NitroxModel/Discovery/GameInstallationFinder.cs
```diff
@@ -13,8 +13,6 @@ public class GameInstallationFinder : IFindGameInstallation
     {
         private readonly List<IFindGameInstallation> finders = new List<IFindGameInstallation>
         {
-            new SteamGameRegistryFinder(),
-            new EpicGamesRegistryFinder(),
```
Sunrunner37: Why were these removed?

NitroxClient/GameLogic/Vehicles.cs
```diff
@@ -123,19 +123,54 @@ private void OnVehiclePrefabLoaded(GameObject prefab, string guid, Vector3 spawn
             Rigidbody rigidBody = gameObject.GetComponent<Rigidbody>();
             rigidBody.isKinematic = false;
             GuidHelper.SetNewGuid(gameObject, guid);
-            if (interactiveChildIdentifiers.IsPresent())
+
+            // Updates names and colours with persisted data .....yeah.....
+            if (CraftData.GetTechType(gameObject) != TechType.None)
```
Sunrunner37: Maybe pass in tech type so you can do `if(techType == TechType.Cyclops)`

NitroxClient/GameLogic/Constructor.cs
```diff
@@ -35,8 +36,40 @@ public void BeginCrafting(GameObject constructor, TechType techType, float durat
...
                 string constructedObjectGuid = GuidHelper.GetGuid(constructedObject);
-                ConstructorBeginCrafting beginCrafting = new ConstructorBeginCrafting(constructorGuid, constructedObjectGuid, techType, duration, childIdentifiers, constructedObject.transform.position, constructedObject.transform.rotation);
+               
+                //Initialize some default values to serialize.
+                Vector3[] HSB = new Vector3[5];
+                Vector3[] Colours = new Vector3[5];
+                Vector4 tmpColour = Color.white;
+                string name = "Cyclops"; // Cant find a way to actually get the Cyclops name.
+
+                if (!vehicle)
```
Sunrunner37: maybe break this into a new function such as `getVehicleAttributes(vehicle, out hsb, out colours, out name)`

NitroxModel/Discovery/GameInstallationFinder.cs
ASimpson87: uhh sorry that shouldn't be there... i have to do that as it causes problems since i have a few different installs. Removing.

ASimpson87:
Added getObjectAttributes function to constructor , parsed techType to prefebloaded. Added Steam and Epic game finder back to gameinstallationfinder.

PR #647, opened 2019-01-20
custom save file name

disseminate:


NitroxServer/App.config
```diff
@@ -10,5 +10,6 @@
         <add key="GameMode" value="Survival"/>
         <add key="DisableConsole" value="false"/>
         <add key="ServerAdminPassword" value=""/>
+        <add key="SaveFile" value="save.nitrox"/>
```
AlexejheroYTB: Maybe make the save name configurable, but enforce the `.nitrox` at the end of the string?

DerTyan: Maybe change the setting to something like world name and then combine the world name and the file extension as the save file.

AlexejheroYTB: Yea thats what I meant

disseminate: Done

PR #652, opened 2019-01-21
Replaced lidgren with litenetlib for connection lost issues

DerTyan:
Replaced lidgren as the networking library with litenetlib as an attempt to resolve the connection lost issue some people encounter, that were not based on port forwarding issues.
After some hours of testing, no random disconnects occurred in our testing environment, when before they occurred approximately every 0.5-5 minutes.
For absolute clarification, if this is the reason for the random disconnects, further testing with the community is required. 

On the technical side:
The code is loosely based on the networking branch.
I had to create a WrapperPacket for the litenetlib serializer because I did not want to refactor the whole packet handling. For better performance (maybe?) it would be possible to refactor the packet handling.

### DIFF
--- a/NitroxClient/Communication/UdpClient.cs
+++ b/NitroxClient/Communication/UdpClient.cs
@@ -1,21 +1,25 @@
-﻿using Lidgren.Network;
-using NitroxClient.Communication.Abstract;
+﻿using NitroxClient.Communication.Abstract;
 using NitroxClient.MonoBehaviours.Gui.InGame;
 using NitroxModel.Logger;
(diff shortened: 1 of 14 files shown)


Sunrunner37:
I think we'll want to refactor this so we can support multiple parallel implementations (at least for the time being).  Ideally, users can select the implementations via the configuration file in the server.  Clients can either try both when connecting or negotiate with a primary and switch to the secondary if required.  Thoughts?

DerTyan:
Parallel implementation in #664

Issue #537, opened 2018-12-28
Vehicles exploding randomly, solved why


WonkaOh:
The vehicle is rendered in 2 times, one for the owner and one for the other player. The vehicle you are driving is also delayed for the non-owner of the vehicle and since that is, he sees the vehicle hitting unnecessary debris and it causes damage to his render of that vehicle. If he doesn't repair his render of YOUR vehicle, your vehicle will explode when it runs out of health on his render.

Agent77326:
No it‘s not synched, at the moment it‘s like two independent simulations of the vehicle.

Agent77326:
Dunno if this belongs really there, but: #118

PR #654, opened 2019-01-22
Upgrade Module Persistence & (Exosuit Syncing Partially Completed)

ASimpson87:
Upgrade module persistence works as a temporary fix until initial packet sync refactor has been completed. 

Only contains the start of the Exosuit Syncing ie movement.


NitroxPower:
Brilliant! That was another really important fix good job.Also i would like to mention this #537 since its very important you may find a solution.

NitroxClient/Communication/Packets/Processors/InitialPlayerSyncProcessor.cs
```diff
@@ -194,13 +196,19 @@ private void SetPlayerGuid(string playerguid)
...
         
-        private void SpawnPlayerEquipment(List<EquippedItemData> equippedItems)
+        private void SpawnPlayerEquipment(List<EquippedItemData> equippedItems, bool vehiclesToSpawn)
         {
             Log.Info("Received initial sync packet with " + equippedItems.Count + " equipment items");
 
-            using (packetSender.Suppress<EquipmentAddItem>())
+            EquipmentItemAdder itemAdder = new EquipmentItemAdder(packetSender, equippedItems);
+
+            if (vehiclesToSpawn)
```
Sunrunner37: This should always be based on base pieces to spawn as we are listening on an even that would never fire if there were no base pieces.

ASimpson87:
Closing. Upgrade module persistence has been handled through another PR and exo syncing is awhile off.

PR #656, opened 2019-01-23
Sending server commands for day/night on DevConsole

DerTyan:
As requested in #645 the day/night commands from the DevConsole now send the ServerCommand packet.

### DIFF
--- a/NitroxModel/Packets/ServerCommand.cs
+++ b/NitroxModel/Packets/ServerCommand.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using NitroxModel.Logger;
 
 namespace NitroxModel.Packets
 {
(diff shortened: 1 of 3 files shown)


NitroxModel/Packets/ServerCommand.cs
```diff
@@ -11,5 +12,11 @@ public ServerCommand(string[] args)
         {
             CmdArgs = args;
         }
+
+        public ServerCommand(string cmd)
+        {
+            CmdArgs = cmd.Split(' ');
+            Log.Info(string.Join(" ", CmdArgs));
```
Sunrunner37: remove

DerTyan: Done

PR #658, opened 2019-01-23
Add Component searching to the SceneDebugger

killzoms:
Note needs exact name

### DIFF
--- a/NitroxClient/Debuggers/SceneDebugger.cs
+++ b/NitroxClient/Debuggers/SceneDebugger.cs
@@ -256,8 +256,6 @@ namespace NitroxClient.Debuggers
                                 }
                             }
 
-                            Log.Info("Primary");
-
                             foreach (GameObject child in showObjects)
                             {
                                 string guiStyle = child.transform.childCount > 0 ? "bold" : "label";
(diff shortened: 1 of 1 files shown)


killzoms:
Would do a typeName contains search but im afraid it may return to many gameObjects and affect performance

NitroxClient/Debuggers/SceneDebugger.cs
```diff
@@ -256,6 +256,8 @@ private void RenderTabHierarchy()
                                 }
                             }
 
+                            Log.Info("Primary");
```
Measurity: Logging still useful?

NitroxClient/Debuggers/SceneDebugger.cs
```diff
@@ -293,7 +295,18 @@ private void RenderTabHierarchy()
 
                     if (string.IsNullOrEmpty(gameObjectSearchPatternInvalidMessage))
                     {
+                        Type type = AppDomain.CurrentDomain.GetAssemblies()
+                        .Select(a => a.GetType(gameObjectSearch))
```
Measurity: Case insensitive search might be good here so you don't have to type the name exactly.

NitroxClient/Debuggers/SceneDebugger.cs
```diff
@@ -305,17 +318,20 @@ private void RenderTabHierarchy()
                 using (GUILayout.ScrollViewScope scroll = new GUILayout.ScrollViewScope(hierarchyScrollPos))
                 {
                     hierarchyScrollPos = scroll.scrollPosition;
-
+                    Log.Info("Secondary");
```
Measurity: Logging still useful?

killzoms:
image

PR #664, opened 2019-01-25
We can now support the parallel implementation of LiteNetLib and Lidgren

DerTyan:
Users can now choose with networking layer they want to use. LiteNetLib and Lidgren are currently implemented.

LiteNetLib implementation based on #652

### DIFF
--- a/NitroxClient/ClientAutoFacRegistrar.cs
+++ b/NitroxClient/ClientAutoFacRegistrar.cs
@@ -3,6 +3,8 @@ using Autofac;
 using NitroxClient.Communication;
 using NitroxClient.Communication.Abstract;
 using NitroxClient.Communication.MultiplayerSession;
+using NitroxClient.Communication.NetworkingLayer.Lidgren;
+using NitroxClient.Communication.NetworkingLayer.LiteNetLib;
 using NitroxClient.Communication.Packets.Processors.Abstract;
 using NitroxClient.GameLogic;
 using NitroxClient.GameLogic.ChatUI;
@@ -32,13 +34,17 @@ namespace NitroxClient
                 .SingleInstance();
 
             containerBuilder.RegisterType<PlayerPreferenceManager>().SingleInstance();
-				
+
             containerBuilder.RegisterType<MultiplayerSessionManager>()
                 .As<IMultiplayerSession>()
                 .As<IPacketSender>()
                 .InstancePerLifetimeScope();
 
-            containerBuilder.RegisterType<UdpClient>()
+            containerBuilder.RegisterType<LidgrenClient>()
+                .As<IClient>()
+                .InstancePerLifetimeScope();
+
+            containerBuilder.RegisterType<LiteNetLibClient>()
                 .As<IClient>()
                 .InstancePerLifetimeScope();
 
@@ -46,7 +52,7 @@ namespace NitroxClient
                 .AsSelf() //Would like to deprecate this registration at some point and just work through an abstraction.
                 .As<ILocalNitroxPlayer>()
                 .InstancePerLifetimeScope();
-            
+
             containerBuilder.RegisterType<PlayerManager>().InstancePerLifetimeScope();
             containerBuilder.RegisterType<PlayerModelManager>().InstancePerLifetimeScope();
             containerBuilder.RegisterType<PlayerVitalsManager>().InstancePerLifetimeScope();
--- a/NitroxClient/ClientAutoFacRegistrar.cs
+++ b/NitroxClient/ClientAutoFacRegistrar.cs
@@ -3,6 +3,7 @@ using Autofac;
 using NitroxClient.Communication;
 using NitroxClient.Communication.Abstract;
 using NitroxClient.Communication.MultiplayerSession;
+using NitroxClient.Communication.NetworkingLayer;
 using NitroxClient.Communication.NetworkingLayer.Lidgren;
 using NitroxClient.Communication.NetworkingLayer.LiteNetLib;
 using NitroxClient.Communication.Packets.Processors.Abstract;
@@ -40,11 +41,7 @@ namespace NitroxClient
                 .As<IPacketSender>()
                 .InstancePerLifetimeScope();
 
-            containerBuilder.RegisterType<LidgrenClient>()
-                .As<IClient>()
-                .InstancePerLifetimeScope();
-
-            containerBuilder.RegisterType<LiteNetLibClient>()
+            containerBuilder.RegisterType<ModulatingClient>()
                 .As<IClient>()
                 .InstancePerLifetimeScope();
 
--- a/NitroxClient/Communication/MultiplayerSession/MultiplayerSessionManager.cs
+++ b/NitroxClient/Communication/MultiplayerSession/MultiplayerSessionManager.cs
@@ -2,8 +2,10 @@
 using System.Collections.Generic;
 using NitroxClient.Communication.Abstract;
 using NitroxClient.Communication.MultiplayerSession.ConnectionState;
+using NitroxClient.Communication.NetworkingLayer.Lidgren;
 using NitroxClient.GameLogic;
 using NitroxModel;
+using NitroxModel.Core;
 using NitroxModel.Helper;
 using NitroxModel.Logger;
 using NitroxModel.MultiplayerSession;
--- a/NitroxClient/MonoBehaviours/Gui/MainMenu/JoinServer.cs
+++ b/NitroxClient/MonoBehaviours/Gui/MainMenu/JoinServer.cs
@@ -7,12 +7,16 @@ using NitroxClient.GameLogic.PlayerModel;
 using NitroxClient.GameLogic.PlayerModel.Abstract;
 using NitroxClient.GameLogic.PlayerModel.ColorSwap;
 using NitroxClient.GameLogic.PlayerModel.ColorSwap.Strategy;
+using NitroxClient.Communication.MultiplayerSession.ConnectionState;
+using NitroxClient.Communication.NetworkingLayer.LiteNetLib;
+using NitroxClient.GameLogic.PlayerModelBuilder;
 using NitroxClient.GameLogic.PlayerPreferences;
 using NitroxClient.Unity.Helper;
 using NitroxModel.Core;
 using NitroxModel.Helper;
 using NitroxModel.Logger;
 using NitroxModel.MultiplayerSession;
+using NitroxModel.Networking;
 using UnityEngine;
 using UnityEngine.UI;
 
@@ -293,7 +297,7 @@ namespace NitroxClient.MonoBehaviours.Gui.MainMenu
             RightSideMainMenu.groups.Add(joinServerMenu);
 
             //Not sure what is up with this menu, but we have to use the RectTransform of the Image component as the parent for our color picker panel.
-            //Most of the UI elements seem to vanish behind this Image otherwise. 
+            //Most of the UI elements seem to vanish behind this Image otherwise.
             RectTransform joinServerBackground = joinServerMenu.GetComponent<Image>().rectTransform;
             joinServerBackground.anchorMin = new Vector2(0.5f, 0.5f);
             joinServerBackground.anchorMax = new Vector2(0.5f, 0.5f);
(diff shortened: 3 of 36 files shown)


NitroxPower:
We can choose? What will be the difference? I thought that LiteNetLib just has more stable connection.Anyways Good work :+1:

NitroxClient/ClientAutoFacRegistrar.cs
```diff
@@ -30,21 +32,23 @@ private static void RegisterCoreDependencies(ContainerBuilder containerBuilder)
...
-				
+
             containerBuilder.RegisterType<MultiplayerSessionManager>()
                 .As<IMultiplayerSession>()
                 .As<IPacketSender>()
                 .InstancePerLifetimeScope();
 
-            containerBuilder.RegisterType<UdpClient>()
-                .As<IClient>()
+            containerBuilder.RegisterType<LidgrenClient>()
```
Measurity: You can register an implementation as a base class:

```CS
containerBuilder.RegisterType<LidgrenClient>().As<NitroxClient>();
```

Better yet would be to conditionally register one or the other based on a configuration file. You can read more about it here (scroll to lower half of page): https://autofaccn.readthedocs.io/en/latest/configuration/modules.html

NitroxClient/Communication/Abstract/IMultiplayerSessionState.cs
```diff
@@ -5,7 +5,7 @@ namespace NitroxClient.Communication.Abstract
 {
     public interface IMultiplayerSessionState
     {
-        IClient Client { get; }
+        IClient Client { get; set; }
```
Measurity: Remove the setter because it should be given by AutoFac.

NitroxClient/Communication/Abstract/IClient.cs
```diff
@@ -11,5 +12,7 @@ public interface IClient
         void Start(string ipAddress, int serverPort);
         void Stop();
         void Send(Packet packet);
+
+        NetworkingType GetNetworkingType();
```
Measurity: Remove this because AutoFac will decide the networking type.

NitroxClient/Communication/MultiplayerSession/MultiplayerSessionManager.cs
```diff
@@ -15,7 +17,7 @@ public class MultiplayerSessionManager : IMultiplayerSession, IMultiplayerSessio
     {
         private readonly HashSet<Type> suppressedPacketsTypes = new HashSet<Type>();
 
-        public IClient Client { get; }
+        public IClient Client { get; set; }
```
Measurity: Remove the setter.

NitroxClient/Communication/MultiplayerSession/MultiplayerSessionManager.cs
```diff
@@ -24,10 +26,10 @@ public class MultiplayerSessionManager : IMultiplayerSession, IMultiplayerSessio
         public IMultiplayerSessionConnectionState CurrentState { get; private set; }
         public MultiplayerSessionReservation Reservation { get; private set; }
 
-        public MultiplayerSessionManager(IClient client)
+        public MultiplayerSessionManager()
         {
             Log.Info("Initializing MultiplayerSessionManager...");
-            Client = client;
+            Client = NitroxServiceLocator.LocateService<LidgrenClient>();
```
Measurity: Keep this code as it was before

NitroxClient/MonoBehaviours/Gui/MainMenu/JoinServer.cs
```diff
@@ -158,6 +161,16 @@ private void StartMultiplayerClient()
             }
             catch (ClientConnectionFailedException)
             {
+                multiplayerSession.Client.Stop();
```
Measurity: Not sure if this works. Seems very hacky to just jump networking implementation on error in the hope it'll work. I think it's better to just let the client crash/close and restart using LiteNetLibClient.

NitroxClient/MonoBehaviours/Gui/MainMenu/JoinServer.cs
```diff
@@ -290,7 +303,7 @@ private void InitializeJoinMenu()
             RightSideMainMenu.groups.Add(joinServerMenu);
 
             //Not sure what is up with this menu, but we have to use the RectTransform of the Image component as the parent for our color picker panel.
-            //Most of the UI elements seem to vanish behind this Image otherwise. 
+            //Most of the UI elements seem to vanish behind this Image otherwise.
```
Measurity: I think your auto formatting settings are weird. It adds a space to every comment on the end. Can you undo these small changes?

NitroxModel/Networking/NetworkingType.cs
```diff
@@ -0,0 +1,8 @@
+namespace NitroxModel.Networking
+{
+    public enum NetworkingType
```
Measurity: Not needed when you use AutoFac to decide implementation.

NitroxModel/Packets/WrapperPacket.cs
```diff
@@ -0,0 +1,16 @@
+namespace NitroxModel.Packets
+{
+    public class WrapperPacket
```
Measurity: Why wrap the packet?

NitroxServer/Communication/NetworkingLayer/LiteNetLib/LiteNetLibServer.cs
```diff
@@ -0,0 +1,147 @@
...
+            server.UnconnectedMessagesEnabled = true;
+            server.UpdateTime = 15;
+            server.UnsyncedEvents = true;
+            server.Start(portNumber);
+
+            isStopped = false;
+        }
+
+        public override void Stop()
+        {
```
Measurity: You're overriding `Stop` but you aren't calling the base code so now it won't do anything. Is this intended?

NitroxServer/Communication/NetworkingLayer/LiteNetLib/LiteNetLibServer.cs
```diff
@@ -0,0 +1,147 @@
...
+            }
+            catch (Exception ex)
+            {
+                Log.Info("Exception while processing packet: " + packet + " " + ex);
+            }
+        }
+
+        public void OnNetworkReceiveUnconnected(IPEndPoint remoteEndPoint, NetDataReader reader, UnconnectedMessageType messageType)
+        {
+            int i = 0;
```
Measurity: Is it useful to keep this method?

NitroxServer/Communication/NetworkingLayer/NitroxConnection.cs
```diff
@@ -0,0 +1,10 @@
+using NitroxModel.Packets;
+using NitroxModel.Packets.Processors.Abstract;
+
+namespace NitroxServer.Communication.NetworkingLayer
+{
+    public abstract class NitroxConnection : IProcessorContext
```
Measurity: Can be an interface instead of abstract class.

NitroxServer/Serialization/ServerConfig.cs
```diff
@@ -32,6 +34,11 @@ public class ServerConfig
             {"freedom", GameModeOption.Freedom},
         };
 
+        private Dictionary<string, NetworkingType> networkingTypeByConfig = new Dictionary<string, NetworkingType>
```
Measurity: Not needed if using AutoFac.

NitroxServer/Serialization/ServerConfig.cs
```diff
@@ -22,6 +22,8 @@ public class ServerConfig
         private const string DEFAULT_PASSWORD_SETTING = "ServerAdminPassword";
         private const string SAVENAME_SETTING = "SaveName";
         private const string DEFAULT_SAVENAME_SETTING = "save";
+        private const string NETWORKING_TYPE_SETTING = "NetworkingType";
+        private const NetworkingType DEFAULT_NETWORKING_TYPE = NetworkingType.LIDGREN;
```
Measurity: Change to be a string.

NitroxServer/Server.cs
```diff
@@ -9,26 +13,42 @@ namespace NitroxServer
...
 
-        public Server(WorldPersistence worldPersistence, World world, UdpServer udpServer, ServerConfig serverConfig)
+        public Server(WorldPersistence worldPersistence, World world, ServerConfig serverConfig)
         {
             Instance = this;
             this.worldPersistence = worldPersistence;
             this.world = world;
 
-            this.udpServer = udpServer;
+            CreateServer(serverConfig.NetworkingType);
```
Measurity: NitroxServer should be passed in from the constructor automatically by AutoFac.

DartHulos:
@NitroxPower
The parallel implementation of both networking types was specifically requested by Sunrunner37 in #652.
And the dual implementation also adds an extra security layer. LiteNetLib may resolve the connectivity issue some users experience, but there always remains the possibility that LiteNetLib may cause problems for other users since it is impossible to perform large scale testing before releasing a new networking type to the public as any users network context is highly individual. If that would be the case users can always fall back to the lidgren implementation until a better solution is found.

NitroxModel/Packets/WrapperPacket.cs
DerTyan: Because of LiteNetLib Serializer. See #652 for more information. It was a bit weird.

NitroxClient/ClientAutoFacRegistrar.cs
DerTyan: The problem is that the client falls back to the other client if no connection was established. So both must be registered. Or what do you think exactly?

NitroxClient/MonoBehaviours/Gui/MainMenu/JoinServer.cs
DerTyan: While testing this works, but I will have a look tomorrow for a better solution.

NitroxClient/ClientAutoFacRegistrar.cs
Measurity: Depends on if you keep the fallback mechanism you have now. If you think it's best like it is then just register them both in autofac. I recommend to still register them as the base class `NitroxClient` though.

MadMaxOfYore: The general rule of thumb is: 

Register your lowest level abstraction within the IoC container. The reason for that is that the less promises we make in our contracts to the outside world, the less we have to refactor when it comes time to make changes. 

Classes, abstract classes, and Interfaces are all contracts. Interfaces only ever promise signatures in their contracts. Classes and abstract classes; however, make a guarantee of both signatures and behavior (at least for sealed members - which all are by default) in theirs.

In this case, that would be the IClient interface. You'll need to read my review of NitroxClient to understand why that definitely should not be registered in IoC.

NitroxClient/Communication/NetworkingLayer/NitroxClient.cs
```diff
@@ -0,0 +1,27 @@
+using NitroxClient.Communication.Abstract;
+using NitroxModel.Logger;
+using NitroxModel.Networking;
+using NitroxModel.Packets;
+
+namespace NitroxClient.Communication.NetworkingLayer
+{
+    public abstract class NitroxClient : IClient
```
MadMaxOfYore: This is a useless abstraction. An abstract class should only ever exist as a promise of default behavior. This class is nearly a functional equivalent of the IClient interface.

NitroxClient/MonoBehaviours/Gui/MainMenu/JoinServer.cs
MadMaxOfYore: Oof - I feel like this misses the mark. A simple state pattern hidden behind the IClient interface is really all that is needed to handle this problem. You could have a state for Lidgren and a state for LitNetLib. The default state would be whatever the user selected, and the states could "fallback" to each other in the event one of them fails.

Hit me up in Discord if you'd like some pointers on this and I'd be happy to work you with. A few simple refactors and you'll be off to the races!

DerTyan:
We implemented a basic state pattern solution, but we don't think that is exactly as you requested it.
Please let us know if there is anything left. If you give us more information on how you would like this to be implemented we would be happy to make further changes.

NitroxModel/packages.config
```diff
@@ -1,5 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
   <package id="Autofac" version="2.6.3.862" targetFramework="net35" />
+  <package id="LiteNetLib" version="0.8.2-beta" targetFramework="net35" />
```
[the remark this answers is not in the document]
DerTyan: In our tests the beta version performed better than the stable 0.7.7.2.

PR #665, opened 2019-01-26
Fixed UnitTesting issues with NitroxTest

killzoms:


NitroxPatcher/Patches/BaseGhost_Finish_Patch.cs
```diff
@@ -32,7 +32,7 @@ public static IEnumerable<CodeInstruction> Transpiler(MethodBase original, IEnum
                      * TransientLocalObjectManager.Add(TransientLocalObjectManager.TransientObjectType.BASE_GHOST_NEWLY_CONSTRUCTED_BASE_GAMEOBJECT, gameObject);
                      */
                     yield return new CodeInstruction(OpCodes.Ldc_I4_1);
-                    yield return original.Ldloc<GameObject>(1);
+                    yield return original?.Ldloc<GameObject>(1);
```
Sunrunner37: These should not be nullable.  We need an exception or they will silently fail if UWE changes them.

NitroxClient/Communication/Packets/Processors/BedEnterProcessor.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using NitroxClient.Communication.Packets.Processors.Abstract;
+using NitroxModel.Packets;
+
+namespace NitroxClient.Communication.Packets.Processors
+{
+    class BedEnterProcessor : ClientPacketProcessor<BedEnter>
```
Sunrunner37: Let's remove and add an issue instead.

NitroxTest/Model/PacketProcessorTest.cs
```diff
@@ -55,8 +55,20 @@ public void RuntimeDetectsAllClientPacketProcessors()
...
-                        $"{processor} has not been discovered by the runtime code!")
+                {
+                    try
+                    {
+                        Assert.IsTrue(clientDependencyContainer.ResolveOptional(processor.BaseType) != null,
+                            $"{processor} has not been discovered by the runtime code!");
+                    }
+                    catch (Autofac.Core.DependencyResolutionException ex)
+                    {
+                        if (ex.InnerException.GetType() != typeof(System.Security.SecurityException))
```
Sunrunner37: What components are throwing e call violations?

NitroxPatcher/Patches/BaseGhost_Finish_Patch.cs
killzoms: Made it that way due too the unitTest passing in a null parameter...

NitroxClient/Communication/MultiplayerSession/MultiplayerSessionManager.cs
```diff
@@ -78,7 +78,6 @@ public void RequestSessionReservation(PlayerSettings playerSettings, Authenticat
         public void ProcessReservationResponsePacket(MultiplayerSessionReservation reservation)
         {
             Reservation = reservation;
-            CurrentState.NegotiateReservation(this);
```
Sunrunner37: Should we be removing this?

MadMaxOfYore: Sorry @killzoms but SR is correct. This is the state that that session manager should be in upon receiving a confirmed reservation:

```suggestion
            CurrentState.NegotiateReservation(this);
```
https://github.com/SubnauticaNitrox/Nitrox/blob/master/NitroxClient/Communication/MultiplayerSession/ConnectionState/AwaitingSessionReservation.cs#L8

killzoms:
figured out how to update local submodules with the latest commit...

killzoms:
Considering the age I will be closing this merge request

PR #679, opened 2019-01-28
Add basic fire extinguisher syncing

micaww:
Basic fire douse/extinguish syncing to nearby players w/ packet rate limiting. 

This surely can be improved over time:
- Player must have the fire streamed in in order to see the effects
- Fire position is used instead of GUID as they're not synced (as far as I know). Fires don't move so this should be fine for most cases, especially those spawned as part of the map. Tested on Aurora.

Updating this to a GUID-based fire resolution would be trivial if that sort of system were in place.

### DIFF
--- a/NitroxClient/ClientAutoFacRegistrar.cs
+++ b/NitroxClient/ClientAutoFacRegistrar.cs
@@ -74,6 +74,7 @@ namespace NitroxClient
             containerBuilder.RegisterType<SeamothModulesEvent>().InstancePerLifetimeScope();
             containerBuilder.RegisterType<EscapePodManager>().InstancePerLifetimeScope();
             containerBuilder.RegisterType<Debugger>().InstancePerLifetimeScope();
+            containerBuilder.RegisterType<FireManager>().InstancePerLifetimeScope();
         }
 
         private void RegisterPacketProcessors(ContainerBuilder containerBuilder)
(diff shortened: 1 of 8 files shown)


NitroxPower:
Keep in mind thow somebody is working on the Cyclops Syncing and he did sync the fire it may cause issues the way you may make it....

micaww:
@NitroxPower this syncs the fire extinguisher used by the players. If the Cyclops fire object spawn points are synced, then these features will work fine together.

micaww:
Actually just noticed his douse syncing implementation. Pretty similar it seems, but won't work any fires except for Cyclops unless the server spawns them or syncs map fire GUIDs with clients.

NitroxPower:
i am not a developer and i dont know how to script i just wanted to inform you just in case it may not work great work thow :+1:

Sunrunner37:
I recently merged in the cyclops fire code.  Should we use the Fires gamelogic class that he introduced?  Any other refactoring we need to do?

Sunrunner37:
Are the Aurora fires spawned by the client of the server? (I'm guessing they are client side).  Is there a way for us to server-side spawn them so the GUIDs will be synced?  Maybe they are virtual entities.

micaww:
@Sunrunner37 yes, spawned by client. Is the Aurora and all of it's child entities part of the batch files that the server already parses? If so, why would the fires be excluded?

Should we close this PR as the code is unnecessary if we sync GUIDs? I would still like to continue discussion on syncing the rest of these unsynced map ents.

Issue #678, opened 2019-01-28
Aurora warning patch and random auroraTime persistence


quimnuss:
`CreateWorld `gets called by `CreateFreshWorld `and also `LoadFromFile`, and it builds a new `EventTriggerer `which recreates random timers for the aurora events. Here: https://github.com/SubnauticaNitrox/Nitrox/blob/master/NitroxServer/Serialization/World/WorldPersistence.cs#L136

Persisting the random `auroraTimer ` will fix this since `serverStartTime `is already persisted.

Additionally, class `AuroraWarnings::Update()` is not patched so the client will trigger the events regardless of the server trigger.

Duplicate of #499
Duplicate of #626

quimnuss:
Also, quick temporary fix without changing the persisted data is to remove the randomness.

aeppacher:
Duplicate of #626

quimnuss:
Correction. The AuroraWarnings Update actually calls Trigger on auroraWarningN StoryGoal which ultimately calls the StoryGoal.Execute() which _is_ patched. These auroraWarning fields seem to be set in the UnityEditor, so I don't know what GoalType they are, which seems important because StoryGoal::Execute handles PDA, Radio and Encyclopedia, but nothing is called there for Story GoalType.

Anyway, what looks certain is that the crashedShipExploder explosion story event is handled on its own monobehaviour.

Another method that needs to be changed is `CrashedShipExploder::Update()`, which is the one actually exploding the aurora. Nitrox takes care of this in the StoryGoal patch and related messages by setting the countdown+explosion time at the moment the explosion packet is received; effectively triggering the explosion at the next frame Update.

Coding-Hen:
@Measurity This can be closed

Issue #626, opened 2019-01-16
Aurora events not persisting?


buinak:
We had a couple of explosions happen past early game for players, including one that cost us the whole inventory. Is the event data not persisting?

Coding-Hen:
@Measurity This can be closed

Issue #499, opened 2018-12-24
Aurora keeps exploding


RylenWolf:
Version 1.0.0.8 Nexusmods

The first time I played, we fixed the aurora (after explosion) and saved the game. On restart, it eventually explodes again at random.

Even if I repair the aurora with the console command, it keeps exploding once every restart (I'm fairly certain it's the same timer that it usually goes on for a new game.. every two hours?)

I have not tried manually repairing before the explosion since the first time (Which was done after the initial explosion) so there is room to test to see if an immediate teleport and manual repair with the tool will stop this behaviour, but it should have just saved the manual repairs from the first save, so I figured I would report this here anyways.

I will test another time if teleport+manual repair before explosion stops this and update with a comment.

Agent77326:
Even so, can reproduce multiple Aurora expolsions when server is restarted.

Coding-Hen:
@Measurity This can be closed

PR #683, opened 2019-01-29
Set explosion to 3.1 days, quickpatch to _disable_ client side aurora story timers

quimnuss:
The client CrashedShipExploder uses a timeMonitor that is Updated at Update. I've set a Patch that sets the timers to a greater number than the server does so that the later triggers it first and overrites those values.

A better approach is to actually disable the timeMonitor time update or the CrashedShipExploder Update altogether and re-activate it the moment the server requests the aurora explosion.

this PR should solve #499 #626 #678


quimnuss:
Setting the timer to a high value will wrong the warnings that use it to lerp the moment they are triggered. Let me fix this before merging.

PR #686, opened 2019-01-30
Added fix for Bootstrappers

killzoms:


NitroxServer/GameLogic/Entities/Spawning/EntityBootstrappers/CrashFishBootstrapper.cs
```diff
@@ -17,7 +17,7 @@ private Entity SpawnChild(Entity parentEntity, DeterministicBatchGenerator deter
         {
             string guid = deterministicBatchGenerator.NextGuid();
 
-            return new Entity(parentEntity.Position, parentEntity.Rotation, new UnityEngine.Vector3(1, 1, 1), techType, parentEntity.Level, parentEntity.ClassId, true, guid);
+            return new Entity(parentEntity.Position, parentEntity.Rotation, new UnityEngine.Vector3(1, 1, 1), techType, parentEntity.Level, "", true, guid);
```
Sunrunner37: Should we pass in the actual class ids for Crash/CrashPowder?

killzoms: We can, although I dont think the classId prefab is different from the TechType prefab

PR #696, opened 2019-02-01
ServerConfig refactoring and enforcement

DKingCN:
Refactor ServerConfig
Enforce MaxConnections limit on NitroxServer
Protect server side with config key ServerPassword (only implemented on server side)

@killzoms mentioned it (MaxConnections) on discord so i guess i need to notify him about this pr.

### DIFF
--- a/NitroxServer/GameLogic/PlayerManager.cs
+++ b/NitroxServer/GameLogic/PlayerManager.cs
@@ -6,6 +6,7 @@ using NitroxModel.Helper;
 using NitroxModel.MultiplayerSession;
 using NitroxModel.Packets;
 using NitroxServer.Communication;
+using NitroxServer.ConfigParser;
 using NitroxServer.GameLogic.Players;
 using NitroxServer.UnityStubs;
 
@@ -16,12 +17,14 @@ namespace NitroxServer.GameLogic
     {
         private readonly Dictionary<Connection, ConnectionAssets> assetsByConnection = new Dictionary<Connection, ConnectionAssets>();
         private readonly PlayerData playerData;
+        private readonly ServerConfig serverConfig;
         private readonly Dictionary<string, PlayerContext> reservations = new Dictionary<string, PlayerContext>();
         private readonly HashSet<string> reservedPlayerNames = new HashSet<string>();
 
-        public PlayerManager(PlayerData playerData)
+        public PlayerManager(PlayerData playerData, ServerConfig serverConfig)
         {
             this.playerData = playerData;
+            this.serverConfig = serverConfig;
         }
 
         public List<Player> GetPlayers()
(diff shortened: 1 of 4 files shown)


NitroxServer/GameLogic/PlayerManager.cs
```diff
@@ -40,24 +43,35 @@ public List<Player> GetPlayers()
...
             {
-                ConnectionAssets assetPackage;
-                assetsByConnection.TryGetValue(connection, out assetPackage);
+                // TODO: ServerPassword in NitroxClient
 
-                if (assetPackage == null)
-                {
-                    assetPackage = new ConnectionAssets();
-                    assetsByConnection.Add(connection, assetPackage);
+                if (serverConfig.ServerPassword != "" && (authenticationContext.ServerPassword.IsEmpty() || (authenticationContext.ServerPassword.Get() != serverConfig.ServerPassword)))
```
MadMaxOfYore: Can you look into what it would take to simply hash the password rather than ship it around in plain text? This is the kind of stuff that kiddos love to exploit simply for the sake of doing so...

NitroxServer/Serialization/ServerConfig.cs
```diff
@@ -166,13 +50,50 @@ public string GenerateRandomString(int size, bool lowerCase)
...
             Configuration config = ConfigurationManager.OpenExeConfiguration(ConfigurationUserLevel.None);
-            config.AppSettings.Settings[DEFAULT_PASSWORD_SETTING].Value = pw;
+            config.AppSettings.Settings[serverAdminPasswordSetting.Name].Value = pw;
             config.Save(ConfigurationSaveMode.Modified);
 
             ConfigurationManager.RefreshSection("appSettings");
         }
+    }
+    
+    internal class ServerConfigItem<T>
```
MadMaxOfYore: This needs to be in its own file.

DKingCN: It was actually nested in ServerConfig till i pull it out. Do we have too extract every helper class? I dont have much experience on teamwork or huge project but as i can see, too much files of internal helper class will just mess devs minds up.

MadMaxOfYore: Nah, most devs use the search facilities within the IDE to find the code they are looking for. I think the default hotkey for VS is pressing sift twice, then you can just type the name of the class you want. Trying to solve the problem of solution clutter in any other way never scales well. 

I don't think that will be too much of a problem here though. It looks like those classes are in their small namespace anyways.

NitroxServer/GameLogic/PlayerManager.cs
MadMaxOfYore: Very well, we just need to encourage people to not use a password that they use anywhere else. Putting this out there amounts to us encouraging them to store their password in clear text on their hard drive - which is borderline irresponsible on our part.

I can take a look at it at a later time. I don't see any good of holding the process up on this unless SR feels differently.

NitroxServer/Serialization/ServerConfig.cs
DKingCN: ok

PR #699, opened 2019-02-02
Disable client-side save confirmation popup

micaww:


killzoms:
was wondering who would take care of this GG

micaww:
Easy peasy. Saw someone mention it in Discord the other day.

PR #705, opened 2019-02-02
Add support for command Help

DKingCN:
the output is split into lines due to text length restriction on the client side.

feature tested.

btw chatlog box is a little bit small for an active chat.

### DIFF
--- a/NitroxServer/ConsoleCommands/HelpCommand.cs
+++ b/NitroxServer/ConsoleCommands/HelpCommand.cs
@@ -12,8 +12,7 @@ namespace NitroxServer.ConsoleCommands
     internal class HelpCommand : Command
     {
         private readonly PlayerData playerData;
-        private Dictionary<Perms, List<string>> helpTexts = new Dictionary<Perms, List<string>>();
-
+        
         public HelpCommand(PlayerData playerData) : base("help", Perms.PLAYER, "", "Display help about supported commands")
         {
             this.playerData = playerData;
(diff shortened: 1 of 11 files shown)


DKingCN:
i think its better now

NitroxServer/ConsoleCommands/HelpCommand.cs
```diff
@@ -0,0 +1,65 @@
...
+            public int Compare(Command x, Command y)
+            {
+                return x.Name.CompareTo(y.Name);
+            }
+        }
+
+        private List<string> GetHelpText(Perms perm)
+        {
+            List<string> helpText;
+            if (helpTexts.TryGetValue(perm, out helpText))
```
Sunrunner37: Do we need to cache?  Seems like something infrequently run so less code / complexity is preferred over optimization.

NitroxServer/ConsoleCommands/HelpCommand.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            this.playerData = playerData;
+        }
+
+        public override void RunCommand(string[] args, Optional<Player> player)
+        {
+            if (player.IsPresent())
+            {
+                List<string> cmdsText = GetHelpText(playerData.GetPermissions(player.Get().Name));
+                cmdsText.ForEach(cmdText => SendServerMessageIfPlayerIsPresent(player, cmdText));
```
Sunrunner37: Instead of sending a packet for every command should we concat the command help text with new lines in the GetText method?  Then we don't deal with lists and just have a string.

DKingCN: The original implement is to concat them. but on client side there is a restriction that chatlog lines <=255 and single line char length <= 255.

NitroxServer/ConsoleCommands/HelpCommand.cs
DKingCN: i glanced through other command class found they pretty lightweightted. except that im unsure about the cost of NitrozServiceLocator. The cache is introduced to reduce attack surface at the initial purpose. Its okay to remove cache if you are sure its better.
Second confirmation please.

NitroxServer/ConsoleCommands/HelpCommand.cs
DKingCN: and the current chatlog box is too small to break/warp a lot. yet i will let the chatlog box that way because i think there might be a remake from the maintainer of it in the future.

NitroxServer/ConsoleCommands/HelpCommand.cs
DKingCN: never mind. i removed it.

PR #718, opened 2019-02-05
This is a gift to you

killzoms:
ClassId: 8d3d3c8b-9290-444a-9fea-8e5493ecd6fe is Reaper Leviathan


NitroxPower:
Does this fix anything that you changed what does it do?

killzoms:
doesnt fix anything but it does slightly change whats done on a specific date. though this likely shouldnt have been done in this spot but a different one

JustinHoeky:
If i'm correct, it will spawn a Reaper Leviathan in front of the player on 1th of April.
I would love to see their faces, but not sure if the admins will allow it haha

killzoms:
@JustinHoeky You see why its a gift?

Sunrunner37:
Best PR thus far.

Dessix:
@JustinHoeky It appears every single entity that would be spawned on that day would be a reaper leviathan. I don't think even the player would spawn at that point...

PR #725, opened 2019-02-15
Massive Changes to Entity spawning to support CellRoots

killzoms:


NitroxClient/GameLogic/Entities.cs
```diff
@@ -106,13 +102,18 @@ private void UpdatePosition(Entity entity)
...
+                opGameObject.Get().transform.localPosition = entity.LocalPosition;
+                opGameObject.Get().transform.localScale = entity.LocalScale;
+                if (entity.LocalPosition != new Vector3(0, 0, 0))
+                {
+                    opGameObject.Get().transform.position = entity.Position;
+                    opGameObject.Get().transform.rotation = entity.Rotation;
+                }
             }
-            else
+            else if (!entity.IsChild)
```
Sunrunner37: Why do we want to ignore logging this error for child entities?

NitroxClient/GameLogic/Entities.cs
```diff
@@ -106,13 +102,18 @@ private void UpdatePosition(Entity entity)
...
             {
-                opGameObject.Get().transform.position = entity.Position;
-                opGameObject.Get().transform.localRotation = entity.Rotation;
-                opGameObject.Get().transform.localScale = entity.Scale;
+                opGameObject.Get().transform.localRotation = entity.LocalRotation;
+                opGameObject.Get().transform.localPosition = entity.LocalPosition;
+                opGameObject.Get().transform.localScale = entity.LocalScale;
+                if (entity.LocalPosition != new Vector3(0, 0, 0))
+                {
+                    opGameObject.Get().transform.position = entity.Position;
```
Sunrunner37: I'm not entirely understanding this area.  We only update the position if there was a local position?  Wouldn't this cause a lot of entities to not be updated?

NitroxClient/GameLogic/Spawning/DefaultEntitySpawner.cs
```diff
@@ -24,20 +24,43 @@ public Optional<GameObject> Spawn(Entity entity, Optional<GameObject> parent)
...
+            if (parent.IsPresent() && parent.Get().name == "CellRoot(Clone)")
+            {
+                LargeWorldEntity ent = gameObject.GetComponent<LargeWorldEntity>();
+                ent.transform.SetParent(parent.Get().transform, true);
+                ent.OnAddToCell();
+
+                gameObject.transform.localPosition = entity.LocalPosition;
+                gameObject.transform.localRotation = entity.LocalRotation;
+            }
+            else if (gameObject.name == "CellRoot(Clone)")
```
Sunrunner37: Should we move this logic into a `TechType.None` entity spawner?

NitroxClient/GameLogic/Spawning/DefaultEntitySpawner.cs
```diff
@@ -24,20 +24,43 @@ public Optional<GameObject> Spawn(Entity entity, Optional<GameObject> parent)
...
-            gameObject.transform.localScale = entity.Scale;
+            gameObject.transform.localScale = entity.LocalScale;
+
+            if (parent.IsPresent() && parent.Get().name == "CellRoot(Clone)")
+            {
+                LargeWorldEntity ent = gameObject.GetComponent<LargeWorldEntity>();
+                ent.transform.SetParent(parent.Get().transform, true);
+                ent.OnAddToCell();
+
+                gameObject.transform.localPosition = entity.LocalPosition;
```
Sunrunner37: minor: looks like these two lines are common so can be out of the if/else if/else

NitroxModel-Subnautica/DataStructures/GameLogic/Entities/SubnauticaUwePrefabFactory.cs
```diff
@@ -40,6 +41,18 @@ public override List<UwePrefab> GetPossiblePrefabs(string biome)
             return prefabs;
         }
 
+        public override List<UwePrefab> GetPrefabForClassId(string classId)
```
Sunrunner37: minor: `getPrefabsForClassId`

NitroxModel/DataStructures/GameLogic/Entity.cs
```diff
@@ -9,7 +10,7 @@ namespace NitroxModel.DataStructures.GameLogic
     [ProtoContract]
     public class Entity
```
Sunrunner37: I think in the future (probably not this PR).  We'll want to break up into different types of entities:

`DroppedEntity` - has serializedGameObject and only basic elements
`LocallyPositionedEntity`
`GloballyPositionedEntity`

NitroxServer-Subnautica/GameLogic/Entities/Spawning/ReefbackBootstrapper.cs
```diff
@@ -23,11 +28,10 @@ public void Prepare(Entity parentEntity, DeterministicBatchGenerator determinist
         {
             for(int spawnPointCounter = 0; spawnPointCounter < LocalCreatureSpawnPoints.Count; spawnPointCounter++)
             {
-                Vector3 localSpawnPosition = LocalCreatureSpawnPoints[spawnPointCounter];
                 float targetProbabilitySum = (float)deterministicBatchGenerator.NextDouble() * creatureProbabiltySum;
                 float probabilitySum = 0;
 
-                foreach (ReefbackEntity creature in SpawnableCreatures)
+                foreach (ReefbackSpawnData.ReefbackCreature creature in SpawnableCreatures)
```
Sunrunner37: Let's refactor a bit to get two methods: `SpawnReefbackPlants` and `SpawnReefbackAnimals`

NitroxServer-Subnautica/GameLogic/Prefab.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using NitroxServer.UnityStubs;
+
+namespace NitroxServer.GameLogic
+{
+    public class Prefab
```
Sunrunner37: Can we use `UwePrefab`?

NitroxServer-Subnautica/Serialization/ResourceAssetsParser.cs
```diff
@@ -58,17 +82,457 @@ public static ResourceAssets Parse()
                                 wei.localScale = new UnityEngine.Vector3(resourcesFile.reader.ReadSingle(), resourcesFile.reader.ReadSingle(), resourcesFile.reader.ReadSingle());
                                 resourceAssets.WorldEntitiesByClassId.Add(wei.classId, wei);
                             }
-                        }
+                            break;
+                        case "ReefbackSlotsData":
```
Sunrunner37: Let's make some new class here.  `WorldEntityDataParser` and `ReefbackSlotsDataParser`.  Then we can have a Dictionary<String, MonoBehaviourParser>`

NitroxServer-Subnautica/Serialization/ResourceAssetsParser.cs
```diff
@@ -58,17 +82,457 @@ public static ResourceAssets Parse()
...
+                                    position = creaturePrefab.position,
+                                    scale = creaturePrefab.scale,
+                                    rotation = creaturePrefab.rotation,
+                                    probability = probability,
+                                    minNumber = minNumber,
+                                    maxNumber = maxNumber
+                                });
+                            }
+                            break;
+                        default:
```
Sunrunner37: Throw into a `DefaultMonobehaviourParser`

NitroxServer-Subnautica/Serialization/ResourceAssetsParser.cs
```diff
@@ -58,17 +82,457 @@ public static ResourceAssets Parse()
...
+                    }
+                }
+            }
+
+            afr.Position = prevPos;
+
+            return null;
+        }
+
+        public static List<UwePrefab> GetPrefabFromPlaceholder(string classId)
```
Sunrunner37: Wasn't this method in the class above?  Shouldn't this class just parse all of the data then have it passed to the consumers later?

NitroxServer-Subnautica/Serialization/ResourceAssetsParser.cs
```diff
@@ -58,17 +82,457 @@ public static ResourceAssets Parse()
...
+                    }
+                }
+            }
+
+
+
+            return null; // This should likely never happen
+        }
+
+        private static AssetFileInfoEx FindComponentOfType(Type type, AssetsFileInstance assetsFileInstance, ulong gameObjectPathId)
```
Sunrunner37: Might want these methods in a parsing helper class or the base of an abstract parser

NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs
```diff
@@ -202,27 +190,99 @@ private IEnumerable<Entity> SpawnEntitiesStaticly(EntitySpawnPoint entitySpawnPo
...
-                foreach(Entity childEntity in spawnedEntity.ChildEntities)
+                for (int i = 0; i < spawnedEntity.ChildEntities.Count - prefabs.Count; i++)
                 {
-                    yield return childEntity;
+                    Entity childEntity = spawnedEntity.ChildEntities[i + prefabs.Count];
                 }
             }
         }
 
+        private IEnumerable<Entity> CreateEntityCellRoot(EntitySpawnPoint entitySpawnPoint, DeterministicBatchGenerator deterministicBatchGenerator)
```
Sunrunner37: Do we need to modify the deserialize patch to delete UWE's existing cell roots (we explicitly prevented it from doing that)

NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs
```diff
@@ -202,27 +190,99 @@ private IEnumerable<Entity> SpawnEntitiesStaticly(EntitySpawnPoint entitySpawnPo
...
+                        spawnedEntity.ChildEntities.AddRange(SpawnEntitiesUsingRandomDistribution(childSpawnPoint, prefabs, deterministicBatchGenerator));
+                    }
+                    else if (childSpawnPoint.ClassId != null)
+                    {
+                        spawnedEntity.ChildEntities.AddRange(SpawnEntitiesStaticly(childSpawnPoint, deterministicBatchGenerator));
+                    }
+                }
+            }
+
+            spawnedEntity.ChildEntities.ForEach(c => c.IsChild = true);
```
Sunrunner37: Does this consider all entities (outside of cell roots) as children?

NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs
```diff
@@ -202,27 +190,99 @@ private IEnumerable<Entity> SpawnEntitiesStaticly(EntitySpawnPoint entitySpawnPo
...
+                    prefab.TechType = wei.Get().TechType;
+
+                    Entity childEntity = new Entity(prefab.LocalPosition,
+                        prefab.LocalRotation,
+                        prefab.LocalScale,
+                        prefab.TechType,
+                        prefab.CellLevel,
+                        prefab.ClassId,
+                        true,
+                        prefab.Guid) { IsChild = true };
```
Sunrunner37: minor: want to add it to the constructor?

NitroxServer-Subnautica/NitroxServer-Subnautica.csproj
```diff
@@ -39,6 +39,9 @@
     <Reference Include="Assembly-CSharp">
```
Sunrunner37: Let's fix these paths, this was my mistake.

NitroxServer-Subnautica/GameLogic/Prefab.cs
killzoms: I think i did switch over to UwePrefab due to the merging

NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs
killzoms: Might want to I just left em as is.

NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs
killzoms: As we havent figured out Batch spawning yea it will just set em all as children of the CellRoot

NitroxClient/GameLogic/Entities.cs
killzoms: Was supposed to be entity.Position...

NitroxClient/GameLogic/Spawning/DefaultEntitySpawner.cs
killzoms: In the event some gameObjects are passed to the TechType.None entity spawner id like to avoid that...

Sunrunner37: I meant for us to move it while keeping the `if` check

NitroxServer-Subnautica/GameLogic/Entities/Spawning/ReefbackSpawnData.cs
```diff
@@ -5,42 +5,30 @@ namespace NitroxServer_Subnautica.GameLogic.Entities.Spawning.EntityBootstrapper
 {
     public static class ReefbackSpawnData
     {
-        public struct ReefbackEntity
+        public struct ReefbackCreature
```
Sunrunner37: Let's refactor into a single struct: `ReefbackEntity`.  This would allow us to simplify the reefback bootstrapper into a single function with two invocations:

```
   spawnEntities(spawnablePlants, plantsProbabilitySum, parentEntity, deterministicBatchGenerator);
   spawnEntities(spawnableCreatures, creatureProbabilitySum, parentEntity, deterministicBatchGenerator);

    private void SpawnPlants(List<ReefbackEntity> possibleEntities, float probabilitySum, Entity parentEntity, DeterministicBatchGenerator deterministicBatchGenerator)
```

NitroxModel/Packets/EntityTransformUpdates.cs
```diff
@@ -42,17 +42,23 @@ public class EntityTransformUpdate
...
                 Rotation = rotation;
+                Scale = scale;
+                LocalPosition = localPosition;
+                LocalRotation = localRotation;
             }
 
             public override string ToString()
             {
-                return "(" + Guid + " " + Position + " " + Rotation + ")";
+                return "(" + Guid + " " + Position + ")";
```
Sunrunner37: minor: Let's add the other items and the class name.

NitroxPatcher/Patches/Persistent/PrefabPlaceholdersGroup_Spawn_Patch.cs
```diff
@@ -0,0 +1,22 @@
...
+using System.Linq;
+using System.Text;
+using Harmony;
+using System.Reflection;
+
+namespace NitroxPatcher.Patches.Persistent
+{
+    public class PrefabPlaceholdersGroup_Spawn_Patch : NitroxPatch
+    {
+        public static bool Prefix()
```
Sunrunner37: Let's add a comment saying what this patch does

NitroxServer-Subnautica/GameLogic/Entities/Spawning/SubnauticaEntitySpawnPointFactory.cs
```diff
@@ -20,21 +20,23 @@ public override List<EntitySpawnPoint> From(AbsoluteEntityCell absoluteEntityCel
...
+                                         gameObject.Parent);
                 }
             }
             else
             {
-                spawnPoints.Add(
-                    new EntitySpawnPoint(absoluteEntityCell, transform.Position, transform.Rotation, transform.Scale, gameObject.ClassId)
-                );
+                EntitySpawnPoint spawnPoint = new EntitySpawnPoint(absoluteEntityCell, transform.Position, transform.Rotation, transform.Scale, gameObject.ClassId, gameObject.Id, gameObject.Parent);
+                if (spawnPoint.Parent == null)
```
Sunrunner37: We only add the spawn point if it has a parent?

NitroxServer-Subnautica/Serialization/ResourceAssets/MonoBehaviourParsers/DefaultMonoBehaviourParser.cs
```diff
@@ -0,0 +1,97 @@
...
+        {
+            AssetsFile resourcesFile = instance.file;
+            AssetsFileReader afr = resourcesFile.reader;
+            AssetsFileTable resourcesFileTable = instance.table;
+
+            if (gameObjectPathId == 0)
+            {
+                return; // Gotta love early breaks
+            }
+            else if (gameObjectPathId != 0 && !checkedGameObjects.Contains(gameObjectPathId))
```
Sunrunner37: minor: we already know gameObjectPathId is != 0

NitroxServer-Subnautica/Serialization/ResourceAssets/MonoBehaviourParsers/DefaultMonoBehaviourParser.cs
```diff
@@ -0,0 +1,97 @@
...
+                    AssetFileInfoEx prefabIdentifierAsset = ResourceAssetsHelper.FindComponentOfType<PrefabIdentifier>(instance, (ulong)gameObjectPathId);
+                    resourcesFile.reader.Position = prefabIdentifierAsset.absoluteFilePos + 28;
+
+                    resourcesFile.reader.ReadCountStringInt32(); // Empty name
+                    string classId = resourcesFile.reader.ReadCountStringInt32(); // ClassId
+
+                    List<UwePrefab> val;
+                    if (!SubnauticaUwePrefabFactory.PrefabsByClassId.TryGetValue(classId, out val)) // should never be false but just incase
+                    {
+                        SubnauticaUwePrefabFactory.PrefabsByClassId.Add(classId, prefabs);
```
Sunrunner37: I think if you use the [] accessor it won't throw an exception if it already exists... that way you can remove the if check.

```
 SubnauticaUwePrefabFactory.PrefabsByClassId[classId] = prefabs;
```

NitroxServer-Subnautica/Serialization/ResourceAssets/MonoBehaviourParsers/ReefbackSlotsDataParser.cs
```diff
@@ -0,0 +1,156 @@
...
+                };
+
+                resourcesFile.reader.Position = prevPos;
+
+
+                int minNumber = resourcesFile.reader.ReadInt32();
+                int maxNumber = resourcesFile.reader.ReadInt32();
+                float probability = resourcesFile.reader.ReadSingle();
+
+                ReefbackSpawnData.SpawnableCreatures.Add(new ReefbackSpawnData.ReefbackCreature()
```
Sunrunner37: If you pass these (`ReefbackSpawnData`) in via the constructor you won't have to statically assign these.

NitroxServer-Subnautica/Serialization/ResourceAssets/ResourceAssetsHelper.cs
```diff
@@ -0,0 +1,211 @@
...
+                    {
+                        afr.Position = prevPos;
+                        return componentAFI;
+                    }
+                }
+            }
+
+
+
+            return null; // This should likely never happen
```
Sunrunner37: Should we throw an exception with debugging information?  It looks like the callers are not checking for null.

NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs
```diff
@@ -202,27 +190,99 @@ private IEnumerable<Entity> SpawnEntitiesStaticly(EntitySpawnPoint entitySpawnPo
...
             IEntityBootstrapper bootstrapper;
             if (customBootstrappersByTechType.TryGetValue(techType, out bootstrapper))
             {
                 bootstrapper.Prepare(spawnedEntity, deterministicBatchGenerator);
 
-                foreach(Entity childEntity in spawnedEntity.ChildEntities)
+                for (int i = 0; i < spawnedEntity.ChildEntities.Count - prefabs.Count; i++)
                 {
-                    yield return childEntity;
+                    Entity childEntity = spawnedEntity.ChildEntities[i + prefabs.Count];
```
Sunrunner37: How does this work?

Also, it looks like we are no longer yielding these.

NitroxServer/GameLogic/Entities/Spawning/EntitySpawnPoint.cs
```diff
@@ -7,35 +7,74 @@ namespace NitroxServer.GameLogic.Entities.Spawning
...
+            Density = density;
+            AllowedTypes = allowedTypes;
+            BiomeType = biomeType;
+            LocalPosition = localPosition;
+            LocalRotation = localRotation;
+
+            EntitySpawnPoint parent = null;
+            if (parentId == null)
+            {
+                parentsById.Add(id, this);
```
Sunrunner37: Setting it as its own parent?

NitroxServer-Subnautica/GameLogic/Entities/Spawning/SubnauticaEntitySpawnPointFactory.cs
killzoms: If the spawnPoint has no parent it will be added to the list of entities to spawn(aka its the root parent of other entities) and other entities will be setting their parent as that entity.(gameObject.Parent = gameObject.ParentId)

NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs
killzoms: ah right thats dead code... The client is Spawning ChildEntities so having them be returned while also existing as a child of an already returned entity will create bloat

NitroxServer/GameLogic/Entities/Spawning/EntitySpawnPoint.cs
killzoms: This would be adding the current entitySpawnPoint as a new possible root spawnPoint(aka it has no parent and thus will either be the root of other spawnPoints or its own gameObject)

NitroxServer-Subnautica/Serialization/ResourceAssets/MonoBehaviourParsers/ReefbackSlotsDataParser.cs
killzoms: Considering the class was static and the places that make use of it converting the class into a constructable class and passing it around via Autofac or otherwise wouldve been more of a hassle than its worth...

NitroxClient/GameLogic/Entities.cs
```diff
@@ -53,7 +53,7 @@ public void Spawn(List<Entity> entities)
                 {
                     Spawn(entity, Optional<GameObject>.Empty());
                 }
-                else
+                else if (alreadySpawnedGuids.Contains(entity.Guid))
```
Sunrunner37: Wouldn't this always be true as the condition above is false?

NitroxClient/GameLogic/SimulationOwnership.cs
```diff
@@ -14,6 +16,8 @@ public class SimulationOwnership
         private readonly IPacketSender packetSender;
         private readonly Dictionary<string, SimulationLockType> simulatedGuidsByLockType = new Dictionary<string, SimulationLockType>();
         private readonly Dictionary<string, LockRequestCompleted> completeFunctionsByGuid = new Dictionary<string, LockRequestCompleted>();
+
+        public static readonly List<AbsoluteEntityCell> LoadedCells = new List<AbsoluteEntityCell>();
```
Sunrunner37: How is this being used?  I only see items being added to it.

NitroxClient/GameLogic/SimulationOwnership.cs
```diff
@@ -76,5 +80,10 @@ public void StopSimulatingGuid(string guid)
         {
             simulatedGuidsByLockType.Remove(guid);
         }
+
+        internal void RegisterEntityForBroadcasting(string guid)
```
Sunrunner37: Can we remove this?

NitroxClient/GameLogic/Spawning/NoneEntitySpawner.cs
```diff
@@ -0,0 +1,71 @@
...
+using NitroxModel.DataStructures.GameLogic;
+using NitroxModel.DataStructures.Util;
+using NitroxModel.Helper;
+using NitroxModel_Subnautica.Helper;
+using UnityEngine;
+using UWE;
+
+namespace NitroxClient.GameLogic.Spawning
+{
+    public class NoneEntitySpawner : IEntitySpawner
```
Sunrunner37: How many entities are we talking here?  Will spawning 'None' be a large performance hit?

NitroxClient/GameLogic/Spawning/NoneEntitySpawner.cs
```diff
@@ -0,0 +1,71 @@
...
+    {
+        /**
+         * Crash fish are spawned by the CrashHome in the Monobehaviours Start method
```
Sunrunner37: Change the comment to something applicable to this.

NitroxModel-Subnautica/NitroxModel-Subnautica.csproj
```diff
@@ -42,7 +42,14 @@
...
       <HintPath>..\packages\Autofac.2.6.3.862\lib\NET35\Autofac.Configuration.dll</HintPath>
     </Reference>
-    <Reference Include="LitJson, Version=1.0.5467.41753, Culture=neutral, processorArchitecture=MSIL" />
+    <Reference Include="LitJson, Version=1.0.5467.41753, Culture=neutral, processorArchitecture=MSIL">
+      <SpecificVersion>False</SpecificVersion>
+      <HintPath>$(SubnauticaManaged)\LitJson.dll</HintPath>
+    </Reference>
+    <Reference Include="Newtonsoft.Json, Version=9.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
+      <SpecificVersion>False</SpecificVersion>
+      <HintPath>F:\SteamLibrary\steamapps\common\Subnautica\Subnautica_Data\Managed\Newtonsoft.Json.dll</HintPath>
```
Sunrunner37: Let's fix the path

NitroxModel/DataStructures/GameLogic/Entity.cs
```diff
@@ -9,7 +11,7 @@ namespace NitroxModel.DataStructures.GameLogic
...
     {
-        public AbsoluteEntityCell AbsoluteEntityCell => new AbsoluteEntityCell(Position, Level);
+        public AbsoluteEntityCell AbsoluteEntityCell => new AbsoluteEntityCell((Position == new Vector3(0, 0, 0)) ? LocalPosition : Position, Level);
```
Sunrunner37: Hmm, is this right?  Should we ever use local position to calculate the AbsoluteEntityCell?  It is probably based off of the parent's position.  Will this mess things up that depend on AbsoluteEntityCell?

NitroxModel/NitroxModel.csproj
```diff
@@ -47,9 +47,16 @@
     <Reference Include="LZ4">
       <HintPath>..\lib\LZ4.dll</HintPath>
     </Reference>
+    <Reference Include="Newtonsoft.Json, Version=9.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
+      <SpecificVersion>False</SpecificVersion>
+      <HintPath>F:\SteamLibrary\steamapps\common\Subnautica\Subnautica_Data\Managed\Newtonsoft.Json.dll</HintPath>
```
Sunrunner37: Fix path

NitroxModel/NitroxModel.csproj
```diff
@@ -59,6 +66,10 @@
     <Reference Include="System.Data.DataSetExtensions" />
     <Reference Include="System.Data" />
     <Reference Include="System.Xml" />
+    <Reference Include="UnityEngine.UI, Version=1.0.0.0, Culture=neutral, processorArchitecture=MSIL">
+      <SpecificVersion>False</SpecificVersion>
+      <HintPath>F:\SteamLibrary\steamapps\common\Subnautica\Subnautica_Data\Managed\UnityEngine.UI.dll</HintPath>
```
Sunrunner37: Same

NitroxServer/Communication/Packets/Processors/CellVisibilityChangedProcessor.cs
```diff
@@ -14,6 +14,9 @@ class CellVisibilityChangedProcessor : AuthenticatedPacketProcessor<CellVisibili
         private readonly EntitySimulation entitySimulation;
         private readonly PlayerManager playerManager;
 
+        private static int simCount = 0;
```
Sunrunner37: Were these for debugging?  Can we remove?

NitroxServer/GameLogic/Entities/EntityData.cs
```diff
@@ -164,12 +192,53 @@ public Optional<Entity> GetEntityByGuid(string guid)
         {
             Entity entity = null;
 
+            TryGetParentOrChild(guid, out entity);
+
+            return Optional<Entity>.OfNullable(entity);
+        }
+
+        private bool TryGetParentOrChild(string guid, out Entity entity)
```
Sunrunner37: It looks like Get/Remove have turned into very expensive and complex operations.  It is possible to keep entitiesByGuid a flat map of all entities (including children) then only return items that are not children in the phasing logic above?  That would allow all of these methods to remain as-is with only a small change above.

NitroxModel-Subnautica/NitroxModel-Subnautica.csproj
killzoms: damn missed that one...

NitroxModel/DataStructures/GameLogic/Entity.cs
killzoms: so far i havent seen any issues with the entitycell being misaligned

NitroxClient/GameLogic/Spawning/NoneEntitySpawner.cs
killzoms: the performance hit ive seen so far is relatively minor

NitroxModel/NitroxModel.csproj
MarijnS95: If my mind serves me correctly, you can just use `Include="Newtonsoft.Json"` and get rid of the `SpecifcVersion` element as well.

NitroxClient/Communication/Abstract/DeferredPacketReceiver.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using NitroxModel.DataStructures.GameLogic;
+using NitroxModel.Packets;
+
+namespace NitroxClient.Communication.Abstract
+{
+    public abstract class DeferredPacketReceiver
```
MarijnS95: Shouldn't the methods in this class be marked `abstract` as well, to enforce them being implemented by the subclass?

I've been out of the loop for too long, but these "default implementations" seem wrong to me.

NitroxModel/DataStructures/GameLogic/Entity.cs
MarijnS95: Just looking at the code, it seems either `Position` _or_ `LocalPosition` is initialized, never at the same time (I haven't checked for external assignments), which must be the reason for this ternary if. Does it make sense to keep them both? This seems somewhat inline with the [split recommendation](https://github.com/SubnauticaNitrox/Nitrox/pull/725/files#r257466085) comment @Sunrunner37 made.

On the same note, would you mind annotating these constructors with their intended usecase? I've been out of the Nitrox loop for a while, and can't easily figure out where either is supposed to be used.

NitroxClient/NitroxClient.csproj
```diff
@@ -46,12 +48,6 @@
       <HintPath>..\packages\LiteNetLib.0.8.2-beta\lib\net35\LiteNetLib.dll</HintPath>
       <Private>True</Private>
     </Reference>
-    <Reference Include="UnityEngine">
```
MarijnS95: As mentioned in another comment, just leave the references like this, instead of the autogenerated one that adds a `Version` field etc, only to do away with it by setting `SpecificVersion` to false.

NitroxModel/DataStructures/GameLogic/Entity.cs
```diff
@@ -18,46 +20,73 @@ public class Entity
...
 
         [ProtoMember(8)]
-        public List<Entity> ChildEntities { get; set; } = new List<Entity>();
+        public int Level { get; set; }
 
         [ProtoMember(9)]
+        public string ClassId { get; set; }
+
+        [ProtoIgnore]
+        public Dictionary<string, Entity> ChildEntitiesByGuid { get { return childEntitiesByGuid; } set { childEntitiesByGuid = value; } }
```
MarijnS95: I can't remember: Does `ProtoIgnore` / `NonSerialized` break with auto properties?

NitroxModel/DataStructures/GameLogic/Entity.cs
```diff
@@ -81,11 +136,49 @@ public Entity(Vector3 position, Quaternion rotation, Vector3 scale, TechType tec
...
+        public void AddChild(Entity child)
+        {
+            ChildEntitiesByGuid.Add(Guid, child);
         }
 
         public override string ToString()
         {
-            return "[Entity Position: " + Position + " TechType: " + TechType + " Guid: " + Guid + " Level: " + Level + " classId: " + ClassId + " ChildEntities: " + ChildEntities + " SpawnedByServer: " + SpawnedByServer + "]";
+            List<Entity> childList = ChildEntitiesByGuid.Values.ToList();
+            if (childList == null)
```
MarijnS95: Pretty sure `ToList()` returns an empty list (not `null`) when there are no items.

If `ChildEntitiesByGuid` were `null`, this would have already thrown a `NullReferenceException` above.

NitroxModel/Logger/Log.cs
```diff
@@ -98,7 +98,11 @@ public static void Debug(object o)
         // Helping method for formatting string correctly with arguments
         private static string Format(string fmt, params object[] arg)
         {
-            return string.Format(fmt, arg);
+            if (arg.Length > 0)
```
MarijnS95: Does `string.Format` really break when there are no items?!

https://docs.microsoft.com/en-us/dotnet/api/system.string.format?view=netframework-3.5#System_String_Format_System_String_System_Object___

> Object[] args
An object array that contains zero or more objects to format.

NitroxServer-Subnautica/GameLogic/Entities/Spawning/SubnauticaEntitySpawnPointFactory.cs
```diff
@@ -20,21 +20,23 @@ public override List<EntitySpawnPoint> From(AbsoluteEntityCell absoluteEntityCel
...
 
-                    spawnPoints.Add(
-                        new EntitySpawnPoint(absoluteEntityCell,
-                                             entitySlotData.localPosition,
-                                             entitySlotData.localRotation,
-                                             stringSlotTypes,
-                                             entitySlotData.density,
-                                             entitySlotData.biomeType.ToString())
-                    );
+                    new EntitySpawnPoint(absoluteEntityCell,
```
MarijnS95: It is very ugly to hide that the constructor adds the item to an internal static dictionary on `gameObject.Parent` here. Any ideas for a slight refactor?

NitroxServer-Subnautica/Serialization/ResourceAssets/MonoBehaviourParsers/ReefbackSlotsDataParser.cs
```diff
@@ -0,0 +1,156 @@
...
+using AssetsTools.NET.Extra;
+using NitroxServer_Subnautica.GameLogic.Entities.Spawning.EntityBootstrappers;
+using NitroxServer_Subnautica.Serialization.ResourceAssets;
+using NitroxServer.UnityStubs;
+
+namespace NitroxServer_Subnautica.Serialization.ResourceAssets.MonoBehaviourParsers
+{
+    public class ReefbackSlotsDataParser : MonoBehaviourParser
+    {
+        public override void Parse(AssetsFileInstance instance, long gameObjectPathId, ResourceAssets resourceAssets)
```
MarijnS95: It looks like most of the code parses the same data structure as `DefaultMonoBehaviourParser`. Would it be possible to merge it (together with the "reefbackCreature" loop below) and decrease duplication as well as future code maintenance?

Even if there are slight differences which I have not noticed, splitting it out into a few common functions goes a long way already.

NitroxServer/Communication/Packets/Processors/CellVisibilityChangedProcessor.cs
MarijnS95: Same for the `Count` variables in `cellEntities` / `ownershipChange`?

NitroxServer/ConsoleCommands/Processor/ConsoleCommandProcessor.cs
```diff
@@ -42,7 +42,7 @@ public ConsoleCommandProcessor(IEnumerable<Command> cmds, PlayerManager playerMa
 
         public void ProcessCommand(string msg, Optional<Player> player, Perms perms)
         {
-            if (string.IsNullOrWhiteSpace(msg))
```
MarijnS95: Shouldn't this stay the way it is?

NitroxServer/GameLogic/Entities/EntitySimulator.cs
```diff
@@ -93,10 +109,17 @@ private List<Entity> AssignForCells(Player player, AbsoluteEntityCell[] added)
...
-                                                simulationOwnershipData.TryToAcquire(entity.Guid, player, DEFAULT_ENTITY_SIMULATION_LOCKTYPE)));                       
+                foreach (Entity entity in entities)
+                {
+                    if (cell.Level <= entity.Level && simulationOwnershipData.TryToAcquire(entity.Guid, player, DEFAULT_ENTITY_SIMULATION_LOCKTYPE))
+                    {
+                        assignedEntities.Add(entity);
+                        assignedEntities.AddRange(
+                        entity.ChildEntitiesByGuid.Values.Where(childEntity => cell.Level <= childEntity.Level &&
+                                                    ((childEntity.SpawnedByServer && serverSpawnedSimulationWhiteList.Contains(childEntity.TechType)) || !childEntity.SpawnedByServer) &&
+                                                    simulationOwnershipData.TryToAcquire(childEntity.Guid, player, DEFAULT_ENTITY_SIMULATION_LOCKTYPE)));
```
MarijnS95: A more general organizational question pops up here: does it make sense to separate parent and child simulation+ownership?

NitroxServer/Serialization/ServerConfig.cs
```diff
@@ -89,7 +89,7 @@ public ServerConfigItem(string itemName, T defaultValue)
                 string text = ConfigurationManager.AppSettings[itemName];
 
                 // Empty string is ignored
-                if (typeof(T) == typeof(string) && (string.IsNullOrEmpty(text) || string.IsNullOrWhiteSpace(text)))
+                if (typeof(T) == typeof(string) && (string.IsNullOrEmpty(text) || string.IsNullOrEmpty(text)))
```
MarijnS95: `string.IsNullOrEmpty(text) || string.IsNullOrEmpty(text)` Is superfluous. But why not just `if (typeof(T) == typeof(string) && string.IsNullOrWhiteSpace(text))`

NitroxClient/Communication/Abstract/DeferredPacketReceiver.cs
Sunrunner37: Also, what is the use-case for having multiple implementations of a DeferredPacketReceiver?  I am only seeing one in the PR but may be missing something.

killzoms:
the use case is rather small but changing the defer code every time we need
to add a packet to the deferred list will get old... got it in place now
for extending to the entity simulation packet to defer it...

Sunrunner37:
Looks like Gharg was able to deprecate deferred packet sending in his PR.  So now you won't need to worry about doing that.  Let me know if you'll get a chance to look at this again.

killzoms:
Been patched in

PR #744, opened 2019-03-10
Fix issue that building some base part is not reported as completed when rejoining the server.

brianpopow:
## Steps to reproduce:

- Start a new session
- Build a base with a hatch
- logout
- login again
- The Hatch will appear as not completed (same for other stuff like Water filtration machine for example).

This PR attempts to fix this issue.

### DIFF
--- a/NitroxClient/GameLogic/Building.cs
+++ b/NitroxClient/GameLogic/Building.cs
@@ -80,39 +80,34 @@ namespace NitroxClient.GameLogic
             
             string guid = GuidHelper.GetGuid(gameObject);
 
-            if (amount < 0.95f) // Construction complete event handled by function below
-            {
-                ConstructionAmountChanged amountChanged = new ConstructionAmountChanged(guid, amount);
-                packetSender.Send(amountChanged);
-            }
+            ConstructionAmountChanged amountChanged = new ConstructionAmountChanged(guid, amount);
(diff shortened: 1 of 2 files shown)


brianpopow:
The main issue was, that in some cases LATEST_CONSTRUCTED_BASE_PIECE was not present. This happened mostly with solarpanels, basehatch and the waterfiltration machine.  In consequence constructionCompleted packet would not be send in those cases, because finishedPiece would have no value.

Reviewing this one week later, i totally agree to you that my changes hardly explain why this should not happen again. I think i have fooled myself into believing my changes where responsible for that. What i have also changed was sending the construction amount always, not only when its below 0.95. I think thats why i did not see the symptoms of this anymore (almost completed base pieces). I think my changes just made it just much more unlikely to happen. 

I have a hard time reproducing this issue with my changes. I have not seen LATEST_CONSTRUCTED_BASE_PIECE to be **NOT** present. I have added a log output in case this happens, but i have not seen this in like 10 attempts. Without my changes it is reproducible.

brianpopow:
@Sunrunner37: I have no longer full confidence in my changes. I will close this PR. What i can say for sure is, that for some reason LATEST_CONSTRUCTED_BASE_PIECE is sometimes not present. I could not find the reason for that yet. Sorry for wasting your time.

PR #751, opened 2019-03-16
Stabilized Network Communication

acidscorch:
Ready for Review

~~!Attention: This needs more rework. After further investigating and long term tests, i've noticed, that after about 10 minutes, connections begin to make weird things. Seems to be related to Lindgren.SendBuffer. I try to solve this and extend this pull request.
I let this pull request Open, because this solutions works already fine. And can be used by those who are testing against other issues (cyclops, etc).~~

Objective: 
- eliminate desync issues caused by packet loss
- eliminate issues caused by processing incoming packets in the wrong order

Changes:
- introduced a packet sequence for every connection between host and clients
- processing packets in the order that they were send and not in the order they were received
- request resend of packets that are missed in sequence
- packet cache for sent packets to be able to resend missing packets
- changed lindgren message read for safe usage
- increased SendBuffer

### DIFF
--- a/NitroxClient/Communication/NetworkingLayer/Lidgren/LidgrenClient.cs
+++ b/NitroxClient/Communication/NetworkingLayer/Lidgren/LidgrenClient.cs
@@ -1,3 +1,4 @@
+﻿using System.Collections.Generic;
 using System.Threading;
 using Lidgren.Network;
 using NitroxClient.Communication.Abstract;
@@ -16,6 +17,13 @@ namespace NitroxClient.Communication.NetworkingLayer.Lidgren
         private NetClient client;
         private AutoResetEvent connectedEvent = new AutoResetEvent(false);
         private readonly DeferringPacketReceiver packetReceiver;
+
+        private long incomingSequence = 0;
+        private long outgoingSequence = 0;
+        private int packetSentCacheCount = 100;
+        private List<Packet> packetSentCache = new List<Packet>();
+        private List<Packet> packetPreReceivedCache = new List<Packet>();
+        private Packet latestSkippedPackageCausedOfResendRequests = null;
 
         public LidgrenClient()
         {
@@ -40,6 +48,36 @@ namespace NitroxClient.Communication.NetworkingLayer.Lidgren
 
         public void Send(Packet packet)
         {
+            outgoingSequence++;
+            packet.Sequence = outgoingSequence;
+            packetSentCache.Add(packet);
+            if (packetSentCache.Count > packetSentCacheCount)
+            {
+                packetSentCache.RemoveAt(0);
+            }
+            SendPreparedPacket(packet);
+        }
+
+        private void ResendPackage(long sequence)
+        {
+            if (packetSentCache.Exists(packet => packet.Sequence == sequence))
+            {
+                Packet packetToResend = packetSentCache.Find(packet => packet.Sequence == sequence);
+                SendPreparedPacket(packetToResend);
+            }
+            else
+            {
+                SendPreparedPacket(new PacketResendRequest(sequence * -1));
+            }
+        }
+
+        private void RequestResendPackage(int sequence)
+        {
+            SendPreparedPacket(new PacketResendRequest(sequence));
+        }
+
+        private void SendPreparedPacket(Packet packet)
+        {
             byte[] bytes = packet.Serialize();
 
             NetOutgoingMessage om = client.CreateMessage();
@@ -53,8 +91,13 @@ namespace NitroxClient.Communication.NetworkingLayer.Lidgren
         {
             client.UnregisterReceivedCallback(ReceivedMessage);
             client.Disconnect("");
+            packetSentCache.Clear();
+            packetPreReceivedCache.Clear();
+            latestSkippedPackageCausedOfResendRequests = null;
+            incomingSequence = 0;
+            outgoingSequence = 0;
         }
-
+    
         public void ReceivedMessage(object peer)
         {
             NetIncomingMessage im;
@@ -77,7 +120,7 @@ namespace NitroxClient.Communication.NetworkingLayer.Lidgren
                         {
                             connectedEvent.Set();
                         }
-                        else if (LostConnectionModal.Instance)
+                        else if (LostConnectionModal.Instance != null)
                         {
                             LostConnectionModal.Instance.Show();
                         }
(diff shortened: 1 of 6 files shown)


MarijnS95:
Hey, thanks for your contribution.

From the looks of it, you're aiming to solve issues with packet loss and ordering, by means of implementing a "reliable" layer on top of Lidgren. However, this kind of boilerplate is already implemented in that library, and supposed to work just fine.

This Lidgren feature is controlled with a so-called `DeliveryMethod` enum; you can read up on these here: https://github.com/lidgren/lidgren-network-gen3/wiki/Basics. For example, movement packets should be tagged `UnreliableSequenced`: loosing a few packets does not matter because a new one is usually just around the corner, but they should be processed in the order they were sent to prevent jittering back between new and old positions, which happens when they are delivered in a different order.
Likewise, most other "event" packets should be tagged `ReliableOrdered`, to make sure they are *all* processed, in the order they were sent.

What I recommend you is to check up on which packets get "out of sync", and make sure their `DeliveryMethod` is set correctly. Looking at Nitrox's source code, only the two aforementioned methods are used, which means receiving packets out of order is already very unlikely.

However, Lidgren also has a concept called [Sequence Channels](https://github.com/lidgren/lidgren-network-gen3/wiki/Sequence-Channels); only packets belonging to the same channel are processed in order with respect to eachother. Packets belonging to different channels are not required to be in order/sequence, as per the definition. Do note that channels are separated per delivery method.

Disclaimer: I've been out of Nitrox development for a long time, and thus don't have anything to relate to WRT network stability. Also, I haven't read through the diff just yet, and based my comments solely on the PR description.

acidscorch:
Hey, thanks for the detailed reply.

We have already tested different DeliveryMethods and verified, that they were correctly set. But are still noticing situations where they are received in the wrong order.

The Method used for these packets is ReliablyOrdered and the sequence is also active with channel 0. 

The other reason for building this was to get a possibilty to generally trace which information was transfered from clients to server and vice versa. To track down some issues where informations from one player are not reaching another one.

MarijnS95:
Thanks for confirming the issue. 

I've heard about cases where Nitrox is sending too many packets, overfilling the send buffer in Lidgren. Afaik this will drop packets (unless it's configured to block) which should never happen for `ReliablyOrdered`. Is this happening on your side? If so, I think we should pursue to fix/improve that, instead of building a reliability layer on top of another reliability layer. Either by decreasing the amount, frequency and size of the packets, as well as optimizing parameters in Lidgrens `NetPeerConfiguration`.
At some point it is safe to consider a client "disconnected" if the server is unable to deliver messages,  instead of piling them up until running out of memory.

Again, going off the current code, it seems only player and vehicle movement packets are sent unreliably ordered, while there are a lot of other packets that should be sent like that, such as creature updates.
Would you mind checking which packets are eligible for such treatment, mark them as such (+ give them a unique channel) and see if it helps any? As far as I remember, clients still simulate creatures in chunks, which can be a real network hog due to the size and frequency of these packets.

Last but not least: I do agree that such debugging information (being able to see sequence numbers) is important in debugging any network issues. Lidgren has debugging info; can we use that instead of manually adding our own?

acidscorch:
Nevertheless i've asked Killzoms to provide a log of his fork.

acidscorch:
I've reproduced it myself. It took quite a few attempts to make it happen again. The logs are from the current master code only extend by my package layer and not from any modified fork. (you can compare my fork).

Please see the logs at:
- client line 341, 342
- server line 152 - 156

[nitrox-2019-03-17_server.log](https://github.com/SubnauticaNitrox/Nitrox/files/2975325/nitrox-2019-03-17_server.log)
[nitrox-2019-03-17_client.log](https://github.com/SubnauticaNitrox/Nitrox/files/2975326/nitrox-2019-03-17_client.log)

MarijnS95:
@acidscorch That is very odd, I'd only expect to see such a thing when the server is multithreaded and the log + send not protected by a lock. Which, looking at the code, isn't the case here since there's only one thread handling messages from all clients.
In any case, would you mind logging the threadId and stringifying the `NetIncomingMessage`, so that we can what sequence number Lidgren gave it? https://github.com/lidgren/lidgren-network-gen3/blob/42693bdb6e39039535e5f5d2da5d510c5cde363e/Lidgren.Network/NetIncomingMessage.cs#L116

Someone needs to spend some time scrutinizing Nitrox, see if there's a possibility for a misconfiguration, concurrent access/race condition or misuse of Lidgren. If not, we should extract a minimal example where reorderings occur when they shouldn't, and use that to fix Lidgren or file an issue at their tracker.

acidscorch:
This is really very odd. In the extended log i can see, that a message is received, but this message is not processed in any way. Line 329. This is strange, because the only code between the hoped next log output is a switch and not more. (i've also extended the log output to pint ticks for possibly wrong write order of the logger)

[nitrox-2019-03-17_2_client.log](https://github.com/SubnauticaNitrox/Nitrox/files/2975421/nitrox-2019-03-17_2_client.log)

image

MarijnS95:
@acidscorch How about printing the `MessageType`? Looks like it's not containing `NetIncomingMessageType.Data` and thus getting skipped entirely.
Apparently they use powers of two, but without it representing a bitmask (https://github.com/lidgren/lidgren-network-gen3/blob/master/Lidgren.Network/NetIncomingMessageType.cs#L32). Let's check if that's really the case, or if one of the bits accidentally got toggled.

Also, upon further inspection these issues with Lidgren are known to Nitrox for a while, which is the reason for switching to LiteNetLib (by means of a properly abstracted parallel implementation). While I could ask you to switch to it in the config file and call it a day, it looks like we're on the right track to solve this issue.

acidscorch:
I've finally managed to trace this thing down. 

It's related to how Lindgren manages it's own incoming packages and how to access the data that is beeing received. Update of PR is underway (need to clean up some things before). 

The PR will furthermore contain my additional solution, for the other mentioned purporses.

MarijnS95:
Interesting. Any idea what makes this read position change? Extra data that is used internally to Lidgren? Alignment?

Anyway, instead of copying the data, I recommend to extend the `packet.Deserialize` method to take the offset and length, and pass it to the `MemoryStream`:
https://docs.microsoft.com/en-us/dotnet/api/system.io.memorystream.-ctor?view=netframework-3.5#System_IO_MemoryStream__ctor_System_Byte___System_Int32_System_Int32_

acidscorch:
I've logged the whole data bytes quite a few times out and noticed that as well as the outgoing message data as the incoming data message is very often longer than the given bytes. I have tried to make sense of the additional bytes, but i'm not certain if they are crc sums or anything else. 
Accessing the .data property directly is a not good idea. 
On the incoming side i've noticed that if messages are incoming in short intervalls they are combined in the .data property separated by some other bytes. I don't know if this is an Lindgren internal buffer behavior. 
Lindgren offers explicit the two methods .WriteBytes and .ReadBytes to get the right information in and out of the messages. I don't think it would be safe usage to push this directly to a stream.

MarijnS95:
> Accessing the .data property directly is not a good idea.

That is completely correct. I don't understand *at all* why Lidgren makes this value public, instead of `internal`. The documentation summary even mentions it:

> Gets or sets the internal data buffer

It is stupid to confuse library users with that.

It makes a lot of sense to send/receive more messages in a single buffer (complete with extra internal data), as that reduces allocation and GC pressure. The actual contents of this buffer don't and shouldn't mean anything to us, it's the `NetIncomingMessage` that guides us to the right data. Since they provide their own "serialization" by means of `Read*` and `Write*`, it's a good idea not to copy the buffer unless explicitly needed, as is the case with `ReadBytes()`. Unfortunately there's no overload without arguments that just retrieves all the contents from start to end.

> And i think it may be better to let the Packet class stable with an well-formed standardized in and out behavior for serialization. (non explicit interface)

I don't think an overload with extra offset and length will do any harm (as long as one properly uses the other). Especially on a "performance critical" path that has a tendency to deal with lots of data at once.

acidscorch:
I share your opinion about the design of the Lindgren access modifiers. 

Extending the packet class is not an option for me. I won't contribute any code, from which i'm not convinced myself. 

My task was to fix the issues related to communications. I've invested a lot of time to trace this and a whole bunch of work to build the fix and test it. As you mentioned likewise, the problem has been around for a very long time now and there was a massive (?workaround?) change to be able to switch to LiteNetLib. So if this would have been an easy thing to fix, why it didn't happen much much earlier?

I'm 41 years old, have 3 children, a house, a fulltime job as project manager and lot's of other things to do. My son asked me to help him out with this, because he wants to play subnautica with some friends. Which is on hold now for some time, because the last release was not realy enjoyable. And as i can see, there are more than 150 issues open, that will need to be adressed and may be fixed. In my opinion optimization can be done after fixing game-breakers and feature completion.

This pull request is as it is. Please don't be sulk, if i won't make any other changes, except ones that i'm convinced of, that they may be harmfull. 

But hey, this is a public project, if someone of the repo-admins decides to pull this PR, feel free to create your own PR afterwards to optimize the code i've provided. I won't do it under my name. 
And on the other hand, if the repo-admins deny my PR, then it is as it is. I'd be fine with this.

MarijnS95:
@acidscorch Thanks for resolving the bug. I'm confident many will appreciate the effort, even though Nitrox has already switched to LiteNetLib.

I don't think this PR will be merged as-is, since the issue is simply a misuse of the `.Data` property. Unless you have compelling reasons to keep the ordering/reliability code, which - again - is already part of Lidgren.

You have made quite a few "statements" and had to pull personal life into that comment: I'll refrain from answering them, as it'll only end in a downwards spiral of drama.
There is one major takeaway though. Next time you're on an issue, please share your thoughts and progress during the debugging and development (for example in the comments section of said issue, or on the official discord server). This community has to work together to make Nitrox succeed, and I'm sure someone would have been able to tell you the above, saving time from writing this reliability layer.

acidscorch:
@MarijnS95 Sorry for the open words and may be directing some of the frustration towards you.

I didn't wanted to mention discord, but as you did, this may be the cause for this. What you say, is exactly what i did on discord. I've asked people and was also in some dm conversations. (you can easily ask Killzoms, InfamousJay, CriticalCookie, DragoXqwas, 4Vanu about our conversations and my ideas)
I've asked what is going on and where i may help, the most other answers were short like "they are working on the spawning thing" or even shorter "they are busy" or (...) "soon"  or even no answer at all (these people i don't name, cause i don't know reasons for this). 
I also did a request in the general chat, that someone please could hand me over to the devs. 
I don't know if you have been on the discord in the last time? My opinion is, that the community is breaking apart. There are no public infos on what is currently in work and who is at which task. So much for "the community has to work together".

I want to thank the above named persons, they all were polite and i thank them very much, that they have reacted on my requests and even gave me some little informations. If you read this > thank you!! 
Killzoms told me about this issue and i started to work on it. He also told me, that my additional logic may be too much to adress this issue. But without knowing, what the problem realy is, this was a way to to solve it.

I agree to you, that the solution i've added is an overlay over the logic of lindgren. And that it may not be needed if at lindgren and litenetlib is everything okay. And i agree to you, that it may need a little bit rework to exclude packets that need no excplicit sequential ordering. 

But i won't throw the solution away because of the following points:
- performance execution impacts if everything goes well, are 4 if statements, 2 method jumps and 2 list reference changes. this is less than rendering a single triangel
- data impact is one additional float field which is transported compressed
- solution may further be extended to cover up short disconnects of users > the cache is currently defined for 1000 packets, i'm not long enough in the project to estimate how long this could catch up if not-sequential packets are skipped, before a full client-resync is required, which i think is much more performance expensive than my solution
- in the current master, lidgren is still the default and not litenetlib

MarijnS95:
The usual Discord problem. We (obviously) have a few private channels to not get buried under general chit-chat and "release wen" questions, and in this case you should have been redirected there.

I'm not that active there anyway (only hop on a few times a week to check for new messages), and have the other channels muted - probably just like the others. I'll see if I can invite you to the other channels such that we can discuss in a way that's easier than GH comments.

Finally, I like the idea of covering up for short disconnects. However, wouldn't that be solved by simply upping the buffer sizes in Lidgren/LiteNetLib? If not, this needs to be refactored to operate over abstract network connections, such that LiteNetLib benefits from it as well. But as said before, let's discuss where to go before building anything.
The default configuration should have litenetlib by default, but then I don't know if this requires the hoster to copy that config file there in the first place - I've been out of Nitrox for too long :wink:.

DerTyan:
@MarijnS95 We can set the default to LiteNetLib and everything will work as before. The client has automatic fallback NitroxClients which try to connect to LiteNetLib or Lidgren. 😄 The hoster has nothing to do, except he wants to switch back to Lidgren.

acidscorch:
After talking with dev-team this will need more abstraction and intense testing. So this will be planned for later, after the next release. 
For the small fixes of lindgren usage and sendbuffer i create another pr.

PR #754, opened 2019-03-20
Persistent vehicle batteries

Ghaarg:
Featured:
 - Synchronization of adding/removing batteries from vehicles
 - Persistent save of said batteries on server side
 - Fixed problems with synching vehicle creation from MobileVehiclePlatform
 - Now sorts vehicles for the initial creation at connection

### DIFF
--- a/NitroxClient/Communication/Packets/Processors/StorageSlotAddItemProcessor.cs
+++ b/NitroxClient/Communication/Packets/Processors/StorageSlotAddItemProcessor.cs
@@ -1,0 +1,38 @@
+﻿using NitroxClient.Communication.Abstract;
+using NitroxClient.Communication.Packets.Processors.Abstract;
+using NitroxClient.GameLogic;
+using NitroxClient.GameLogic.Helper;
+using NitroxClient.GameLogic.Spawning;
+using NitroxModel.DataStructures.GameLogic;
+using NitroxModel.Logger;
+using NitroxModel.Packets;
+using UnityEngine;
+
+namespace NitroxClient.Communication.Packets.Processors
+{
+    class StorageSlotAddItemProcessor : ClientPacketProcessor<StorageSlotItemAdd>
+    {
+        private readonly IPacketSender packetSender;
+        private readonly StorageSlots storageSlots;
+
+        public StorageSlotAddItemProcessor(IPacketSender packetSender, StorageSlots storageSlots)
+        {
+            this.packetSender = packetSender;
+            this.storageSlots = storageSlots;
+        }
+
+        public override void Process(StorageSlotItemAdd packet)
+        {
+            Log.Debug("Process storage slot add");
+            ItemData itemData = packet.ItemData;
+            GameObject item = SerializationHelper.GetGameObject(itemData.SerializedData);
+
+            // Mark this entity as spawned by the server
+            item.AddComponent<NitroxEntity>();
+            item.SetNewGuid(itemData.Guid);
+
+            storageSlots.AddItem(item, itemData.ContainerGuid);
+            Log.Debug("Process storage slot add finished");
+        }
+    }
+}
--- a/NitroxClient/Communication/Packets/Processors/StorageSlotAddItemProcessor.cs
+++ b/NitroxClient/Communication/Packets/Processors/StorageSlotAddItemProcessor.cs
@@ -22,8 +22,7 @@ namespace NitroxClient.Communication.Packets.Processors
         }
 
         public override void Process(StorageSlotItemAdd packet)
-        {
-            Log.Debug("Process storage slot add");
+        {            
             ItemData itemData = packet.ItemData;
             GameObject item = SerializationHelper.GetGameObject(itemData.SerializedData);
 
@@ -31,8 +30,7 @@ namespace NitroxClient.Communication.Packets.Processors
             item.AddComponent<NitroxEntity>();
             item.SetNewGuid(itemData.Guid);
 
-            storageSlots.AddItem(item, itemData.ContainerGuid);
-            Log.Debug("Process storage slot add finished");
+            storageSlots.AddItem(item, itemData.ContainerGuid);            
         }
     }
 }
--- a/NitroxClient/Communication/Packets/Processors/StorageSlotAddItemProcessor.cs
+++ b/NitroxClient/Communication/Packets/Processors/StorageSlotAddItemProcessor.cs
@@ -27,7 +27,11 @@ namespace NitroxClient.Communication.Packets.Processors
             GameObject item = SerializationHelper.GetGameObject(itemData.SerializedData);
 
             // Mark this entity as spawned by the server
-            item.AddComponent<NitroxEntity>();
+            if (item.GetComponent<NitroxEntity>() == null)
+            {
+                item.AddComponent<NitroxEntity>();
+            }
+            
             item.SetNewGuid(itemData.Guid);
 
             storageSlots.AddItem(item, itemData.ContainerGuid);            
--- a/NitroxClient/Communication/Packets/Processors/StorageSlotAddItemProcessor.cs
+++ b/NitroxClient/Communication/Packets/Processors/StorageSlotAddItemProcessor.cs
@@ -24,13 +24,7 @@ namespace NitroxClient.Communication.Packets.Processors
         public override void Process(StorageSlotItemAdd packet)
         {            
             ItemData itemData = packet.ItemData;
-            GameObject item = SerializationHelper.GetGameObject(itemData.SerializedData);
-
-            // Mark this entity as spawned by the server
-            if (item.GetComponent<NitroxEntity>() == null)
-            {
-                item.AddComponent<NitroxEntity>();
-            }
+            GameObject item = SerializationHelper.GetGameObject(itemData.SerializedData);            
             
             item.SetNewGuid(itemData.Guid);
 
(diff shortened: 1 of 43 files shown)


Ghaarg:
There is still the ToDo to save the actual battery amount on the server, as they will not be updated at the moment.
I plant to implement that in a bigger update, that synchronizes/reworks power for all sides.

NitroxClient/Communication/Packets/Processors/StorageSlotAddItemProcessor.cs
```diff
@@ -0,0 +1,36 @@
...
+            this.storageSlots = storageSlots;
+        }
+
+        public override void Process(StorageSlotItemAdd packet)
+        {            
+            ItemData itemData = packet.ItemData;
+            GameObject item = SerializationHelper.GetGameObject(itemData.SerializedData);
+
+            // Mark this entity as spawned by the server
+            item.AddComponent<NitroxEntity>();
```
Sunrunner37: Will this add multiple of the same component if a player stores it, picks it up, then stores it again?

NitroxClient/GameLogic/Constructor.cs
```diff
@@ -62,11 +64,50 @@ public void BeginCrafting(GameObject constructor, TechType techType, float durat
...
+                constructedObject.AddComponent<NitroxEntity>();
+                SendBatterySlots(constructedObject, childIdentifiers);
             }
             else
             {
                 Log.Error("Could not send packet because there wasn't a corresponding constructed object!");
             }
         }
+
+        private void SendBatterySlots(GameObject constructedObject, List<InteractiveChildObjectIdentifier> childIdentifiers)
```
Sunrunner37: Why do we need to send the battery slots after the object begins crafting?  Shouldn't these already be taken care of by `InteractiveChildObjectIdentifiers`?

NitroxClient/GameLogic/InitialSync/StorageSlotInitialSynchProcessor.cs
```diff
@@ -0,0 +1,72 @@
...
+using NitroxClient.GameLogic.Spawning;
+using NitroxModel.DataStructures.GameLogic;
+using NitroxModel.DataStructures.Util;
+using NitroxModel.Logger;
+using NitroxModel.Packets;
+using UnityEngine;
+
+namespace NitroxClient.GameLogic.InitialSync
+{
+    class StorageSlotInitialSynchProcessor : InitialSyncProcessor
```
Sunrunner37: Typo in the name

NitroxClient/GameLogic/StorageSlots.cs
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+
+        public void BroadcastItemAdd(InventoryItem item, GameObject gameObject)
+        {
+            string guid = GuidHelper.GetGuid(gameObject);
+            NitroxEntity entity = gameObject.GetComponent<NitroxEntity>();
+            if (entity == null)
+            {
+                return;
```
Sunrunner37: I'm not sure that we can rely on NitroxEntity (btw this seems like a silent failure case).  It's original purpose was to mark items so we know what was authoritatively spawned but has kind've scope creeped to all entities.  It may be something we want to deprecate because it is losing meaning.

NitroxClient/GameLogic/StorageSlots.cs
```diff
@@ -0,0 +1,131 @@
...
+            }
+
+            StorageSlotItemRemove slotItemRemove = new StorageSlotItemRemove(guid);
+            packetSender.Send(slotItemRemove);
+        }
+
+        public void AddItem(GameObject item, string containerGuid, bool silent = false)
+        {
+            GameObject owner = GuidHelper.RequireObjectFrom(containerGuid);
```
Sunrunner37: This should throw an exception if it does not get a game object back.  You don't have to check for null.

NitroxClient/GameLogic/StorageSlots.cs
```diff
@@ -0,0 +1,131 @@
...
+            if (entity == null)
+            {
+                return;
+            }
+
+            StorageSlotItemRemove slotItemRemove = new StorageSlotItemRemove(guid);
+            packetSender.Send(slotItemRemove);
+        }
+
+        public void AddItem(GameObject item, string containerGuid, bool silent = false)
```
Sunrunner37: What are the use-cases for wanting to / not wanting to play sound?  Can we pick one or the other instead of switching?

NitroxClient/GameLogic/StorageSlots.cs
```diff
@@ -0,0 +1,131 @@
...
+                return;
+            }
+            // only need to watch EnergyMixin slots for now (only other type will be propulsion cannon)
+            Optional<EnergyMixin> opEnergy = Optional<EnergyMixin>.OfNullable(owner.GetComponent<EnergyMixin>());
+            if (opEnergy.IsPresent())
+            {
+                EnergyMixin mixin = opEnergy.Get();
+                StorageSlot slot = (StorageSlot)mixin.ReflectionGet("batterySlot");
+
+                // Maybe need to suppress batterysound play
```
Sunrunner37: Seems like we are doing this below?

NitroxServer/Communication/Packets/Processors/StorageSlotRemoveItemProcessor.cs
```diff
@@ -0,0 +1,28 @@
...
+
+        public StorageSlotRemoveItemProcessor(PlayerManager playerManager, InventoryData inventoryData)
+        {
+            this.playerManager = playerManager;
+            this.inventoryData = inventoryData;
+        }
+
+        public override void Process(StorageSlotItemRemove packet, Player player)
+        {
+            // Only need to send this, if it is a shared owner
```
Sunrunner37: What do you mean by shared owner?

NitroxClient/Communication/Packets/Processors/StorageSlotAddItemProcessor.cs
Ghaarg: I "stole" that piece of code from ItemContainerAddProcessor. But we can test that. If it does, then we can change it on both parts

NitroxClient/GameLogic/Constructor.cs
Ghaarg: SendBatterySlots will fill the StorageSlots with default batteries for the children, that are of EnergyMixin/BatterySource type. This needs to be done because i suppress the default batteries, that will be spawned for every vehicle.
I suppress them cause default batteries will be spawned for someone who connects. And these default batteries would cause many battery packets that will cause errors on other clients (cause guids are not synched) and also these default batteries shall not override the real batteries

NitroxClient/GameLogic/InitialSync/StorageSlotInitialSynchProcessor.cs
Ghaarg: You mean StorageSlots?

NitroxClient/GameLogic/StorageSlots.cs
Ghaarg: We could leave it out. I use it for the initial synch so you are not hearing est. 200 battery switching sounds

NitroxClient/GameLogic/StorageSlots.cs
Ghaarg: Its for skipping the default batteries. We could info log that.
I need a way to distiguish synched batteries from unsynched ones (default spawned batteries issue). Maybe we have a better way doing that but i do not know how.

NitroxClient/GameLogic/StorageSlots.cs
Ghaarg: It will anyway cause of the validate inside requireObject. I can leave that out.
Again i stole the code a bit from ItemContainer

NitroxServer/Communication/Packets/Processors/StorageSlotRemoveItemProcessor.cs
Ghaarg: Bad wording. It means if someone tries to remove a battery, that is not there in the first place, it will not send the info to other players (cause slot is empty anyways)

NitroxClient/Communication/Packets/Processors/StorageSlotAddItemProcessor.cs
Ghaarg: I testet it and it seems the NitroxEntity component will not be serialized with the item.
Nontheless i build a statement around it so it will never add a second NitroxEntity

NitroxClient/GameLogic/StorageSlots.cs
acidscorch: Maybe an overload to the GuidHelper.GetGuid can help. I had the same problem.
The GetGuid generates a new uniqueguid if it can't find one. 
Suggestion:
.GetGuid(gameobject, bool allowEmpty) > Returns emtpy.string if no guid found. So you can see if it is a synced one.

NitroxClient/GameLogic/InitialSync/VehicleInitialSyncProcessor.cs
```diff
@@ -17,6 +18,7 @@ public VehicleInitialSyncProcessor(Vehicles vehicles)
 
         public override void Process(InitialPlayerSync packet)
         {
+            packet.Vehicles.Sort(new VehicleModelComparer());
```
Sunrunner37: Why are we doing this sorting?

PR #755, opened 2019-03-20
Simple Lidgren Fix

acidscorch:
- changed retrieving of Message.data
- changed .SendBuffer Size
~~- switched default Client-Communication to LiteNetLib and Lidgren as fallback~~


DerTyan:
If we want to set LiteNetLib as default, then you have to change the default in the server config file as well.

acidscorch:
You are right. I removed the switch from the pr and let this for someone else if needed. I have only tested with lidgren, So i don't know if it is safe or not.

PR #756, opened 2019-03-22
Fabricator Multiplayer Usage + Fix duplicate Item drop

acidscorch:
Objective:
- extend Fabricators synced in Multiplayer
- fix crafted Item duplication on nearby remote Players
- supress remote player usage of currently used Fabricator/Workbench
- supress remote player deconstruction of currently used Fabricator/Workbench 

Changes:
- removed former try_pickup variant
- refactored GameLogic.Crafting to handle all local and remote Events in one Place
- consolidated crafting relevant Patch-Event-Class-Files in one File for each original patched Class 
- extended Log for output milliseconds

Open ToDos:
- translate Display-Text

fixes#749


acidscorch:
needs rework in case of deconstructing the fabricator

NitroxClient/GameLogic/Crafting.cs
```diff
@@ -9,27 +12,453 @@ namespace NitroxClient.GameLogic
 {
     public class Crafting
     {
+        internal static Crafting Instance;
```
Sunrunner37: We should not use a singleton here, instead opting to use our dependency injection framework AutoFac.  Please review how the NItroxServiceLocater is used in other places.

NitroxClient/GameLogic/Crafting.cs
```diff
@@ -9,27 +12,453 @@ namespace NitroxClient.GameLogic
...
         {
+            Instance = this;
             this.packetSender = packetSender;
+            timer.AutoReset = true;
+            timer.Elapsed += new System.Timers.ElapsedEventHandler(OnTimerEvent);
+            timer.Start();
+        }
+
+        //Applies to Fabricator and Workbench
+        public void GhostCrafter_Post_OnHandHover(GameObject gameObject, GUIHand hand)
```
Sunrunner37: Please look at our contributing guidelines and make sure all code is consistent with the rest of the code base:
https://github.com/SubnauticaNitrox/Nitrox/blob/master/CONTRIBUTING.md

Just a few observations in this method:
* No snake case for method naming; this is only acceptable for patch class names.
* No need to fully qualify static calls - use imports to reduce noise.
* No need to pragma debug log statements as we can disable debug log level for releases.  This will reduce the amount of noise.
* Avoid "_" in local names as this is not consistent with the rest of the code base.

NitroxClient/Communication/Packets/Processors/CraftingProcessor.cs
```diff
@@ -0,0 +1,85 @@
...
+using NitroxClient.Communication.Packets.Processors.Abstract;
+using NitroxClient.GameLogic.Helper;
+using NitroxClient.Unity.Helper;
+using NitroxModel.Core;
+using NitroxModel.Packets;
+using UnityEngine;
+
+namespace NitroxClient.Communication.Packets.Processors
+{
+    public class CrafterStartUseProcessor : ClientPacketProcessor<CrafterStartUse>
```
Sunrunner37: What is the reason for moving from Fabricator to Crafter?  Is crafter what is being used internally?  I prefer we stay with fabricator as that is what users will intuitively search for.

NitroxPatcher/Patches/BuilderTool_Patch.cs
```diff
@@ -4,19 +4,69 @@
...
+        }
+
+        public override void Patch(HarmonyInstance harmony)
+        {
+            PatchPostfix(harmony, TARGET_METHOD);
+        }
+    }
+
+
     public class BuilderTool_HandleInput_Patch : NitroxPatch
```
Sunrunner37: Our pattern is a single file per method patch.  This makes it easy for people to see if their given method has already been patched before.

NitroxClient/Communication/Packets/Processors/CraftingProcessor.cs
```diff
@@ -0,0 +1,85 @@
...
+            this.packetSender = packetSender;
+        }
+
+        public override void Process(CrafterStartUse packet)
+        {
+            GameLogic.Crafting.Instance.GhostCrafter_Remote_OnStartUse(packet.CrafterGuid, packet.PlayerName);
+        }
+    }
+
+    public class CrafterEndUseProcessor : ClientPacketProcessor<CrafterEndUse>
```
Sunrunner37: Our pattern is a single file per packet type.  This allows users to easily jump to the file they need to edit without hunting through various files.

NitroxClient/GameLogic/Crafting.cs
acidscorch: Okay, i will change this.

NitroxClient/GameLogic/Crafting.cs
acidscorch: Sorry, for violating the guidelines. I've been taught naming rules different than used here (but that's quite a while ago now).

- No need to fully qualify > Okay, i will change this.
- Avoid "_" in local names > Okay, i will change this.

To the other two Points, i would like to explain, why i changed the way of naming and usage.

Snake case naming:

For my understanding, the main purpose of this mod is to handle events from the local game-logic and the remote game-logic (server or other players). Events from the local game-engine come from patched methods and can be Pre, Post or Injected via transpiler. The patched method itself is settled in a base-class of the Subnautica-Assembly. Looking at the used method names i can tell at a glance, when this method is called and where it comes from. The first part (class name) and the last part (patched method) are exactly the same as used in Subnautica.

Pragma debug:

That's another thing i've been thaught different. And for me, this way of using dev infos is the best i've ever seen. Why?

I've started with the master from 3 weeks ago and on every log file check i've hundreds of lines from the spawning investigation which are scattered around many files. They relate to investigations of an other Contributor and are not really relevant to me. And i can't easily swith them off. I must go to each line and uncomment it. 
Placing log-statements for investigation and bug fixing is necessary. To remove them to have clean code is not necessary. If any Issue is opened related to this logic part, the bugfixing developer starts all over again and puts in again many lines of logging and removes it after solving the issue. It's work that doesn't need to be done. But the log-output should only conentrate on the investigated theme(s).  
Using normal log statements will have the need to make them filterable (not only debug or trace on/off), which means to extend the logger-class with some filtering-logic and maintain the filters that can be used. Filtering is especially necessary when having OnHover logging or similar.
Not all dev-logging is intended to be in the release logging. Even not if debug log is activated at the customer side. 
When using pragma in the right way:
- Every other developer can take the classes and build them and nothing comes out to logging by default. He has to explicitly activate the pragmas before build. In normal use, no need to manage and apply any filters first for blending out the log-statements of other functions. 
- Additional defines can be easily set by everyone without touching logger-classes and they have no effect on someone others logging needs (the only thing to keep in mind is, that the stated pragmas in the .csproj file must not be included in a pull request). 
- No need to concern about which information is logged in a normal debug or release build, because these lines don't get even compiled in these builds
- Normal logging statements are always executed, even if they are filtered later. This may have impacts e.g. if every second hover action or runtime methods are logged.

NitroxClient/Communication/Packets/Processors/CraftingProcessor.cs
acidscorch: The processing of these packets relate to both fabricator and workbench. The underlying base logic is the crafter logic from which both are derived. The class that does the magic in nitrox is also named Crafting.

NitroxPatcher/Patches/BuilderTool_Patch.cs
acidscorch: On investigating the crafter logic i've notived that correctly following, which event is when triggered it is necessary to jump through many different files. And taking over logic from other classes at generating new method patches also needs to jump to other files. The other point is, that for a logic process not only one patched method of Subnautica class is needed rather all of them, to see how which method is modified and which event it triggers at which logic class.

NitroxClient/Communication/Packets/Processors/CraftingProcessor.cs
acidscorch: Not to have to jump through files was the intended goal. All processors that are related to the Crafting-Logic are concentrated here. 
As already mentioned before the point of view for this architecture is:

[Pach-Classes] >> CallMethods > [Logic-Class with the magic (e.g. Crafting)] < CallMethods << [Processors (e.g. CraftingProcessor)]

NitroxClient/GameLogic/Crafting.cs
```diff
@@ -9,27 +12,453 @@ namespace NitroxClient.GameLogic
...
+        {
+#if TRACE && GAMEEVENTCRAFTING
+            NitroxModel.Logger.Log.Debug("GhostCrafter_Post_OnHandHover");
+#endif
+            //if Crafter is in use by remote Player > Display
+            string _crafterGuid = GuidHelper.GetGuid(gameObject);
+            bool _inUseByOther = false;
+            string _playerName = string.Empty;
+            lock (fabricatorsInUseByRemotePlayers)
```
Sunrunner37: You don't need locks on the client side as unity always runs everything in a single thread included incoming packets.

NitroxClient/GameLogic/Crafting.cs
Sunrunner37: For the two contention points:

Snake Case - Only patch classes should be snake case to easily differentiate the class and method that are being patched.  Having gamelogic understand that it is being invoked by a patch is inappropriate coupling.  GameLogic classes, such as this Crafter class, should have no knowledge of being invoked via pre/post as they may also be invoked by a packet processor or some other process.  Ideally these methods be made more generic like `RegisterCrafterInUse()`

Pragma - My professional background is not in c# so I had to do a little research.  I'm actually not able to find sources advocating this style of logging even after expanding my search to c++.  To me the trade-off is quite clear, having 3 lines per log statement(two of which are not indented) will be unreadable if you were to have several log statements in a single method.  If desired, the same behavior could be achieved via an enumeration with similar benefits excluding the mostly negotiable compile time benefits.  However, coming from a language without pragmas, and an environment where we log almost every detail, I've never really seen the need but wouldn't protest if we did that.

NitroxClient/GameLogic/Crafting.cs
```diff
@@ -9,27 +12,453 @@ namespace NitroxClient.GameLogic
...
+
+                if (_inUseByOther && _crafterGuid != string.Empty)
+                {
+                    return false;
+                }
+
+                return true;
+            }
+            else
+            { return true; }
```
Sunrunner37: minor: formatting

NitroxClient/GameLogic/Crafting.cs
```diff
@@ -9,27 +12,453 @@ namespace NitroxClient.GameLogic
...
+                lock (fabricatorsInUseByRemotePlayers)
+                {
+                    _inUseByOther = fabricatorsInUseByRemotePlayers.ContainsKey(_crafterGuid);
+                    if (_inUseByOther)
+                    {
+                        _playerName = fabricatorsInUseByRemotePlayers[_crafterGuid].PlayerName;
+                    }
+                }
+
+                if (_inUseByOther && _crafterGuid != string.Empty)
```
Sunrunner37: The guid shouldn't be empty as GuidHelper should assign it a guid.

NitroxModel/Packets/FabricatorEndCrafting.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+using NitroxModel.DataStructures;
+
+namespace NitroxModel.Packets
+{
+    [Serializable]
+    public class FabricatorEndCrafting : Packet
```
Sunrunner37: Should this also be renamed to Crafter?

NitroxPatcher/Patches/BuilderTool_Patch.cs
```diff
@@ -4,19 +4,69 @@
...
         public static readonly OpCode INJECTION_OPCODE = OpCodes.Callvirt;
-        public static readonly object INJECTION_OPERAND = typeof(Constructable).GetMethod("SetState", BindingFlags.Public | BindingFlags.Instance);
-        
+        public static readonly object INJECTION_OPERAND = typeof(Constructable).GetMethod("SetState", BindingFlags.Public | BindingFlags.Instance);
+
+        public static bool Prefix(BuilderTool __instance)
+        {
+            bool _result = true;
+
+            _result = NitroxServiceLocator.LocateService<Crafting>().BuilderTool_Pre_HandleInput(__instance.gameObject);
```
Sunrunner37: As stated above, we should give all of these good, descriptive names.  From here it is not clear what the result entails.

NitroxModel/Packets/FabricatorEndCrafting.cs
acidscorch: Yes, you are right, this would be coherent.

NitroxClient/GameLogic/Crafting.cs
acidscorch: Yes. This is a relict of having the GetGuid method overloaded, so that it could return an empty string, to see if guid transfering and assigning has properly happend or something went wrong and the guid wasn't already set.

NitroxClient/GameLogic/Crafting.cs
acidscorch: okay

NitroxClient/GameLogic/Crafting.cs
acidscorch: The lock is for the use of the System.Timer instead of a WindowsForms.Timer. I went to the System.Timer because WindowsForms is not part of the nitroxclient. System.Timer could be connected to a SynchronizingObject. But i don't know if there is one available which implements ISynchronizeInvoke from Subnautica and if using it would interfere with the GUI and Rendering logic from Subnautica. 
The System.Timer is relative easy to implement, but creates a WorkerThread. Thats the reason for the lock. 

An alternate approach would be, to not use the timer. Instead hook on the event when a remote player is disconnected. That also wouldn't need a "well defined" wait time for the duration until the usage is freed from the crafter. (which is always hard to define, because may be someone is reaching multiple storages from the crafter spot and can possibly use the crafter for a while, when crafting multiple items one after another).
In case something went wrong with unlocking the crafter after normal use, the player which is still   displayed as blocking user will then have to disconnect and reconnect, or use the crafter another time and free it again, which should be no problem for the players.

Issue #691, opened 2019-01-30
Overall Building Bugs


ScriptedAnar:
There are many building bugs which make the game really unplayable 

- [ ] 1.Reinforcements arent Sync with other players

- [ ] 2.Many things that you build when you rejoin they arent full constructed example moonpolls some Multiporpuse rooms Reinforcements and etc

- [ ] 3.Fabricators that are in bases when you rejoin they are unpowered and you must reconstruct them every time you join the server (That may happens with Modification Stations too)

- [ ] 4.Every time you join the game the power is always 0 

- [ ] 5.Placing posters in a wall it isnt visible to other players 

- [ ] 6.Bio/nuclear reactors (known bug) 

- [ ] 7.Solar Panels are "flying" from the base #668 

- [ ] 8.Upgrade Console and water filtration are super buggy and position is bugged when you rejoin #688 

Probably there are more these are what i have found for now.

hanawovino:
I can confirm these issues, especially multipurpose rooms. If you rejoin the server, they are back at 97% or something. Items (like the water filtration) you placed inside of them are now glitched (show up as red) and can't be destructed or rebuilt anymore.

Tinkertee:
I have the same issue on my end. I restarted the server and Subnautica and it's all screwed up for me aswell. ( Happened while my husband was out resource gathering and I was building - He was bugged and couldnt see most of the buildings, upon reset/restart it was buggy for me aswell )

FlameCraft1234:
This is very true it happens to me.

acidscorch:
Hello NitroxPower, 

are these all building related known issues? Could you please check again all issues and reference/concentrate them here? I would like to have this issue as single work-list.

For the following points i have fixes ready, but would like to finish first all issues before creating a PR.
Fixes:
- BaseLayout save/load (structures, inferiour, furniture saved/reloaded at the correct location)
- BuiltOrder correct save and ordering logic for InitialSync after reconnect: base-main-structures (e.g. corridors, foundations, rooms) before base-attachments (e.g. reinforcements, windows) before interiour before furniture  (fixes flying solar-panels, because they are now placed after the grounding base-structure)
[on my todo list: relation-list for reinforcements, to place them directly after the related base-main-structure is placed, to avoid structure-leaks at initial sync e.g. when moonpool is placed before reinforcements]
- Deconstruction is now synced and saved for nearly all objects [on my todo list: filtermachine, reactors, waterpark)
[...]

PR #782, opened 2019-04-07
Fixed building all kinds of tubes

CtrlAltDavid01:
This SEEMS to fix the building problem, let me know if I broke something.

- The tubes are now building properly
- Deconstruction is not fixed
- When deconstructing other things can deisappear too

I'll try to fix deconstruction too
(I tested it with multiple clients)


acidscorch:
The problem is more complex:
- duplicated or overlapping items in save file (with same positions, so the multilplayer builder can't place them which causes guid messed up)
- base_geometry patch reaplies guids from deconstructed objects (which also causes guid messed up)
- begin deconstructing hook doesn't transfer guids in all cases to deconstructable
- bases spawn when deconstructing a middle Element of long reaching corridors which needs reassing of new Base guid to objects
- deconstructed objects are not correctly removed from build History
- Initialplayer sync is Not ordered so attached objects (like solarpanel are placed before the grounding Base piece)

CtrlAltDavid01:
I noticed it while trying to debug the thing, but this fixed the building part, I reloaded the server multiple times, I had 2 clients online and while I didn't deconstruct anything it worked.
I hope I'll understand the problem.

acidscorch:
I've already quite a lot of tracing output in my test-code. I will test your change the next days and have a look, if it also works with more complex use cases.

acidscorch:
Hmm... at first review, i can't see a difference to the original patch for the constructor. The Method "MakeCellDeconstructable" is called directly after the "Constructor". 
Have a look at my log and search for "BaseDeconstructable_Post_Constructor" (there may be 2 or 3 lines for this method call, if a guid needs to be updated) and directly after that the "MakeCellDeconstructable" is called. 
https://www.dropbox.com/s/d8kck16kax5yshe/nitrox-2019-04-07.log?dl=0

Maybe i'm missing something, but for me it seems that both ways lead to the same endresult after calling.

CtrlAltDavid01:
Okay, I'll have a look, I thoughtthe same thing, but no matter how many times I tested, the corridors finnishd building correctly with this patch and without it it broke instantly.
Maybe I just didn't find the real reason for the problem.

CtrlAltDavid01:
I know what Idid differently, I  store the transfor gameobject, not the BaseDeconstructable.
And I  forgot to rename the .cs file.

acidscorch:
Hmm, sounds interesting. Besides the log shows the same guids for BaseDeconstructable.GameObject and Geometry.GameObject, so i think they are the same. Okay, i will make more and deeper tests tomorrow.

NitroxPower:
Here i had reported all the building issues i had found so far if you want you can take a look #691

CtrlAltDavid01:
It seemed to fix the reinforments and the windowsthatcan be added, but not 100% sure about that

acidscorch:
Hi, i've tried different variants and also broken save games. 
You are right that "MakeCellDeconstructable" is not called for all interiour base pieces that are connected to a base piece (windows, reinforcements, ladders, reactors, watermachine, modulstation, watertank and so on).
So the guid transfer for outer objects is no more interfered by this method.
I need to check now deeper how to make the guid transfer for these objects or if we even need the transfer here. Maybe they can be handled like normal constructables.

acidscorch:
I think that this change can be pulled. For now it skips the problematic objects and skips also the problem, that e.g. ladders or tanks generate two basedeconstructables instead of one. Maybe we have to change back later to constructor, when the other objects are correct handled.
But for now it's a little improvement to the current state and shouldn't brake something.

CtrlAltDavid01:
I'll rename the file correctly  and then it canbe pulled

acidscorch:
Hello CtrlAltDavid01,

could you have a look at Issue#691 and maybe test the solution?

CtrlAltDavid01:
Okay, It fixed point 1,2. Posters place, but they are falling infinitely.

Sunrunner37:
Closing this out as there has been a lot of separate movement on this issue in https://github.com/SubnauticaNitrox/Nitrox/issues/951

PR #793, opened 2019-04-15
Cleanup Visual Studio Solution

acidscorch:
- removed dynamic dependencies for stabilisation (static package libs are now in lib folder, if an update of a package.dll is needed, the new .dll has to be placed in lib folder)
- updated hint-paths for use of lib folder (all dependencies should be referenced from there now)
- removed unneeded projects harmony and lidgren
- deactivated compile of NitroxTest-Project (don't know if this is still needed)

Reason:
- solution is now compile-ready after rebasing from master without the need to adjust references


Ghaarg:
The dlls in packages are from nuget. As i need for my udp punch server that uses the dotnet core version of litenetlib, this solution wont work for me.

I also never got any errors after setting DevVars (i may needed to hit compile multiple times cause VS can be an idiot, but thats it)

acidscorch:
Updates on external Packages could break current implementations, so i would prefer to stay on the current version and only update the libs, if there is a special reason or need to do so. 

As the punch server isn't listed here, should it be no problem to use the core version of litenetlib only in this separate project/branch until it is finished for release or github? The communication layer should work with a static .dll on the counter side of nitrox too. 

The DevVars try to grab the .dlls from the Subnautica Path (and yes or any other defined source-path). If there is a fresh clean Installation of Subnautica with no previous Nitrox or Launcher Installation, the Solution wouldn't find some files at all.

MarijnS95:
This makes no sense at all.

The entire reason for package managers to exist is to manage dependencies and versioning, without having to set up a distribution pipeline or blow up the repo with binaries.

The libraries are no "dynamic dependencies" (whatever that means). They are statically defined in the various `packages.config` files, and downloaded by the de facto standard, built-in NuGet package manager. Their versions are pinned and will only be bumped up when needed, after being proven working.

Harmony requires a few modifications to be suitable for Nitrox, which is why it is included as a separate repo (submodule). This does not only allow any contributor to see, improve or add code to the Harmony fork, it also removes any doubts about a precompiled binary from an unknown source.

Lidgren is included as a submodule because it's outdated on NuGet. Again, we'll not include a binary from unknown origin but instead compile it from source.

The "DevVars" system doesn't just copy game libraries for compiling; it also deploys Nitrox binaries back to the game folder to allow quick iteration. This way the entire setup is automated and resilient to game updates. There should be no issue with a fresh install of Subnautica, unless it's in a different path than the default (before you ask: we're not going to scan the registry for it).

Finally, I don't see why NitroxTest has to be removed from compiling with Debug on AnyCPU. This project contains a ton of automated tests to "guarantee" correctness, which unfortunately doesn't seem to be used by all contributors these days... :disappointed: 

Takeaway: After cloning Nitrox (which should automatically clone the submodules on modern versions of git) and eventually setting a custom Subnautica path if the installation is located elsewhere, everything should work out of the box. If not, you're probably doing something wrong. Feel free to ask around and post logs: we're happy to help and improve the build system where necessary.

Issue #794, opened 2019-04-15
Global Y-Coordinate missmatch between Clients - Help Wanted


acidscorch:
The Y-Coordinate (height level) isn't correctly between clients. Data transfers at packets for the value are correct. Maybe culture format related at client side. Noticed with players from different countrys. 

This causes remote player positioning half room-height below or above current player level and leads to misplaced Objects at building.

NitroxPower:
Its not the reason that we from different country it was happening with my friends too when we was playing our survival (version 1.3.0.1).

acidscorch:
It seems like, that someone else noticed the Y-mismatch a while ago at first time spawning the player in the escape pod and simply tried to solve it via moving the spawning player a little bit up. That's also the reason, why i'm falling a little, when connecting with a new player to the multiplayer game.

For me the values should be (0, -0.2f, 0) to result in the same coordinates after the local player has spawned and finished falling without performing any movement. I'm searching further, why the client-world-coordinates differ from client to client. 

![image](https://user-images.githubusercontent.com/48223659/56286332-bdeccc00-6119-11e9-9d39-eacf0c396c1b.png)

Issue #128, opened 2018-02-04
[Bases / Vehicles] Fix player sinking in subroots


Sunrunner37:

RabidCrab:
Currently working on player sinking/sliding in the Cyclops. Walking/running is currently functional, but jumping and the cyclops turning is still a work in progress.

PR #798, opened 2019-04-20
Fix player sinking and relative rotation

MarijnS95:
Fixes https://github.com/SubnauticaNitrox/Nitrox/issues/128
Fixes https://github.com/SubnauticaNitrox/Nitrox/issues/794

Address Y-offset issues by removing the broken "interpolation"/"smoothing".
Also, make rotation relative to the subroot (just like positioning). This is to keep the player properly oriented when the subroot moves without (timely) receiving position updates from other players in this subroot.

### Next up
We need to make a light-weight "RigidBody" controller, since it's only used to integrate velocities after all. The current implementation will keep interpolating the velocity causing players to fly away when no packets are received. The new implementation should apply some kind of smoothing towards the end and hold the player stationary when no packets are received for a prolonged amount of time.
Thereafter smoothing for remote controlled Cyclopses needs to be addressed. Players in this cyclops seem to stutter since the game probably doesn't account for the cyclops moving "on it's own" (without the player fixed in the piloting seat). This can cause a player to fall out when standing close to the wall or inside the helm.

### DIFF
--- a/NitroxClient/GameLogic/MovementHelper.cs
+++ b/NitroxClient/GameLogic/MovementHelper.cs
@@ -43,11 +43,9 @@ namespace NitroxClient.GameLogic
                 // This should be a one-off teleport.
                 gameObject.transform.position = remotePosition;
             }
-            // overcorrections can cause jitter when standing still.
-            else if (distance > 0.001f)
+            else
             {
-                float maxAdjustment = (float)Math.Log10(distance) * 4f;
-                remoteVelocity += MathUtil.ClampMagnitude(velocityToMakeUpDifference - remoteVelocity, maxAdjustment, -maxAdjustment);
+                remoteVelocity = velocityToMakeUpDifference;
             }
 
             return remoteVelocity;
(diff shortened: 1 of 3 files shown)


acidscorch:
For Cyclops:
I've tried a little bit with setting the withdriving player in LockedMode as long as the cyclops is performing the move commands. Fall offs didn't occure anymore and every movement of the cyclops was directly transfered to the player without hops/shutter or anything else.
 The Problem is, that the camera is also resetted by this Method. So i tried to just set the lockmode directly and attach the cyclops transform as parent of the player transform, so that every change on the cyclops transform is inherited to the player transform. But it seems that on the direct approach something is missing versus the LockMode Method, because hops and fall offs occured again.

acidscorch:
To completely skip over the velocity of a remote player, there is also the method .MoveToPosition, which also creates the right animations at swimming or entering a base or using tools (like seaglide, etc)

MarijnS95:
Sorry @acidscorch, completely forgot to answer.

I don't think `LockedMode` is the way to go. It is used specifically to lock the player camera, while we only want the parent-transform part. A better solution will be investigated.

What's the advantage of `MoveToPosition`? All animations seem work just fine when using velocity, if that's what you mean.
That function will introduce little teleports instead of smooth player movements, which isn't elegant nor realistic. A velocity-based system really isn't that much more work for all the extra gain.

Issue #796, opened 2019-04-15
SaveGame-Management - Enhancement Request - Help Wanted


acidscorch:
Task Package 1:

- [ ] deactivate ingame Save-Option from Menu (screen1 below)

Task Package 2:

- [ ] stop server game simulation when last active player disconnects
- [ ] start server game simulation when no player is connected and a first player connects to server

Task Package 3:

Extend the Client-Login Screen for Save-Game Management (background: some people use dedicated servers and don't run them locally via launcher, additionaly this is a nicer way to handle local save-games then editing and copying save.nitrox file)

- [ ] after connect to server show Save-Game Screen (screen 2, it is only a suggestion)
- [ ] options for new, rename, delete and stop simu require to type in server password 
- [ ] if server is currently simulating a game (active players in game) > the current active save game is colored/highlighted/etc. and rename/delete is inactive for this selection, join game is only active if this line is selected and not for all when server is not active. Edit-options should be active for the other lines, so editing could be done without having the need to stop the current game
- [ ] if server is not currently simulating a game (logged in user is the first player), a savegame can be selected and with join game the server loads this game (see Task-Package2)

Notes:

If you like this Enhancement Request, please leave a thumbs up, so it can be seen if there is interest in this Enhancement.

If you are a Dev and like to work on the tasks, please leave a short comment, so that it can be seen who is working on it and can avoid not necessary doubled parallel work. 

Greetings, AcidScorch

![screen1](https://user-images.githubusercontent.com/48223659/56164075-3778a300-5fd0-11e9-999c-189cf9cf7939.JPG)

![screen2](https://user-images.githubusercontent.com/48223659/56164089-40697480-5fd0-11e9-9894-7919e82a5b1d.JPG)

Jannify:
The option to select save games is nice 👍 but probably better to have in the launcher.

dartasen:
Save managing is #958 related

PR #808, opened 2019-05-21
Removed removed save button and stopped save on quit

Coding-Hen:


Coding-Hen:
Removed inline with request #796

PR #827, opened 2019-08-04
Split World into several files

killzoms:
Renaming the Save name to "world" felt better...


killzoms:
Capture

killzoms:
Besides helping locate the portions that should be optimized more, it has a
better aesthetic feel and lets others transfer "important" portions of
their save without keeping the whole world

NitroxServer/Serialization/World/PersistedWorldData.cs
```diff
@@ -55,16 +37,16 @@ public bool IsValid()
                 return false;
             }
 
-            return (ParsedBatchCells != null) &&
-                   (ServerStartTime != null) &&
+            return (WorldData.ParsedBatchCells != null) && // Always returns false on empty saves
```
Measurity: Maybe add method `IsValid()` to WorldData class as well so this gets shorter and easier to read.

Code then becomes:
```
return WorldData.IsValid() && BaseData != null && ...
```

NitroxServer/Serialization/World/WorldData.cs
```diff
@@ -0,0 +1,40 @@
...
+namespace NitroxServer.Serialization.World
+{
+    [ProtoContract]
+    public class WorldData
+    {
+        [ProtoMember(1)]
+        public List<Int3> ParsedBatchCells { get; set; }
+
+        [ProtoMember(2)]
+        public DateTime ServerStartTime { get; set; }
```
Measurity: DateTime is a struct so it won't be null.
```suggestion
        public DateTime? ServerStartTime { get; set; }
```

MarijnS95: I don't think you ever want this to be `null`? Though it shouldn't be a wrong or invalid date either.

Measurity: Yup I should have said a bit more on why I wanted the change. But right now, because it can't be `null`, it'll potentially be a bad `DateTime` if serialization has no value for it. Either that or it throws an exception.

killzoms: @Measurity I believe the date is January 1st some year... I cant remember the year though

killzoms: it will default to 01/01/0001

Measurity: It'd still prefer it would be `DateTime?` because 01/01/0001 is a valid date but we want the value to say "I'm not a valid date" hence, `null`. See: https://stackoverflow.com/questions/4821560/better-datetime-or-use-defaultdatetime-for-null

killzoms: unfortunately it will never be null through protobuf... i am rather bad at memory but i believe it handles DateTime differently

MarijnS95: That seems a good concession to make, given that it will either be assigned to something useful or remain `null`. Using the nullable type forces this to be explicitly checked on retrieval as well.

MarijnS95: @killzoms the nullable type forces it to be wrapped in something that can represent null. It seems obvious that a datetime is converted to an integer or timestamp-string of some sorts.

Measurity: Yeah it's a struct so it can't be null. I thought that protobuf handled Nullable<T> too but if that's not the case then `default(DateTime)` it is.

PR #839, opened 2019-09-29
Update WorldPersistence.cs

Bl4ckSkull666:
Add Backups from save files before it rewrite. And move save path from main to saves directory.


SavageJayGatsby:
I know Killzoms wants to do a little project with the saves and splitting them apart so it i'm sure they will be getting their own folder for it. Pull request 827 https://github.com/SubnauticaNitrox/Nitrox/pull/827

PR #842, opened 2019-10-19
Bugfix: since the change from guid to NitroxId the server was unable to load EntityData resulting in a "Persisted state is not valid" error

Praevalens:
Since commit (cacf262) the server was unable to save EntityData.
To fix this the thread lock was bypassed. Since protobuff-net is thread safe but the protobuf metadata inspector is not, we have to prepare the metadata in a static constructor.


Praevalens:
Did not do thorough enough testing

killzoms:
wait does protobuf serialize dictionaries?

killzoms:
also you on the discord by any chance?

PR #843, opened 2019-10-19
Several server bugfixes of bugs created by the guid to NitroxId change

Praevalens:
See commits for more information

### DIFF
--- a/NitroxClient/GameLogic/InitialSync/EscapePodInitialSyncProcessor.cs
+++ b/NitroxClient/GameLogic/InitialSync/EscapePodInitialSyncProcessor.cs
@@ -15,7 +15,7 @@ namespace NitroxClient.GameLogic.InitialSync
 
         public override void Process(InitialPlayerSync packet)
         {
-            EscapePodModel escapePod = packet.EscapePodsData.Find(x => x.Id == packet.AssignedEscapePodId);
+            EscapePodModel escapePod = packet.EscapePodsData.Find(x => x.Id.Equals(packet.AssignedEscapePodId));
             escapePodManager.AssignPlayerToEscapePod(escapePod);
             escapePodManager.SyncEscapePodIds(packet.EscapePodsData);
         }
--- a/NitroxClient/GameLogic/InitialSync/EscapePodInitialSyncProcessor.cs
+++ b/NitroxClient/GameLogic/InitialSync/EscapePodInitialSyncProcessor.cs
@@ -15,7 +15,7 @@ namespace NitroxClient.GameLogic.InitialSync
 
         public override void Process(InitialPlayerSync packet)
         {
-            EscapePodModel escapePod = packet.EscapePodsData.Find(x => x.Id.Equals(packet.AssignedEscapePodId));
+            EscapePodModel escapePod = packet.EscapePodsData.Find(x => x.Id == packet.AssignedEscapePodId);
             escapePodManager.AssignPlayerToEscapePod(escapePod);
             escapePodManager.SyncEscapePodIds(packet.EscapePodsData);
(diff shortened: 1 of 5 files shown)


NitroxServer/GameLogic/Entities/EntityData.cs
```diff
@@ -11,7 +11,7 @@ namespace NitroxServer.GameLogic.Entities
     [ProtoContract]
     public class EntityData
     {        
-        [ProtoMember(1)]
```
killzoms: Is this change needed anymore now that NitroxId has been updated?

NitroxClient/GameLogic/InitialSync/EscapePodInitialSyncProcessor.cs
```diff
@@ -15,7 +15,7 @@ public EscapePodInitialSyncProcessor(EscapePodManager escapePodManager)
 
         public override void Process(InitialPlayerSync packet)
         {
-            EscapePodModel escapePod = packet.EscapePodsData.Find(x => x.Id == packet.AssignedEscapePodId);
```
killzoms: I would recommend leaving as "==" considering that the "==" operator was added

NitroxServer/GameLogic/Entities/EntityData.cs
Praevalens: I checked, it isn't so i reverted it

NitroxClient/GameLogic/InitialSync/EscapePodInitialSyncProcessor.cs
Praevalens: Yes, forgot to change that back, should i close this pr adn resubmit with the changes?

NitroxServer/GameLogic/Entities/EntityData.cs
```diff
@@ -11,7 +11,7 @@ namespace NitroxServer.GameLogic.Entities
...
+        [ProtoIgnore]
```
Sunrunner37: The diff still shows these as changed.  You may want to rebase everything and resubmit the pr.

killzoms:
```
discord: btw this is still there...

![](https://cdn.discordapp.com/attachments/635059612000845835/635100777706160149/unknown.png)

You got one file tho
 
PraevalensToday at 9:04 AM
Yes thats because that actually fixed an issue with the entities not being deserialised```

MarijnS95: > The diff still shows these as changed. You may want to rebase everything and resubmit the pr.

The changes only seem to be reverted in `EscapePodData`, not `EntityData`. Keep in mind that you can (and should!) always (force)push to the same branch _without_ opening a separate PR, losing all this review context.

But if dictionary serialization "works properly" now it is probably best to clean this up for every serializable structure in a separate commit. After all the dictionary itself is still serialized, but with a lock around it. @Sunrunner37 what was the reason for these locks again? A packet should be built and sent on a single thread (if not, add locking where sharing occurs) as this seems pretty overkill to me unless I'm missing something important.

Praevalens: So I figured the locking had something to do with it being necessary for saving (serializing) while also continuing to receive packets and save those to the dictionary. But I'm just guessing.

In the mean time I've noticed that the game still doesn't load properly after a server reset so I'll see what i can do about that.

Coding-Hen: @Praevalens any update on this? This issue does really effect working on other issues

Issue #834, opened 2019-09-19
Events and story mod seem to bug


paberthier15:
Since the beginning, we have to reach a place on the surface, 300m approximately from our survival capsule but when we arrive to, we have to come back and it starts again. On the surface, there are some floating trashes and one purple thing. Don't know what it means.
I don't know if the events and the story mod work correctly, the Aurora explode each time after launching the session. The radio works sometime but not often like in solo mod.

HarveyBB:
That is not a story point it is just the server's 'player marker'
The 'floating trashes and one purple thing' is the servers player model so that it doesn't get in the way of gameplay.

If I were you I'd just recolour it in the ping tab so you know which mark is a story and which is the server

Coding-Hen:
The aurora exploding each time and other events is currently something I am working on. It's due to the timers for events (Aurora being one of them) restarting everyone the server is loaded and not saved to file

paberthier15:
Thank you Coding-Hen, I thought it was a timer problem but I wasn't sure.

Coding-Hen:
@Measurity This can be closed

Issue #814, opened 2019-05-31
All of the bugs that me & my friends have encountered that are effecting game play majorly


ImStormzTV:
Me & 2 other friends are playing on a Multiplayer server & these are the biggest bugs that we've encountered:
-Aurora re-exploding randomly
-Base components disappearing
-Sandshark becoming bigger and smaller rapidly
-Items duping in storage
-Items disappearing from inventory
-Seamoth not moving at all
-Teleporting to moonpool when entering seamoth
-Fabricator deconstructing and reconstructing and then consistently replaying the sound of making an item
-Bodies of other players and beacons of players not correctly displaying
-Seamoth inching forward with no driver, then teleporting roughly 20 meters every few seconds
-Base being flooded for some people and not for others
If there are any solutions that you can think of that would help, please reply with them.

Zaannaah:
i too have had  similar issues. 
-the aurora explosion  
-base components disappeareing (and reappearing. i'll deconstruct something and i'll exit load back in and it will have come back and i'll have to deconstruct again.) 
-issues with fabricators showing "not powered" and i have to deconstruct and reconstruct. 
-items disappearing. 
-a huge issue with other players disappearing so we cant see each other. (and sometimes their marker would just start moving in a random direction and keep going farther and farther away)
-got randomly teliported in a seamoth back to the moonpool and i couldnt get back out. i was stuck in the seamoth until i restarted the game. 
-seamoths would randomly explode 
-the classic keypad on the aurora issue. 
-one bug where i got in a seamoth and my body just trailed out behind me in an ever extending line. that one was quite hilarious looking. 
-base being flooded. 
-distorted wreckage. 
-was unable to cut through certain laser cuttable doors.

Measurity:
Gonna close this as this issue is too broad and not specific enough for us to track and fix. Make new GitHub issues for unique issues, please.

Issue #770, opened 2019-04-04
A Few Bugs


SamstarGaming1:
Here is a list of bugs my friends and I have experienced, I've added some gifs to help the developers.  (Sorry if any of these bugs have already been reported):

- Upon starting the server some habitat components deconstruct out of nowhere.
 https://media.giphy.com/media/5QMG8yIJmEqxgsOxUq/giphy.gif
- Bioreactors and nuclear reactors do the same as above although both have to be fully deconstructed before reconstructing it. All the fuel previously in the reactors disappear requiring them to be refueled.
https://giphy.com/gifs/878nV3H4m4c3bp0XT0
- Upwards of 20+ Reefbacks spawn in certain biomes, causing major lag spikes.
https://giphy.com/gifs/cXHyqlcnothvAC3gvo
- The Aurora detonates multiple times.
- Sometimes shipwrecks spawn without anything inside. Relogging is a temporary fix although only cosmetic models spawn inside. This has made it almost impossible to find any fragments as they don't spawn. Even after deleting and creating a new world twice, there are still no fragments inside any of the wrecks. (We had no other option but to use console commands to unlock all the blueprints.)
https://giphy.com/gifs/SSQ4ilRciwPbXkzR4d 
- When re-entering the server all fabricators and modification stations become unpowered despite there being enough power.
https://giphy.com/gifs/61XTidgqIk0Dn1c9mz
- Radio messages sometimes do not play.
https://giphy.com/gifs/ZNeht3iL77V7u0GSdG
- If the radio messages do play the waypoints are only given to the player that played the message.
- The amount of power in a base does not sync between clients, leading to a lot of issues. Fuel in the reactors doesn't sync either, requiring every player to get their own fuel for one base. 
- Time also doesn't sync. This ties in with the issue above since the time isn't the same for each client, solar panels produce different amounts of energy.
- Seamoths get stuck when trying to enter Cyclops
https://giphy.com/gifs/ZEYgrcQ4sTdCeKvlkU
- When connecting back to the server some habitat components act a bit glitchy.
https://giphy.com/gifs/1SG5qD4lVBr5gzWpUv 
- Items in planters don't sync.
- Water filtration machines just disappear.
- When relogging some previously deconstructed components reappear, essentially duplicating materials.

docquantum:
Thanks for the detailed bug report!

[...]

PR #845, opened 2019-10-30
Event timers save and reload timings only

### LINKED ISSUE #499: Aurora keeps exploding

### LINKED ISSUE #626: Aurora events not persisting?

### LINKED ISSUE #678: Aurora warning patch and random auroraTime persistence

Coding-Hen:
Currently saves the timers on the first save and reload.

Open for people to review

Fixes: #834, fixes #499, fixes #626, fixes #678  

Partly fixes:
#814, #770,

### DIFF
--- a/NitroxServer/GameLogic/EventData.cs
+++ b/NitroxServer/GameLogic/EventData.cs
@@ -1,0 +1,75 @@
+﻿using System;
+using System.Collections.Generic;
+using NitroxModel.Logger;
+using ProtoBufNet;
+
+namespace NitroxServer.GameLogic
+{
+    [ProtoContract]
+    public class EventData
+    {
+        [ProtoMember(1)]
+        public List<NitroxTimer> SerialisedTimers
+        {
+            get
+            {
+                lock (timers)
+                {
+                    return timers;
+                }
+            }
+            set { timers = value; }
+        }
+
+        [ProtoIgnore]
+        private List<NitroxTimer> timers = new List<NitroxTimer>();
+
+        internal bool HasTimers()
+        {
+            return timers.Count > 0;
+        }
+
+        public void AddTimer(NitroxTimer timer)
+        {
+            lock (timers)
+            {
+                timers.Add(timer);
+            }
+        }
+
+        public void PauseTimers()
+        {
+            lock (timers)
+            {
+                //TODO: remove, Logs the count of timers
+                Log.Info("Paused " + timers.Count);
+                foreach(NitroxTimer timer in timers)
+                {
+                    timer.Stop();
+                    //TODO: remove, how long left till event
+                    Log.Info(timer.TimeLeftDate);
+                }
+            }
+        }
+
+        public void StartTimers()
+        {
+            lock (timers)
+            {
+                //TODO: remove, Logs the count of timers
+                Log.Info("Started " + timers.Count);
+                foreach (NitroxTimer timer in timers)
+                {
+                    timer.Start();
+                    //TODO: remove, how long left till event
+                    Log.Info("Started " + timer.TimeLeftDate);
+                }
+            }
+        }
+
+        public override string ToString()
+        {
+            return timers.Count.ToString();
+        }
+    }
+}
--- a/NitroxServer/GameLogic/EventData.cs
+++ b/NitroxServer/GameLogic/EventData.cs
@@ -1,5 +1,4 @@
-﻿using System;
-using System.Collections.Generic;
+﻿using System.Collections.Generic;
 using NitroxModel.Logger;
 using ProtoBufNet;
 
--- a/NitroxServer/GameLogic/EventData.cs
+++ b/NitroxServer/GameLogic/EventData.cs
@@ -40,13 +40,9 @@ namespace NitroxServer.GameLogic
         {
             lock (timers)
             {
-                //TODO: remove, Logs the count of timers
-                Log.Info("Paused " + timers.Count);
                 foreach(NitroxTimer timer in timers)
                 {
                     timer.Stop();
-                    //TODO: remove, how long left till event
-                    Log.Info(timer.TimeLeftDate);
                 }
             }
         }
@@ -55,13 +51,9 @@ namespace NitroxServer.GameLogic
         {
             lock (timers)
             {
-                //TODO: remove, Logs the count of timers
-                Log.Info("Started " + timers.Count);
                 foreach (NitroxTimer timer in timers)
                 {
                     timer.Start();
-                    //TODO: remove, how long left till event
-                    Log.Info("Started " + timer.TimeLeftDate);
                 }
             }
         }
--- a/NitroxServer/GameLogic/EventTriggerer.cs
+++ b/NitroxServer/GameLogic/EventTriggerer.cs
@@ -1,17 +1,28 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Timers;
 using NitroxModel.Logger;
 using NitroxModel.Packets;
+using NitroxServer.GameLogic.Bases;
+using ProtoBufNet;
 
 namespace NitroxServer.GameLogic
 {
     public class EventTriggerer
     {
+
         PlayerManager playerManager;
-        public EventTriggerer(PlayerManager playerManager)
+        EventData eventData;
+
+        public EventTriggerer(PlayerManager playerManager, EventData eventData)
         {
             this.playerManager = playerManager;
-            SetupEventTimers();
+            this.eventData = eventData;
+            if (!eventData.HasTimers())
+            {
+                SetupEventTimers();
+            }
+            eventData.StartTimers();
         }
 
         public void SetupEventTimers()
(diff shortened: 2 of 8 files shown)


NitroxServer/GameLogic/EventData.cs
```diff
@@ -0,0 +1,74 @@
...
+using NitroxModel.Logger;
+using ProtoBufNet;
+
+namespace NitroxServer.GameLogic
+{
+    [ProtoContract]
+    public class EventData
+    {
+        [ProtoMember(1)]
+        public List<NitroxTimer> SerialisedTimers
```
Sunrunner37: Do these serialize cleanly while extending system timer?

NitroxServer/GameLogic/EventData.cs
```diff
@@ -0,0 +1,74 @@
...
+        {
+            lock (timers)
+            {
+                //TODO: remove, Logs the count of timers
+                Log.Info("Paused " + timers.Count);
+                foreach(NitroxTimer timer in timers)
+                {
+                    timer.Stop();
+                    //TODO: remove, how long left till event
+                    Log.Info(timer.TimeLeftDate);
```
Sunrunner37: Comment says to remove

NitroxServer/GameLogic/EventData.cs
```diff
@@ -0,0 +1,74 @@
...
+                }
+            }
+        }
+
+        public void StartTimers()
+        {
+            lock (timers)
+            {
+                //TODO: remove, Logs the count of timers
+                Log.Info("Started " + timers.Count);
```
Sunrunner37: Same with these

NitroxServer/GameLogic/EventTriggerer.cs
```diff
@@ -1,41 +1,46 @@
...
-            CreateTimer(auroraTimer + 24000, StoryEventType.Extra, "Story_AuroraExplosion");
+            eventData.AddTimer(CreateTimer(auroraTimer * 0.2d, StoryEventType.PDA, "Story_AuroraWarning1"));
+            eventData.AddTimer(CreateTimer(auroraTimer * 0.5d, StoryEventType.PDA, "Story_AuroraWarning2"));
+            eventData.AddTimer(CreateTimer(auroraTimer * 0.8d, StoryEventType.PDA, "Story_AuroraWarning3"));
+            eventData.AddTimer(CreateTimer(auroraTimer, StoryEventType.PDA, "Story_AuroraWarning4"));
+            eventData.AddTimer(CreateTimer(auroraTimer + 24000, StoryEventType.Extra, "Story_AuroraExplosion"));
         }
 
-        public Timer CreateTimer(double time, StoryEventType eventType, string key)
+        public NitroxTimer CreateTimer(double time, StoryEventType eventType, string key)
```
Sunrunner37: Might be clean to make this a static method factory on NitroxTimer.... i.e. `NitroxTimer.From(x, y, z)`

NitroxServer/GameLogic/NitroxTimer.cs
```diff
@@ -0,0 +1,89 @@
...
+        [ProtoMember(3)]
+        public double InitialInterval;
+        [ProtoMember(4)]
+        public PlayerManager PlayerManager;
+        [ProtoMember(5)]
+        public StoryEventType EventType;
+        [ProtoIgnore]
+        private double? timeLeft;
+        [ProtoIgnore]
+        private DateTime m_dueTime;
```
Sunrunner37: minor: no Hungarian notation.

NitroxServer/GameLogic/EventTriggerer.cs
Coding-Hen: Is there a need for this? Usually you would only do static method factory if you default values. Other than the reset value which isnt strickly needed anyway. Unless you are suggesting that we have a method for each of the event timers?

NitroxServer/GameLogic/NitroxTimer.cs
```diff
@@ -0,0 +1,92 @@
...
+using System.Text;
+using NitroxModel.Logger;
+using NitroxModel.Packets;
+using ProtoBufNet;
+
+namespace NitroxServer.GameLogic
+{
+    [Serializable]
+    [ProtoContract]
+    public class NitroxTimer : System.Timers.Timer
```
Measurity: Probably better to just wrap the `System.Timers.Timer` instead of inheriting it. You won't have to fight the existing Timer's methods/state and any possible issues with timer serialization will also be solved then.

NitroxServer/GameLogic/NitroxTimer.cs
```diff
@@ -0,0 +1,92 @@
...
+        [ProtoMember(4)]
+        public PlayerManager PlayerManager;
+        [ProtoMember(5)]
+        public StoryEventType EventType;
+        [ProtoIgnore]
+        private double? timeLeft;
+        [ProtoIgnore]
+        private DateTime dueTime;
+
+        public NitroxTimer() : base()
```
Measurity: Mandatory field data should be passed in through the constructor OR make it a private constructor with a static `NitroxTimer.From(...)` method that calls the private ctor as Sunrunner said.

NitroxServer/GameLogic/NitroxTimer.cs
```diff
@@ -0,0 +1,92 @@
...
+        {
+            Log.Info("Triggering event type " + EventType.ToString() + " at time " + DateTime.Now.AddMilliseconds(Interval).ToString(@"hh\:mm\:ss") + " with param " + Key.ToString());
+            PlayerManager.SendPacketToAllPlayers(new StoryEventSend(EventType, Key));
+            if (AutoReset)
+            {
+                dueTime = DateTime.Now.AddMilliseconds(InitialInterval);
+            }
+        }
+
+        public Delegate returned()
```
Measurity:
```suggestion
        public Delegate Returned()
```
Dunno what this is but if it's needed then use CamelCase because it's public.

NitroxServer/GameLogic/NitroxTimer.cs
MarijnS95: FYI there has been a discussion going on dropping this inheritance entirely. For posterity, let me wrap up the discussion points:

- `System.Timers.Timer` is likely not properly serializable (and/or very large), causing all kinds of trouble (applies to it being a member too);
- Timed events have a predefined (static) duration, and can (should) be solely based on the time the game has been running.
  I consider durations (like `double auroraTimer = RandomNumber(2.3d, 4d)...`) the to be static enough as well, in that it is only defined once when a new save is started, and easily persisted as a number in one of the save files.

The resulting code could do away with the `NitroxTimer` entirely, as nothing needs to be serialized. When the server loads the save it only needs to take into account the elapsed time since the save has been created, which can be stored in one of the base files (in case it isn't already). Knowing whether an event should have triggered already is as simple as checking if the trigger time is less than the elapsed game time.
Special care needs to be taken when the game is saved right as an event is supposed to fire, such that it is either processed before being saved, or processed (scheduled with a very short duration) when the save is loaded again.
Though I don't know exactly how current events hold up against this, perhaps the guarantee that it has either been triggered according to the elapsed time (thus requiring the game to be in a certain state) or scheduled to be triggered at some point in the future is enough for the game to function properly.
In much the same way this is advantageous on clients too; they can reconstruct specific parts of the game state (has the Aurora exploded, etc) by simply looking at the elapsed time received in the initial "state synchronization" packet while loading.

Keeping track of the game time on the server is incredibly easy and doesn't require a game loop either (like that game that apparently uses `update()` to `+=` the update delta to a float). The following pseudo-code would do, discussed previously as well:
```C#
public Duration ElapsedTimeThisSession => DateTime.Now() - timeServerStarted;
public Duration ElapsedTime => elapsedTimeAccordingToSave + ElapsedTimeThisSession;
```
In the event that we want the server to support pause-ing (when no clients are connected, idling while waiting for a new connection) it is only required to add `ElapsedTimeThisSession` to `elapsedTimeAccordingToSave` (and perhaps save to disk for a proper restore point), and reset `timeServerStarted` as soon as the "game world" continues simulating. The system timers need to be restarted as well.

PR #854, opened 2019-11-16
Begin BatchRoot spawning update

killzoms:
This is still unfinished and needs a few bugs to be ironed out


NitroxClient/GameLogic/Entities.cs
```diff
@@ -50,18 +52,32 @@ public void Spawn(List<Entity> entities)
         {
             foreach (Entity entity in entities)
             {
+                GameObject batchRoot = EnsureBatchRoot(entity);
                 if (!alreadySpawnedIds.Contains(entity.Id))
                 {
-                    Spawn(entity, Optional<GameObject>.Empty());
+                    Spawn(entity, Optional<GameObject>.Empty(), batchRoot);
```
Sunrunner37: If we were to remove alreadySpawnedIds would that fix the disappearing issue?

killzoms: unfortunately it needs changes on both server and client but it would(still an issue of the gameObjects getting assigned to the Waiters)

killzoms:
Alright Spawning Is Working now However there is still the issue of Entities "disappearing" in the eyes of the server

killzoms:
```ArgumentNullException: Value cannot be null.
Parameter name: key
  at System.Collections.Generic.Dictionary`2[TKey,TValue].TryInsert (TKey key, TValue value, System.Collections.Generic.InsertionBehavior behavior) [0x00008] in <599589bf4ce248909b8a14cbe4a2034e>:0 
  at System.Collections.Generic.Dictionary`2[TKey,TValue].set_Item (TKey key, TValue value) [0x00000] in <599589bf4ce248909b8a14cbe4a2034e>:0 
  at NitroxClient.MonoBehaviours.NitroxIdentifier.OnProtoDeserializeObjectTree (ProtobufSerializer _) [0x00034] in <d2a197d1127746bf9a9df18f0a539c3d>:0 
  at ProtobufSerializer+<DeserializeObjectTreeAsync>d__30.MoveNext () [0x000e6] in <08e3a801b15c454eadae7919261dd9cc>:0 
``` Just one final error to solve

**Comment by killzoms** [comment]
FIXED!!! It works!!! Now... Its also extendable Soooo Enjoy having it all set up

**Comment by killzoms** [comment]
ah crap now that I think of it did I leave logging in in places?

File: NitroxClient/GameLogic/Entities.cs [hunk 2]
```diff
@@ -50,23 +52,52 @@ public void Spawn(List<Entity> entities)
         {
             foreach (Entity entity in entities)
             {
+                LargeWorldStreamer.main.cellManager.UnloadBatchCells(ToInt3(entity.AbsoluteEntityCell.CellId)); // Just in case
+                EntityCell entityCell = EnsureCell(entity);
+                entityCell.Initialize();
```
Sunrunner37: Is this safe to call again each time?  Should we call once when it is instantiated below?

NitroxClient/GameLogic/Entities.cs
```diff
@@ -20,6 +21,7 @@ public class Entities
         private readonly DefaultEntitySpawner defaultEntitySpawner = new DefaultEntitySpawner();
         private readonly SerializedEntitySpawner serializedEntitySpawner = new SerializedEntitySpawner();
         private readonly Dictionary<TechType, IEntitySpawner> customSpawnersByTechType = new Dictionary<TechType, IEntitySpawner>();
+        private readonly Dictionary<NitroxModel.DataStructures.Int3, EntityCell> EntityCells = new Dictionary<NitroxModel.DataStructures.Int3, EntityCell>();
```
Sunrunner37: Aren't entity cells also dictated by level and not just Int3?

NitroxPatcher/Patches/Persistent/CellManager_TryLoadCacheBatchCells_Patch.cs
```diff
@@ -0,0 +1,51 @@
...
+using NitroxModel.DataStructures;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection.Emit;
+
+// TODO: Temporarily persistent to run before everything.  When we migrate the patch hook to an early point then make this non-persistent
+namespace NitroxPatcher.Patches.Persistent
+{
+    class CellManager_TryLoadCacheBatchCells_Patch : NitroxPatch, IPersistentPatch
```
Sunrunner37: Let's go through each patch in this PR and add documentation at the top.  Similar to https://github.com/SubnauticaNitrox/Nitrox/blob/master/NitroxPatcher/Patches/Dynamic/Base_SpawnPiece_Patch.cs#L16

Also, we'll want to add some samples of the code that we are injecting.

NitroxPatcher/Patches/Persistent/PAXTerrainController_LoadAsync_Patch.cs
```diff
@@ -34,34 +64,46 @@ public static IEnumerable<CodeInstruction> Transpiler(MethodBase original, IEnum
 
         public override void Patch(HarmonyInstance harmony)
         {
-            PatchTranspiler(harmony, getMethod());
+            PatchTranspiler(harmony, GetMethod());
+        }
+
+        private static FieldInfo GetStateField(IReflect type)
```
Sunrunner37: I believe there is a GetField in reflection helper

NitroxClient/Helpers/NitroxProtobufSerializer.cs
```diff
@@ -0,0 +1,119 @@
...
+            foreach (Type type in Assembly.GetExecutingAssembly().GetTypes())
+            {
+                bool hasUweProtobuf = (type.GetCustomAttributes(typeof(ProtoContractAttribute), true).Length > 0);
+
+                if (hasUweProtobuf)
+                {
+                    // As of the latest protobuf update they will automatically register detected attributes.
+                    model.Add(type, true);
+                    knownTypes[type] = int.MaxValue; // UWE precompiled is going to pass everything to us
+                    NitroxTypes[type] = int.MaxValue;
```
Sunrunner37: minor: extract common code into a method

NitroxClient/GameLogic/Entities.cs
killzoms: Moved into EnsureCell

NitroxClient/GameLogic/Entities.cs
killzoms: Added New class named CellRoot to differentiate EntityCells

NitroxPatcher/Patches/Persistent/PAXTerrainController_LoadAsync_Patch.cs
killzoms: That GetField in reflection helper unfortunately wont help with Async methods

NitroxClient/Helpers/NitroxProtobufSerializer.cs
killzoms: Moved to AddType method

killzoms:
Note: CrashFish are not fixed with this

killzoms:
Now they're fixed kinda

Sunrunner37:
Per our previous conversation, in what instance does Subnautica serialize the batch cells (resulting in our identifier being serialized)?  Is it doing local caching to the disk?  This must be some new feature that they implemented because we were not previously having serialization issues.  Ideally, we do not want to touch their local serialization.

killzoms:
yea local caching is being done. trying to change the implementation would mean we likely need to notify the server to send new spawn data each time we change the visible cells although this implementation might not function ideally in the case of gameobjects that are stored into another gameobject as for the stacktrace... ive narrowed down where the code is getting called and its... easiest to simply patch the serializer... patching another coroutine is likely only going to add excess overhead and last until changes happen reason for patching the serializer is to limit excess overhead and keep the patch functional for as long as possible

NitroxModel/DataStructures/CellRoot.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace NitroxModel.DataStructures
+{
+    public struct CellRoot
```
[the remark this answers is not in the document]
killzoms: its less data but essentially is the same(something we may want to look into to decrease byte sizes when uploading entities)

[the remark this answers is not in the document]
killzoms: the server needs the helpers and extra data in the AbsoluteEntityCell the client less so in fact it adds more to what the client has to download for each entity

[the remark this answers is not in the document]
killzoms: can likely adjust the implementation to ignore fields when serialization occurs as excess upload/download sizes add extra time needed for each packet to be sent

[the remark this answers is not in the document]
MarijnS95: > can likely adjust the implementation to ignore fields when serialization occurs as excess upload/download sizes add extra time needed for each packet to be sent

Yes please. Especially in the context of Nitrox I would consider deduplicating (thus decreasing maintenance and new-contributor "but how does this work?" hurdles) at least as much of a win as saving a couple bytes off a structure.

Though, reading up on `AbsoluteEntityCell`, the serialized variables are already identical? The rest is _useful_ helper functions and getters (or is protobuf stupid enough to serialize the pure getters too??).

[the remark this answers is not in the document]
killzoms: windows serialization is used if memory serves for packets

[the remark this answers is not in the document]
killzoms: and hm... first time i looked through the absoluteEntityCell class i thought j saw position being a normal field not a get property... looked through it a moment ago and looks like i was incorrect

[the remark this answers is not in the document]
killzoms: Changed to AbsoluteEntityCell

NitroxPatcher/Patches/Persistent/CellManager_TryLoadCacheBatchCells_Patch.cs
killzoms: Added Some documentation

PR #864, opened 2019-12-14
Unity 2019 Upgrade

Shirkie01:
Updates the projects to .Net Framework 4.0 and removes references to libraries that no longer exist.


Shirkie01:
The last piece to get this playable again is to fix the exception thrown in BaseDeconstructable_Deconstruct_Patch.cs on the line
```
yield return original.Ldloc<GameObject>(1);
```

Listing the local variables results in 
```
Base (0)
UnityEngine.Vector3 (1)
ConstructableBase (2)
BaseGhost (3)
```

Should the ConstructableBase type replace GameObject?

PR #866, opened 2019-12-22
Persist inventory through multiple client sessions

Shirkie01:
Changes the Player Id from a ushort to a NitroxId, and removes the PlayerGameObjectId.

### DIFF
--- a/NitroxModel/DataStructures/NitroxId.cs
+++ b/NitroxModel/DataStructures/NitroxId.cs
@@ -13,8 +13,6 @@ namespace NitroxModel.DataStructures
         [ProtoMember(1)]
         private Guid guid;
 
-        public static NitroxId Empty { get; } = new NitroxId(Guid.Empty);
-
         public NitroxId()
         {
             guid = Guid.NewGuid();
@@ -23,16 +21,6 @@ namespace NitroxModel.DataStructures
         public NitroxId(byte[] bytes)
         {
             guid = new Guid(bytes);
-        }
-
-        public NitroxId(Guid guid)
-        {
-            this.guid = guid;
-        }
-
-        public static NitroxId Parse(string input)
-        { 
-            return new NitroxId(Guid.Parse(input));
         }
 
         protected NitroxId(SerializationInfo info, StreamingContext context)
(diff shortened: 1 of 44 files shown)


Sunrunner37:
Is this something that we want to do?  It effectively increases every packet size that contains a player id from 16 bits to 16 bytes.

Shirkie01:
I was working on getting inventory to persist across logins. Right now a new Player.GameObjectId is created every time a player joins so their stuff is lost. Perhaps instead of changing everything to use NitroxId, I can try including the GameObjectId as part of the PersistedPlayerData.

NitroxModel/DataStructures/NitroxId.cs
```diff
@@ -52,5 +52,15 @@ public override string ToString()
...
+
+        public static bool operator ==(NitroxId id1, NitroxId id2)
+        {
+            return id1?.guid == id2?.guid;
+        }
+
+        public static bool operator !=(NitroxId id1, NitroxId id2)
+        {
+            return !(id1 == id2);
+        }
```
Coding-Hen: Is this needed anymore as you are not using/changing the NitroxId anywhere else?

NitroxServer/Player.cs
```diff
@@ -23,13 +23,13 @@ public class Player : IProcessorContext
...
 
-        public Player(PlayerContext playerContext, NitroxConnection connection, Vector3 position, Optional<NitroxId> subRootId)
+        public Player(PlayerContext playerContext, NitroxConnection connection, Vector3 position, Optional<NitroxId> subRootId, NitroxId gameObjectId)
         {
             PlayerContext = playerContext;
             this.connection = connection;
             Position = position;
             SubRootId = subRootId;
-            GameObjectId = new NitroxId();
+            GameObjectId = gameObjectId;
```
Sunrunner37: Do we need to have the overhead of persisting this data?  Can we make it deterministic.... maybe based on the player name?  I believe we used to do that but it fell off at some point in time.

NitroxModel/DataStructures/NitroxId.cs
Shirkie01: The new direction is to use a deterministic NitroxId, so I am changing the class now. I would like to keep the operators as there are lots of places in the code where == is used to check equality.

NitroxServer/Player.cs
Shirkie01: I've changed the code to use a deterministic NitroxId following the RFC guidelines for GUID version 5.

Shirkie01:
Closing as apparently we don't care about "the overhead of persisting this data" and "making the NitroxId deterministic based on the player name" any more as seen in pull request #902.

PR #883, opened 2020-01-10
Fixing cyclops horn

Jannify:
Fixed horn processor and changed range to 100-130 meters.

### DIFF
--- a/NitroxClient/Communication/Packets/Processors/CyclopsActivateHornProcessor.cs
+++ b/NitroxClient/Communication/Packets/Processors/CyclopsActivateHornProcessor.cs
@@ -13,6 +13,8 @@ namespace NitroxClient.Communication.Packets.Processors
     {
         private readonly IPacketSender packetSender;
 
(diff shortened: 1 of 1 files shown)


NitroxClient/Communication/Packets/Processors/CyclopsActivateHornProcessor.cs
```diff
@@ -21,7 +23,12 @@ public override void Process(CyclopsActivateHorn hornPacket)
             GameObject cyclops = NitroxIdentifier.RequireObjectFrom(hornPacket.Id);
             CyclopsHornControl horn = cyclops.RequireComponentInChildren<CyclopsHornControl>();
 
-            Utils.PlayEnvSound(horn.hornSound, horn.hornSound.gameObject.transform.position, 20f);
+            FieldInfo field = typeof(FMOD_CustomEmitter).GetField("evt", BindingFlags.Instance | BindingFlags.NonPublic);
```
Measurity: Store the `FieldInfo` on a `private static readonly` field.

Measurity:
I noticed a sound cut-off issue if you repeatedly press it. I'll have a look if I can fix it. Maximum range works fine but could be quite a bit higher imo.

PR #887, opened 2020-01-12
Added autofac dependency to NitroxPatcher

Measurity:
Patches are now loaded through AutoFac.

I want patches to use proxy classes that make the reflection calls. Removing the need for the `ReflectionHelper` class. The proxy classes will use proper caching to improve performance on future calls. This PR is the first step to allow that.


NitroxPatcher/Patches/Dynamic/uGUI_SignInput_OnDeselect_Patch.cs
```diff
@@ -1,12 +1,12 @@
...
 using Harmony;
-using NitroxModel.Core;
 using NitroxClient.GameLogic;
+using NitroxModel.Core;
 
-namespace NitroxPatcher.Patches
+namespace NitroxPatcher.Patches.Dynamic
 {
-    public class uGUI_SignInput_OnDeselect_Patch : NitroxPatch
+    public class uGUI_SignInput_OnDeselect_Patch : NitroxPatch, IDynamicPatch
```
Sunrunner37: Do we still need NitroxPatch now that we have INitroxPatch?

Measurity: I could have used NitroxPatch directly instead of INitroxPatch. Though I liked INitroxPatch with just the `Patch` and `Restore` methods on it because that's all the `NitroxPatcher.Execute` needed.

Issue #832, opened 2019-09-03
Deserializing the PersistedWorldData doesn't reload EntityData


Coding-Hen:
When the savefile is deserialized it doesn't have the SerializableEntitiesById populated this means that it is not persistant data and resets the savefile each time the server is started.
![image](https://user-images.githubusercontent.com/8798074/64210431-2525d580-ce9b-11e9-8cb6-98ef666eeea2.png)

killzoms:
now that i look... i dont think ProtoBuf actually supports dictionary serialization and if thats the case... a work-around would have to be implemented...

killzoms:
i have one in mind but its just a theory at this point in time(i cant open computer to code atm because i gotta wake up for work at 3AM its 8PM now

Coding-Hen:
Glad I wasn't the only one who found this. Yes I think a work around would
be easier bearing in mind how baked in it is.

Coding-Hen:
@killzoms I have had a look and seems that when the NitorxId was introduced replacing the String GUID it caused this issue. commit cacf262d4d2a4c8718a90a3de97cf3ed39a53105

PR #888, opened 2020-01-18
Added functional world versioning and Fixes EntityData serialization

killzoms:
~~Likely more Dictionaries are Included in everything we are serializing. We should work towards switching all of them over to lists for serialization~~ Edit: The Previous message is no longer valid...


Coding-Hen:
Believe should fix #832

Measurity:
Need to add proper locking because fields are sometimes used with and sometimes without locks. Also should prevent using **any** field that can change to be used as a lock object. Either need a dedicated lock object (for example):

`private readonly object getEntitiesLock = new object()`

Or use custom thread-safe list/dictionaries that can also be serialized by protobuf. The latter would simplify the code a lot but will impact performance a bit more than custom lock objects.

killzoms:
after closer examination it would appear other things are still broken

PR #894, opened 2020-01-25
Launcher globalization

iCleeem:
Will be easier to change values and maybe implement different localizations


Coding-Hen:
I'm wondering if we are going down the route of language extraction if its work implementing the switching of language if the systems is different therefore implementing globalisation?

MarijnS95:
This PR includes a ton of line-ending changes from `lf` to `crlf`, making the diff (as shown on GitHub) effectively useless.

Would you mind to (force) push a cleanup with the line endings reverted to their original `lf`? I assume Nitrox is using `core.autocrlf=true` across the board, which ensures all line endings are appropriately normalized to `lf` for consistency, and converted to `crlf` when checked out on Windows machines.

While at it, please also remove unnecessary trailing whitespace that has been introduced.

Finally, upon quick inspection of [WPF Globalization and Localization Overview](https://docs.microsoft.com/en-us/dotnet/framework/wpf/advanced/wpf-globalization-and-localization-overview) it doesn't look like this problem is (generally) "solved" with merging resource dictionaries.
@Ghaarg You are the initial author of the WPF based Launcher; what's your view/feedback on localization?

MarijnS95:
@iCleeem Note that you can (and are kindly requested to) push new changes to this PR by pushing them to the same branch. This makes it so that review comments and any other info related to these changes are not lost when opening a new one.

iCleeem:
The line endings and the whitespaces should be ok now, didn't think about the line endings 😢

iCleeem:
Both line endings and whitespaces should be ok now, I didn't think to configure my git with **core.autocrlf**

About the microsoft article on the globalization, I've implemented that things in a projet in my school 2 years ago and I really hate the way the did that, That's why I decided to go with merged dictionnaries cause it's the easiest way and efficient.

Next week I will try to finished to implement what I started to do, if it's not pleasant, I'll check to implement localization with another method

MarijnS95:
@iCleeem Awesome, the diff now only shows exactly those changes you have made. There are still a couple trailing newlines though, particularly in `App.xaml`. The easiest way to chase this is to configure Visual Studio to trim trailing whitespace on save and re-save every file, and going even further: configure Visual Studio to render (trailing) whitespace to make it visible (IIRC it can do both... Otherwise an editor like visual Studio Code will do).
It might also be the auto-formatter that takes care of these cases.

I think I agree on the globalization. Quickly reading over the docs didn't make it immediately clear how to implement it though. Still, I'd like to hear from a seasoned developer (preferably with a deployed application in a professional environment) what implementations and drawbacks they came across, but that's likely not possible so this obvious and easy-to-use solution will do. It can always be swapped out in the unlikely event that it doesn't suit our needs.
That raises the following question though: Can this dictionary be swapped out or inserted at runtime based on either:
- (Preferably) Language detection on the target machine
- A drop-down to change the language

I would **guess** that the "builtin" globalization takes care of these kinds of things, which is necessary to take any sort of advantage of extracting text to a separate file.

iCleeem:
@MarijnS95 I'll check that, thank you very much

Upon quick inspection on [](https://docs.microsoft.com/en-us/dotnet/api/system.globalization.cultureinfo.currentculture?view=netframework-4.8) it's easy to localy change the CultureInfo of the app, now i'll try to figure if it's possible to replace the dictionnary at the runtime, I guess it's possible

MarijnS95:
@iCleeem I would say that changing the merged dictionary at startup is a core requirement to make globalization work in the first place :sweat_smile: 

Changing it later at runtime is optional. I don't think a language change option is necessary, nor should we expect the OS language to change.

iCleeem:
@Coding-Hen The default language will always be the english one which implies that he must be up to date. In `App.xaml` we can see that the dictionnary for english is already included by default, when the launcher starts, if the OS language is different we try to switch the ressource dictionnary to the corresponding one, if he does exist and have the right version, if not we return.

Maybe the error message is not relevant to the user?

Coding-Hen:
> 
> 
> @Coding-Hen The default language will always be the english one which implies that he must be up to date. In `App.xaml` we can see that the dictionnary for english is already included by default, when the launcher starts, if the OS language is different we try to switch the ressource dictionnary to the corresponding one, if he does exist and have the right version, if not we return.
> 
> Maybe the error message is not relevant to the user?

I understand about getting the correct version. I just don't understand the version. Say you add a new button called Play and you only update the english version not the french, etc. then you try to load french the version will fall back to english correct? My question is why we cant load the french for all the values and then just load in the french or default the Play button when the language file does not have the value it is trying to find?

iCleeem:
@Coding-Hen Actually, If I remove the version condition it already does that ! Thank's to the merged dictionnaries, 2 languages can complete each other dynamically.
Just sounds not really clean to me, that's why I just decided to fall back to the default version instead of letting the merged dictionnaries freestyling on their own 🔞 
But I can remove my version condition if everyone agree on that !

NitroxLauncher/OptionPage.xaml
```diff
@@ -4,40 +4,46 @@
       xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006" 
       xmlns:d="http://schemas.microsoft.com/expression/blend/2008" 
       xmlns:local="clr-namespace:NitroxLauncher"
-      mc:Ignorable="d" 
-      d:DesignHeight="470" d:DesignWidth="719"
+      mc:Ignorable="d" d:DesignHeight="470" d:DesignWidth="719"
```
MarijnS95: Is the autoformatter putting everything on the same line?

NitroxLauncher/Localization/Localization-EN.xaml
```diff
@@ -0,0 +1,80 @@
...
+    <sys:String x:Key="NitroxGameSinglerplayerButtonTooltip">Play alone</sys:String>
+
+    <sys:String x:Key="NitroxGameMultiplayerButton">MULTIPLAYER</sys:String>
+    <sys:String x:Key="NitroxGameMultiplayerButtonTooltip">Launch the game with multiplayer enabled</sys:String>
+
+    
+    <!-- Options Page -->
+
+    <sys:String x:Key="NitroxOptionTitle">Option Page</sys:String>
+    <sys:String x:Key="NitroxOTitle">Settings</sys:String>
```
MarijnS95: Let's not use odd abbreviations. Instead, give these both a proper name (including `NitroxOption` written in full) that explains the (usage) difference between the two.

NitroxLauncher/OptionPage.xaml
```diff
@@ -4,40 +4,46 @@
...
-                    <TextBox FontFamily="{StaticResource Roboto}" Width="427" Height="43" FontSize="16" MaxLines="1" Foreground="White" TextAlignment="Left" Padding="9" Background="{x:Null}" BorderBrush="{x:Null}" Text="{Binding PathToSubnautica, RelativeSource={RelativeSource FindAncestor, AncestorType=Page}}" SelectionOpacity="-1" ToolTip="Subnautica install location"/>
+                <TextBlock FontFamily="{StaticResource Roboto}" FontSize="12" Foreground="White" Text="{DynamicResource ResourceKey=NitroxOptionGameDirectory}"/>
+                <Border HorizontalAlignment="Left" Height="43" BorderBrush="#33FFFFFF" Background="#9928292C" BorderThickness="1" CornerRadius="4,4,4,4" Margin="0,4,0,0">
+                    <TextBox IsReadOnly="True" FontFamily="{StaticResource Roboto}" Width="427" Height="43" FontSize="16" MaxLines="1" Foreground="White" TextAlignment="Left" Padding="9" Background="{x:Null}" BorderBrush="{x:Null}" Text="{Binding PathToSubnautica, Mode=OneWay}" SelectionOpacity="-1" ToolTip="{DynamicResource ResourceKey=NitroxOptionGameDirectoryTooltip}" />
```
MarijnS95: Why is this converted to readonly/OneWay? Users should be able to change the path, especially when the default is chosen because no game is found using the likely limited detection methods.

NitroxLauncher/OptionPage.xaml
```diff
@@ -4,40 +4,46 @@
...
+                    <TextBlock FontFamily="{StaticResource Roboto}" Padding="0,1,0,0" FontSize="16" Foreground="White" Text="{DynamicResource ResourceKey=NitroxOptionSubmit}" />
                 </Button>
             </StackPanel>
+
+            <StackPanel Grid.Row="4" Margin="0,30,0,0">
+                <TextBlock FontFamily="{StaticResource Roboto}" FontSize="12" Foreground="White" Text="{DynamicResource ResourceKey=NitroxOptionLanguage}"/>
+                <Border HorizontalAlignment="Left" Height="43" BorderBrush="#33FFFFFF" Background="#9928292C" BorderThickness="1" CornerRadius="4,4,4,4" Margin="0,4,0,0" ToolTip="{DynamicResource ResourceKey=NitroxOptionLanguageTooltip}">
+                    <TextBox IsReadOnly="True" FontFamily="{StaticResource Roboto}" Width="427" Height="43" FontSize="16" MaxLines="1" Foreground="White" TextAlignment="Left" Padding="9" Background="{x:Null}" BorderBrush="{x:Null}" Text="{Binding Lang, Mode=OneWay}" SelectionOpacity="-1"/>
```
MarijnS95: Readonly for the chosen language? With no ability to change the language at runtime I don't see a reason to provide this `TextBox` in the first place.

NitroxLauncher/Localization/Localization-EN.xaml
```diff
@@ -0,0 +1,80 @@
...
+    <sys:String x:Key="NitroxOptionLanguageTooltip">Change the localization of the launcher</sys:String>
+
+    <sys:String x:Key="NitroxOptionSubmit">Change</sys:String>
+    <sys:String x:Key="NitroxOptionSubmitTooltip">Confirm the configuration changings</sys:String>
+    
+    <!-- Server Page -->
+
+    <sys:String x:Key="NitroxServerTitle">Server Page</sys:String>
+    
+    <sys:String x:Key="NitroxServerSTitle">Server Options</sys:String>
```
MarijnS95: Same here, the `S` fumbled inbetween doesn't mean anything.

NitroxLauncher/ServerPage.xaml
```diff
@@ -1,338 +1,58 @@
...
+      xmlns:Themes="clr-namespace:Microsoft.Windows.Themes;assembly=PresentationFramework.Aero2"
+      x:Class="NitroxLauncher.ServerPage"
+      mc:Ignorable="d" d:DesignHeight="470" d:DesignWidth="719"
       DataContext="{Binding RelativeSource={RelativeSource Self}}"
-      Title="Server Page" Margin="0,0,0,0">
-    <Page.Resources>
+      Title="{DynamicResource ResourceKey=NitroxServerTitle}"
+      Margin="0">
 
-        <Style x:Key="FocusVisual">
```
MarijnS95: Why is this considered "dead code"? Doesn't it apply any styling (or just the default)? Even if close to the default, I would assume there's purpose behind including it here.

NitroxLauncher/ServerPage.xaml
```diff
@@ -1,338 +1,58 @@
...
                         <ColumnDefinition Width="*" />
                         <ColumnDefinition Width="300" />
                     </Grid.ColumnDefinitions>
+
                     <StackPanel Grid.Column="0" VerticalAlignment="Center">
-                        <TextBlock FontFamily="{StaticResource Roboto}" Foreground="White" FontSize="22"><Run Text="SERVER TYPE"/></TextBlock>
-                        <RadioButton Name="RBIsDocked" GroupName="ServerType" Content="Docked console" Margin="5" FontFamily="{StaticResource Roboto}" Foreground="LightGray" FontSize="14" HorizontalAlignment="Left" VerticalAlignment="Top" Click="RBServer_Clicked"/>
-                        <RadioButton Name="RBIsExternal" GroupName="ServerType" Content="External command prompt" Margin="5" FontFamily="{StaticResource Roboto}" Foreground="LightGray" FontSize="14" HorizontalAlignment="Left" VerticalAlignment="Top" Click="RBServer_Clicked"/>
+                        <TextBlock FontFamily="{StaticResource Roboto}" Foreground="White" FontSize="22">
+                            <Run Text="SERVER TYPE"/>
```
MarijnS95:
```suggestion
                            <Run Text="{DynamicResource ResourceKey=NitroxServerServerType}"/>
```

NitroxLauncher/ServerPage.xaml
iCleeem: None of the style define in this `<Page.Resources>` is used, the combobox were replaced by radio buttons few PR ago. In my opinion keeping 200 hundreds line of codes just for a simple combobox was a bit too much, wasn't it ?

NitroxLauncher/OptionPage.xaml
iCleeem: Because the textbox isn't implemented in TwoWay so the textbox is only used to display the value of the path when you change throught the button

NitroxLauncher/NitroxLauncher.csproj
```diff
@@ -88,6 +89,18 @@
       <SubType>Designer</SubType>
       <Generator>MSBuild:Compile</Generator>
     </Page>
+    <Content Include="Localization\Localization-en.xaml">
+      <Generator>MSBuild:Compile</Generator>
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
```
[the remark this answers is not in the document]
MarijnS95: Is the new `<Resource>` not referenceable by key, or something else generated at compile time?

Does the new `Uri` consistently reference the language file inside the assembly? Is there a possibility to create the `ResourceDictionary` from a `Assembly.GetExecutingAssembly()` reference? There are functions to get a file stream and perhaps other ways of accessing the file, that should prevent accidentally loading files from a relative path outside the scope of the assembly.

Finally, getting to the point I originally hinted at: How are these resource files generally dealt with when specified directly in a xaml page? Are they precompiled into something smaller and more efficiently parse-able? If so, is it possible to do the same with these localization files too, and reference them by a statically compiled key of some sort?

NitroxLauncher/OptionPage.xaml
MarijnS95: @iCleeem This is a legitimate question, not a "please revert this".

NitroxLauncher/ServerPage.xaml
MarijnS95: Thanks for clarifying.

NitroxLauncher/OptionPage.xaml
iCleeem: @MarijnS95 No the autoformatter did not put everything on the same line, I decided to modify this on my own

PR #898, opened 2020-01-26
Launcher server (console) page ui enhancements

grandoth:
Here are some proposed changes to the Launcher server (console) pages ui. I admire what the Nitrox team has done/is doing. 

It's my first PR. Go easy :).


Sunrunner37:
Thanks for the first PR!  Since this changes some stylistic elements, can you post a screenshot?

NitroxLauncher/ServerPage.xaml.cs
```diff
@@ -9,27 +9,40 @@ namespace NitroxLauncher
     public partial class ServerPage : Page
     {
         public string Version => "NITROX ALPHA " + Assembly.GetAssembly(typeof(Extensions)).GetName().Version.ToString(3);
-        private bool embeddedServer;
+        private bool suppressFeeback;
```
Coding-Hen: Is this meant to be suppressFeedback?

grandoth: Yes. Oops.

grandoth:
```suggestion
        private bool suppressFeedback;
```

grandoth:
image

NitroxLauncher/ServerPage.xaml.cs
```diff
@@ -39,10 +52,13 @@ private void StartServer_Click(object sender, RoutedEventArgs e)
 
         private void OnSelectionChange(object sender, SelectionChangedEventArgs e)
         {
-            ComboBox box = (ComboBox)sender;
-            ComboBoxItem item = (ComboBoxItem)box.SelectedValue;
-            
-            embeddedServer = item.Tag.ToString() == "embedded";
+            if (!suppressFeeback)
```
grandoth:
```suggestion
            if (!suppressFeedback)
```

NitroxLauncher/ServerPage.xaml.cs
```diff
@@ -9,27 +9,40 @@ namespace NitroxLauncher
     public partial class ServerPage : Page
     {
         public string Version => "NITROX ALPHA " + Assembly.GetAssembly(typeof(Extensions)).GetName().Version.ToString(3);
-        private bool embeddedServer;
+        private bool suppressFeeback;
         private readonly LauncherLogic logic;
 
         public ServerPage(LauncherLogic logic)
         {
+            suppressFeeback = true;
```
grandoth:
```suggestion
            suppressFeedback = true;
```

NitroxLauncher/ServerPage.xaml.cs
```diff
@@ -9,27 +9,40 @@ namespace NitroxLauncher
...
+            foreach (ComboBoxItem boxItem in CBox.Items)
+            {
+                if ( (Properties.Settings.Default.IsEmbeddedServer && boxItem.Tag.ToString() == "embedded") ||
+                    (!Properties.Settings.Default.IsEmbeddedServer && boxItem.Tag.ToString() != "embedded"))
+                {
+                    CBox.SelectedItem = boxItem;
+                }
+            }
+
+            suppressFeeback = false;
```
grandoth:
```suggestion
            suppressFeedback = false;
```

grandoth:
Right now the settings default is Command Prompt Window based on it being the first in the list. I matched that behavior with the setting. However, do we want to encourage people to use the in launcher console instead? I could swap the order and also make the default the in launcher console...lemme know.

Measurity:
I like the default option to be the most comfortable to the user. If the embedded server matches/exceeds the useability of a console window then it can be the default. **But** I think that's not the case just yet.

One of the issues with the embedded server is the performance when it has a lot of log lines. Also its features are quite limited other than displaying text and we can do more with it.

NitroxLauncher/ServerPage.xaml.cs
grandoth: Commit 56e78c2

grandoth:
Comments resolved. @Measurity. I will work on on making Launcher resizeable next, but I'll make that a separate PR.

grandoth:
@Coding-Hen On the server console page, I've replaced the text buttons with icons for send and stop (exit). Hopefully, now it won't be ambiguous to exiting the app...I'll get the changes checked in hopefully tomorrow.

In terms of UI, I thought it most important that the send button be immediately following the Command Line, and that the command line line up vertically with the console history for monospaced text alignment.

grandoth:
Latest:
image
image

NitroxLauncher/NitroxLauncher.csproj
```diff
@@ -149,6 +149,16 @@
...
+    <Resource Include="Fonts\RobotoMono-Bold.ttf" />
+    <Resource Include="Fonts\RobotoMono-BoldItalic.ttf" />
+    <Resource Include="Fonts\RobotoMono-Italic.ttf" />
+    <Resource Include="Fonts\RobotoMono-Light.ttf" />
+    <Resource Include="Fonts\RobotoMono-LightItalic.ttf" />
+    <Resource Include="Fonts\RobotoMono-Medium.ttf" />
+    <Resource Include="Fonts\RobotoMono-MediumItalic.ttf" />
+    <Resource Include="Fonts\RobotoMono-Regular.ttf" />
+    <Resource Include="Fonts\RobotoMono-Thin.ttf" />
+    <Resource Include="Fonts\RobotoMono-ThinItalic.ttf" />
```
Sunrunner37: Do we need to retain the license in source?

grandoth: Short answer:
I'd say yes, and it's already done.

Long answer:
Per my understanding of the Apache License v2: yes and I would say we have. I confirmed both fonts, "Roboto" and "Roboto Mono", included have the same license (the former's license was there already). I performed a binary compare between the two license files confirming they matched (summarized in 661de77 comments). IMHO a common license file in the same folder applying to both is a reasonable interpretation of the structure. We could include two identical license files with different names or org in a subfolder if you think it's necessary.

What I haven't checked is if we actually distribute the license file in the "install", which Apache requires as I understand. Also, sometimes a copyright notice is at the top of the file license file and/or included in a summary of open source technologies in the distribution...I don't see Google having listed themselves as the copyright holder anywhere of the fonts, but I probably am just not looking in the right place.

<hr>

The above is for reference only and not be construed as legal advice...

grandoth: Is this adequate and should I resolve this review?

NitroxLauncher/ServerConsolePage.xaml
```diff
@@ -16,25 +16,34 @@
...
+                </Button>
+                <Button TabIndex="2" DockPanel.Dock="Right" Margin="5,0,0,0" Width="42" Height="30" Style="{StaticResource BTNCORNER}" Foreground="White" FontSize="16" x:Name="CommandButton" ToolTip="Send command" Click="CommandButton_OnClick" RenderTransformOrigin="0.852,0.559">
                     <Button.Background>
                         <LinearGradientBrush EndPoint="0.5,1" StartPoint="0.5,0">
                             <GradientStop Color="#FF00B1E9" Offset="0"/>
                             <GradientStop Color="#FF006BE9" Offset="1"/>
                         </LinearGradientBrush>
                     </Button.Background>
-                    Send
+                    <Image Source="Images/material-design-icons/baseline_send_white_24dp.png"/>
```
Sunrunner37: Since we are working on localization in a separate PR should we keep this text?

NitroxLauncher/ServerPage.xaml
```diff
@@ -307,36 +307,28 @@
...
-                        </Border>
+                        <TextBlock FontFamily="{StaticResource Roboto}" Foreground="White" FontSize="22"><Run Text="SERVER TYPE"/></TextBlock>
+                        <RadioButton Name="RBIsDocked" GroupName="ServerType" Content="Docked console" Margin="5" FontFamily="{StaticResource Roboto}" Foreground="LightGray" FontSize="14" HorizontalAlignment="Left" VerticalAlignment="Top" Checked="RBServer_Checked"/>
+                        <RadioButton Name="RBIsExternal" GroupName="ServerType" Content="External command prompt" Margin="5" FontFamily="{StaticResource Roboto}" Foreground="LightGray" FontSize="14" HorizontalAlignment="Left" VerticalAlignment="Top" Checked="RBServer_Checked"/>
+                        <!--
+                        <CheckBox Name="ChkIsExternal" Content="External command prompt process" Margin="0,5,0,5" FontFamily="{StaticResource Roboto}" Foreground="LightGray" FontSize="16"/>
+                        <TextBlock FontFamily="{StaticResource Roboto}" Foreground="LightGray" FontSize="14" TextWrapping="Wrap"><Run Text="Launches server in external command prompt process."/></TextBlock>
+                        -->
```
Sunrunner37: Let's remove dead code.

NitroxLauncher/NitroxLauncher.csproj
Sunrunner37: That sounds good to me.

NitroxLauncher/ServerConsolePage.xaml
grandoth: I removed the "Send" text and replaced it with an icon mostly to take up less space on the console command line and to match the added server "stop" button, also no text. I wasn't really thinking about localization, although I see how it's related. There is still a ToolTip that ought to be localized for each.

I can't keep both the image and the text (compile error). If you don't want the icons (at all) or prefer they be implemented as part of the localization PR instead, say the word.

NitroxLauncher/ServerPage.xaml
grandoth: Whoops. Got it. Related to dead code...although I'm brand new to WPF, there is a huge block in the "Page Resources" tag that I think is now OBE since the ComboBox is gone...I'll look into that next.

PR #901, opened 2020-01-27
Localhost gets admin permissions by default

Measurity:


NitroxServer/Communication/Packets/Processors/PlayerJoiningMultiplayerSessionProcessor.cs
```diff
@@ -33,36 +35,44 @@ public override void Process(PlayerJoiningMultiplayerSession packet, NitroxConne
...
             PlayerJoinedMultiplayerSession playerJoinedPacket = new PlayerJoinedMultiplayerSession(player.PlayerContext, techTypes);
             playerManager.SendPacketToOtherPlayers(playerJoinedPacket, player);
 
+            // Make players on localhost admin by default.
+            if (IPAddress.IsLoopback(connection.Endpoint.Address))
+            {
+                world.PlayerData.SetPermissions(player.Name, Perms.ADMIN);
+            }
+            Perms initialPerms = world.PlayerData.GetPermissions(player.Name);
+            Log.Info($"Changed player '{player.Name}' on '{connection.Endpoint.Address}' permissions to '{initialPerms.ToString()}'");
```
Sunrunner37: Do we always want to log?

Coding-Hen: In my opinion, this would just be a given when you host the server so I would say we probably don't need to log it at all

Measurity: Sounds good to me.

PR #902, opened 2020-01-28
Added PlayerID saving to fix Inventory loading

Mathew3000:
Added the Players NitroxID to PersistedPlayerdata. This results in Players inventory being assigned correctly when loading a saved gamestate.


NitroxServer/GameLogic/Players/PlayerData.cs
```diff
@@ -244,6 +278,19 @@ public PersistedPlayerData(string playerName, ushort playerId)
                 PlayerName = playerName;
                 PlayerId = playerId;
             }
+
+            /// <summary>
+            /// Overload for creating new PersistedPlayerData if a nitroxID is present
+            /// </summary>
+            /// <param name="playerName">the players name</param>
+            /// <param name="playerId">the players id</param>
+            /// <param name="nitroxId">the players NitroxID</param>
```
MarijnS95: IMO the documentation for these params don't add anything that isn't implied by their name already.

NitroxServer/Player.cs
```diff
@@ -32,6 +32,24 @@ public Player(PlayerContext playerContext, NitroxConnection connection, Vector3
...
         }
 
+        /// <summary>
+        /// This overload is used when a NitroxId is already present
+        /// </summary>
+        /// <param name="playerContext"></param>
+        /// <param name="connection"></param>
+        /// <param name="position">the stored position</param>
+        /// <param name="playerId">the stored playerId for the new player</param>
+        /// <param name="subRootId"></param>
```
MarijnS95: Can we omit empty documentation?

Besides, I don't think `position` and `playerId` are required to come from persisted storage; it is up to the use of this constructor to fill it in with that data. I don't see it necessary to assume that this constructor is only getting called when loading a player from persisted data.

NitroxServer/GameLogic/Players/PlayerData.cs
```diff
@@ -182,6 +182,33 @@ public Optional<NitroxId> GetSubRootId(string playerName)
...
+                PersistedPlayerData playerPersistedData = GetOrCreatePersistedPlayerData(playerName);
+                return playerPersistedData.PlayerNitroxId;
+            }
+        }
+
+        /// <summary>
+        /// Sets a NitroxID corresponding to a given PlayerName in persisted playerdata
+        /// </summary>
+        /// <param name="playerName">the playername to set the ID for</param>
+        /// <param name="id">the NitroxID to set</param>
```
MarijnS95: While I appreciate comments and documentation, these are only paraphrasing the function name and arguments. Is there anything specific that can be said about the implementation, that is relevant to the caller? Something the documentation _adds on top of_ the method signature?

NitroxModel/DataStructures/NitroxId.cs
```diff
@@ -17,7 +17,17 @@ public NitroxId()
         {
             guid = Guid.NewGuid();
         }
-        
+
+        /// <summary>
+        /// This is used to "recreate" a NitroxID from a String
+        /// The XML-serializer needs this after deserializing to convert from "Helper-Types" to internal data Types
```
MarijnS95: Not a fan of "This is used to", how about "Recreate a NitroxID from a String"?. Besides, I don't see a need to explain where this function is currently being used. Seems obvious to me that it's just a useful helper constructor.

NitroxServer/GameLogic/Players/PlayerData.cs
```diff
@@ -182,6 +182,33 @@ public Optional<NitroxId> GetSubRootId(string playerName)
...
+        public void SetPlayerNitroxID(string playerName, NitroxId id)
+        {
+            lock (playersByPlayerName)
+            {
+                playersByPlayerName[playerName].PlayerNitroxId = id;
```
MarijnS95: Is there a possibility for `playerName` to not be in `playersByPlayerName` (yet)? Given that this is called when the player is constructed and an ID was _not_ found in the save data?

NitroxServer/GameLogic/PlayerManager.cs
```diff
@@ -100,7 +100,24 @@ public Player CreatePlayer(NitroxConnection connection, string reservationKey, o
...
+                if (id != null)
+                {
+                    player = new Player(playerContext, connection, position, id, subRootId);
+                }
+                // If no NitroxId was found then just create a new ID and then save it to persisted data
+                else
+                {
+                    player = new Player(playerContext, connection, position, subRootId);
+                    playerData.SetPlayerNitroxID(playerContext.PlayerName, player.GameObjectId);  
+                }
```
MarijnS95: I'd say this `if`/`else` is obvious enough to not require two sentences of explanation. A single line above the `if` explaining why either path is taken would be plenty.

NitroxServer/Player.cs
```diff
@@ -32,6 +32,15 @@ public Player(PlayerContext playerContext, NitroxConnection connection, Vector3
             GameObjectId = new NitroxId();
         }
 
+        public Player(PlayerContext playerContext, NitroxConnection connection, Vector3 position, NitroxId playerId, Optional<NitroxId> subRootId)
```
MarijnS95: Can you move this constructor above the other one,  and let the other one use this one? That would look like:

```C#
public Player(PlayerContext playerContext, NitroxConnection connection, Vector3 position, Optional<NitroxId> subRootId)
        : this(playerContext, connection, position, new NitroxId(), subRootId)
        {
        }
```

NitroxServer/GameLogic/Players/PlayerData.cs
```diff
@@ -244,6 +266,13 @@ public PersistedPlayerData(string playerName, ushort playerId)
                 PlayerName = playerName;
                 PlayerId = playerId;
             }
```
MarijnS95: Same here, can you reuse the larger constructor? `: this(playerName, playerId, null)`.

NitroxServer/Player.cs
Mathew3000: I did not think about that in that moment

NitroxServer/GameLogic/Players/PlayerData.cs
Mathew3000: Doing that would also make the "SetPlayerNitroxID" obsolete as the player is created with the ID when calling "GetOrCreatePersistedPlayerData"

Mathew3000: And that would also mean the PlayerID cant be "null" in PlayerManager and that would make the If statement obsolete

PR #904, opened 2020-01-30
SceneDebugger update

Jannify:
A few changes to the SceneDebugger:

- Changed global position and rotation to a read-only field
- Changed rotation value fields from quaternion to euler-angles
- Added an option to filter (System/mscorlib and UnityEngine) the "invokable" methods
- Reordered transform values
- Position of all Debuggers resets if the F7 key is pressed to prevent loosing the window
- Bug fixes

grafik
grafik

### DIFF
--- a/NitroxClient/MonoBehaviours/NitroxDebugManager.cs
+++ b/NitroxClient/MonoBehaviours/NitroxDebugManager.cs
@@ -1,5 +1,4 @@
-﻿using System;
-using System.Collections.Generic;
+﻿using System.Collections.Generic;
 using NitroxClient.Debuggers;
 using NitroxModel.Logger;
 using UnityEngine;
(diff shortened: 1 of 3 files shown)


NitroxClient/MonoBehaviours/NitroxDebugManager.cs
```diff
@@ -84,7 +84,8 @@ public void ToggleDebugging()
             {
                 UWE.Utils.PopLockCursor();
                 HideDebuggers();
-            }
+                Debuggers.ForEach(debugger => debugger.ResetWindowPosition());
```
Measurity: Use `foreach` to because functional-style programming should not cause side-effects

MarijnS95: @Measurity In this case I beg to differ; `ForEach` has no return value (unlike `.Select` or other aggregation helpers) meaning it is really only useful when causing side-effects.

Measurity: Because the designers of the C# language think it was a bad design on their part to include `ForEach` which is why it is not on an `IEnumerable<>`.

https://docs.microsoft.com/en-us/archive/blogs/ericlippert/foreach-vs-foreach

MarijnS95: But assigning to a temporary or writing a (usually "long") LINQ expression within a `foreach ()` doesn't look that neat IMO, though in this case where no other transforms are used I agree.

Besides, I'm certain Nitrox has functional code _somewhere_ (outside of `.ForEach`) that causes side-effects - likely written by me. Convenience and readability dominate paradigm :man_shrugging:

PR #905, opened 2020-01-30
Allow rebuilding data for version changes

Coding-Hen:
Rebuilds the data that the version has changed on without having to recreate the whole save file when certian areas don't deserialise.


Measurity:
We should be able to serialize/deserialize without having a "save incompatibility". As long as we're only adding new fields (ProtoMember) after existing fields.

Can you test if `Serializer.SerializeWithLengthPrefix(stream, saveHeader, PrefixStyle.Base128, 1);` should work for us (since we're serializing multiple times per stream).

In short: Try read saves anyway (by using different serialize method) even if version doesn't match up.

PR #906, opened 2020-02-01
Fixed manually registering Peeper

killzoms:
Wasn't what I originally had in mind when opening this branch but it works nonetheless.


Sunrunner37:
Can you provide some context behind the change?

killzoms:
Used GetMembers instead of specifically using GetField and GetProperty as it more closely matches the usage inside protobuf

As for the Protobuf-net.dll I changed an if statement in MetaType.AddField() to instead have 
```
	else if (member != null)
	{
		foreach (MemberInfo memberInfo2 in member)
		{
			if (memberInfo2.DeclaringType == this.type)
			{
				memberInfo = memberInfo2;
				break;
			}
		}
	}
```
 tacked on to allow having overridden members when manually registering. This is already handled in ApplyDefaultBehaviour when the classes are automatically registered

PR #910, opened 2020-02-04
Begin Changes for PrefabPlaceholders

killzoms:
~~As of this moment this is unfinished and still has some bugs~~ 
this is functional in almost every instance of its use ***Im looking at you Mushroom forest***


killzoms:
Alright Most things are positioned correctly however it would appear the Matrix multiplication is failing to give a correct value thus nested Matrices are not "supported" at this time ie wrecks still broken

killzoms:
Positioning is now working as intended next is adding localChanges to Position and Rotation

PR #921, opened 2020-02-10
Stops/Starts periodic saving when last/first player (dis)connects

Coding-Hen:


NitroxServer/GameLogic/PlayerManager.cs
```diff
@@ -135,6 +140,12 @@ public void PlayerDisconnected(NitroxConnection connection)
                 }
 
                 assetsByConnection.Remove(connection);
+
+                if(ConnectedPlayers().Count == 0)
```
Measurity: Minor: No space after if statement; apply formatting to this file.

Coding-Hen: Ty I missed that. I thought our editor config picked this up?

NitroxServer/GameLogic/PlayerManager.cs
```diff
@@ -82,6 +82,11 @@ public List<Player> GetPlayers()
                 reservations.Add(reservationKey, playerContext);
                 assetPackage.ReservationKey = reservationKey;
 
+                if (ConnectedPlayers().Count > 0 && reservedPlayerNames.Count == 1)
```
Measurity: The additional `reservedPlayerNames.Count == 1` seems to be redundant.

NitroxServer/GameLogic/PlayerManager.cs
Measurity: Yes it should. If the code style is wrong then we should fix that.

NitroxServer/GameLogic/PlayerManager.cs
Coding-Hen: I did this because to prevent when having multiple players connect that it keeps starting the timer

Measurity: Can do ConnectedPlayers().Count == 1 then right?

Sunrunner37:
This may have an unintended side effect of forcing players to manually save when exiting the server.  Is there a problem with leaving this running? (outside of needing to suppress the timer when players are not on)

Coding-Hen:
> 
> 
> This may have an unintended side effect of forcing players to manually save when exiting the server. Is there a problem with leaving this running? (outside of needing to suppress the timer when players are not on)

Is there a problem with causing them to save upon the last player leaving? If they are going to close the server anyway it will save and I would have though we would probably want a save to be written if nobody else is connected as then if the server crashes then we have the last save of when someone was connected.

NitroxServer/GameLogic/PlayerManager.cs
```diff
@@ -135,6 +140,12 @@ public void PlayerDisconnected(NitroxConnection connection)
                 }
 
                 assetsByConnection.Remove(connection);
+
+                if (ConnectedPlayers().Count == 0)
+                {
+                    Server.Instance.Save();
+                    Server.Instance.DisablePeriodicSaving();
```
Measurity, line 147: Probably should swap these so that the timer is stopped before saving in case the timer triggers at the same time as this save call.

PR #924, opened 2020-02-11
Added "==" and "!=" operator to NitroxId. (Fixes repairability of escape pod)

Her0brineCreeper:
This patch makes the secondary systems and the radio of the escape pod repairable again.
The == operator doesn't seem to work for comparing the NitroxIds so I changed it to .Equals() to make the repairing work.


killzoms:
Since we are here already do you want to add the "==" and "!=" operators to NitroxId to avoid future issues?

Her0brineCreeper:
Here you go.
I added the "==" and "!=" operators to NitroxId and changed the ".Equals()" back to "==".

PR #925, opened 2020-02-12
Log version at patcher and server start

grandoth:
-Log version at patcher and server start
-Add version to ReflectionHelper (couldn't get XAML to bind to it perhaps due to needing PropertyChanged interface?)
-Remove ServerPage ComboBox deadcode (Win 7 style) since ComboBox is removed and Microsoft ended extended support for Windows 7.
-Add multiple res Nitrox logos in a proper ico file until logo redesign; better than .NET/Windows resizing. Add icon to NitroxLauncher.exe

A few questions/comments:
1. So now we can version all the assemblies with single AssemblyInfoCommon.cs allowing them to roll together. So should we simply use Assembly.GetExecutingAssembly().GetName().Version for everything instead of getting the assembly that contains the Extensions type?
2. I didn't replace with the ReflectionHelper version property here since it was a stub...
.\NitroxModel\Packets\MultiplayerSessionPolicy.cs:
3. As mentioned above, the XAML UI page refs to version have their own properties. Is there some magic that can help make this work? Clues that I found seem to reference needing the PropertyChanged interface which seemed like overkill for something that is not changing at runtime...


grandoth:
Understand your comment. Will get to it when I can (hopefully in the next week). We're on day 2 of n for our house remodel...Time is short right now.

grandoth:
This is OBE.

PR #926, opened 2020-02-12
Feature/80cmdude/#913 n log

80cmdude:
Nice **_small_** first pull request. 👍 

Removed log4net and replaced it with NLog instead. Updated all references to the old static class and replaced with a new signleton instance of Nlog. All the writes to both console and file are handled async. 
Ive basically just copied what there was already. NLog is really extensible, you can do crazy things like write to mongo db etc, but i thought anything more than what is currently there is overkill.

**Acceptance Criteria**
- Performance: Should be easily be able to handle over 1k writes batched without affecting performance too much (see https://github.com/nlog/NLog/wiki/AsyncWrapper-target)
- Sensitive data done through a method as I really dont want to use Regex 🤢 
- Simple #ifdebug to ingnore Trace and Debug on release build
- _hopefully_ easy to understand format.

**Questions**
1. Do we need to make the singleton creation for NLog thread safe? I was thinking about just doing it but it might hit performance if we send loads of async logs and have to check a lock.
2. Are we happy with format?
`Console: |Date|Type|Message|Exception`
`File: Date|Type|Message|Exception|StackTrace`


80cmdude:
- region syntax okay :+1: , just my personal preference to manage code but I don't mind removing I'll use comments :+1:
- static class okay I can revert that but I think when it comes to performance it's not a huge difference because it's a singleton instance that holds the Nlog instance directly. Personally I'm just used to ignoring statics because of DI but I'm fine for that :) I am aware the debug attribute is nice for statics.
- I'm tracing log sensitive because I want to locally log the real values in the console but ignore the values when I put them into the log file. I actually specifically set up the Nlog file to allow this. Unless you are happy to never log sensitive info (console or file)? That's the reason I did it this way :) 
- I would quite like to keep logcategory. Although if I'm refactoring to a static class I'll be changing to log.logmessage(catogoery,message). This is the one I really wouldn't like to budge on because I want to keep the log class as clean as possible. I understand it's harder for Devs in code but it's only one property and saves maintaining loads of log.debugmessage() log.tracemessage() etc that all do the same thing. However if you really would like I can revert. If the SDK was more complicated that it currently is then I would agree with you.

Questions
- do you not want the log file in bin? Sorry I assumed you would want it there. I missed that when I made the change.

 Sorry if this formatting is crap im on a train on my phone

80cmdude:
Yeah so screw everything i said about log catogorey.. after changing it back to static and thinking about it it does make more sense to split them all out. Not worked on a open source project before and I rarely use static classes. ^^

Ive not finished everything im just waiting on your answer about the sensitive logging :)

Measurity:
The way you handled sensitive logging is very good. Using trace for this is fine by me.

For now I think it's best to match NLog configuration to how logging was with log4net and then change it later in another PR if needed. You probably want to configure NLog from code since the Nitrox project basically has different working directories (launcher, server and game) and won't all have access to the NLog config file.

Also didn't notice a log file the first time (could have missed it). Could be worth it to add the base path to the filename in case other ppl get the same issue: https://stackoverflow.com/a/16485855/1277156

80cmdude:
@Measurity moved the nlog file and added the base dir, also cleaned up the class so its nice and readable 👍

NitroxPatcher/Main.cs
```diff
@@ -99,7 +99,7 @@ private static void InitPatches()
 
             Multiplayer.OnBeforeMultiplayerStart += Apply;
             Multiplayer.OnAfterMultiplayerEnd += Restore;
-            Log.Info("Completed patching using " + Assembly.GetExecutingAssembly().FullName);
+            Log.DebugSensitive("Completed patching using {0}", Assembly.GetExecutingAssembly().FullName);
```
Measurity: This can be kept as info and assembly name is not sensitive information.

NitroxServer/ConsoleCommands/ChangeAdminPasswordCommand.cs
```diff
@@ -28,7 +28,7 @@ public override void RunCommand(string[] args, Optional<Player> player)
             }
             catch (Exception ex)
             {
-                Log.Error("Error attempting to change admin password: " + args[0], ex);
+                Log.Exception("Error attempting to change admin password:", ex);
```
Measurity:
```suggestion
                Log.Exception("Error attempting to change admin password", ex);
```

NitroxServer/ConsoleCommands/ChangeServerPasswordCommand.cs
```diff
@@ -40,7 +40,7 @@ public override bool VerifyArgs(string[] args)
         private void ChangeServerPassword(string password, string name)
         {
             serverConfig.ChangeServerPassword(password);
-            Log.Info($"Server password changed to \"{password}\" by {name}");
+            Log.DebugSensitive("Server password changed to {0} by {1}", password, name);
```
Measurity: Keep quotes around `password` and add them to `name`

NitroxServer/Serialization/World/WorldPersistence.cs
```diff
@@ -209,10 +209,10 @@ private World CreateFreshWorld()
                                                               NitroxServiceLocator.LocateService<Dictionary<string, List<PrefabAsset>>>());
 
             Log.Info("World GameMode: " + gameMode);
-
-            Log.Info("Server Password: " + (string.IsNullOrEmpty(config.ServerPassword) ? "None. Public Server." : config.ServerPassword));
-            Log.Info("Admin Password: " + config.AdminPassword);
-
+            var serverPass = (string.IsNullOrEmpty(config.ServerPassword) ? "None. Public Server." : config.ServerPassword);
+            Log.DebugSensitive("Server Password: {0} ", serverPass);
```
Measurity: Should be kept as info because in RELEASE mode debug is disabled and then logs will not print.

Sunrunner37:
Let's keep the log naming the same `Info/Debug/Error` as it likely isn't worth nuking our easy access to git blame to change and the names are shorter/intuitive already.

NitroxModel/Logger/Log.cs
```diff
@@ -1,148 +1,114 @@
...
-        public static void Debug(string fmt, params object[] arg)
+        public static void RegisterInGameLogger(InGameLogger gameLogger)
         {
-            log.Debug(Format(fmt, arg));
+            logger.Info("Registered InGameLogger");
+            inGameLogger = gameLogger;
         }
 
-        public static void Debug(object o)
+        public static void SetInGameMessagesEnabled(bool enabled)
```
Measurity: Why change this method when it's only ever enabled?

NitroxServer/Serialization/World/WorldPersistence.cs
80cmdude: Yup thats a miss on my part there moving from info to debug

NitroxModel/Logger/Log.cs
80cmdude: So back to your earlier comment about not changing things that already exist.. I noticed this method is called to enable but never called to disable.. I thought about removing this but then in the future someone might want to disable it. I just didnt want to lost functionality that already existed

NitroxServer/ConsoleCommands/ChangeAdminPasswordCommand.cs
```diff
@@ -40,7 +40,7 @@ public override bool VerifyArgs(string[] args)
         private void ChangeAdminPassword(string password, string name)
         {
             serverConfig.ChangeAdminPassword(password);
-            Log.Info($"Admin password changed to \"{password}\" by {name}");
+            Log.DebugSensitive("Admin password changed to {0} by {1}", password, name);
```
Measurity: Use info here too.

NitroxServer/ConsoleCommands/ChangeServerPasswordCommand.cs
```diff
@@ -40,7 +40,7 @@ public override bool VerifyArgs(string[] args)
         private void ChangeServerPassword(string password, string name)
         {
             serverConfig.ChangeServerPassword(password);
-            Log.Info($"Server password changed to \"{password}\" by {name}");
+            Log.DebugSensitive("Server password changed to \"{0}\" by \"{1}\"", password, name);
```
Measurity: Use info instead of debug.

NitroxModel/Logger/Log.cs
Measurity: In this case it was designed to only be enbled and never disabled. It limits things that can go wrong (accidental disable and don't know why game messages stop working).

It's best to not support features if we don't need them at all to reduce technical debt.

80cmdude: Okay that's fair i shall revert this and remove the boolean

dartasen:
@80cmdude Up ? This could be an amazing PR

brianpopow:
I think nlog would be a really nice feature. I think this can make debugging much easier, if  structured logging is used. I want to try that out and for that it would be good if this PR is merged.

@80cmdude do you mind if i take over and finish this up in another PR?

Measurity:
This is being continued here #1031 I'll leave this open until one of the PRs is merged.

Sunrunner37:
https://github.com/SubnauticaNitrox/Nitrox/pull/1121

PR #927, opened 2020-02-14
Refactored serializable classes

killzoms:
This should solve the issue of things being lost at deserialization

### DIFF
--- a/NitroxServer/GameLogic/Bases/BaseData.cs
+++ b/NitroxServer/GameLogic/Bases/BaseData.cs
@@ -23,7 +23,13 @@ namespace NitroxServer.GameLogic.Bases
                     return serializableBasePiecesById;
                 }
             }
-            set { serializableBasePiecesById = basePiecesById = value; }
+            set
+            {
+                lock (basePiecesById)
+                {
+                    serializableBasePiecesById = basePiecesById = value;
+                }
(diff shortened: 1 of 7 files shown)


NitroxServer/GameLogic/Bases/BaseData.cs
```diff
@@ -11,36 +11,11 @@ namespace NitroxServer.GameLogic.Bases
...
-            }
-            set { basePiecesById = value; }
-        }
 
-        [ProtoMember(2)]
-        public List<BasePiece> SerializableCompletedBasePieceHistory
-        {
-            get
-            {
-                lock (completedBasePieceHistory)
```
Sunrunner37: Is it safe to remove the locks?  All of the functions below leverage these locks too.  For example, I believe AddBasePiece can be called from individual client threads while the saving can happen on the main thread based on a timer (at the same time)

killzoms: It "MIGHT" run into issues depending on how protobuf-net is programmed but It would appear Protobuf-net is thread safe

killzoms: https://github.com/protobuf-net/protobuf-net/wiki/Capabilities

PR #928, opened 2020-02-14
Optional refactor

killzoms:
This allows Optional to be serialized by Protobuf and removes the abstract classes
Note: I Believe this might fix some base issues however that has yet to be tested. Testing requires PR #927 to be merged

### DIFF
--- a/NitroxModel/DataStructures/Optional.cs
+++ b/NitroxModel/DataStructures/Optional.cs
@@ -100,6 +100,16 @@ namespace NitroxModel.DataStructures.Util
             return !optional.HasValue;
         }
 
+        public static T Get<T>(this Optional<T> optional)
+        {
+            if (optional.IsEmpty())
+            {
+                throw new OptionalEmptyException<T>();
+            }
+
+            return optional.Value;
+        }
+
         public static T OrElse<T>(this Optional<T> optional, T elseValue)
         {
             if (optional.IsEmpty())
--- a/NitroxModel/DataStructures/Optional.cs
+++ b/NitroxModel/DataStructures/Optional.cs
@@ -67,11 +67,6 @@ namespace NitroxModel.DataStructures.Util
             return new Optional<T>(value);
         }
 
-        public T Get()
-        {
-            return Value;
-        }
-
         public static implicit operator Optional<T>(T obj)
         {
             return new Optional<T>(obj);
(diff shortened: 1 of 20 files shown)


NitroxModel/DataStructures/Optional.cs
```diff
@@ -1,106 +1,101 @@
...
-    public class HasValueOptional<T> : Optional<T>
+    [ProtoContract]
+    public class Optional<T>
     {
-        private readonly T value;
-
-        public HasValueOptional(T value)
-        {
-            this.value = value;
+        public T Value
```
Sunrunner37: What is the benefit of attribute vs function?  Is the switch worth nuking all of our easy access to git blame?

killzoms: In our previous implementation we had to convert from an Optional to a non optional value which may or may not fall apart at runtime(Proto issues)

killzoms: Also yea I should have left .Get() instead of just changing everything like this

MarijnS95: All other reasoning aside, the "rename" changes can be moved to a separate commit and listed in a rev-ignore file. Pass the file to [`git blame --ignore-revs-file`](https://git-scm.com/docs/git-blame#Documentation/git-blame.txt---ignore-revs-fileltfilegt) or set `blame.ignoreRevsFile` to the filename.

Granted, everyone needs to do this on their own, locally.

NitroxTest/Model/OptionalTest.cs
```diff
@@ -78,6 +78,14 @@ public void OptionalValueTypeOrElseValidValue()
             Assert.AreEqual(1, op.OrElse(2));
         }
 
+        [TestMethod]
+        public void OptionalSetValue()
+        {
+            Optional<int> op = Optional<int>.Of(0);
+            op = null;
+            Assert.IsTrue(op.IsEmpty());
```
Sunrunner37: This behavior seems like a bug.  Optional should never be null - if it is, a developer should get a null pointer exception so they know about it instead of silently failing.  

With that being said, that  change likely loses the benefit for having OptionalExtentions.  We may want to just put that code in the optional class unless we have other reasons for keeping it.

MarijnS95: We could still throw a detailed NPE from the extension method?

Sunrunner37: We could do that, but what is the purpose of the extension methods?

MarijnS95: In that case it becomes as useless as embedding the function directly in the Optional class, except for a more obvious error message. One might accidentally see an NPE on `op.xyz()` as the value not being set, when `op` itself is in fact `null`.

I haven't settled on making `Optional` a value-type that is default-null though. Right now it seems like we're being explicit on top of being explicit: both when setting _and_ getting a possibly-optional value.

killzoms: The purpose of the extension methods was for when someone sets the Optional to null when assigning a value to it say

```csharp
Optional<randomType> opt = new Optional<randomType(randomTypeValue);
if (opt.IsPresent())
{
    Console.WriteLine("Optional has a value! {0}", opt.Value);
}

opt = null;
if (opt.IsEmpty())
{
    Console.WriteLine("Optional is Empty!");
}

opt = newRandomTypeValue;

if (opt.IsPresent())
{
    Console.WriteLine("Optional Is Present! {0}", opt.Value);
}
```

PR #938, opened 2020-02-22
Serialization fixes

killzoms:


NitroxClient/GameLogic/EscapePodManager.cs
```diff
@@ -28,27 +28,37 @@ public EscapePodManager(IPacketSender packetSender, IMultiplayerSession multipla
             this.multiplayerSession = multiplayerSession;
         }
 
-        public void AssignPlayerToEscapePod(EscapePodModel escapePod)
+        public void AssignPlayerToEscapePod(Optional<EscapePodModel> escapePod)
```
Sunrunner37: Why do we want this to be optional?  There should always be an escape pod for the player.

NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs
```diff
@@ -89,6 +89,18 @@ public List<Entity> LoadUnspawnedEntities(Int3 batchId)
                 Log.Info("Spawning " + entities.Count + " entities from " + spawnPoints.Count + " spawn points in batch " + batchId);
             }
 
+
+            for (int x = 0; x < entities.Count; x++)
```
Sunrunner37: Was this for testing?  If not, let's break into a new function.  Also, it might be cleaner/faster to index in a map then loop through that.

NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs
```diff
@@ -270,6 +283,11 @@ private void AssignPlaceholderEntitiesIfRequired(Entity entity, string classId,
                         continue;
                     }
 
+                    if (prefab.EntitySlot.IsPresent())
+                    {
+                        //SpawnEntitiesUsingRandomDistribution(new EntitySpawnPoint(entity.AbsoluteEntityCell, transform.LocalPosition, transform.LocalRotation, )) // Unimplemented
```
Sunrunner37: dead code

NitroxServer/GameLogic/Entities/NitroxEntitySlot.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.Text;
+
+namespace NitroxServer.GameLogic.Entities
+{
+    public class NitroxEntitySlot
+    {
+        public string[] AllowedTypes;
+
+        public string BiomeType;
+
```
Sunrunner37: It doesn't look like these are used.  Should they be?  Or are we just doing this to ensure the parsing cursor moves?

NitroxClient/GameLogic/EscapePodManager.cs
killzoms: should always be one however there are instances where a packet may be partially sent or drserialized when we get it(throwing that null exception) did check if it made a difference and the answer... the player was able to log into the server fine with the npe not being thrown hell even werewolfs could connect

NitroxServer/GameLogic/Entities/NitroxEntitySlot.cs
killzoms: this is in place for something im finishing in the near future same for the dead code above to allow EntitySlot spawning under prefab placeholders

NitroxClient/GameLogic/EscapePodManager.cs
Sunrunner37: I don't think this is a transmission problem as there are parity checks and the packet should always come in with a good state.  Most likely we have a bug that is causing the player's assigned id to not be sent or incorrectly loading the data.  It could also be a silent failure in deserialization. Irrespective of the cause, we should identify the root problem and fix it; maybe if someone is able to consistently replicate it then identify what piece is missing.

MarijnS95: And even if we were to take "advantage" of protobufs partial sending / merge / "unpack from stream", the packet should still only be processed when it's complete.

Indeed, this sounds like a mistake elsewhere.

killzoms: Back when testing this... I was able to consistently replicate the issue however I don't remember if It was a protobuf serialization issue(likely) or an issue with other information however as of making this PR it would appear to have been fixed, Having it optional is just for IsEmpty check to use Log.Error

NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs
Sunrunner37: Let's do this one as well.

NitroxClient/GameLogic/EscapePodManager.cs
```diff
@@ -30,8 +30,13 @@ public EscapePodManager(IPacketSender packetSender, IMultiplayerSession multipla
 
         public void AssignPlayerToEscapePod(EscapePodModel escapePod)
         {
+            if (escapePod != null)
```
Sunrunner37: Do we consider this a release blocker?  How often is it happening?

NitroxClient/GameLogic/EscapePodManager.cs
```diff
@@ -30,8 +30,13 @@ public EscapePodManager(IPacketSender packetSender, IMultiplayerSession multipla
...
+            {
+                throw new NullReferenceException("ESCAPE POD NOT FOUND!!!");
```
Sunrunner37: We can do `Validate.NotNull(escapePod, "Escape pod was null");`

NitroxClient/GameLogic/EscapePodManager.cs
killzoms: It doesn't happen often so... Not really a Release blocker

killzoms:
Something's Breaking at the moment so not ready for Merge

NitroxClient/GameLogic/Entities.cs
```diff
@@ -122,6 +128,11 @@ private IEntitySpawner ResolveEntitySpawner(Entity entity)
...
             TechType techType = entity.TechType.Enum();
 
+            if (entity.ClassId == "55d7ab35-de97-4d95-af6c-ac8d03bb54ca")
+            {
+                return new CellRootSpawner();
+            }
+
             if (customSpawnersByTechType.ContainsKey(techType))
             {
                 return customSpawnersByTechType[techType];
```
Sunrunner37: Would it make sense to make this customSpawnersByClassId?  Then we can put those hard coded class ids into an enum so it would be more readable.

NitroxClient/GameLogic/Spawning/CellRootSpawner.cs
```diff
@@ -0,0 +1,34 @@
...
+using NitroxModel.DataStructures.GameLogic;
+using NitroxModel.DataStructures.Util;
+using UnityEngine;
+using NitroxInt3 = NitroxModel.DataStructures.Int3;
+
+namespace NitroxClient.GameLogic.Spawning
+{
+    public class CellRootSpawner : IEntitySpawner
+    {
+        public Optional<GameObject> Spawn(Entity entity, Optional<GameObject> parent, EntityCell cellRoot)
```
Sunrunner37: Is this ever called with a parent?

NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs
```diff
@@ -282,9 +314,33 @@ private void AssignPlaceholderEntitiesIfRequired(Entity entity, string classId,
                                             deterministicBatchGenerator.NextId(),
                                             entity);
 
+                    if (prefab.EntitySlot.IsPresent())
+                    {
+                        Entity possibleEntity = SpawnEntitySlotEntities(prefab.EntitySlot.Get(), transform, deterministicBatchGenerator, entity);
+                        if (possibleEntity != null)
+                        {
+                            entity.ChildEntities.Add(possibleEntity);
```
Sunrunner37: Should we add all of the entities that are spawned in `SpawnEntitySlotEntities`?  From your logic, it looks like you only add the first.  Is there significance to that?

NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs
```diff
@@ -218,11 +231,30 @@ private IEnumerable<Entity> CreateEntityWithChildren(EntitySpawnPoint entitySpaw
 
             yield return spawnedEntity;
 
-            // Children are yielded as well so they can be indexed at the top level (for use by simulation 
-            // ownership and various other consumers).  The parent should always be yielded before the children
-            foreach (Entity childEntity in spawnedEntity.ChildEntities)
+            if (parentEntity == null)
```
Sunrunner37: Why does the spawnedEntity need to be parentless to emit its children?  Does this ensure that we only emit at the very top level (cell root)?  If that is the case, can spawnedEntity above be yielded twice if its cell root were to yield it as a child?  Let's clarify the exact behavior of this condition in a comment.

NitroxClient/GameLogic/Spawning/CellRootSpawner.cs
killzoms: CellRoots never have a parent, if they did UWE made a big mistake somewhere

NitroxClient/GameLogic/Entities.cs
killzoms: while we can do that it also adds more complexity

NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs
killzoms: there should only be ONE entity being returned which is always the first one, the rest of the returned entities are children

PR #940, opened 2020-02-28
Unauthenticated packet handler fix

killzoms:


Sunrunner37:
This doesn't look like the right fix.  The client shouldn't receive packets that are unable to be resolved to a type.  This is simply ignoring the problem which is a silent failure.

Sunrunner37:
The reason these were going to unauthenticated has been fixed.  Closing out.

PR #946, opened 2020-03-01
Optional change

killzoms:


NitroxTest/Model/Packets/PacketsSerializableTest.cs
```diff
@@ -39,5 +43,59 @@ public void AllPacketsAreSerializable()
...
+                                                       !t.IsAbstract);
+            foreach (Type type in surrogates)
+            {
+                ISerializationSurrogate surrogate = (ISerializationSurrogate)Activator.CreateInstance(type);
+                Type surrogatedType = type.BaseType.GetGenericArguments()[0];
+                surrogateSelector.AddSurrogate(surrogatedType, streamingContext, surrogate);
+            }
+
+            // For completeness, we could pass a StreamingContextStates.CrossComputer.
+            BinaryFormatter serializer = new BinaryFormatter(surrogateSelector, streamingContext);
```
Sunrunner37: In a follow-up, we may want to pull this out into a factory class as it is a duplicate of the code in `Packet`

PR #952, opened 2020-03-10
Various Entity spawning changes

killzoms:
Might be one or two slight issues to fix in the future but for now FPS FIXES!!!

### DIFF
--- a/NitroxClient/GameLogic/Spawning/DefaultEntitySpawner.cs
+++ b/NitroxClient/GameLogic/Spawning/DefaultEntitySpawner.cs
@@ -1,6 +1,7 @@
 ﻿using NitroxClient.MonoBehaviours;
 using NitroxModel.DataStructures.GameLogic;
 using NitroxModel.DataStructures.Util;
+using NitroxModel.Logger;
 using NitroxModel_Subnautica.Helper;
 using UnityEngine;
 using UWE;
@@ -35,10 +36,23 @@ namespace NitroxClient.GameLogic.Spawning
             {
                 gameObject.transform.SetParent(parent.Get().transform, true);
             }
+            else if (entity.ParentId != null)
+            {
+                NitroxEntity.AddToParent(entity.ParentId, gameObject);
+            }
 
             if (parent.IsPresent() && parent.Get().GetComponent<LargeWorldEntityCell>() != null)
             {
                 LargeWorldEntity.Register(gameObject); // This calls SetActive on the GameObject
+            }
+            else if (gameObject.GetComponent<LargeWorldEntity>() != null && gameObject.transform.parent == null)
+            {
+                gameObject.transform.SetParent(cellRoot.liveRoot.transform, true);
+                LargeWorldEntity.Register(gameObject);
+            }
+            else if (gameObject.transform.parent != null && parent.Get().GetComponent<LargeWorldEntityCell>() != null)
+            {
+                LargeWorldEntity.Register(gameObject);
             }
             else
             {
--- a/NitroxClient/MonoBehaviours/NitroxEntity.cs
+++ b/NitroxClient/MonoBehaviours/NitroxEntity.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using NitroxModel.DataStructures;
+using NitroxModel.DataStructures.GameLogic;
 using NitroxModel.DataStructures.Util;
 using NitroxModel.Helper;
 using ProtoBuf;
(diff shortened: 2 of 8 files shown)


NitroxClient/MonoBehaviours/NitroxEntity.cs
```diff
@@ -61,6 +94,27 @@ public static void SetNewId(GameObject gameObject, NitroxId id)
...
+                    childGameObject.transform.SetParent(gameObject.transform, true);
+                    if (childGameObject.GetComponent<LargeWorldEntity>() != null)
+                    {
+                        LargeWorldEntity.Register(childGameObject);
+                    }
+                }
+
+                gameObjectsWaitingForParent.Remove(id);
+            }
+
```
Sunrunner37: I don't think this logic belong here.  We do not need to do this look up every time we add an id - we add ids all the time to game objects.

NitroxClient/MonoBehaviours/NitroxEntity.cs
```diff
@@ -14,6 +15,8 @@ public class NitroxEntity : MonoBehaviour, IProtoTreeEventListener
     {
         private static Dictionary<NitroxId, GameObject> gameObjectsById = new Dictionary<NitroxId, GameObject>();
 
+        private static Dictionary<NitroxId, List<GameObject>> gameObjectsWaitingForParent = new Dictionary<NitroxId, List<GameObject>>();
```
Sunrunner37: Let's move this logic out into a new class.  We can call it right before spawning a new entity.

NitroxClient/MonoBehaviours/NitroxEntity.cs
```diff
@@ -61,6 +94,27 @@ public static void SetNewId(GameObject gameObject, NitroxId id)
                 entity = gameObject.AddComponent<NitroxEntity>();
             }
 
+            List<GameObject> childGameObjects = new List<GameObject>();
+
+            if (gameObjectsWaitingForParent.TryGetValue(id, out childGameObjects))
+            {
+                foreach (GameObject childGameObject in childGameObjects)
+                {
+                    if (childGameObject == null)
```
Sunrunner37: How will this code interact with the simulation framework?  Will we still keep the request to simulate if an entity is pending?

NitroxPatcher/Patches/Dynamic/ProtobufSerializer_GetOrAddComponent_Patch.cs
```diff
@@ -0,0 +1,59 @@
...
+using System.Linq;
+using System.Reflection;
+using System.Reflection.Emit;
+using System.Text;
+using Harmony;
+using NitroxModel.DataStructures.Util;
+
+namespace NitroxPatcher.Patches.Dynamic
+{
+    class ProtobufSerializer_GetOrAddComponent_Patch : NitroxPatch, IDynamicPatch
```
Sunrunner37: Let's add some robust documentation at the top to specify the purpose/function of this patch.

NitroxClient/GameLogic/Spawning/DefaultEntitySpawner.cs
```diff
@@ -35,11 +36,24 @@ public Optional<GameObject> Spawn(Entity entity, Optional<GameObject> parent, En
...
             if (parent.IsPresent() && parent.Get().GetComponent<LargeWorldEntityCell>() != null)
             {
                 LargeWorldEntity.Register(gameObject); // This calls SetActive on the GameObject
             }
+            else if (gameObject.GetComponent<LargeWorldEntity>() != null && gameObject.transform.parent == null)
+            {
+                gameObject.transform.SetParent(cellRoot.liveRoot.transform, true);
+                LargeWorldEntity.Register(gameObject);
+            }
+            else if (gameObject.transform.parent != null && parent.Get().GetComponent<LargeWorldEntityCell>() != null)
```
Sunrunner37: Isn't this case already handled in the top 'if' statement:

```            
if (parent.IsPresent() && parent.Get().GetComponent<LargeWorldEntityCell>() != null)
```

Sunrunner37:
I was thinking a bit more on this.  To segment the logic, it might make sense to have an event in NitroxEntity so classes can subscribe to new entity addition events.  This would allow us to extract out the parent checking to a new class to decouple the logic.

Sunrunner37:
Included in  https://github.com/SubnauticaNitrox/Nitrox/commit/638973024ae0f9b7e49952ee5c0ad80908d960ab

PR #953, opened 2020-03-10
Added error if Nitrox Launcher is started from temp directory

Measurity:
Should reduce issues in help channel on discord down the line by running Nitrox Launcher from zip without extracting it first.

### DIFF
--- a/NitroxLauncher/App.xaml.cs
+++ b/NitroxLauncher/App.xaml.cs
@@ -1,4 +1,7 @@
 ﻿using System;
+using System.IO;
+using System.Reflection;
+using System.Timers;
 using System.Windows;
 using System.Windows.Controls;
 using System.Windows.Documents;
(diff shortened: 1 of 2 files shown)


NitroxLauncher/App.xaml.cs
```diff
@@ -21,6 +24,14 @@ protected override void OnStartup(StartupEventArgs e)
                 DefaultValue = FindResource(typeof(Page))
             });
             
+            // Error if running from a temporary directory because Nitrox Launcher won't be able to write files directly to zip/rar
+            // Tools like WinRAR do this to support running EXE files while it's still zipped.
+            if (Directory.GetCurrentDirectory().StartsWith(Path.GetTempPath(), StringComparison.OrdinalIgnoreCase))
+            {
+                MessageBox.Show("Nitrox launcher should not be executed from a temporary directory. Install Nitrox launcher properly by extracting and moving it to a dedicated location on your PC.", "Invalid working directory", MessageBoxButton.OK, MessageBoxImage.Error);
```
Sunrunner37: let's say , "by extracting ALL files" so someone doesn't just extract the launcher.

PR #959, opened 2020-03-12
Added beacon name sync

Measurity:
- Added beacon sync
- Added beacon persist to server for new players or reconnecting players
- Fixes a locking issue
- Fixed equipment helper checking for null on component types instead of doing it 'the unity way'
- Removed unused GenericNode class
- Fixed code styling for array initializers
- Refactored and added defensive checks to player stats UI so that it doesn't crash if elements go missing (due to gamemode change)

Suppressed the ping/beacon rename packet on initial sync so that connecting players don't fire beacon rename packets.

Initially I used `PingManager.onRenamed` event but doing it this way would deviate from the standard patching mechanism so I went with a patch instead.


Measurity:
Bump

NitroxClient/Communication/Packets/Processors/InitialPlayerSyncProcessor.cs
```diff
@@ -36,19 +39,21 @@ public override void Process(InitialPlayerSync packet)
         
         private IEnumerator ProcessInitialSyncPacket(object sender, EventArgs eventArgs)
         {
-            bool moreProcessorsToRun;
-
-            do
+            // Some packets should not fire during game session join but only afterwards so that initialized/spawned game objects don't trigger packet sending again. 
+            using (packetSender.Suppress<PingRenamed>())
```
Sunrunner37, line 43: Should we add this closer to the code that invokes this?

NitroxClient/Communication/Packets/Processors/PingRenamedProcessor.cs
```diff
@@ -0,0 +1,51 @@
...
+            if (!beacon)
+            {
+                // This can be ok if origin of ping instance component was not from a beacon (but from signal or other).
+                return;
+            }
+            if (beacon.GetComponent<Player>())
+            {
+                // Skip over beacon component on player GameObjects
+                return;
+            }
```
Sunrunner37, line 42: log the ones that are error?

NitroxClient/GameLogic/ChatUI/PlayerChat.cs
```diff
@@ -60,8 +60,12 @@ public void HideLog()
 
         public void ShowChat()
         {
-            ShowLog();
+            if (!InputField.CanEnable())
```
Sunrunner37, line 63: What does this check do?

NitroxClient/Unity/Helper/GameObjectHelper.cs
```diff
@@ -55,5 +57,45 @@ public static GameObject RequireGameObject(string name)
...
         }
+
+        /// <summary>
+        ///     Calls the selector if the Unity object is alive. Returns null without call if not alive.
+        /// </summary>
+        /// <param name="obj"></param>
+        /// <param name="selector"></param>
+        /// <typeparam name="TIn"></typeparam>
+        /// <typeparam name="TOut"></typeparam>
+        /// <returns></returns>
```
Sunrunner37, line 68: These are not providing value.  Same below.

NitroxServer/Player.cs
```diff
@@ -137,15 +137,13 @@ public void SendPacket(Packet packet)
...
+            if (obj == null)
             {
                 return false;
             }
 
-            Player player = (Player)obj;
-
-            return player.Id == Id;
+            Player player = obj as Player;
+            return player != null && player.Id == Id;
```
Sunrunner37, line 146: We already checked for null above.

NitroxModel/Packets/PingRenamed.cs
```diff
@@ -0,0 +1,39 @@
...
+    public class PingRenamed : Packet
+    {
+        [ProtoMember(1)]
+        public NitroxId Id { get; }
+
+        [ProtoMember(2)]
+        public string Name { get; }
+        
+        [ProtoMember(3)]
+        public byte[] BeaconGameObjectSerialized { get; }
```
Sunrunner37, line 18: Do all pings have beacons?

Measurity: No, all beacons have pings.

NitroxClient/Communication/Packets/Processors/PingRenamedProcessor.cs
Measurity: None of these have to be errors. And this information doesn't seem useful most of the time. I wish we had some `.Trace()` or just disable logs from this class by default.

PR #961, opened 2020-03-14
Admin Mode Patch for launcher

dartasen:
Detect when launcher is running inside C:\Program Files and warn that the launcher likely needs admin permissions to properly install Nitrox into Subnautica. Provide an option in the popup to restart the launcher with admin permissions


NitroxLauncher/App.xaml.cs
```diff
@@ -30,6 +30,31 @@ protected override void OnStartup(StartupEventArgs e)
             {
                 MessageBox.Show("Nitrox launcher should not be executed from a temporary directory. Install Nitrox launcher properly by extracting ALL files and moving these to a dedicated location on your PC.", "Invalid working directory", MessageBoxButton.OK, MessageBoxImage.Error);
                 Environment.Exit(1);
+                Shutdown();
```
Measurity: Environment.Exit will kill the app before this call so `Shutdown()` is unreachable.

Either remove the `Environment.Exit(1)` or remove `Shutdown()`.

NitroxLauncher/App.xaml.cs
```diff
@@ -30,6 +30,31 @@ protected override void OnStartup(StartupEventArgs e)
...
+            }
+
+            if (!RoleDetection.isAppRunningInAdmin()) {
+
+                if (Directory.GetCurrentDirectory().StartsWith(Environment.GetFolderPath(Environment.SpecialFolder.ProgramFiles), StringComparison.OrdinalIgnoreCase))
```
Measurity: This should check for subnautica install path. Not the launcher working directory. If Subnautica is installed in Program Files then the launcher likely needs admin.

Coding-Hen: You should be able to just get the subnautica directory from the launcherlogic instance
```suggestion
                if (LauncherLogic.Instance.SubnauticaPath.StartsWith(Environment.GetFolderPath(Environment.SpecialFolder.ProgramFiles), StringComparison.OrdinalIgnoreCase))
```

dartasen: He won't work, LauncherLogic isn't up yet while entering the application

dartasen:
Well it's working as you requested to but as long as ProgramFiles is is most of the cases the directory used for playing Subnautica, will it not be better to always ask to have admin rights throught the app.manifest instead of often running "hacky code" to restart the launcher ?

NitroxLauncher/Patching/AppHelper.cs
```diff
@@ -0,0 +1,60 @@
...
+    {
+        public static string ProgramFileDirectory = Environment.ExpandEnvironmentVariables("%ProgramW6432%");
+
+        public static bool IsAppRunningInAdmin()
+        {
+            WindowsPrincipal wp = new WindowsPrincipal(WindowsIdentity.GetCurrent());
+            return wp.IsInRole(WindowsBuiltInRole.Administrator);
+        }
+
+        public static void RestartAsAdmin(string path)
```
Measurity: Could you separate the logic from this method out into two?
`RestartAsAdmin` sounds good if it only restarted the app but it does more than that: checks the path of program files. 

You can remove the comment in the `SubnauticaPath` property if you change the code to something like:
```cs
if (!AppHelper.IsRunningAsAdmin() && Path.GetFullPath(path).StartsWith(ProgramFileDirectory, StringComparison.OrdinalIgnoreCase))
{
    AppHelper.RestartAsAdmin();
}
```

NitroxLauncher/LauncherLogic.cs
```diff
@@ -102,6 +100,7 @@ public async Task<string> SetTargetedSubnauticaPath(string path)
                         nitroxEntryPatch.Remove();
                     }
                     nitroxEntryPatch = new NitroxEntryPatch(path);
+                    AppHelper.RestartAsAdmin(path); //If the change path to ProgramFiles, then we ask for admin perm
```
Measurity: Change this code so comment is not needed (see other comment).

PR #963, opened 2020-03-14
No alloc yield

Measurity:
See: https://stackoverflow.com/a/39269012/1277156

> Both yield return 0 and yield return null yields for a single frame. The biggest difference is that yield return 0 allocates memory because of boxing and unboxing of the 0 that happens under the hood, but yield return null does not allocate memory. Because of this, it is highly recommended to use yield return null if you care about performance.


Sunrunner37:
Cool did not know this.

PR #964, opened 2020-03-14
entity debugger and debug arrow

nesrak1:
That was scary but looks like we have debug stuff now so yay

### DIFF
--- a/NitroxClient/Debuggers/EntityDebugger.cs
+++ b/NitroxClient/Debuggers/EntityDebugger.cs
@@ -1,0 +1,41 @@
+﻿using System.Collections.Generic;
+using NitroxClient.MonoBehaviours;
+using NitroxModel.DataStructures;
+using UnityEngine;
+
+namespace NitroxClient.Debuggers
+{
+    public class EntityDebugger : BaseDebugger
+    {
+        public EntityDebugger() : base(200, null, KeyCode.E, true, false, false, GUISkinCreationOptions.DERIVEDCOPY)
+        {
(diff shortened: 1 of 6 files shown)


NitroxClient/Debuggers/EntityDebugger.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        private void RenderEntityDebugger()
+        {
+        }
+
+        public override void OnGUIExtra()
+        {
+            foreach (KeyValuePair<NitroxId, GameObject> gameObjectPairs in NitroxEntity.GetGameObjects())
+            {
+                NitroxId id = gameObjectPairs.Key;
```
Jannify: Do you need to initialize a new variable for a one time use?

NitroxClient/Debuggers/EntityDebugger.cs
```diff
@@ -0,0 +1,41 @@
...
+                GameObject gameObject = gameObjectPairs.Value;
```
Jannify: If it's just a reference I think it's better to use `GameObject gameObject => gameObjectPairs.Value;`

NitroxClient/Debuggers/EntityDebugger.cs
nesrak1: I guess not? Just seemed more consistent to have both vars up front.

Jannify: Then I think it would be the best to use `=>`

Issue #957, opened 2020-03-12
Add "ALPHA" text in main menu and during loading screen


Measurity:
We want to communicate to players that Nitrox is still in alpha and that they should expect bugs.

Add UI text to the existing canvas in game. Can use scene debugger to search for the canvas object in-game (press F7). Should be able to add atext element to it like so: https://docs.unity3d.com/ScriptReference/Canvas.html

Jannify:
See https://github.com/SubnauticaNitrox/Nitrox/pull/969

PR #969, opened 2020-03-16
Added a alpha version text to the loading screen.

Jannify:
unknown


keyboardfox:
I have one super small feedback, perhaps you could try to add some margins to the top, so the text is aligned with the left side? :D

Sunrunner37:
> I have one super small feedback, perhaps you could try to add some margins to the top, so the text is aligned with the left side? :D

A little margin would be good if possible.

Should we make persistent like the version number / alpha used to be across subnautica before release?  We may be able to re-use that component.

Jannify:
I will add a margin but not as big as the loading progress one (one line).

@Sunrunner37 What about a persistent version number but the "warning text" fades away?

Jannify:
grafik
grafik

NitroxClient/MonoBehaviours/Gui/MainMenu/LoadingScreenExtras.cs
```diff
@@ -0,0 +1,46 @@
...
+namespace NitroxClient.MonoBehaviours.Gui.MainMenu
+{
+    public class LoadingScreenExtras
+    {
+        private static string assemblyVersion => Assembly.GetAssembly(typeof(LoadingScreenExtras)).GetName().Version.ToString();
+        private static GameObject loadingTextGameObject => uGUI.main.loading.loadingText.gameObject;
+
+        private static Dictionary<string, uGUI_TextFade> nitroxLoadingExtrasGameObject = new Dictionary<string, uGUI_TextFade>();
+
+        public static void Initialize()
```
Sunrunner37: Do items in this class need to be static?  Can we model this as a monobehaviour?

NitroxClient/MonoBehaviours/Gui/MainMenu/LoadingScreenExtras.cs
```diff
@@ -0,0 +1,46 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using UnityEngine;
+
+namespace NitroxClient.MonoBehaviours.Gui.MainMenu
+{
+    public class LoadingScreenExtras
```
Sunrunner37: Maybe something more explicit like `LoadingScreenVersionText`  I'm not sure that the generic map will be used for other things.

Coding-Hen:
Resolves #957

NitroxClient/MonoBehaviours/Gui/MainMenu/LoadingScreenExtras.cs
Jannify: Because monobehaviour is not needed I think it would be better for performance to not make this a monobehaviour .

Jannify:
> Just a few comments. Also, it looks like the code may not match the in-game screenshot. I don't see anything in the PR that makes the text persist into the game. Maybe I am missing something.

The warning-text fades out after the loading screen, the version number stays. That's the trick.

Issue #823, opened 2019-07-18
Keeps opening in VR


FTWGaming0:
The way I made subnautica open in the epic games launcher was to add "-vrmode none" to the additional command line arguments, but the launcher doesn't appear to give me that option... can anyone show me a way of doing this with it?

Praevalens:
Bit late, but I've seen people fixing this issue by adding the game to steam as a non-steam game with the "-vrmode none" arguments which worked for them.

HarveyBB:
I want to know how to play **with** VR tbh

BlockArchitech:
Simply, Unplug your VR headset, or turn off your switch box. Worked for me!

HMD: Oculus Rift S

FTWGaming0:
I don't want to have to crawl behind my computer and unplug my headset every time I want to play this game tho,

Coding-Hen:
@FTWGaming0 @Praevalens Is this only a problem with the epic? Or is it steam and epic?

Measurity:
Closing because of #972

PR #972, opened 2020-03-17
Vrmode none

dartasen:
https://www.reddit.com/r/subnautica/comments/a67m2c/subnautica_from_epic_games_store_force_launching/

### DIFF
--- a/NitroxLauncher/LauncherLogic.cs
+++ b/NitroxLauncher/LauncherLogic.cs
@@ -258,7 +258,7 @@ namespace NitroxLauncher
 
             if (PlatformDetection.IsEpic(subnauticaPath))
             {
-                startInfo.Arguments = "-EpicPortal";
+                startInfo.Arguments = "-EpicPortal -vrmode none";
             }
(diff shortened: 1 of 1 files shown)


NitroxLauncher/LauncherLogic.cs
```diff
@@ -258,7 +258,7 @@ private Process StartSubnautica()
 
             if (PlatformDetection.IsEpic(subnauticaPath))
             {
-                startInfo.Arguments = "-EpicPortal";
+                startInfo.Arguments = "-EpicPortal -vrmode none";
             }
```
Sunrunner37: Do we need to have this argument for all platforms?  Should we make it a setting in the launcher?

Coding-Hen:
Fixes #823

NitroxLauncher/LauncherLogic.cs
killzoms: Its not an issue with Steam, could be an issue with other platforms however but don't quote me

PR #973, opened 2020-03-17
Improved entity debugger

nesrak1:
Lots of text and lots of boxes

### DIFF
--- a/NitroxClient/Debuggers/EntityDebugger.cs
+++ b/NitroxClient/Debuggers/EntityDebugger.cs
@@ -7,9 +7,17 @@ namespace NitroxClient.Debuggers
 {
     public class EntityDebugger : BaseDebugger
     {
+        private static Color labelBgColor = new Color(0.2f, 0.2f, 0.2f, 0.5f);
+        private static Color labelFgColor = Color.white;
+        private static float textXOffset = 20f;
+        private static Texture2D lineTex;
+
+        private static List<Rect> rects;
+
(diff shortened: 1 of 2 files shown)


NitroxClient/Debuggers/EntityDebugger.cs
```diff
@@ -18,24 +26,93 @@ private void RenderEntityDebugger()
...
-                        GUIStyle style = GUI.skin.label;
-                        GUIContent textContent = new GUIContent("ID " + id.ToString());
-                        Vector2 size = style.CalcSize(textContent);
-                        GUI.Label(new Rect(screenPos.x, Screen.height - screenPos.y, size.x, size.y), textContent);
+                        bool finished = true;
+                        foreach (Rect rect in rects)
+                        {
+                            if (rect.Overlaps(drawSize))
+                            {
+                                drawSize.x = rect.x;
```
nesrak1: with this line: https://cdn.discordapp.com/attachments/528915769107415040/689549996638339197/unknown.png
without this line: https://cdn.discordapp.com/attachments/528915769107415040/689550007598448643/unknown.png

not sure whether to keep or remove, tell me your thoughts

Jannify: Think the first will give a bit more "organized mess" of NitroxIDs :D

dartasen: I'm for the first too

PR #982, opened 2020-03-20
Fixes issue where chat opens when another input field has focus

Measurity:
image

This fixes the issue where Y was opening chat when another input field in game already had focus.


NitroxClient/Communication/Packets/Processors/ChatMessageProcessor.cs
```diff
@@ -32,20 +32,19 @@ public override void Process(ChatMessage message)
...
             Optional<RemotePlayer> remotePlayer = remotePlayerManager.Find(message.PlayerId);
-
-            if (remotePlayer.IsPresent())
+            if (!remotePlayer.IsPresent())
             {
-                RemotePlayer remotePlayerInstance = remotePlayer.Get();
-                ChatLogEntry chatLogEntry = new ChatLogEntry(remotePlayerInstance.PlayerName, message.Text, remotePlayerInstance.PlayerSettings.PlayerColor);
-                playerChat.WriteChatLogEntry(chatLogEntry);
-                playerChat.ShowLog();
+                return;
```
Sunrunner37: Is this a happy case when a disconnected play sends a message and they arrive out of order?  Should we log a warn?

Measurity: Having more info when an error occurs sounds good to me.

PR #986, opened 2020-03-20
Fixed player stats being 0 on initial connection

Coding-Hen:


NitroxServer/GameLogic/PlayerManager.cs
```diff
@@ -116,7 +116,7 @@ public Player PlayerConnected(NitroxConnection connection, string reservationKey
                 {
                     if (!allPlayersByName.TryGetValue(playerContext.PlayerName, out player))
                     {
-                        player = new Player(playerContext.PlayerId, playerContext.PlayerName, playerContext, connection, NitroxVector3.Zero, new NitroxId(), Optional<NitroxId>.Empty(), Perms.PLAYER, new PlayerStatsData(), new List<EquippedItemData>(), new List<EquippedItemData>());
```
Sunrunner37: Do we want this object to be null?  Should we default to good values instead?

Coding-Hen:
Closing in favour of #990

PR #989, opened 2020-03-22
Base fixes

killzoms:
Bases Stay deleted now :D The parenting issue still exists however

### DIFF
--- a/NitroxModel/DataStructures/Optional.cs
+++ b/NitroxModel/DataStructures/Optional.cs
@@ -73,7 +73,8 @@ namespace NitroxModel.DataStructures.Util
 
         public override string ToString()
         {
-            return "Optional Contains: " + Value.ToString();
+            string str = Value != null ? Value.ToString() : "Nothing";
+            return $"Optional Contains: {str}";
         }
 
         private Optional(SerializationInfo info, StreamingContext context)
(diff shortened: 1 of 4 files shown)


NitroxModel/DataStructures/Optional.cs
```diff
@@ -71,6 +71,11 @@ public static Optional<T> OfNullable(T value)
             return new Optional<T>(value);
         }
 
+        public override string ToString()
+        {
+            return "Optional Contains: " + Value.ToString();
```
Measurity: .ToString() is not needed and should check for null

NitroxPatcher/Patches/Dynamic/Base_CopyFrom_Patch.cs
```diff
@@ -0,0 +1,32 @@
...
+{
+    class Base_CopyFrom_Patch : NitroxPatch, IDynamicPatch
+    {
+        public readonly MethodInfo METHOD = typeof(Base).GetMethod(nameof(Base.CopyFrom), BindingFlags.Public | BindingFlags.Instance);
+
+        public static void Prefix(Base __instance, Base sourceBase)
+        {
+            NitroxEntity entity = sourceBase.GetComponent<NitroxEntity>();
+
+            if (sourceBase.GetComponent<BaseRoot>() == null)
```
Measurity: Don't use `== null` checks on Unity objects because they can be dead but not null.
```suggestion
            if (!sourceBase.GetComponent<BaseRoot>())
```

PR #995, opened 2020-03-24
Compile entire project & run server on Linux

MarijnS95:
These patches allow compiling the entire project (currently minus `NitroxTest` and `NitroxLauncher`) under Linux using `xbuild`/`msbuild`. The server starts just fine but I am unable time-wise to test any client with it.

### TODO:
- Figure out an msbuild compiler barrier for `NitroxLauncher`
- Convert `NitroxTest` to platform/VS-agnostic test framework
- Fix `Resources/` capitalization in AssetTools, where I've been told this should be addressed.

Note that there'll be loads more fun when Subnautica Experimental drops to stable :wink:. Dependency capitalization changed once again, aside a hoist of other changes.

### How-to:
This assumes have all the necessary packages installed (`mono`, `xbuild`/`msbuild`, `nuget`) for your favorite distribution. In the cloned folder, run:
```sh
nuget restore
msbuild # or the deprecated xbuild
# Start the server
mono ./NitroxServer-Subnautica/bin/Debug/NitroxServer-Subnautica.exe
# Or start the server using wine:
./NitroxServer-Subnautica/bin/Debug/NitroxServer-Subnautica.exe
```


Jannify:
If it should not be merged why not making it a draft?

MarijnS95:
@Jannify Because you cannot do so after clicking the button slightly too quickly?

Ghaarg:
I am also migrating to linux and had some thoughts for linux migration some time ago.
An idea i had to make the launcher also usable on linux/mac was to migrate most of it to something like [ElektronNet](https://github.com/ElectronNET/Electron.NET) or as a lightweight version [webview-cs](https://github.com/zenakuten/webview-cs) (preferred on my side cause i hate npm).
This approach  would be more work but may be worth it and better than writing another app that just patches, starts and unpatches afterwards.
But ofc, there may be other cross platform front-end solutions for C# i do not know and may be better suited.
At the moment i don't think we come around mono and that is a problem with the launcher solutions above. Harmony and some reflections do not seem to be available for the dotnet framework (the cross platform one), which both front-ends use.

At the moment i don't know any elegant way to solve the above problems. Maybe a mono app that just patches, runs the client and unpatches is the best way until we can use Harmony with dotnet.

nesrak1:
This fix is really only to compile and run the server. The launcher is being rewritten in electron at the moment (looks like it will be written in js and call .net to patch and start the game.) Also, unity uses mono for all platforms afaik so there shouldn't be any problems with Harmony.

MarijnS95:
Rebased this PR on top of master to solve conflicts, fix more capitalization issues :see_no_evil: and implement a compile-time barrier for NitroxTest and NitroxLauncher. The barrier is simple but cannot be applied to a `<Project>` as a whole, unfortunately.

MarijnS95:
Build-tested on Windows, still works fine. Let the review process commence :grin:

Sunrunner37:
This looks fine to me although I don't know anything about linux compilation.  If you can update then we can merge.

MarijnS95:
@Sunrunner37 Unfortunately this needs validation on Windows again. Multiple issues and capitalization needed to be fixed even if it was just a few patches behind. I expect Linux building is going to be broken 90% of the time if work persists like this.

Referring back to an earlier rant about transitive dependencies not properly including `LitJSON.dll` in the output dir, the [commit that "fixes" this](https://github.com/SubnauticaNitrox/Nitrox/pull/995/commits/b26e5ff0d29488a940d1d9caa9dc5d368e09ea64) is no longer needed and the server can now properly start again without having `LitJSON.dll` in the same folder as `NitroxModel-Subnautica.dll`. Looks like that got fixed by installing a Subnautica update which includes both files in `Subnautica_Data/Managed`. Should we rely on their version though?

MarijnS95:
> I expect Linux building is going to be broken 90% of the time if work persists like this.

The simple solution is obviously to run the tests on Linux as well using GH Actions. Smooth sailing so far (not), the only thing left is making the depot downloader cross-compatible and then everything should work. Let's do that in a different PR though.

PR #996, opened 2020-03-25
Fixed: No air if spawned inside base

TheClockTwister:
Players that have been spawned inside a base were swimming and lacking oxygen. Upon exiting the base to get oxygen, the "entering animation" was played and they actually entered the base, instead of exiting. This was patched. Now, if you spawn inside a base, you have oxygen and on exit via a hatch, you actually exit the base.

### DIFF
--- a/NitroxClient/GameLogic/InitialSync/PlayerPositionInitialSyncProcessor.cs
+++ b/NitroxClient/GameLogic/InitialSync/PlayerPositionInitialSyncProcessor.cs
@@ -43,7 +43,7 @@ namespace NitroxClient.GameLogic.InitialSync
                         // If player is not swimming
                         Player.main.SetCurrentSub(root);
 
-                        if (!root.isBase) // Additionally, if player is not in base (has to ba a vehicle)
+                        if (!root.isBase) // Additionally, if player is not in base (has to be a vehicle)
                         {
                             Quaternion vehicleAngle = root.transform.rotation;
                             position = vehicleAngle * position;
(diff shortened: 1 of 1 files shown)


NitroxClient/GameLogic/InitialSync/PlayerPositionInitialSyncProcessor.cs
```diff
@@ -38,13 +38,18 @@ public override IEnumerator Process(InitialPlayerSync packet, WaitScreen.ManualW
                 {
                     SubRoot root = sub.Get().GetComponent<SubRoot>();
                     // Player position is relative to a subroot if in a subroot
-                    if (root != null && !root.isBase)
-                    {                        
+                    if (root != null)
+                    {
+                        // If player is not swimming
                         Player.main.SetCurrentSub(root);
```
TheClockTwister: This line had to be executen whenever a player is nit in the water (so whenever root != null). But it was in the combined if clause so it was only executen when the player was in a vehicle, so I separated the if statements

NitroxClient/GameLogic/InitialSync/PlayerPositionInitialSyncProcessor.cs
```diff
@@ -38,13 +38,18 @@ public override IEnumerator Process(InitialPlayerSync packet, WaitScreen.ManualW
...
-                        Quaternion vehicleAngle = root.transform.rotation;
-                        position = vehicleAngle * position;
-                        position = position + root.transform.position;
-                        Player.main.SetPosition(position);
+
+                        if (!root.isBase) // Additionally, if player is not in base (has to ba a vehicle)
```
Coding-Hen: If being really picky
```suggestion
                        if (!root.isBase) // Additionally, if player is not in base (has to be a vehicle)
```

PR #1004, opened 2020-03-29
NitroxModel: Forward unity logs to "Nitrox Logs/"

MarijnS95:
Important information like uncaught exceptions that bubble all the way
to the engine do not end up in our logs. They can however be caught from
Unity and forwarded to our log wrapper for convenience.

While an alternative might be to omit Nitrox's logfile alltogether and
rely on Player.log this has the additional benefit of proper formatting;
the log tag includes the time and source (Server/Client/Unity) making
it much easier on the eyes.

--- 

As discussed on Discord this might be handy to have. I was astounded to not see any exceptions in the log while they were definitely occuring. Am I using the wrong logging facility (with `Player.log` and the console of an attached debugger as alternatives) and are these files simply unused?

TODO:
- Should we remove `LogType.Log`? The game is rather spammy here
- The `ConsoleAppender` indirectly forwards the logs to `Player.log`, making them appear twice. Any ideas to solve that, perhaps a filter on `LoggerName` or changing the layout of the loggers?

### DIFF
--- a/NitroxModel/Logger/Log.cs
+++ b/NitroxModel/Logger/Log.cs
@@ -8,6 +8,7 @@ using log4net.Filter;
 using log4net.Layout;
 using log4net.Repository.Hierarchy;
 using NitroxModel.Discovery;
+using UnityEngine;
 
 namespace NitroxModel.Logger
 {
@@ -16,6 +17,7 @@ namespace NitroxModel.Logger
         private static bool inGameMessages;
 
         private static readonly ILog log = LogManager.GetLogger(GetLoggerName());
+        private static readonly ILog unityLog = LogManager.GetLogger("Unity");
         public static InGameLogger InGameLogger { get; set; }
 
         static Log()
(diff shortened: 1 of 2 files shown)


Measurity:
Lets remove the consoleappender if the log is running from game (so that it doesn't get added to player.log).

NitroxModel/Logger/Log.cs
```diff
@@ -142,6 +144,35 @@ private static void Setup()
...
+                        break;
+                    case LogType.Assert:
+                        unityLog.Error($"Assertion failed: {condition}");
+                        break;
+                    case LogType.Warning:
+                        unityLog.Warn(condition);
+                        break;
+                    case LogType.Log:
+                        unityLog.Debug(condition);
+                        break;
```
Measurity: Agreed on LogType.Log. Too much information that's better handled by our logging methods on the spots we need. Best remove it.

MarijnS95:
> Lets remove the consoleappender if the log is running from game (so that it doesn't get added to player.log).

I have been told that some users attach a debugger (indirectly through the launcher), which would break that functionality.

As discussed before, let's wait a bit for https://github.com/SubnauticaNitrox/Nitrox/pull/926 before deciding where to take this.

MarijnS95:
@dartasen Was this addressed in a different PR? Would you mind linking it for posterity, instead of closing without reason?

dartasen:
@MarijnS95 It was not, I thought it was abandoned, sorry

MarijnS95:
@dartasen Well, I guess it's my responsibilty to update this then, as promised.

PR #1008, opened 2020-03-29
Implement kick reason

dartasen:


NitroxClient/MonoBehaviours/Gui/InGame/PlayerKickedModal.cs
```diff
@@ -17,13 +17,13 @@ public class PlayerKickedModal : MonoBehaviour
         private static GameObject playerKickedSubWindow;
         public static PlayerKickedModal Instance { get; private set; }
 
-        public void Show()
+        public void Show(ref string reason)
```
Measurity: Prefer not to use `ref` keyword when possible.

NitroxClient/MonoBehaviours/Gui/InGame/PlayerKickedModal.cs
```diff
@@ -17,13 +17,13 @@ public class PlayerKickedModal : MonoBehaviour
...
         {
             FreezeTime.Begin("NitroxDisconnected");
-            StartCoroutine(Show_Impl());
+            StartCoroutine(Show_Impl(reason));
         }
 
-        private static void InitSubWindow()
+        private static void InitSubWindow(ref string reason)
```
Measurity: Same here, do not use `ref` keyword when possible.

NitroxClient/MonoBehaviours/Gui/InGame/PlayerKickedModal.cs
```diff
@@ -45,7 +45,7 @@ private static void InitSubWindow()
                 GameObject header = playerKickedSubWindow.FindChild("Header"); //Message Object
 
                 Text messageText = header.GetComponent<Text>();
-                messageText.text = "You have been kicked from the server";
+                messageText.text = reason;
```
Measurity: Might be good idea to set a default text if packet message was null or empty.
```suggestion
                messageText.text = string.IsNullOrWhiteSpace(reason) ? "You've been kicked from the server" : reason;
```

PR #1011, opened 2020-04-01
Fixed base construction ReadyForReview

TheGuy920:


NitroxServer/GameLogic/Bases/BaseManager.cs
```diff
@@ -78,7 +126,7 @@ public void BasePieceConstructionCompleted(NitroxId id, NitroxId baseId)
                         // a fully constructed piece.  Therefor, we always update this attribute to make sure it
                         // is the latest.
                         basePiece.BaseId = baseId;
-                        basePiece.ParentId = Optional.OfNullable(baseId);
+                        basePiece.ParentId = Optional<NitroxId>.OfNullable(baseId);
```
dartasen, line 129: Optional<Type> is no longer required with the new implementation of Optional class

NitroxServer/GameLogic/Bases/BaseManager.cs
```diff
@@ -68,8 +108,16 @@ public void BasePieceConstructionCompleted(NitroxId id, NitroxId baseId)
...
-                    basePiece.ConstructionCompleted = true;
+                    if (basePiece.ConstructionAmount > 0.4f)
+                    {
+                        basePiece.ConstructionAmount = 1.0f;
+                        basePiece.ConstructionCompleted = true;
+                    }
+                    else
+                    {
+                        basePiece.ConstructionAmount = 1.0f;
+                        basePiece.ConstructionCompleted = true;
```
dartasen, line 119: Let's not split up the code into an if/else with same content in it, unfinish maybe ?

NitroxServer/GameLogic/Bases/BaseManager.cs
```diff
@@ -5,8 +5,10 @@
 using System.Linq;
 using NitroxModel.DataStructures.Util;
 
+
 namespace NitroxServer.GameLogic.Bases
 {
+    [System.Runtime.InteropServices.ComVisible(true)]
```
dartasen, line 11: Let's remove this, we have better keyword for managing visibility inside a class. Besides the default value of this attribute is true (cf [ComVisibleAttribute MicrosoftDocs](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.comvisibleattribute))

NitroxServer/GameLogic/Bases/BaseManager.cs
```diff
@@ -131,7 +179,7 @@ public void UpdateBasePieceMetadata(NitroxId id, BasePieceMetadata metadata)
 
                 if (basePiece != null)
                 {
-                    basePiece.Metadata = Optional.OfNullable(metadata);
+                    basePiece.Metadata = Optional<BasePieceMetadata>.OfNullable(metadata);
```
dartasen, line 182: Same as previous comments

NitroxServer/GameLogic/Bases/BaseManager.cs
```diff
@@ -33,13 +35,44 @@ public List<BasePiece> GetPartiallyConstructedPieces()
...
+                partiallyConstructedPiecesById.Add(basePiece.Id, basePiece);
+                basePiece.ConstructionAmount = 1.0f;
+                basePiece.ConstructionCompleted = true;
+            }
+            if (basePiece.ConstructionAmount > 0.4f)
+            {
+                basePiece.ConstructionAmount = 1.0f;
+                basePiece.ConstructionCompleted = true;
+            }
+
```
dartasen, line 75: Maybe add some comments to the code in order to allows us to understand what the idea you had behind that code

NitroxPatcher/Patches/Dynamic/BaseGhost_Place_Patch.cs
```diff
@@ -0,0 +1,41 @@
+﻿using System;
+using System.Reflection;
+using Harmony;
+using NitroxClient.GameLogic.Helper;
+using NitroxModel.Logger;
+using static NitroxClient.GameLogic.Helper.TransientLocalObjectManager;
+
+namespace NitroxPatcher.Patches.Dynamic
+{
+    public class BaseGhost_Place_Patch : NitroxPatch, IDynamicPatch
```
Sunrunner37, line 10: It looks like you are trying to merge deleted files.  Can you verify that you were originally testing on the tip of master?

NitroxServer/GameLogic/Bases/BaseManager.cs
```diff
@@ -33,29 +35,83 @@ public List<BasePiece> GetPartiallyConstructedPieces()
...
-
         public void AddBasePiece(BasePiece basePiece)
         {
-            lock (partiallyConstructedPiecesById)
+            //anything that is not a window, hatch, or reinforcement get run normally
+            if (basePiece.TechType.Name != "BaseReinforcement" | basePiece.TechType.Name != "BaseHatch" | basePiece.TechType.Name != "BaseWindow")
+            {
+                //just in case some how a base piece is spawned at max completion
+                //the base piece will get added to at least one directory
+                if(basePiece.ConstructionAmount < 1.0f)
```
Sunrunner37, line 45: Can we validate that this is the behavior in creative?  (along with the types of game objects that have this behavior)  We would also need to validate that these items also don't send construction complete events.  However, it could be a valid edge case given these two things check out.

NitroxServer/GameLogic/Bases/BaseManager.cs
```diff
@@ -33,29 +35,83 @@ public List<BasePiece> GetPartiallyConstructedPieces()
...
+                }
+                else
+                {
+                    lock (partiallyConstructedPiecesById)
+                    {
+                        completedBasePieceHistory.Add(basePiece);
+                    }
+                }
+            }
+            //when adding base peices this tries to eliminate partially constructed parts that were once fully constructed
```
Sunrunner37, line 60: how do you know that they were once fully constructed?

NitroxServer/GameLogic/Bases/BaseManager.cs
```diff
@@ -33,29 +35,83 @@ public List<BasePiece> GetPartiallyConstructedPieces()
...
+                basePiece.ConstructionCompleted = true;
+            }
+            if (basePiece.TechType.Name == "BaseHatch" && completedBasePieceHistory.Contains(basePiece))
+            {
+                basePiece.ConstructionCompleted = false;
+                partiallyConstructedPiecesById.Add(basePiece.Id, basePiece);
+                basePiece.ConstructionAmount = 1.0f;
+                basePiece.ConstructionCompleted = true;
+            }
+            if (basePiece.TechType.Name == "BaseReinforcement" && completedBasePieceHistory.Contains(basePiece))
```
Sunrunner37, line 82: I don't think this is the right solution.  Generally we don't hard code tech types outside of very specific cases.  It is also not clear why we want to mark their construction completed immediately upon placing.

NitroxServer/GameLogic/Bases/BaseManager.cs
```diff
@@ -33,29 +35,83 @@ public List<BasePiece> GetPartiallyConstructedPieces()
...
+                if(basePiece.ConstructionAmount < 1.0f)
+                {
+                    lock (partiallyConstructedPiecesById)
+                    {
+                        partiallyConstructedPiecesById.Add(basePiece.Id, basePiece);
+                    }
+                }
+                else
+                {
+                    lock (partiallyConstructedPiecesById)
```
Sunrunner37, line 54: This is the wrong lock

NitroxServer/GameLogic/Bases/BaseManager.cs
```diff
@@ -33,29 +35,83 @@ public List<BasePiece> GetPartiallyConstructedPieces()
...
+            {
+                basePiece.ConstructionCompleted = false;
+                partiallyConstructedPiecesById.Add(basePiece.Id, basePiece);
+                basePiece.ConstructionAmount = 1.0f;
+                basePiece.ConstructionCompleted = true;
+            }
+            if (basePiece.TechType.Name == "BaseReinforcement" && completedBasePieceHistory.Contains(basePiece))
             {
+                basePiece.ConstructionCompleted = false;
                 partiallyConstructedPiecesById.Add(basePiece.Id, basePiece);
```
Sunrunner37, line 85: We shouldn't interact with these datastructures outside the context of a lock.

NitroxServer/GameLogic/Bases/BaseManager.cs
```diff
@@ -33,29 +35,83 @@ public List<BasePiece> GetPartiallyConstructedPieces()
...
                     basePiece.ConstructionAmount = constructionAmount;
-
+                    /*
+                     // this if defaults to true, so if construction is completed, it gets set to flase so its not completed
+                     // I could be wrong but this might cause issues
                     if (basePiece.ConstructionCompleted)
                     {
                         basePiece.ConstructionCompleted = false;
                     }
+                    */
```
Sunrunner37, line 107: Excluding out-of-order packet edge cases, I think if you are receiving amount changed packets it is either constructing or deconstructing.

NitroxServer/GameLogic/Bases/BaseManager.cs
```diff
@@ -33,29 +35,83 @@ public List<BasePiece> GetPartiallyConstructedPieces()
...
+                }
+                //This ensures that any pecice over 95% is completed and added to the completed base piece history
+                if (basePiece.ConstructionAmount >= 0.95f && completedBasePieceHistory.Contains(basePiece) == false)
+                {
+                    basePiece.ConstructionAmount = 1.0f;
+                    basePiece.ConstructionCompleted = true;
+                    completedBasePieceHistory.Add(basePiece);
```
Sunrunner37, line 114: We should rely on the completed packet to tell the server it is done.

NitroxServer/GameLogic/Bases/BaseManager.cs
```diff
@@ -83,10 +138,13 @@ public void BasePieceConstructionCompleted(NitroxId id, NitroxId baseId)
 
                     partiallyConstructedPiecesById.Remove(id);
 
+                    //this task is now completed by BasePieceConstructionAmountChanged()
+                    /*
                     lock (completedBasePieceHistory)
                     {
                         completedBasePieceHistory.Add(basePiece);
                     }
+                    */
```
Sunrunner37, line 147: We don't recommend leaving dead code in PRs.  We expect PRs to be merge worthy unless marked as "DO NOT MERGE"

NitroxServer/GameLogic/Bases/BaseManager.cs
```diff
@@ -98,19 +156,31 @@ public void BasePieceDeconstructionBegin(NitroxId id)
             lock (completedBasePieceHistory)
             {
                 basePiece = completedBasePieceHistory.Find(piece => piece.Id == id);
-
+                if (basePiece.ConstructionAmount < 1f && partiallyConstructedPiecesById.ContainsKey(basePiece.Id))
```
Sunrunner37, line 159: This statement looks like it would break deconstruction.  It will only remove the base piece from completed list (and set the construction amount/flag) if the construction amount was less than the completed value.

NitroxServer/GameLogic/Bases/BaseManager.cs
Sunrunner37: Why is this comment marked resolved?

NitroxServer/GameLogic/Bases/BaseManager.cs
```diff
@@ -33,29 +35,83 @@ public List<BasePiece> GetPartiallyConstructedPieces()
...
+                basePiece.ConstructionCompleted = true;
+            }
+            //I am not sure if this helps at all, but it is supposed to ensure that the windows, hatches, and reinforcements don't go missing
+            //and that they they are built to 100%
+            if (basePiece.TechType.Name == "BaseWindow" && completedBasePieceHistory.Contains(basePiece))
+            {
+                basePiece.ConstructionCompleted = false;
+                partiallyConstructedPiecesById.Add(basePiece.Id, basePiece);
+                basePiece.ConstructionAmount = 1.0f;
+                basePiece.ConstructionCompleted = true;
```
Sunrunner37, line 73: I'm not sure what the purpose was to see it to false when subsequently overriding it to true.

NitroxServer/GameLogic/Bases/BaseManager.cs
TheGuy920: I don't, but there is an issue in which base pieces that I personally constructed to 100% get reloaded at 89% or 95% or something else. Rather than having to deal with that I believe it would be more beneficial to just any base piece above a certain threshold, ON RELOAD. I personally do not construct a base piece and leave it at 50% or 80% and reload the world, I always finish my build beforehand and I think a lot of people can agree with me.

NitroxServer/GameLogic/Bases/BaseManager.cs
TheGuy920: oops

NitroxServer/GameLogic/Bases/BaseManager.cs
TheGuy920: I will mark it as do not merge now, once I figure out how

NitroxServer/GameLogic/Bases/BaseManager.cs
TheGuy920: thankyou

NitroxServer/GameLogic/Bases/BaseManager.cs
TheGuy920: me neither...

NitroxServer/GameLogic/Bases/BaseManager.cs
TheGuy920: I meant to set that to false

NitroxServer/GameLogic/Bases/BaseManager.cs
TheGuy920: Does it need to be removed or replaced, I did not add this, this was there when I got my hands on this code

NitroxPatcher/Patches/Dynamic/BaseGhost_Place_Patch.cs
TheGuy920: Does it need to be removed or replaced or slightly changed, I did not add this, this was there when I got my hands on this code

NitroxServer/GameLogic/Bases/BaseManager.cs
TheGuy920: Yes when build in creative and speed is above 1 or the fastbuild cheat is active it goes from 0% to 100%

NitroxServer/GameLogic/Bases/BaseManager.cs
TheGuy920: noted

NitroxPatcher/Patches/Dynamic/BaseGhost_Place_Patch.cs
TheGuy920: i didn't even edit this file

Sunrunner37:
We appreciate your desire to contribute but I don't think this PR is heading in the right direction.  We shouldn't be hard coding that base pieces are complete when they may be not.  Additionally, this area of the code should not have special exceptional cases for the various base pieces (hard coded piece names).  I think that you'll find that the experience on master is maturing pretty well (outside of a few face-based parts, such as windows).  Given this, I am going to close this PR.

PR #1013, opened 2020-04-02
Move and Refactor ServerConfig

dartasen:
First step in order to implement custom ServerConfig inside the launcher


NitroxModel/Helper/StringHelper.cs
```diff
@@ -0,0 +1,23 @@
...
+using System.Text;
+
+namespace NitroxModel.Helper
+{
+    public static class StringHelper
+    {
+        public static string GenerateRandomString(int size)
+        {
+            StringBuilder builder = new StringBuilder();
+            Random random = new Random();
```
Measurity, line 11: It's best to create a `Random` instance once and reuse it so that when this method is called in a loop it will generate new numbers and not repeat itself. See: https://stackoverflow.com/a/3854950/1277156

NitroxServer/Server.cs
```diff
@@ -12,8 +14,11 @@ public class Server
         private Communication.NetworkingLayer.NitroxServer server;
         private readonly World world;
         private readonly WorldPersistence worldPersistence;
+        private readonly ServerConfig serverConfig;
+
         public bool IsRunning { get; private set; }
-        private bool isSaving;
+        public bool isSaving { get; private set; }
```
Measurity, line 20: Use PascalCasing for `isSaving` because it's now public.

NitroxServer/Server.cs
```diff
@@ -22,16 +27,24 @@ public Server(WorldPersistence worldPersistence, World world, ServerConfig serve
...
+            this.world = world;
+            Instance = this;
+
             // TODO: Save once after last player leaves then stop saving.
             saveTimer = new Timer();
             saveTimer.Interval = serverConfig.SaveInterval;
             saveTimer.AutoReset = true;
-            saveTimer.Elapsed += delegate { Save(); };
+            saveTimer.Elapsed += delegate
+            { Save(); };
```
Measurity, line 47: Leave this on the same line

NitroxModel/Server/ServerConfigItem.cs
```diff
@@ -0,0 +1,81 @@
...
+            Validate.NotNull(itemName);
+
+            Name = itemName;
+            value = defaultValue; //not Value, we don't want to rewrite the config to default value again and again
+        
+            try
+            {
+                TypeConverter converter = TypeDescriptor.GetConverter(typeof(T));
+
+                if (converter == null)
```
Measurity, line 52: Not your code but this whole if can be removed since converter can't be null

PR #1019, opened 2020-04-05
Fixed and added additional badge

Coding-Hen:
Fixed the current build badge and added an aditional one for the licence

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,7 @@
 # Subnautica Nitrox
 An open-source, multiplayer modification for the game <a href="https://unknownworlds.com/subnautica/">Subnautica</a>.
 
+[![Release](https://github.com/SubnauticaNitrox/Nitrox/workflows/Building%20Nitrox%20Release/badge.svg)](https://github.com/SubnauticaNitrox/Nitrox/actions?query=workflow%3A%22Building+Nitrox+Release%22)
 [![Build](https://github.com/SubnauticaNitrox/Nitrox/workflows/Building%20Nitrox%20Pull%20Request/badge.svg)](https://github.com/SubnauticaNitrox/Nitrox/actions?query=workflow%3A%22Building+Nitrox+Pull+Request%22)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -5,7 +5,8 @@
 # Subnautica Nitrox
 An open-source, multiplayer modification for the game <a href="https://unknownworlds.com/subnautica/">Subnautica</a>.
 
-![](https://github.com/SubnauticaNitrox/Nitrox/workflows/Building%20Nitrox%20Release/badge.svg)
+[![Build](https://github.com/SubnauticaNitrox/Nitrox/workflows/Building%20Nitrox%20Pull%20Request/badge.svg)](https://github.com/SubnauticaNitrox/Nitrox/actions?query=workflow%3A%22Building+Nitrox+Pull+Request%22)
```
Jannify, line 8: The problem is that this link will show the result of the newest pull request and that is not what we want (I think). The current "image not found"-image is just because there hasn't been a release yet.

Coding-Hen: You may be right but as far as I am aware we should still have the image available from last time it ran should it not? Do we even have a action for the release?

Jannify: Yeah this workflow is only for releases: https://github.com/SubnauticaNitrox/Nitrox/blob/master/.github/workflows/buildingRelease.yml
If the action has never run (and that's the case because different workflow) there will be no badge.

Jannify:
I think the addition to the release and the license badge is nice, would like a pull request with it...

Jannify:
Sorry for being so picky (and probably misleading English) but (in my opinion) a badge for pull request (which is still in this pull request) is misleading if it's on the front page of the mod. 
Maybe you can extend the triggers of the "Building Nitrox Release"-workflow with `push` (=> `on: [release, push]`) and have two badges for release and master. As far as I remember you can filter the trigger event of an action in the badge link (see [badges](https://help.github.com/en/actions/configuring-and-managing-workflows/configuring-a-workflow#adding-a-workflow-status-badge-to-your-repository) for an example).

PR #1023, opened 2020-04-05
Fix server persist

Measurity:
- Added thread safe collections that are serializable to assist server persist
 - Added debug logs for when items are added to containers and change packet is send to server
 - Added utility for getting scene path of GameObject
 - Fixed unit tests except the connection unit tests (these take more time to properly understand the featureset and write unit test for, again)
 - Added 'summary' command to server commands to display general information about the save. Not all is printed yet but it helps for debugging.
 - Changed the IP printing stuff to be release only to reduce unnecessary network requests in development.
- Fixed case where server couldn't start listening on socket but was ignored since LiteNetLib eats the exception and just prints it.

### DIFF
--- a/NitroxClient/GameLogic/ItemContainers.cs
+++ b/NitroxClient/GameLogic/ItemContainers.cs
@@ -73,7 +73,7 @@ namespace NitroxClient.GameLogic
             Optional<ItemsContainer> opContainer = InventoryContainerHelper.GetBasedOnOwnersType(owner);
             if (!opContainer.HasValue)
             {
-                Log.Error($"Could not find item container behaviour on object '{owner.name}' with Nitrox id '{ownerId}'");
+                Log.Warn($"Could not find item container behaviour on object '{owner.name}' with Nitrox id '{ownerId}'");
                 return;
             }
 
--- a/NitroxClient/GameLogic/ItemContainers.cs
+++ b/NitroxClient/GameLogic/ItemContainers.cs
@@ -16,62 +16,32 @@ namespace NitroxClient.GameLogic
     public class ItemContainers
     {
         private readonly IPacketSender packetSender;
-        private readonly LocalPlayer localPlayer;
 
-        public ItemContainers(IPacketSender packetSender, LocalPlayer localPlayer)
+        public ItemContainers(IPacketSender packetSender)
         {
             this.packetSender = packetSender;
-            this.localPlayer = localPlayer;
         }
 
-        public void BroadcastItemAdd(Pickupable pickupable, Transform ownerTransform)
+        public void BroadcastItemAdd(Pickupable pickupable, Transform containerTransform)
         {
-            NitroxId ownerId = null;
-            bool isCyclopsLocker = Regex.IsMatch(ownerTransform.gameObject.name, @"Locker0([0-9])StorageRoot$", RegexOptions.IgnoreCase);
-            bool isEscapePodStorage = ownerTransform.parent.name.StartsWith("EscapePod");
-            if (isCyclopsLocker)
-            {
-                ownerId = GetCyclopsLockerId(ownerTransform);
-            }
-            else if (isEscapePodStorage)
-            {
-                ownerId = GetEscapePodStorageId(ownerTransform);
-            }
-            else
-            {
-                ownerId = NitroxEntity.GetId(ownerTransform.transform.parent.gameObject);
-            }
-
             NitroxId itemId = NitroxEntity.GetId(pickupable.gameObject);
             byte[] bytes = SerializationHelper.GetBytes(pickupable.gameObject);
 
-            ItemData itemData = new ItemData(ownerId, itemId, bytes);
-            ItemContainerAdd add = new ItemContainerAdd(itemData);
-            packetSender.Send(add);
+            ItemData itemData = new ItemData(GetOwner(containerTransform), itemId, bytes);
+            if (packetSender.Send(new ItemContainerAdd(itemData)))
+            {
+                Log.Debug($"Sent: Added item '{pickupable.GetTechType()}' to container '{containerTransform.gameObject.GetHierarchyPath()}'");
+            }
         }
 
-        public void BroadcastItemRemoval(Pickupable pickupable, Transform ownerTransform)
+        public void BroadcastItemRemoval(Pickupable pickupable, Transform containerTransform)
         {
-            NitroxId ownerId = null;
-
-            bool isCyclopsLocker = Regex.IsMatch(ownerTransform.gameObject.name, @"Locker0([0-9])StorageRoot$", RegexOptions.IgnoreCase);
-            bool isEscapePodStorage = ownerTransform.parent.name.StartsWith("EscapePod");
-            if (isCyclopsLocker)
-            {
-                ownerId = GetCyclopsLockerId(ownerTransform);
-            }
-            else if (isEscapePodStorage)
-            {
-                ownerId = GetEscapePodStorageId(ownerTransform);
-            }
-            else
-            {
-                ownerId = NitroxEntity.GetId(ownerTransform.transform.parent.gameObject);
-            }
-
             NitroxId itemId = NitroxEntity.GetId(pickupable.gameObject);
-            ItemContainerRemove remove = new ItemContainerRemove(ownerId, itemId);
-            packetSender.Send(remove);
+            if (packetSender.Send(new ItemContainerRemove(GetOwner(containerTransform), itemId)))
+            {
+                Log.Debug($"Sent: removed item '{pickupable.GetTechType()}' from container '{containerTransform.gameObject.GetHierarchyPath()}'");
+                Log.Debug(Environment.StackTrace);
+            }
         }
 
         public void AddItem(GameObject item, NitroxId containerId)
@@ -79,16 +49,16 @@ namespace NitroxClient.GameLogic
             Optional<GameObject> owner = NitroxEntity.GetObjectFrom(containerId);
             if (!owner.HasValue)
             {
-                Log.Info("Unable to find inventory container with id: " + containerId);
+                Log.Warn("Unable to find inventory container with id: " + containerId);
                 return;
             }
             Optional<ItemsContainer> opContainer = InventoryContainerHelper.GetBasedOnOwnersType(owner.Value);
             if (!opContainer.HasValue)
             {
-                Log.Error("Could not find container field on object " + owner.Value.name);
+                Log.Warn("Could not find container field on object " + owner.Value.name);
                 return;
             }
-            
+
             ItemsContainer container = opContainer.Value;
             Pickupable pickupable = item.RequireComponent<Pickupable>();
             using (packetSender.Suppress<ItemContainerAdd>())
@@ -96,7 +66,7 @@ namespace NitroxClient.GameLogic
                 container.UnsafeAdd(new InventoryItem(pickupable));
             }
         }
-        
+
         public void RemoveItem(NitroxId ownerId, NitroxId itemId)
         {
             GameObject owner = NitroxEntity.RequireObjectFrom(ownerId);
@@ -104,10 +74,10 @@ namespace NitroxClient.GameLogic
             Optional<ItemsContainer> opContainer = InventoryContainerHelper.GetBasedOnOwnersType(owner);
             if (!opContainer.HasValue)
             {
-                Log.Error("Could not find container field on object " + owner.name);
+                Log.Error($"Could not find item container behaviour on object '{owner.name}' with Nitrox id '{ownerId}'");
                 return;
             }
-            
+
             ItemsContainer container = opContainer.Value;
             Pickupable pickupable = item.RequireComponent<Pickupable>();
             using (packetSender.Suppress<ItemContainerRemove>())
@@ -120,25 +90,39 @@ namespace NitroxClient.GameLogic
         {
             string LockerId = ownerTransform.gameObject.name.Substring(7, 1);
             GameObject locker = ownerTransform.parent.gameObject.FindChild("submarine_locker_01_0" + LockerId);
-            
-            if (locker != null)
+            if (!locker)
             {
-                StorageContainer SC = locker.GetComponentInChildren<StorageContainer>();
-                if (SC != null)
-                {
-                    return NitroxEntity.GetId(SC.gameObject);
-                }
-                throw new Exception("Could not find StorageContainer From Object: submarine_locker_01_0" + LockerId);
-
+                throw new Exception("Could not find Locker Object: submarine_locker_01_0" + LockerId);
             }
-            throw new Exception("Could not find Locker Object: submarine_locker_01_0" + LockerId);
-
+            StorageContainer storageContainer = locker.GetComponentInChildren<StorageContainer>();
+            if (!storageContainer)
+            {
+                throw new Exception($"Could not find {nameof(StorageContainer)} From Object: submarine_locker_01_0{LockerId}");                
+            }
+            
+            return NitroxEntity.GetId(storageContainer.gameObject);
         }
 
         public NitroxId GetEscapePodStorageId(Transform ownerTransform)
         {
             StorageContainer SC = ownerTransform.parent.gameObject.RequireComponentInChildren<StorageContainer>();
             return NitroxEntity.GetId(SC.gameObject);
+        }
+
+        private NitroxId GetOwner(Transform ownerTransform)
+        {
+            bool isCyclopsLocker = Regex.IsMatch(ownerTransform.gameObject.name, @"Locker0([0-9])StorageRoot$", RegexOptions.IgnoreCase);
+            if (isCyclopsLocker)
+            {
+                return GetCyclopsLockerId(ownerTransform);
+            }
+            bool isEscapePodStorage = ownerTransform.parent.name.StartsWith("EscapePod");
+            if (isEscapePodStorage)
+            {
+                return GetEscapePodStorageId(ownerTransform);
+            }
+
+            return NitroxEntity.GetId(ownerTransform.parent.gameObject);
         }
     }
 }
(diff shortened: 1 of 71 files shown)


NitroxServer/GameLogic/PlayerManager.cs
```diff
@@ -15,38 +15,32 @@ namespace NitroxServer.GameLogic
...
         }
 
-        public List<Player> GetAllPlayers()
+        public IEnumerable<Player> GetAllPlayers()
         {
-            lock(allPlayersByName)
-            {
-                return new List<Player>(allPlayersByName.Values);
-            }
+            return allPlayersByName.Values;
```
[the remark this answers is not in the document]
Measurity: Yes

NitroxServer/GameLogic/Items/InventoryManager.cs
```diff
@@ -1,69 +1,54 @@
...
         {
-            lock (inventoryItemsById)
-            {
-                inventoryItemsById[itemData.ItemId] = itemData;
-            }
+            inventoryItemsById[itemData.ItemId] = itemData;
+
+#if DEBUG
+            Log.Debug($"Received inventory item '{itemData.ItemId}' to container '{itemData.ContainerId}'. Total items: {inventoryItemsById.Count}");
+#endif
```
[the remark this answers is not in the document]
Measurity: Removed it, should just be debug so that we can decide by changing log levels.

NitroxServer/GameLogic/Unlockables/StoryGoalData.cs
```diff
@@ -8,94 +9,22 @@ namespace NitroxServer.GameLogic.Unlockables
...
-        }
-
-        [ProtoIgnore]
-        private List<string> completedGoals = new List<string>();
+        public ThreadSafeCollection<string> CompletedGoals { get; } = new ThreadSafeCollection<string>();
         
-        [ProtoIgnore]
-        private List<string> radioQueue = new List<string>();
+        [ProtoMember(2)]
+        public ThreadSafeCollection<string> RadioQueue { get; } = new ThreadSafeCollection<string>();
```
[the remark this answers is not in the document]
Measurity: Added tests and the server saves it properly. I think it's solved because this is a custom dictionary marked with ProtoContract.

NitroxServer/Server.cs
```diff
@@ -34,6 +36,24 @@ public Server(WorldPersistence worldPersistence, World world, ServerConfig serve
...
+                StringBuilder builder = new StringBuilder();
+                builder.AppendLine($" - Game mode: {world.GameMode}");
+                builder.AppendLine($" - Inventory items: {world.InventoryManager.GetAllInventoryItems().Count}");
+                builder.AppendLine($" - Storage slot items: {world.InventoryManager.GetAllStorageSlotItems().Count}");
+                builder.AppendLine($" - Known tech: {world.GameData.PDAState.KnownTechTypes.Count}");
+                builder.AppendLine($" - Radio messages stored: {world.GameData.StoryGoals.RadioQueue.Count}");
+                builder.AppendLine($" - Story goals unlocked: {world.GameData.StoryGoals.GoalUnlocks.Count}");
+                builder.AppendLine($" - Story goals completed: {world.GameData.StoryGoals.CompletedGoals.Count}");
+                builder.AppendLine($" - Encyclopedia entries: {world.GameData.PDAState.EncyclopediaEntries.Count}");
+                return builder.ToString();
```
Coding-Hen, line 53: Would it be possible to have the total that it could be out of?

NitroxModel/DataStructures/ThreadSafeCollection.cs
```diff
@@ -0,0 +1,295 @@
...
+using System.Diagnostics;
+using System.Linq;
+using ProtoBufNet;
+
+namespace NitroxModel.DataStructures
+{
+    [DebuggerDisplay("Items = {" + nameof(collection) + "}")]
+    [ProtoContract]
+    [Serializable]
+    public class ThreadSafeCollection<T> : IList<T>
```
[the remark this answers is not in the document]
Measurity: HashSet has different performance charactaistics like a O(1) `.Contains()` check where list is O(n). But I'll do a check if it's worth it just to be sure.

NitroxServer/Serialization/ServerProtobufSerializer.cs
```diff
@@ -9,14 +9,11 @@ namespace NitroxServer.Serialization
 {
     public class ServerProtobufSerializer
     {
-        private readonly RuntimeTypeModel model;
 
-        protected RuntimeTypeModel Model { get { return model; } }
+        protected RuntimeTypeModel Model { get; } = TypeModel.Create();
```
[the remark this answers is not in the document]
Measurity: Has to be protected because it's used by SubnauticaServerProtobufSerializer. Want me to change it to new protected method to add to the `model` so that this can be private?

NitroxServer/Server.cs
Measurity: I'll leave this for a new PR since this can be fleshed out a ton from what it is now.

NitroxServer-Subnautica/Program.cs
```diff
@@ -24,13 +25,19 @@ private static void Main(string[] args)
...
             try
             {
                 server = NitroxServiceLocator.LocateService<Server>();
+                Log.Info($"Loaded save\n{server.SaveSummary}");
                 server.Start();
             }
+            catch (SocketException e)
+            {
+                Log.Error("Unable to start server", e);
+                server?.Stop();
```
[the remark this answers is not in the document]
Measurity: LiteNetLib actually eats the exception so I changed it again to just abort and log.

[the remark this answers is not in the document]
Measurity: I don't think we want to crash on any error since most of the time it will just fail to save/persist something which isn't bad enough to crash the whole thing imo.

NitroxModel/DataStructures/ThreadSafeDictionary.cs
```diff
@@ -0,0 +1,184 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using ProtoBufNet;
+
+namespace NitroxModel.DataStructures
+{
+    [DebuggerDisplay("Items = {" + nameof(Entries) + "}")]
+    [ProtoContract]
```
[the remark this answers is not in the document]
Measurity: Actually not sure how it serializes. Maybe Protobuf uses the properties for it to get/set data. I know that it saves correctly though 😅

NitroxModel/DataStructures/ThreadSafeDictionary.cs
```diff
@@ -0,0 +1,184 @@
...
+    public class ThreadSafeDictionary<TKey, TValue> : IDictionary<TKey, TValue>
```
brianpopow, line 11: Forgive my ignorance, im new to the source code of Nitrox, but would `ConcurrentDictionary` not be on option instead of re-implementing things?

Measurity: I remade it so that we have more control over serialization and how the data is actually processed internally. This allows us to optimize for Nitrox.

For general purpose applications I would have used ConcurrentDictionary.

NitroxModel/DataStructures/GameLogic/VehicleModel.cs
```diff
@@ -23,7 +23,7 @@ public class VehicleModel
         public Quaternion Rotation { get; set; }
 
         [ProtoMember(5)]
-        public List<InteractiveChildObjectIdentifier> InteractiveChildIdentifiers { get; set; } = new List<InteractiveChildObjectIdentifier>();
+        public ThreadSafeCollection<InteractiveChildObjectIdentifier> InteractiveChildIdentifiers { get; }
```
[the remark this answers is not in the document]
Measurity: Added unit test for ThreadSafeCollection as well.

NitroxServer/GameLogic/Unlockables/StoryGoalData.cs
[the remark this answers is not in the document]
Measurity: I verified that the encyclopedia entries are correctly persisted now.

NitroxModel/DataStructures/GameLogic/AbsoluteEntityCell.cs
```diff
@@ -88,30 +127,14 @@ public Int3 GetCellSize(Int3 blocksPerBatch)
...
-                case 0:
-                    return 10;
-                case 1:
-                case 2:
-                case 3:
-                    return 5;
-                default:
-                    throw new Exception();
-            }
+            return Equals(BatchId, other.BatchId) && Equals(CellId, other.CellId) && Level == other.Level;
```
Sunrunner37, line 137: Can other be null here?  For example, if this code path is invoked from == or != operators?

NitroxModel/DataStructures/ThreadSafeCollection.cs
```diff
@@ -0,0 +1,318 @@
...
+            {
+                foreach (T item in collection)
+                {
+                    if (predicate(item))
+                    {
+                        return item;
+                    }
+                }
+            }
+            throw new ArgumentNullException(nameof(predicate), "No item in collection matches select predicate.");
```
Sunrunner37, line 287: minor: seems a weird exception type if predicate is not null.

NitroxModel/DataStructures/ThreadSafeDictionary.cs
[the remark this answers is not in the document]
Sunrunner37: It might be because it extends IDictionary. 

For the future, I don't think we want this to be protobug enabled because it has notoriously bad handling of dictionaries.  There should only be 1 or 2 places that actually persist a dictionary (such as PDA which hasn't been migrated yet).  We have moved everything else over to store only lists and have the indexing as an implementation feature of the class that needs fast lookups.  For now we can leave it.

NitroxModel/DataStructures/GameLogic/AbsoluteEntityCell.cs
Measurity: Technically can be but it's protected sn only called by the overridden public equals against type `object`.

NitroxPatcher/Patches/Dynamic/ItemsContainer_NotifyAddItem_Patch.cs
```diff
@@ -10,12 +10,18 @@ public class ItemsContainer_NotifyAddItem_Patch : NitroxPatch, IDynamicPatch
...
         public static readonly Type TARGET_CLASS = typeof(ItemsContainer);
         public static readonly MethodInfo TARGET_METHOD = TARGET_CLASS.GetMethod("NotifyAddItem", BindingFlags.NonPublic | BindingFlags.Instance);
+        private static ItemContainers itemContainers;
 
         public static void Postfix(ItemsContainer __instance, InventoryItem item)
         {
+            if (itemContainers == null)
+            {
+                itemContainers = NitroxServiceLocator.LocateService<ItemContainers>();
+            }
```
Sunrunner37, line 20: Don't these already cache internally?  Also, will this hang on to a dead reference if we refresh the service locator (such as when reconnecting)?

NitroxModel/DataStructures/ThreadSafeCollection.cs
Measurity: Yikes, read the code wrong here. Good catch.

NitroxServer/Communication/NetworkingLayer/LiteNetLib/LiteNetLibServer.cs
```diff
@@ -35,9 +33,14 @@ public override void Start()
             server.UnconnectedMessagesEnabled = true;
             server.UpdateTime = 15;
             server.UnsyncedEvents = true;
-            server.Start(portNumber);
+            if (!server.Start(portNumber))
+            {
+                return false;
+            }
 
+            Log.Info("Using LiteNetLib as networking library");
```
Sunrunner37, line 41: minor: I think we can remove now.

NitroxModel/DataStructures/ThreadSafeCollection.cs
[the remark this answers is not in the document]
Measurity: Had once case where it changed to list. Fixed now.

NitroxServer/GameLogic/Entities/EntitySimulator.cs
```diff
@@ -92,11 +91,10 @@ private List<Entity> AssignForCells(Player player, AbsoluteEntityCell[] added)
             foreach (AbsoluteEntityCell cell in added)
             {
                 List<Entity> entities = entityManager.GetEntities(cell);
-
                 assignedEntities.AddRange(
                     entities.Where(entity => cell.Level <= entity.Level &&
-                                                ((entity.SpawnedByServer && serverSpawnedSimulationWhiteList.Contains(entity.TechType)) || !entity.SpawnedByServer) &&
-                                                simulationOwnershipData.TryToAcquire(entity.Id, player, DEFAULT_ENTITY_SIMULATION_LOCKTYPE)));                       
+                                             (entity.SpawnedByServer && serverSpawnedSimulationWhiteList.Contains(entity.TechType) || !entity.SpawnedByServer) &&
```
Sunrunner37, line 96: I think we should leave this as-is to not have any ambiguity.

NitroxServer/Serialization/ServerProtobufSerializer.cs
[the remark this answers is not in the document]
Sunrunner37: k we'll leave it

NitroxClient/GameLogic/ItemContainers.cs
```diff
@@ -16,98 +16,67 @@ namespace NitroxClient.GameLogic
...
-            packetSender.Send(remove);
         }
 
         public void AddItem(GameObject item, NitroxId containerId)
         {
             Optional<GameObject> owner = NitroxEntity.GetObjectFrom(containerId);
             if (!owner.HasValue)
             {
-                Log.Info("Unable to find inventory container with id: " + containerId);
+                Log.Warn("Unable to find inventory container with id: " + containerId);
```
Sunrunner37, line 51: Should these be errors?

NitroxPatcher/Patches/Dynamic/ItemsContainer_NotifyAddItem_Patch.cs
Measurity: I don't think they do because had this once inside the protobuf patches and it slowed the game down to a crawl. I'll check the decompiled code now to make sure.

Measurity: Nvm, it does cache internally but it does some casts and checks. I'll revert the changes in methods that aren't called so often.

NitroxServer/GameLogic/Entities/EntitySimulator.cs
Measurity: I prefer not to have redundant parenthesis in the code because the auto formatting will clean it up and changing it to never clean it up also seems undesirable.

How about this?

```cs
bool shouldBeOwnershipSimulated = entity.SpawnedByServer && serverSpawnedSimulationWhiteList.Contains(entity.TechType) || !entity.SpawnedByServer;
bool canAquireLockForPlayer = simulationOwnershipData.TryToAcquire(entity.Id, player, DEFAULT_ENTITY_SIMULATION_LOCKTYPE);
return cell.Level <= entity.Level && shouldBeOwnershipSimulated && canAquireLockForPlayer;
```

NitroxServer/GameLogic/Entities/EntitySimulator.cs
```diff
@@ -92,11 +91,14 @@ private List<Entity> AssignForCells(Player player, AbsoluteEntityCell[] added)
...
-
                 assignedEntities.AddRange(
-                    entities.Where(entity => cell.Level <= entity.Level &&
-                                                ((entity.SpawnedByServer && serverSpawnedSimulationWhiteList.Contains(entity.TechType)) || !entity.SpawnedByServer) &&
-                                                simulationOwnershipData.TryToAcquire(entity.Id, player, DEFAULT_ENTITY_SIMULATION_LOCKTYPE)));                       
+                    entities.Where(entity =>
+                    {
+                        bool isSpawnedByServerAndWhitelisted = entity.SpawnedByServer && serverSpawnedSimulationWhiteList.Contains(entity.TechType);
+                        bool shouldBeOwnershipSimulated = isSpawnedByServerAndWhitelisted || !entity.SpawnedByServer;
+                        bool canAquireLockForPlayer = simulationOwnershipData.TryToAcquire(entity.Id, player, DEFAULT_ENTITY_SIMULATION_LOCKTYPE);
```
Sunrunner37, line 99: This needs to be the last statement because we do not want to try to acquire a lock without checking to see if we should.

NitroxServer/GameLogic/Entities/EntitySimulator.cs
```diff
@@ -92,11 +91,14 @@ private List<Entity> AssignForCells(Player player, AbsoluteEntityCell[] added)
...
                 List<Entity> entities = entityManager.GetEntities(cell);
-
                 assignedEntities.AddRange(
-                    entities.Where(entity => cell.Level <= entity.Level &&
-                                                ((entity.SpawnedByServer && serverSpawnedSimulationWhiteList.Contains(entity.TechType)) || !entity.SpawnedByServer) &&
-                                                simulationOwnershipData.TryToAcquire(entity.Id, player, DEFAULT_ENTITY_SIMULATION_LOCKTYPE)));                       
+                    entities.Where(entity =>
+                    {
+                        bool isSpawnedByServerAndWhitelisted = entity.SpawnedByServer && serverSpawnedSimulationWhiteList.Contains(entity.TechType);
+                        bool shouldBeOwnershipSimulated = isSpawnedByServerAndWhitelisted || !entity.SpawnedByServer;
```
Sunrunner37, line 98: minor: maybe `isEligibleForSimulation`?

PR #1024, opened 2020-04-06
Redesign player chat and NitroxUnity structure

Jannify:
- Moved NitroxUnity from submodule to Nitrox directly to make it easier to change/update asset bundles (with the updated gitignore there is now need for separate repository for it)
- Updated chatlog and added chatkeyhint asset bundle to/from new design of @keyboardfox 

Screenshot_1
Screenshot_2
Screenshot_3
Screenshot_4


NitroxClient/MonoBehaviours/Gui/Chat/PlayerChat.cs
```diff
@@ -0,0 +1,156 @@
...
+
+        private PlayerChatManager playerChatManager;
+        private CanvasGroup canvasGroup;
+        private HorizontalOrVerticalLayoutGroup[] layoutGroups;
+        private GameObject logEntryPrefab;
+        private Image[] backgroundImages;
+        private bool transparent;
+        private InputField inputField;
+
+        private Dictionary<ChatLogEntry, GameObject> entries = new Dictionary<ChatLogEntry, GameObject>();
```
Measurity, line 27: Looks like you don't need a `Dictionary` here. I think a `Queue<GameObject>` would be best or add a `GameObject` field to `ChatLogEntry` and pass the gameobject to it so you can access it later if needed.

NitroxClient/MonoBehaviours/Gui/Chat/PlayerChat.cs
```diff
@@ -0,0 +1,156 @@
...
+            float alpha = transparent ? 1f : TOGGLED_TRANSPARENCY;
+            transparent = !transparent;
+
+            foreach (Image backgroundImage in backgroundImages)
+            {
+                backgroundImage.CrossFadeAlpha(alpha, 0.5f, false);
+            }
+        }
+
+        private IEnumerator ToggleTransparencyCanvasGroup(bool fadeIn)
```
Measurity, line 136: Rename to indicate it will fade the whole chat. Not just change transparency (it will hide or show all the way). Also `canvasGroup` isn't clear enough imo.

Example:
```suggestion
        private IEnumerator ToggleChatFade(bool fadeIn)
```

NitroxClient/Communication/Packets/Processors/ChatMessageProcessor.cs
```diff
@@ -37,18 +42,19 @@ private void LogClientMessage(ChatMessage message)
...
+                Log.Error($"Tried to add chat message for remote player that could not be found with id '${message.PlayerId}' and message: '{message.Text}'.\nAll remote players right now:\n{playerTableFormatted}");
                 throw new Exception($"Tried to add chat message for remote player that could not be found with id '${message.PlayerId}' and message: '{message.Text}'.\nAll remote players right now:\n{playerTableFormatted}");
             }
-            
+
             RemotePlayer remotePlayerInstance = remotePlayer.Value;
-            playerChat.AddMessage(remotePlayerInstance.PlayerName, message.Text, remotePlayerInstance.PlayerSettings.PlayerColor);
-            playerChat.ShowLog();
+            playerChatManager.AddMessage(remotePlayerInstance.PlayerName, message.Text, remotePlayerInstance.PlayerSettings.PlayerColor);
+            playerChatManager.ShowChat();
```
Measurity, line 51: When client sends a message it should always refocus the input field so that client can keep typing.
```suggestion
            playerChatManager.SelectChat();
```

NitroxClient/MonoBehaviours/Gui/Chat/PlayerChat.cs
```diff
@@ -0,0 +1,156 @@
...
+        {
+            chatLogEntry.MessageText = SanitizeMessage(chatLogEntry.MessageText);
+
+            if (entries.Count == MESSAGE_LIMIT)
+            {
+                Destroy(entries.Values.First());
+                entries.Remove(entries.Keys.First());
+            }
+
+            GameObject newChatLogEntry = Instantiate(logEntryPrefab, logEntryPrefab.transform.parent);
```
Measurity, line 66: New entry should not keep their world pos.
```suggestion
            GameObject newChatLogEntry = Instantiate(logEntryPrefab, logEntryPrefab.transform.parent, false);
```

NitroxClient/MonoBehaviours/Gui/Chat/PlayerChat.cs
```diff
@@ -0,0 +1,156 @@
...
+            newChatLogEntry.GetComponent<PlayerChatLogItem>().ApplyOnPrefab(chatLogEntry);
+            entries.Add(chatLogEntry, newChatLogEntry);
+
+            StartCoroutine(UpdateChatEntrySpacing());
```
Measurity, line 70: I believe this is no longer needed if you pass the `false` to `Instantiate` a few lines up. I removed this check and chat seems to work fine. What does this fix exactly?

NitroxClient/GameLogic/ChatUI/PlayerChatManager.cs
```diff
@@ -0,0 +1,163 @@
...
+
+        private PlayerChat playerChat;
+        public Transform PlayerChaTransform => playerChat.transform;
+        private readonly IMultiplayerSession multiplayerSession;
+        private static readonly string assemblyPath = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
+
+        private bool finishedLoading;
+        private bool showChatAfterLoading;
+        private bool selectChatAfterLoading;
+        private readonly List<ChatLogEntry> entriesBeforeFinishedLoading = new List<ChatLogEntry>();
```
Measurity, line 31: This could be refactored away to use `entries`. When the chat finishes loading it can add the entries up to that point and continue using entries for all new messages.

NitroxClient/GameLogic/ChatUI/PlayerChatManager.cs
```diff
@@ -0,0 +1,163 @@
...
+
+        public void ShowChat()
```
Measurity, line 33: This can be refactored to use `IEnumerator` so that it can wait to finish loading and show the chat instead of using a bool flag.

NitroxClient/MonoBehaviours/Gui/Chat/PlayerChat.cs
```diff
@@ -0,0 +1,156 @@
...
+using UnityEngine;
+using UnityEngine.EventSystems;
+using UnityEngine.UI;
+
+namespace NitroxClient.MonoBehaviours.Gui.Chat
+{
+    public class PlayerChat : uGUI_InputGroup
+    {
+        private const int LINE_CHAR_LIMIT = 255;
+        private const int MESSAGE_LIMIT = 64;
```
Measurity, line 15: Rename to indicate it refers to multiple message limit not just 'a' message limit.

```suggestion
        private const int MESSAGES_LIMIT = 64;
```

NitroxClient/MonoBehaviours/Gui/Chat/PlayerChatInputField.cs
```diff
@@ -1,132 +1,63 @@
...
-                        ChatEnabled = false;
-                        break;
-                }
-            }
+            timeLeftUntilAutoClose = PlayerChat.CHAT_VISIBILITY_TIME_LENGTH;
+            FreezeTime = false;
         }
 
-        private void SetGUIStyle()
+        private void LateUpdate()
```
Measurity, line 38: Change to use Update() if possible.

```suggestion
        private void Update()
```

NitroxClient/MonoBehaviours/Gui/Input/KeyBindings/Actions/ChatKeyBindingAction.cs
```diff
@@ -7,8 +7,10 @@ public class ChatKeyBindingAction : KeyBindingAction
     {
         public override void Execute()
         {
-            PlayerChat chatManager = NitroxServiceLocator.LocateService<PlayerChat>();
-            chatManager.ShowChat();
+            if (FPSInputModule.current.lastGroup == null)
```
Measurity, line 10:
```suggestion
            // If no other UWE input field is currently active then allow chat to open.
            // Fixes issue where input field opens on chat hotkey while player is renaming a locker/beacon
            if (FPSInputModule.current.lastGroup == null)
```

NitroxClient/MonoBehaviours/Gui/Chat/PlayerChat.cs
Jannify: There are cases where Unity doesn't update the `LayoutGroup` and from my experience it's better to have that just to be save.

NitroxClient/MonoBehaviours/Gui/Chat/PlayerChat.cs
Jannify: The problem is that the Unity component is called `CanvasGroup`, so don't know how to name it. better without adding useless info

Jannify:
Screenshot_1
Screenshot_2

NitroxClient/GameLogic/ChatUI/PlayerChatManager.cs
```diff
@@ -0,0 +1,87 @@
...
+        public Transform PlayerChaTransform => playerChat.transform;
+        private readonly IMultiplayerSession multiplayerSession;
+
+        public void ShowChat() => Player.main.StartCoroutine(ShowChatAsync());
+        private IEnumerator ShowChatAsync()
+        {
+            while (PlayerChat.IsLoading)
+            {
+                yield return null;
+            }
```
Sunrunner37, line 28: Instead of these while loops can we use: https://docs.unity3d.com/ScriptReference/WaitUntil.html (if so, please update all of the while loops in the PR)

NitroxClient/MonoBehaviours/Gui/Chat/PlayerChat.cs
```diff
@@ -0,0 +1,144 @@
...
+            }
+
+            ChatLogEntry chatLogEntry;
+            GameObject chatLogEntryObject;
+            if (entries.Count != 0 && entries.Last().PlayerName == playerName)
+            {
+                chatLogEntry = entries.Last();
+                chatLogEntry.MessageText += "\n" + message;
+                chatLogEntry.UpdateTime();
+                chatLogEntryObject = chatLogEntry.EntryObject;
```
Sunrunner37, line 77: Would it be better to make each chat entry immutable instead of appending to the players last message?   Might simplify the logic here and be more intuitive.

NitroxClient/GameLogic/ChatUI/PlayerChatManager.cs
```diff
@@ -0,0 +1,87 @@
...
+            playerChat.Select();
+        }
+
+        private IEnumerator LoadChatLogAsset()
+        {
+            CoroutineWithData coroutineWD = new CoroutineWithData(Player.main, AssetBundleLoader.LoadUIAsset("chatlog", "PlayerChatCanvas", true));
+            yield return coroutineWD.Coroutine;
+
+            playerChat = ((GameObject)coroutineWD.Result).gameObject.AddComponent<PlayerChat>();
+            yield return playerChat.SetupChatComponents();
```
Sunrunner37, line 82: Does this need to be a coroutine?  It introduces a bit of complexity with the new class CoroutineWithData.  Just validating we need the code complexity.

Jannify: We need `CoroutineWithData` if we want to unify the UI load (https://github.com/SubnauticaNitrox/Nitrox/blob/cd644980a009ef29fb7e8d2becd4ae8b5594921e/NitroxClient/Unity/Helper/AssetBundleLoader.cs#L35-L48). The normal coroutine has no option to return a value which is needed in this case. I could remove it and use the lines  for both assets in `PlayerChatManager`.

I think a coroutine of some sort is needed because loading and modifying the asset is a heavy task and shouldn't have the possibly to freeze the game.

NitroxClient/MonoBehaviours/Gui/Chat/PlayerChat.cs
Jannify: Yeah I can do this with two different prefabs (one for messages with names and one without).

NitroxClient/GameLogic/ChatUI/PlayerChatManager.cs
Measurity: You can use a callback instead of ienumerator with data class. Ex: Func<TData> as argument to the IEnumerator, invoke it after the yield and then do something with the data at the caller.

NitroxClient/MonoBehaviours/Gui/Chat/PlayerChat.cs
Measurity: Not appending to the last message means that Unity needs to do more work handling the UI objects. I think it's worth it to append.

Issue #951, opened 2020-03-08
[current master] base pieces are not deconstructable


Her0brineCreeper:
While testing the current version of the Nitrox mod compiled from master, we noticed that deconstructed base pieces respawned after rejoining, so I started trying to fix this issue.
The first thing I did was adding some code to BaseManager.cs in order to make deconstruction work on the server side (I just copied the code from "DeconstructionBegin" to "BasePieceConstructionAmountChanged"), because the "DeconstructionBegin" packet isn't send by the client (yet?). After this deconstruction of lonely base pieces worked fine, but the ones in bases were sill not deconstructable. After more testing and searching I found the exact cause of this issue:
Whenever a base piece is finished constructing, the other ones in its base will also be "reconstructed" and get a new id. This leads to two problems:
- The ids of the old pieces change so when they get deconstructed, the id sent to the server to deconstruct isn't the same one the server has assigned to this piece.
- The gameObjects come in in a random? order, so that the latest one to which the id of the ghost is assigned, isn't the right one most of the time.

I found this out by writing the ids and the type of base piece to the log.
Here is an example log output where I added a L pice to an existing base:
```
[00:14:56 Client INFO]: Constructed 9272f502-3d4e-4728-9dd1-b1aa0e59bf70; /BaseCorridorL(Clone)/BaseGhost/BaseCell(Clone)/BaseCorridorLShape(Clone)
[00:14:56 Client INFO]: Constructed e1a8433b-d576-4ff5-9c4b-7ccff07fe193; /BaseCorridorL(Clone)/BaseGhost/BaseCell(Clone)/BaseCorridorLShape(Clone)
[00:14:56 Client INFO]: Constructed 2fd5f653-e801-4507-a7b9-33a8ea545082; /BaseCorridorL(Clone)/BaseGhost/BaseCell(Clone)/BaseCorridorLShape(Clone)
[00:14:59 Client INFO]: Constructed e006aa91-1653-460c-b653-7f6ec6ea27f3; /BaseCorridorL(Clone)/BaseGhost/BaseCell(Clone)/BaseCorridorLShape(Clone)
[00:15:00 Client INFO]: Constructed 341360fe-baa5-45f7-a9f7-82f0c4516e5d; /BaseCorridorL(Clone)/BaseGhost/BaseCell(Clone)/BaseCorridorLShape(Clone)
[...]

PR #1027, opened 2020-04-07
Client-Base-Building-logs/Base-Building-Issue

TheGuy920:
Creates a log file of the current base building status for the client machine, great for debugging!
(Epic Games or Steam)\subnautica\ClientLog.txt

If you use this you will see that there is a client base building issue
any base piece that is added after the first base piece that is added, does not get registered by the client machine at full completion, hence the server does not register it at full completion on reload

using something like a live log viewer and having it open on a second screen is advised
[Live_Log_Viewer.zip](https://github.com/SubnauticaNitrox/Nitrox/files/4441610/Live_Log_Viewer.zip)


Her0brineCreeper:
The biggest base building problems are fixed on the current master (and the version of the repository you use for your pull request) as you can read here: #951 
The reason why the building is not working on your branch is this part of your code: (NitroxClient/GameLogic/Building.cs: 203)
`string basestring = baseId.ToString();`
Starting from the second base piece the "baseId" is null and when trying to convert null to string the function fails and because of this the "constructionCompleted" packet isn't send to the server.
So the problem is just your code and not the Nitrox mod (anymore). For more information please read my issue linked above. Sunrunner37 and me were and are already working on all the building issues.

PR #1029, opened 2020-04-08
Loading screen version indicator now fades out

NeoNX01:
Tiny edit to correct a small oversight (at least I think it is one? if not, my bad!) that made the version number displayed while loading the game persist ingame.

I've made it fade out with the rest of the text.


Measurity:
LGTM but not sure what the original intent was so we should wait on Jannify to review this.

Jannify:
Think that @Sunrunner37 wanted it to persist like the original alpha/beta version text from Subnautica.

NeoNX01:
Ah fair enough, I figured it was a mistake due to the file's name (`LoadingScreen[...]`).

I suggest making it more discreet or a toggleable setting though, as of now it somewhat sticks out. Bit of a shame, considering how important immersion is in Subnautica!

I'm closing this, thank you all for the mod, keep up the great work!

PR #1031, opened 2020-04-10
Use Nlog logging

brianpopow:
This PR finishes what was started with #926: Removes log4net and replaces it with NLog.


NitroxClient/Communication/MultiplayerSession/MultiplayerSessionManager.cs
```diff
@@ -56,11 +56,11 @@ public void ProcessSessionPolicy(MultiplayerSessionPolicy policy)
             switch (localVersion.CompareTo(SessionPolicy.NitroxVersionAllowed))
             {
                 case -1:
-                    Log.InGame($"Your Nitrox installation is out of date. Server: {SessionPolicy.NitroxVersionAllowed}, Yours: {localVersion}.");
+                    Log.ShowInGameMessage($"Your Nitrox installation is out of date. Server: {SessionPolicy.NitroxVersionAllowed}, Yours: {localVersion}.");
```
Measurity, line 59: This should be kept as short as possbile like it was before `Log.InGame`

NitroxClient/GameLogic/Crafting.cs
```diff
@@ -30,7 +30,7 @@ public void FabricatorItemPickedUp(GameObject gameObject, TechType techType)
 
             FabricatorItemPickup fabricatorItemPickup = new FabricatorItemPickup(crafterId, techType.Model());
             packetSender.Send(fabricatorItemPickup);
-            Log.Debug(fabricatorItemPickup);
+            Log.Debug($"{fabricatorItemPickup}");
```
Measurity, line 33: No need for interpolated string here.

NitroxClient/GameLogic/InitialSync/InventoryItemsInitialSyncProcessor.cs
```diff
@@ -51,7 +51,7 @@ public override IEnumerator Process(InitialPlayerSync packet, WaitScreen.ManualW
                     }
                     catch (Exception ex)
                     {
-                        Log.Error("Error deserializing item data " + itemdata.ItemId + " " + ex.Message);
+                        Log.Exception("Error deserializing item data " + itemdata.ItemId, ex);
```
Measurity, line 54: I think it's best to have only `Log.Error` and have an overload that accepts exception object. This makes it consistent with the other methods that are also named after the log level they output on.

NitroxLauncher/LauncherLogic.cs
```diff
@@ -358,6 +358,16 @@ private void SyncAssembliesBetweenSubnauticaManagedAndLib()
 
             List<string> ignoreNoBinaries = new List<string>();
             CopyAllAssemblies(libDirectory, subnauticaManagedPath, ignoreNoBinaries);
+
+            CopyLogConfig(libDirectory);
+        }
+
+        private void CopyLogConfig(string libDirectory)
+        {
+            string logConfig = "NLog.config";
```
Measurity, line 367: If possible it would be better to use NLog API to set the configuration of the logger(s). Relying on the launcher to copy this file over is less certain.

NitroxModel/packages.config
```diff
@@ -2,5 +2,5 @@
 <packages>
   <package id="Autofac" version="2.6.3.862" targetFramework="net35" />
   <package id="LiteNetLib" version="0.8.2-beta" targetFramework="net35" />
-  <package id="log4net" version="2.0.8" targetFramework="net35" />
+  <package id="NLog" version="4.6.8" targetFramework="net40" />
```
Measurity, line 5: Should be net35 for compatibility reasons of NitroxModel.

NitroxServer/Communication/Packets/Processors/SubRootChangedPacketProcessor.cs
```diff
@@ -16,7 +16,7 @@ public SubRootChangedPacketProcessor(PlayerManager playerManager)
 
         public override void Process(SubRootChanged packet, Player player)
         {
-            Log.Info(packet);
+            Log.Info($"{packet}");
```
Measurity, line 19: Do not use interpolated strings when not necessary.

NitroxServer/IpLogger.cs
```diff
@@ -41,7 +41,7 @@ private static void PrintIfHamachi(NetworkInterface netInterface)
             IEnumerable<string> ips = netInterface.GetIPProperties().UnicastAddresses
                 .Select(address => address.Address.ToString())
                 .Where(address => !address.ToString().Contains("fe80::"));
-            Log.Info("If using Hamachi, use this IP: " + string.Join(" or ", ips));
+            Log.DebugSensitive("If using Hamachi, use this IP: {0}", string.Join(" or ", ips));
```
Measurity, line 44: This should be kept as Info

NitroxServer/IpLogger.cs
```diff
@@ -62,7 +62,7 @@ private static void PrintIfLan(NetworkInterface netInterface)
 
                 if (splitIpParts[0] == "10" || splitIpParts[0] == "192" && splitIpParts[1] == "168" || splitIpParts[0] == "172" && secondPart > 15 && secondPart < 32) //To get if IP is private
                 {
-                    Log.Info("If playing on LAN, use this IP: " + eachIp.Address);
+                    Log.DebugSensitive("If playing on LAN, use this IP: {0}", eachIp.Address);
```
Measurity, line 65: Should be kept as Info

NitroxServer/Serialization/BatchCellsParser.cs
```diff
@@ -63,7 +63,7 @@ public void ParseFile(Int3 batchId, string pathPrefix, string prefix, string suf
 
             if (!File.Exists(fileName))
             {
-                //Log.Debug("File does not exist: " + fileName);
+                //Log2.Instance.Log(NLogType.Debug, "File does not exist: " + fileName);
```
Measurity, line 66: Remove the comment here

Measurity:
I'm working on making some improvements to this. I'll apply the changes or make a PR for this PR when I'm ready.

brianpopow:
> I'm working on making some improvements to this. I'll apply the changes or make a PR for this PR when I'm ready.

i gave you access to modify the branch

Measurity:
Alright, almost done. Just need to:

1. Find a way to add the playername to the client log layout dynamically with padding (spaces around the player name). This is useful when running multiple clients at once and you want to know which client logged what.
2. Double check the code

brianpopow:
> 1. Find a way to add the playername to the client log layout dynamically with padding (spaces around the player name). This is useful when running multiple clients at once and you want to know which client logged what.

i think this can be achieved by adding the player name to the global log context. Then all logs would contain the playername.

Measurity:
I know. But I don't want the extra spaces in the layout when the player name is not in the global context properties. Could you add a trim + pad left/right to the player name? I think then this PR is good to go in.

brianpopow:
> I know. But I don't want the extra spaces in the layout when the player name is not in the global context properties. Could you add a trim + pad left/right to the player name? I think then this PR is good to go in.

I have tried to achieve this with 85bb43d96dfedb9230a54e534131d36d74ae1ead
Im a first time nlog user so im not sure if there is maybe a more elegant way of doing this.

More importantly the original config was looking like this:

```
<targets async="true">
    <target name="logfile" xsi:type="File" fileName="${basedir}/logfile.txt" layout="---${newline}${longdate}|${level:uppercase=true}|${message}|${exception}|${newline}${stacktrace}" />
    <target name="logconsole" xsi:type="Console" layout="${longdate}|${level:uppercase=true}|${message}|${exception}"/>
  </targets>
```

Note the `targets async="true"`. While changing to a programmatic configuration this was omitted.
I have now added a AsyncTargetWrapper for the file logger. More info about it [here](https://github.com/nlog/NLog/wiki/AsyncWrapper-target)

Im not sure if we should do that for the console logger, too. What do you think @Measurity ?

NitroxClient/MonoBehaviours/NitroxDebugManager.cs
```diff
@@ -158,17 +158,17 @@ private void OnDisable()
 
         private void SceneManager_sceneLoaded(Scene scene, LoadSceneMode loadMode)
         {
-            Log.Debug($"Scene '{scene.name}' loaded as {loadMode}");
+            Log.Debug("Scene {sceneName} loaded as {$loadMode}", scene.name, loadMode);
```
Sunrunner37, line 161: Why switch over to tokens like this?  Looks like the old format is more readable.

Measurity: I agree but then there is no structured logging (NLog) feature. Which I can use to hide private data or do additional processing based on the name passed in.

Also, we can push these logs to a server with, for example, elasticsearch and query them based on these names instead of having to regex parse them as a preprocessing step.

Measurity:
@brianpopow Console logger isn't as performance critical as the file logger so it's fine. though I wonder if there is a performance issue with using the console logging on the client. We should look into disabling that just to be sure.

NitroxClient/MonoBehaviours/NitroxDebugManager.cs
brianpopow: > I agree but then there is no structured logging (NLog) feature. Which I can use to hide private data or do additional processing based on the name passed in.
> 
> Also, we can push these logs to a server with, for example, elasticsearch and query them based on these names instead of having to regex parse them as a preprocessing step.

Structured logging is actually the reason i wanted this nlog feature done in the first place.
It can help alot in finding bugs. I have a proof of concept branch ready which uses nlog together with [seq](https://datalust.co/seq) as a logging sink. Of course elasticsearch combined with kibana is also an option.

Sunrunner37: I haven't been keeping up with the conversations on NLog so I might need additional context.  

Before talking about the implementation, we should agree that bulk log aggregation is one of our use-cases.  We don't currently do large-scale aggregation of client logs and I'm not sure if it should be done due to privacy reasons.  Furthermore, I'm fuzzy on if this is an appropriate application of the technology. Bulk log aggregation is very helpful for large-scaled, transactional applications; however, the characteristics of Nitrox bugs tend to be out of sync issues resulting from improper gameobject states or bad assumptions on game behavior. Generally, I can think of a few use-cases but wanted to take a step back and ask if we are over-killing.

Measurity: Reasons for this feature:
1. We don't want players that send logs out to us to contain their private data because it might be snooped on and it's none of our business anyway.
2. We can makes the logs smarter due to the tags/names we give to parts of the information which can be filtered out by log-viewing software.
3. Log4Net wasn't particularly fast in writing logs and the client-side game needs all the performance wins it can get. It could have be optimized more but NLog is faster out of the box.
4. NLog has active development going for it unlike the framework we used before. See: https://github.com/apache/logging-log4net/commits/master compared to https://github.com/nlog/NLog/

Sunrunner37: To be clear, I have no problem switching to NLog.  We can anchor this conversation in the structured logging feature, effectively removing 3/4 from your list.  In regards to the other two items:

1) If this is the major use-case, I'm hesitant that we are solving in a pragmatic way.  Should the underlying logging framework be cognizant about the tokens that are sensitive?  It seems the caller should have this context.  Should we write to the console directly or invoke a special sensitive-aware api?  Also, it seems that post processing every log event to do tokenization and looking up for sensitive tokens will mitigate the performance you want to gain.  

2)  I'm trying to understand if this is a real use-case per my discussion points in the last response.  The main use-case presented was bulk log aggregation.

Measurity: We already seperate the logs between sensitive or not for performance reasons. Handing sensitivity of logs messages on the callsite will make the code more complex by having to replace potential private data from it. This code would then be (somewhat) copy pasted between them all.

I'm not sure if we want to do bulk log aggregation right now. Though I can see use cases for having structured logging for filtering out the logs client-side in other software. It does enable opportunities for supporting a larger community through better analytics if we ever want/need it.

Imo this needs to be fleshed out more but I personally thing this is worth the inconvience of having to make our logs less readable just based on the opportunities of having a better grasp at problems occurring in Nitrox.

brianpopow: > 2. I'm trying to understand if this is a real use-case per my discussion points in the last response.  The main use-case presented was bulk log aggregation.

I would not say bulk log aggregation is the main use-case of structured logging. Its just a neat thing you basically get for free when you use nlog with structured logging. I also dont see data privacy issues. The aggregation of logs would be clearly ment only for developers who opt in for this. For all other this would be disabled (for example only configured in DEBUG mode)

The main advantage of structured logging is that you can easier find information in your logs and aggregate logs based on log properties. For example, if you have a log entry for constructing anything: you could add the construction `Amount` as a log property and then be able to find all logs with `Amount > 0.9` or `Amount >0.9 AND TechType == X`

Or the message `Entity was already spawned but not found (is it in another chunk?)`
You could simple find out if thats true, if you attach log context to the message. Here is an example how that could look like:

entity_was_spawned

@Sunrunner37 let me know what you think about structured logging. If you dont think its a useful feature (which is perfectly fine), i will not invest more time in it. It will be a big change, because all log messages will be affacted.

brianpopow:
> @brianpopow Console logger isn't as performance critical since it's used by the server. Though I wonder if there is a performance issue with using the console logging on the client even though we don't see the logs. We should look into disabling that just to be sure.

There is a feature to disable it when there is no console, see: 95a56f37a66a73781bbf9b62afb9dd31fa0d3362

NitroxClient/MonoBehaviours/NitroxDebugManager.cs
Sunrunner37: I let this PR bake a bit while we were getting out our latest release.  Now that I am revisiting PRs to merge, coming back to this one.  Sorry for the delay.

I'm fine with using the feature for our concrete use-cases such as the private data and maybe if we want easy access to globals (such as the player name).  However, I'm not totally convinced that we will end up using log aggregation in such a way that makes changing everything to conform to this style useful.  I would be perfectly fine re-visting this discussion if we do have a ready tool or strong use-case for large aggregation; however, as it currently stands, I think that the vast majority of development is simply tracing a log file in one instance of the Player.log using simple search commands.

killzoms:
Bit off topic but:

Ey @brianpopow you happen to be on the Nitrox Discord by any chance?

brianpopow:
> Bit off topic but:
> 
> Ey @brianpopow you happen to be on the Nitrox Discord by any chance?

@killzoms I was on the nitrox discord for a very few occasions, but im not a regular visitor.

NitroxModel/packages.config
```diff
@@ -1,6 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
   <package id="Autofac" version="2.6.3.862" targetFramework="net35" />
-  <package id="LiteNetLib" version="0.8.2-beta" targetFramework="net35" />
-  <package id="NLog" version="4.7.0" targetFramework="net40" />
+  <package id="NLog" version="4.7.0" targetFramework="net40" />  <package id="log4net" version="2.0.8" targetFramework="net35" />
```
Jannify, line 4: Is it right to add this log4net package?

NitroxServer/Serialization/World/WorldPersistence.cs
```diff
@@ -229,7 +231,10 @@ private World CreateFreshWorld()
                 Log.Info("Server has no password set.");
             }
             Log.InfoSensitive("Admin Password: {password}", config.AdminPassword);
-            Log.Info("To get help for commands, run {commandName} in console or {chatCommandName} in chatbox", "help", "/help");
+            Log.Info($"Admin Password: {config.AdminPassword}");
```
Jannify: Is this "double-call" correct?

NitroxModel/packages.config
brianpopow: no, it is replaced with nlog. Its likely a merge mistake.

NitroxServer/Serialization/World/WorldPersistence.cs
brianpopow: again likely a merge mistake, its removed now

Sunrunner37:
https://github.com/SubnauticaNitrox/Nitrox/pull/1121

PR #1033, opened 2020-04-11
Basic Aurora timer syncing

nesrak1:
Don't merge yet but let me know how you think this should go


nesrak1:
Mainly, just wondering about the weirdness of having a stopwatch follow along with the timers, and dealing with the weird saving code.

NitroxServer/GameLogic/EventTriggerer.cs
```diff
@@ -1,33 +1,61 @@
...
 namespace NitroxServer.GameLogic
 {
     public class EventTriggerer
     {
-        PlayerManager playerManager;
-        public EventTriggerer(PlayerManager playerManager)
+        private PlayerManager playerManager;
+        private GameData gameData;
+        private Stopwatch stopWatch;
+        private double elapsedTime;
```
nesrak1, line 15: we could just directly use the StoryTiming variables here

NitroxServer/GameLogic/StoryTimingData.cs
```diff
@@ -0,0 +1,36 @@
...
+{
+    [Serializable]
+    [ProtoContract]
+    public class StoryTimingData
+    {
+        [ProtoMember(1)]
+        public double ElapsedTime { get; set; }
+
+        [ProtoMember(2)]
+        public double? AuroraExplosionTime { get; set; }
```
Sunrunner37, line 18: Do we need to store this?  Can it be inferred by the ElapsedTime?

NitroxServer/Serialization/World/WorldPersistence.cs
```diff
@@ -188,7 +188,7 @@ private World CreateFreshWorld()
 
             world.SimulationOwnershipData = new SimulationOwnershipData();
             world.PlayerManager = new PlayerManager(players, config);
-            world.EventTriggerer = new EventTriggerer(world.PlayerManager);
+            world.EventTriggerer = new EventTriggerer(world.PlayerManager, gameData);
```
Sunrunner37, line 191: All of the managers here should be decoupled from persistence; ie, we shouldn't pass in gameData and have them understand serialization.  Would recommend pulling data off of the event trigger when saving like we do for everything else above (line 33-37).

NitroxServer/GameLogic/StoryTimingData.cs
nesrak1: No, because of the random explode time, even if I were to use a "time remaining", it still affects the multiplication stuff. We can hardcode explode time but this is more accurate with what sn does.

Coding-Hen:
Fixes: #834, fixes #499, fixes #626, fixes #678

Partly fixes:
#814, #770,

PR #1034, opened 2020-04-11
Catch BadFormatImage Invalid DOS signature to display a better understandable text

dartasen:
This error is related to corrupted .exe or .dll, this could be Nitrox or more likely that Subnautica is corrupted and they need to check the game files

### DIFF
--- a/NitroxLauncher/Pages/LaunchGamePage.xaml.cs
+++ b/NitroxLauncher/Pages/LaunchGamePage.xaml.cs
@@ -35,6 +35,11 @@ namespace NitroxLauncher.Pages
             {
                 await LauncherLogic.Instance.StartMultiplayerAsync();
             }
+            catch(BadImageFormatException ex)
+            {
+                //This exception is throw with "Invalid DOS Signature" which is a clue that some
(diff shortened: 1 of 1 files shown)


Measurity:
So this looks like it is an issue with patching? Or how the dlls are moved over? Can we fix that instead so that we don't need to tell uses to reverify their game?

nesrak1:
It might help if we could see what the corrupted file looked like.

dartasen:
> 
> 
> So this looks like it is an issue with patching? Or with how the dlls are moved over? Can we fix that instead so that we don't need to tell users to reverify their game?

Indeed this is an issue with patching, however we don't have enough informations to patch this one yet. Maybe we can log more informations at patching to figure out what can go wrong ?

NitroxLauncher/Pages/LaunchGamePage.xaml.cs
```diff
@@ -35,6 +35,11 @@ private async void MultiplayerButton_Click(object sender, RoutedEventArgs e)
             {
                 await LauncherLogic.Instance.StartMultiplayerAsync();
             }
+            catch(BadImageFormatException ex)
+            {
+                //This exception is throw with "Invalid DOS Signature" which is a clue that some
+                MessageBox.Show($"Try to check your subnautica files throught Steam/EpicGames in order to fix the game: {ex.Message}", "Error, Seems like your game is corrupted :(", MessageBoxButton.OK, MessageBoxImage.Error);
```
Coding-Hen, line 41: Might be better than "try to check" to be verify

PR #1035, opened 2020-04-12
Modifications to launcher

EtienneDx:
I moved the _path.txt_ config into the _Launcher.exe.config_ file, and stored the original _AssemblyCSharp.dll_ file, as well as the Hash of the modified dll, so it loads faster if someone is only playing multiplayer (no need to re-patch every time).


EtienneDx:
Depends, if they open from the launcher, we do the same as currently: check if applied and if so remove it. If they open from steam directly then I guess it could be a problem, but no more than the current system

EtienneDx:
Idk if I should close it directly but there's a problem with Linux here, since the server binary requires the path.txt file that I moved into the launcher.exe.config

NitroxLauncher/App.config
```diff
@@ -18,10 +18,16 @@
...
-      <setting name="IsExternalServer" serializeAs="String">
-        <value>True</value>
-      </setting>
-    </NitroxLauncher.Properties.Settings>
+      <NitroxLauncher.Properties.Settings>
+          <setting name="IsExternalServer" serializeAs="String">
+              <value>True</value>
+          </setting>
+          <setting name="SubnauticaPath" serializeAs="String">
+              <value>False</value>
```
Measurity, line 26: Should be empty string

NitroxLauncher/Properties/Settings.settings
```diff
@@ -5,5 +5,11 @@
     <Setting Name="IsExternalServer" Type="System.Boolean" Scope="User">
       <Value Profile="(Default)">True</Value>
     </Setting>
+    <Setting Name="SubnauticaPath" Type="System.String" Scope="User">
+      <Value Profile="(Default)">False</Value>
```
Measurity, line 9: Should be empty string

NitroxLauncher/Properties/Settings.Designer.cs
```diff
@@ -1,18 +1,18 @@
 ﻿//------------------------------------------------------------------------------
 // <auto-generated>
-//     This code was generated by a tool.
-//     Runtime Version:4.0.30319.42000
+//     Ce code a été généré par un outil.
```
Measurity, line 3: Can something be done to prevent system language making unncessary changes like this?

NitroxModel/Discovery/InstallationFinders/ConfigFileGameFinder.cs
```diff
@@ -9,26 +9,36 @@ namespace NitroxModel.Discovery.InstallationFinders
...
+
+        public static void AddResolver(ConfigResolver resolver)
+        {
+            getConfig += resolver;
+        }
 
         public Optional<string> FindGame(List<string> errors = null)
         {
-            if (!File.Exists(FILENAME))
+            if(getConfig == null)
```
Measurity, line 23: Use space after if, should be done automatically if you code format with VS

NitroxModel/Discovery/InstallationFinders/ConfigFileGameFinder.cs
```diff
@@ -9,26 +9,36 @@ namespace NitroxModel.Discovery.InstallationFinders
...
-                errors?.Add($@"Game installation directory config file is not set. Create a '{FILENAME}' in directory: '{Directory.GetCurrentDirectory()}' with the path to the Subnautica installation directory.");
+                // should only happen outside of the launcher
+                errors?.Add("Game installation config resolver is not set.");
                 return Optional.Empty;
             }
+            string config = getConfig.Invoke();
 
-            string path = File.ReadAllText(FILENAME).Trim();
-            if (string.IsNullOrEmpty(path))
+            if (config == "" || config == "False")
```
Measurity, line 31: Instead of checking for false it should check if it is a directory with subnautica exe

NitroxLauncher/Properties/Settings.Designer.cs
Jannify: As far as I know you need the English language pack installed.

PR #1040, opened 2020-04-15
Add null check on encyclopedia, pda, partialpda, knowntech and pda lo…

griphos:
Add null check on encyclopedia, pda, partialpda, knowntech and pda log entries.
To prevent load hanging when a null value is present ( it can happen ).

I think a check on loading in WorldPersistence should be added too

### DIFF
--- a/NitroxClient/GameLogic/InitialSync/PdaInitialSyncProcessor.cs
+++ b/NitroxClient/GameLogic/InitialSync/PdaInitialSyncProcessor.cs
@@ -52,7 +52,10 @@ namespace NitroxClient.GameLogic.InitialSync
             {
                 foreach (string entry in entries)
                 {
-                    PDAEncyclopedia.Add(entry, false);
+                    if( null != entry)
+                    {
+                        PDAEncyclopedia.Add(entry, false);
+                    }
                 }
             }
         }
@@ -63,7 +66,10 @@ namespace NitroxClient.GameLogic.InitialSync
 
             foreach (TechTypeModel item in pdaEntryComplete)
             {
-                complete.Add(item.Enum());
+                if (null != item)
+                {
+                    complete.Add(item.Enum());
+                }
             }
 
             Log.Info("PDAEntryComplete Save:" + pdaEntryComplete.Count + " Read Partial Client Final Count:" + complete.Count);
@@ -76,7 +82,11 @@ namespace NitroxClient.GameLogic.InitialSync
 
             foreach (PDAEntry entry in entries)
             {
-                partial.Add(new PDAScanner.Entry { progress = entry.Progress, techType = entry.TechType.Enum(), unlocked = entry.Unlocked });
+                if( null != entry)
+                {
+                    partial.Add(new PDAScanner.Entry { progress = entry.Progress, techType = entry.TechType.Enum(), unlocked = entry.Unlocked });
+                }
+                
             }
 
             Log.Info("PDAEntryPartial Save :" + entries.Count + " Read Partial Client Final Count:" + partial.Count);
@@ -90,8 +100,11 @@ namespace NitroxClient.GameLogic.InitialSync
             {
                 foreach (TechTypeModel techType in techTypes)
                 {
-                    HashSet<TechType> complete = (HashSet<TechType>)(typeof(PDAScanner).GetField("complete", BindingFlags.NonPublic | BindingFlags.Static).GetValue(null));
-                    KnownTech.Add(techType.Enum(), false);
+                    if (null != techType)
+                    {
+                        HashSet<TechType> complete = (HashSet<TechType>)(typeof(PDAScanner).GetField("complete", BindingFlags.NonPublic | BindingFlags.Static).GetValue(null));
+                        KnownTech.Add(techType.Enum(), false);
+                    }
                 }
             }
         }
@@ -106,7 +119,7 @@ namespace NitroxClient.GameLogic.InitialSync
 
                 foreach (PDALogEntry logEntry in logEntries)
                 {
-                    if (!entries.ContainsKey(logEntry.Key))
+                    if (null != logEntry && !entries.ContainsKey(logEntry.Key))
                     {
                         PDALog.EntryData entryData;
                         PDALog.GetEntryData(logEntry.Key, out entryData);


NitroxClient/GameLogic/InitialSync/PdaInitialSyncProcessor.cs
```diff
@@ -52,7 +52,10 @@ private void SetEncyclopediaEntry(List<string> entries)
             {
                 foreach (string entry in entries)
                 {
-                    PDAEncyclopedia.Add(entry, false);
+                    if( null != entry)
+                    {
+                        PDAEncyclopedia.Add(entry, false);
+                    }
```
dartasen, line 58: Minor : `PDAEncyclopedia?.Add(entry, false);` does the same

NitroxClient/GameLogic/InitialSync/PdaInitialSyncProcessor.cs
```diff
@@ -63,7 +66,10 @@ private void SetPDAEntryComplete(List<TechTypeModel> pdaEntryComplete)
 
             foreach (TechTypeModel item in pdaEntryComplete)
             {
-                complete.Add(item.Enum());
+                if (null != item)
+                {
+                    complete.Add(item.Enum());
+                }
```
dartasen, line 72: Minor: 
- Same here use a `?`
- Invert item and null also

NitroxClient/GameLogic/InitialSync/PdaInitialSyncProcessor.cs
```diff
@@ -76,7 +82,11 @@ private void SetPDAEntryPartial(List<PDAEntry> entries)
 
             foreach (PDAEntry entry in entries)
             {
-                partial.Add(new PDAScanner.Entry { progress = entry.Progress, techType = entry.TechType.Enum(), unlocked = entry.Unlocked });
+                if( null != entry)
+                {
+                    partial.Add(new PDAScanner.Entry { progress = entry.Progress, techType = entry.TechType.Enum(), unlocked = entry.Unlocked });
+                }
+
```
dartasen, line 89: Minor: Same here use a `?`

NitroxClient/GameLogic/InitialSync/PdaInitialSyncProcessor.cs
```diff
@@ -106,7 +119,7 @@ private void SetPDALog(List<PDALogEntry> logEntries)
 
                 foreach (PDALogEntry logEntry in logEntries)
                 {
-                    if (!entries.ContainsKey(logEntry.Key))
+                    if (null != logEntry && !entries.ContainsKey(logEntry.Key))
```
dartasen, line 122: Minor: invert null and logEntry

griphos:
Hi,
The issue happened to me when i moved from the nitrox.rux.gg featured version to the head version. I had several pda partial entries which where received null on client side ( thus the fix ) causing the SetPDAEntryPartial method to crash with no logs ( no null object exception ).
Thus hanging the PdaInitialSyncProcessor as it does not yield results.

I added this check in the method and widen to the other method of the PdaInitialSyncProcessor  class has i thought it could be a good idea.

To give you an idea of the data involved

On server side: WorldPersistence.LoadFromFile
'
persistedData.WorldData.GameData.PDAState.PartiallyUnlockedByTechType

keys[8] {{BaseWaterParkFragment}, {BaseWaterPark}, {StasisRifleFragment}, {DiamondBlade}, {CyclopsEngineFragment}, {BaseFiltrationMachine}, {FiltrationMachine},{BaseFiltrationMachineFragment}} 

Values[8]{null,null,null,null,null,null,null,null}
'

On client side PdaInitialSyncProcessor .SetPDAEntryPartial 
'entries[8] {null,null,null,null,null,null,null,null }'

Maybe in the LoadFromFileMethod it's the intended behaviour to have null as value for the PartiallyUnlockedByTechType dictionnary and i should dive deeper server side. But if it's not some self healing or sanitization could be implemented.

For the "invert null and logEntry" sorry an old habit of mine from my old C++ days. This prevent miss assignation in case you type to fast and miss the '!'. If you type logentry != null it's ok, but if you type logentry = null (as a typo ) the compiler won't catch the error you have an assignation in the test. So the test is always true or false ( depends of the value of the assignation ).

Regards

griphos:
On server side in the 
PDAStateData.GetInitialPDAData

An InitialPDAData object is created with PartiallyUnlockedByTechType.Values thus why i get null value on client side in my case.

griphos:
Deserialization mismatch
1.2.0.1
[https://imgur.com/a/m5hxKsw](url)

Head
[https://imgur.com/a/UanV6ym](url)

griphos:
As the data is recoverable from version 1 maybe a migration routine could be implemented.
Could be an external tool or directly inside the server.
Or maybe at this stage of development it's not required and a simple sanitization could be used.
If a PartiallyUnlockedByTechType entry contains a null value then it's removed from the PartiallyUnlockedByTechType  dictionnary.

Or simply increment the WorldData.Version and the users should start from scratch

Your call

dartasen:
> 
> 
> As the data is recoverable from version 1 maybe a migration routine could be implemented.
> Could be an external tool or directly inside the server.
> Or maybe at this stage of development it's not required and a simple sanitization could be used.
> If a PartiallyUnlockedByTechType entry contains a null value then it's removed from the PartiallyUnlockedByTechType dictionnary.
> 
> Or simply increment the WorldData.Version and the users should start from scratch
> 
> Your call

Alright, got it, Imo a migration routine for a single field at this stage of development isn't really needy, we might need to add null check like you did but instead of "silent failure" logging the error so the user is warned. We also might need to increment the WorldData.Version indeed

griphos:
If you increment WorldData.Version it will prevent old data from loading and create a brand new world. Thus a null check is not necessary.
I don't think a head client could safely connect to a 1.2.0.1 server so a null check won't be of use either.
The only good thing could be a sanitization but as the migration routine, is could not be worth.
Tell me and i change my pull request and only increment the WorldData.Version.

dartasen:
@griphos Yes we might need to increment WorldData.Version, Maybe customize the error message related to version which is "{file} is too old" will be a good idea too

NitroxClient/GameLogic/InitialSync/PdaInitialSyncProcessor.cs
```diff
@@ -52,7 +52,10 @@ private void SetEncyclopediaEntry(List<string> entries)
             {
                 foreach (string entry in entries)
                 {
-                    PDAEncyclopedia.Add(entry, false);
+                    if( null != entry)
```
0xF6, line 55:
```suggestion
                    if(!string.IsNullOrEmpty(entry))
```

dartasen:
Thank you for your interest in this mod, however thi pull request is on standby for almost a month now without news, I'm closing this one.

PR #1047, opened 2020-04-19
Command rework

dartasen:
- Moved Alias/Parameters logic into proper methods
- Argument managing is now encapsulated into proper methods, we can't access anymore to the original splitted list of arguments while performing a command
- Argument check has been removed from class command, the process is now done throught the new Parameter specification on command creation
- Improved readability for /help command IG
- Added more feedbacks while using commands to be more user friendly
- Added informations over autosave while server is starting
- Added Optional boolean on `/save` command to manage periodic saving (persisted in ServerConfig)
- Added `/changerservergamemode {gamemode}` command
- Added `/whois {name}` command to get stats/informations of a player
- `/time` command can now be used without parameter, it will just skip time (day ->night or night->day)


Measurity:
Make sure to stick to the code style for method names. If you're using an IDE it should already point this out with a squiggly line below the names that don't match code style.

NitroxServer/ConsoleCommands/Abstract/Command.cs
```diff
@@ -65,7 +166,25 @@ public void SendMessageToPlayer(Optional<Player> player, string message)
...
+            else
+            {
+                Log.Info(message);
+            }
+        }
+
+        /// <summary>
+        ///     Send a message to both console and an existing player
+        /// </summary>
+        public void SendMessageToBoth(Optional<Player> player, string message)
```
Sunrunner37, line 187: "Both" is ambiguous without reading the summary.  when originally reading the code I was thinking this was a message sent to both players.  I'm not sure that we need to call out that this writes to the log (as we do that throughout).  It is likely fine to remove the else from the method above and allow all to be written to info.

NitroxServer/Exceptions/IllegalArgumentException.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System;
+using System.Runtime.Serialization;
+
+namespace NitroxServer.Exceptions
+{
+
+    [Serializable]
+    internal class IllegalArgumentException : Exception
```
Sunrunner37, line 8: We use ArgumentException and ArgumentNullException throughout.  You can use Validate class as a shortcut to these as well.

NitroxServer/ConsoleCommands/Abstract/Command.cs
```diff
@@ -1,62 +1,163 @@
...
+            catch (IllegalArgumentException e)
+            {
+                SendMessage(sender, $"Error: {e.Message}");
+            }
+            catch (Exception e)
+            {
+                Log.Error("Fatal error while trying to execute the command", e);
+            }
+
+            Args = null;
```
Sunrunner37, line 69: Do we need to make this class stateful?  Can this be passed into Execute instead?

NitroxServer/ConsoleCommands/Abstract/Type/TypeBoolean.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        private static readonly IList<string> yesValues = new List<string>
+        {
+            bool.TrueString,
+            "yes",
+            "on"
+        };
+
+        object IParameter<object>.DefaultValue => null;
+        object IParameter<object>.Read(string arg) => Read(arg);
```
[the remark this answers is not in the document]
dartasen: Indeed `Read` is already defined in the abstract class but to be added in our generic list of parameters, we need bool, float, int to inherit from `Parameter<object>`, Multiple inheritance isn't allowed in C# so we need to introduce a covariant interface

NitroxServer/ConsoleCommands/Abstract/Command.cs
dartasen: Yes It used to be passed into Execute() but I decided in this PR to remove the ability to iterate on the original string[] produced by the Processor

NitroxServer/ConsoleCommands/Abstract/Parameter.cs
```diff
@@ -0,0 +1,37 @@
...
+        public abstract bool IsValid(string arg);
+        public abstract T Read(string arg);
+
+        public override string ToString()
+        {
+            return $"{(IsRequired ? '{' : '[')}{Name}{(IsRequired ? '}' : ']')}";
+        }
+    }
+
+    public interface IParameter<out T>
```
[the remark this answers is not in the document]
dartasen: Any idea ?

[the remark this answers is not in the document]
Measurity: Yes we need it to make it covariant:
image

NitroxServer/ConsoleCommands/Abstract/Type/TypeBoolean.cs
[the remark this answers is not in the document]
Measurity: It's needed because it needs to box value-types like `int` and `bool` to be a reference so that it can be stored in the `List<IParameter<object>>` variable. However, it's not needed for types that already have `object` as a base type.

NitroxServer/ConsoleCommands/Abstract/Command.cs
Measurity: I'm working on refactoring this to use `CallArgs` subclass that will be the single argument passed to `Command.Execute` containing all the properties (`Sender` and `Args`) with `ReadAt(int)` methods.

Measurity: Done

PR #1049, opened 2020-04-20
Handled canvas/camera NRE

Measurity:


NitroxClient/MonoBehaviours/Gui/HUD/RemotePlayerVitals.cs
```diff
@@ -76,7 +76,11 @@ public void LateUpdate()
             waterBar?.UpdateVisual();
 
             // Make canvas face camera.
-            canvas.transform.forward = Camera.main.transform.forward;
+            Camera camera = Camera.main;
```
Jannify, line 79: Maybe move this outside the loop? Camera.main is a little resource heavy as far as I know...

Measurity: True, but the camera changes so can't really do this effectively. Already tried to do this.

PR #1064, opened 2020-04-24
Fixed NRE in ping rename

Measurity:


NitroxPatcher/Patches/Dynamic/PingManager_NotifyRename_Patch.cs
```diff
@@ -16,11 +19,20 @@ public class PingManager_NotifyRename_Patch : NitroxPatch, IDynamicPatch
...
             {
                 return;
             }
-            NitroxServiceLocator.LocateService<IPacketSender>().Send(new PingRenamed(NitroxEntity.GetId(instance.gameObject), instance.GetLabel(), SerializationHelper.GetBytes(instance.gameObject)));
+            NitroxId nitroxId = NitroxEntity.GetId(instance.gameObject);
+            if (nitroxId == null)
+            {
+                Log.Error($"Missing Nitrox id on beacon: {instance.GetHierarchyPath()}");
+                return;
+            }
```
Sunrunner37, line 32: I don't think this is possible.  The function above will always return an id - if one is not set, it will set one,

Measurity: Might be good to keep it in debug mode then?

Sunrunner37: I think the assumption that this always returns a value is throughout the code base.

Measurity: Yeah just thought the same. If it's not working it will fail everywhere. Fixed.

PR #1065, opened 2020-04-25
Fix cyclops initsync

Measurity:
Tested that the lights keep their on/off state when rejoining.

### DIFF
--- a/NitroxClient/GameLogic/Vehicles.cs
+++ b/NitroxClient/GameLogic/Vehicles.cs
@@ -434,11 +434,7 @@ namespace NitroxClient.GameLogic
         {
             VehicleModel vehicle;
             vehiclesById.TryGetValue(vehicleId, out vehicle);
-            if (vehicle == null)
-            {
-                return Optional.Empty;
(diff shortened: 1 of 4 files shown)


NitroxClient/GameLogic/Vehicles.cs
```diff
@@ -428,5 +429,16 @@ public bool RemoveVehicle(NitroxId id)
...
+
+        public Optional<T> TryGetVehicle<T>(NitroxId vehicleId) where T : VehicleModel
+        {
+            VehicleModel vehicle;
+            vehiclesById.TryGetValue(vehicleId, out vehicle);
+            if (vehicle == null)
+            {
+                return Optional.Empty;
+            }
+            return Optional.Of((T)vehicle);
```
Sunrunner37, line 441: You can do OfNullable

Measurity: I wanted to get the cast exception in case it's not null but wrong/unexpected type so I have to do checks seperately.

PR #1066, opened 2020-04-25
Suppress volatile packets while game is shutting down

Measurity:
TODO: 
 - Find a good way to detect all cases when game is shutting down or player is disconnecting.
 - Swap packet sender handler to non-volatile.


NitroxClient/Communication/MultiplayerSession/MultiplayerSessionManager.cs
```diff
@@ -97,13 +98,7 @@ public void Disconnect()
 
         public bool Send(Packet packet)
         {
-            Type packetType = packet.GetType();
-            if (!suppressedPacketsTypes.Contains(packetType))
-            {
-                Client.Send(packet);
```
acidscorch, line 103: Hi @Measurity,

wouldn't it be less complex, to just check here the current connectionstate and supress any client.send that is not related to authentification and connectionhandling? In case of getting disconnected, just stop sending any other packages.

Measurity: Not sure if we can make it that simple because we still need to differentiate from connection handling packet and "game logic" packets. This PR is kinda on hold untill we can find a good way to intercept alt+f4 command before the game tries to destory its objects.

acidscorch: Okay, i understand. I had something in mind like:
if( (connected && notclosing) || (packet is "communicationtype") )
{ 
  client.send 
}

But i also have no idea on how to get triggered before an alt+f4 exit of the game .

Jannify: Maybe we can use `OnApplicationQuit`. If we find out the script execution order from uwe we could execute some code before the other `OnApplicationQuit` and `OnDestroy`.  Just some thoughts, no testing or research.

acidscorch: Maybe this can help, but haven't tested if also triggered by alt+f4:

https://docs.unity3d.com/ScriptReference/Application-wantsToQuit.html

UnityEngine.Application.wantsToQuit += ...
Is described as this is executed before quit and may even be used to cancel quitting.

UnityEngine.Application.quitting += ...
Is described as the final quit event. But maybe this is to late and objects are already not available anymore.

acidscorch: @Measurity 
I didn't understand why this is needed at all. Shouldn't there be no PacketSends when the application closes? In which cases are packets sent that are not wanted? If it only happens to a specific type of Packets maybe it is better to handle it there. 

Besides this, for the current base building pr i also needed a info if the "current state is in initialsyncing" and forwarded the info if initalsyncing to the building class. Maybe it is in general better to extend the MultiplayerSessionConnectionStage with "Initialsyncing" and "ApplicationClosing" (maybe later also "Reconnecting"), so the classes themself can decide wether they are sending packets in this state or not, or even ignore it at all if it is not nessecary to decide.

Measurity: I did it like this to be on the safe side. There might be cases where some packets are being send that are important like dropping items or chat messages.

Those events won't work. Other stuff is called before our handler is invoked. Might be able to add to the 'wantsToQuit' before other code but haven't tried that.

acidscorch: Just checked it and yes it works. Have to hit alt+f4 twice to close subnautica:

image

acidscorch: Could you please give me a ping when you have discussed where to handle this event? And eventually provide an nitrox internal eventhandler to hang on. 

I need it also in the building class, because i need to destroy all base objects before leaving subnautica. Otherwise they get saved in the PreloadCache which causes the "PreloadingBase" to get longer and longer with every multiplayer session, because we always build the entire bases again and again on every initialsync. (I don't know if this can be suppressed elsewhere.)

Measurity: Just checked it and yes it works. Have to hit alt+f4 three times to close Subnautica.

Is the world still functioning when you hit alt+f4 1 time and you block it? In my case some objects are actually destroyed. Even though the game won't quit.

acidscorch: After first hitting alt+f4 the UI overlay is gone and i can't access pda or use any tools anymore. Base objects are still there and i can move around or go outside and swim. 

But i assume keeping subnautica running after this event wasn't the goal, just to have a trigger to do some cleaning up or suppress packages. 

Maybe it also works better when our listener is attached earlier or via a persistent patch. My test with the building class generates the listener very late in the start process and events are triggered in the order they are attached to the listener. It seems that other functions also use this trigger to do their cleaning ups before closing the game, but are registered earlier to the eventqueue.

acidscorch: Ah i understand, the OnDestroyEvent for the objects comes before this event and that's your problem.

acidscorch: If attaching the listener earlier (maybe on uGUI.Awake) even doesn't help, we could as a last resort patch the UnityEngine itself and do what we need to do, before every other listener gets executed:

image

PR #1073, opened 2020-04-27
Add hardcore perma death

Coding-Hen:
Adds perma death to any players who die during the game running on hardcore mode
Adds a message on player connect to reject any player who has previously died

### DIFF
--- a/NitroxServer/Communication/Packets/Processors/PlayerDeathEventProcessor.cs
+++ b/NitroxServer/Communication/Packets/Processors/PlayerDeathEventProcessor.cs
@@ -1,0 +1,29 @@
+﻿using NitroxModel.Packets;
+using NitroxModel.Server;
+using NitroxServer.Communication.Packets.Processors.Abstract;
+using NitroxServer.GameLogic;
+
+namespace NitroxServer.Communication.Packets.Processors
+{
+    class PlayerDeathEventProcessor : AuthenticatedPacketProcessor<PlayerDeathEvent>
+    {
+        private readonly PlayerManager playerManager;
+        private readonly ServerConfig serverConfig;
+        public PlayerDeathEventProcessor(PlayerManager playerManager, ServerConfig serverConfig)
+        {
+            this.playerManager = playerManager;
+            this.serverConfig = serverConfig;
+        }
+
+        public override void Process(PlayerDeathEvent packet, Player player)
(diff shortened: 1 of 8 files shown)


NitroxServer/Communication/Packets/Processors/PlayerDeathEventProcessor.cs
```diff
@@ -0,0 +1,29 @@
...
+        }
+
+        public override void Process(PlayerDeathEvent packet, Player player)
+        {
+            if(serverConfig.GameMode == "Hardcore")
+            {
+                player.IsPermaDeath = true;
+                PlayerKicked playerKicked = new PlayerKicked("Permanent death from hardcore mode");
+                player.SendPacket(playerKicked);
+            }
```
Sunrunner37, line 25: Should we send a disconnect packet to other players?

Coding-Hen: If we do that then it will kick all the other players though? 
My thoughts were that we only kick the player that has died permanently

NitroxServer/Communication/Packets/Processors/PlayerDeathEventProcessor.cs
```diff
@@ -0,0 +1,29 @@
...
+        private readonly ServerConfig serverConfig;
+        public PlayerDeathEventProcessor(PlayerManager playerManager, ServerConfig serverConfig)
+        {
+            this.playerManager = playerManager;
+            this.serverConfig = serverConfig;
+        }
+
+        public override void Process(PlayerDeathEvent packet, Player player)
+        {
+            if(serverConfig.GameMode == "Hardcore")
```
dartasen, line 20: Maybe we should modify `ServerConfig` in order to add a way to manipulate `ServerGameMode` enum instead of an hardcoded value

NitroxServer/Communication/Packets/Processors/PlayerDeathEventProcessor.cs
Sunrunner37: I wasn't suggesting to kick the other players.  I was saying to let players know that the player was kicked.

Though on second glance, this should happen anyways as PlayerKicked will send a disconnect automatically.

PR #1077, opened 2020-05-03
Added patch to remove feedback to developers

Coding-Hen:
Not sure a persistent patch is the best way to do this but the method is called before our dynamic patching is called


Sunrunner37:
I'm fine making this a persistent patch.  Even if we allowed both SP and MP we are still running under the context of Nitrox so may impact SP experience unintentionally.

PR #1084, opened 2020-05-05
Added a popup if there is a new github latest release

dartasen:


Coding-Hen:
I think great idea. My only concern is if we have updated the release on GitHub but hasn't been uploaded to the website. Maybe worth talking with Rux and seeing if we can API the website instead for an update?

Measurity:
A delay might also help so that release can be on website. Then we don't need to API the website.

PR #1085, opened 2020-05-06
Implement remaining hardcore mode

Coding-Hen:


NitroxServer/ConsoleCommands/ChangeServerGamemodeCommand.cs
```diff
@@ -18,9 +18,23 @@ public ChangeServerGamemodeCommand(ServerConfig serverConfig) : base("changeserv
...
         {
             ServerGameMode sgm = args.Get<ServerGameMode>(0);
+            bool currentGamemodeHardcore = serverConfig.IsGamemodeHardcore();
 
             serverConfig.GameModeEnum = sgm;
             SendMessage(args.Sender, $"Server gamemode changed to \"{sgm}\" by {args.SenderName}");
+            
+            if (serverConfig.IsGamemodeHardcore())
+            {
+                serverConfig.DisableAutoSave = true;
```
Sunrunner37, line 28: Why are we disabling auto save for hardcore?

Coding-Hen: I wasn't sure if to do this or not. In the base game using the save menu isn't allowed just the save and quit. Happy to remove

Jannify: As far as I remember vanilla SN "saves" when you enter a base and that isn't handled by us.

NitroxModel/Server/ServerConfig.cs
```diff
@@ -182,5 +182,14 @@ public ServerGameMode GameModeEnum
 
         public PlayerStatsData DefaultPlayerStats => new PlayerStatsData(oxygenSetting.Value, maxOxygenSetting.Value, healthSetting.Value, foodSetting.Value, waterSetting.Value, infectionSetting.Value);
         #endregion
+
+        public bool IsGamemodeHardcore()
```
Jannify, line 186: Minor/Nitpicking: For naming consistency I would prefer `IsGameModeHardcore`.

NitroxModel/Server/ServerConfig.cs
```diff
@@ -182,5 +182,14 @@ public ServerGameMode GameModeEnum
...
+        {
+            if(GameMode == "Hardcore")
```
Jannify, line 188: Minor: No if statement required if returning the outcome.

NitroxServer/ConsoleCommands/ChangeServerGamemodeCommand.cs
Coding-Hen: Taken from the subnautica game modes "In addition, you cannot save the game anytime like in the other modes; the only option is to Save & Quit" so I assume that in hardmode mode it doesn't save inside the base

Sunrunner37: I think there will always be many ways to circumvent this.  It doesn't look like we are suppressing the save command.  Also, I think this would more likely cause an issue if something crashed.  Would prefer to keep it and just blacklist the dead players.

NitroxServer/Communication/Packets/Processors/PlayerDeathEventProcessor.cs
```diff
@@ -17,9 +17,10 @@ public PlayerDeathEventProcessor(PlayerManager playerManager, ServerConfig serve
 
         public override void Process(PlayerDeathEvent packet, Player player)
         {
-            if(serverConfig.GameMode == "Hardcore")
+            if(serverConfig.IsGameMode("Hardcore"))
             {
                 player.IsPermaDeath = true;
+                playerManager.UpdatePlayer(player);
```
Sunrunner37, line 23: Is this required?  I think the player object already comes from PlayerManager: https://github.com/SubnauticaNitrox/Nitrox/blob/master/NitroxServer/Communication/Packets/PacketHandler.cs#L27

Coding-Hen: Possibly not but I found if permadeath was updated on the player to ispermadeath = true then it was never being persistantly set as the playermanager list was never updated

Sunrunner37: I think there might be another bug at play here then.  We have other packet processors that update the player object.  So if the persisted information is using a different reference then we should track down what is going on.  Can you test once again to verify that this is the case?  I spot checked the code and the reference looks to be the same.

Coding-Hen: Recheck you are correct this was due to earlier when the player death wasnt being handled properly

PR #1086, opened 2020-05-07
Added UnityEngine.Object alive check to Optional indirectly

Measurity:
NitroxPatcher now adds a new HasValue check at runtime for types of UnityEngine.Object to check if the object is alive before returning true.

Every time a new type is given to `Optional<TheType>` then it'll loop through the registered global conditions for HasValue and fetch all conditions that are assignable to the given type. This means that `Optional<GameObject>` will also have (at least) the checks that `Optional<UnityEngine.Object>` has.

Atm it won't refetch and cache the hasvalue conditions for a type if they change due to `Optional.AddHasValueCondition<T>()` calls. These calls should be done before the first Optional<T> is created where T is a type assignable to any condition types. As done here: https://github.com/SubnauticaNitrox/Nitrox/pull/1086/files#diff-4e5e5240065741957a35125a1975652eR30

### DIFF
--- a/NitroxModel/DataStructures/Optional.cs
+++ b/NitroxModel/DataStructures/Optional.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Runtime.Serialization;
 using System.Security.Permissions;
 using ProtoBufNet;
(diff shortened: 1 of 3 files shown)


NitroxModel/DataStructures/Optional.cs
```diff
@@ -137,16 +169,20 @@ public struct OptionalEmpty
...
-        public static Optional<T> OfNullable<T>(T value)
+        /// <summary>
+        ///     Adds a condition to the optional of the given type that is checked whenever <see cref="Optional{T}.HasValue" /> is checked.
+        /// </summary>
+        /// <param name="hasValueCondition">Condition to add to the <see cref="Optional{T}.HasValue"/> check.</param>
+        /// <param arg="T">Type that should have the extra condition. The given type will also apply to more specific types than itself.</param>
+        public static void ApplyHasValueCondition<T>(Func<T, bool> hasValueCondition) where T : class
         {
-            return Optional<T>.OfNullable(value);
+            ValueConditions.Add(typeof(T), o => hasValueCondition(o as T));
```
Sunrunner37, line 185: Is it possible to move the type looping into this method?  Or is it done at runtime due to later dll loading?

Measurity: It's done lazily so that any more specific types (say `Object` and then `GameObject`) can be added before the checks are set in 'stone'. Doing it here would mean that other calls to `ValueConditions.Add` need to invalidate the checks cache for all optionals everytime or I need to add a "add multiple" method and do it immediatly after that call.

PR #1087, opened 2020-05-08
Fixed multiplayer vehicle synchronization failing if a valid docking bay could not be found

looterz:
This would lead to an infinite loading screen whenever a vehicle with a dockingBayId was unable to find or create the dockingBayBase or dockingBay. Fixed NitroxServer-Subnautica missing the LitJSON dependency as well.

### DIFF
--- a/NitroxClient/GameLogic/InitialSync/CyclopsInitialAsyncProcessor.cs
+++ b/NitroxClient/GameLogic/InitialSync/CyclopsInitialAsyncProcessor.cs
@@ -24,7 +24,7 @@ namespace NitroxClient.GameLogic.InitialSync
         {
             this.waitScreenItem = waitScreenItem;
 
-            Log.Info($"Starting cyclops vehicle sync");
+            Log.Debug($"Starting cyclops vehicle sync");
 
             vehicles.VehicleCreated += OnVehicleCreated;
 
@@ -34,7 +34,7 @@ namespace NitroxClient.GameLogic.InitialSync
                 {
                     cyclopsStillLoading++;
 
-                    Log.Info($"Synchronizing cyclops vehicle {vehicle.Id} {vehicle.Name}");
+                    Log.Debug($"Synchronizing cyclops vehicle {vehicle.Id} {vehicle.Name}");
 
                     vehicles.CreateVehicle(vehicle);
                 }
--- a/NitroxClient/GameLogic/InitialSync/CyclopsInitialAsyncProcessor.cs
+++ b/NitroxClient/GameLogic/InitialSync/CyclopsInitialAsyncProcessor.cs
@@ -24,8 +24,6 @@ namespace NitroxClient.GameLogic.InitialSync
         {
             this.waitScreenItem = waitScreenItem;
 
-            Log.Debug($"Starting cyclops vehicle sync");
-
             vehicles.VehicleCreated += OnVehicleCreated;
 
             foreach (VehicleModel vehicle in packet.Vehicles)
@@ -42,7 +40,7 @@ namespace NitroxClient.GameLogic.InitialSync
 
             totalCyclopsToLoad = cyclopsStillLoading;
 
-            yield return new WaitUntil(() => cyclopsStillLoading <= 0);
+            yield return new WaitUntil(() => cyclopsStillLoading == 0);
         }
 
         private void OnVehicleCreated(GameObject gameObject)
@@ -52,7 +50,7 @@ namespace NitroxClient.GameLogic.InitialSync
             cyclopsStillLoading--;
 
             // After all cyclops are created
-            if (cyclopsStillLoading <= 0)
+            if (cyclopsStillLoading == 0)
             {
                 vehicles.VehicleCreated -= OnVehicleCreated;
             }            
--- a/NitroxClient/GameLogic/InitialSync/CyclopsInitialAsyncProcessor.cs
+++ b/NitroxClient/GameLogic/InitialSync/CyclopsInitialAsyncProcessor.cs
@@ -1,6 +1,7 @@
 ﻿using System.Collections;
 using NitroxClient.GameLogic.InitialSync.Base;
 using NitroxModel.DataStructures.GameLogic;
+using NitroxModel.Logger;
 using NitroxModel.Packets;
 using NitroxModel_Subnautica.Helper;
 using UnityEngine;
@@ -23,6 +24,8 @@ namespace NitroxClient.GameLogic.InitialSync
         {
             this.waitScreenItem = waitScreenItem;
 
+            Log.Info($"Starting cyclops vehicle sync");
+
             vehicles.VehicleCreated += OnVehicleCreated;
 
             foreach (VehicleModel vehicle in packet.Vehicles)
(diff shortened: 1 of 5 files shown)


dartasen:
Hi @looterz, thank you for your interest !

However I don't know if it's the right fix for that problem.
Indeed this avoid silent failure and gives more logs about the failure, but at the end, loading screen will stay stuck on the VehicleSyncProcessor because we didn't reach the event broadcasting of VehicleCreated at the end of the `OnVehiclePrefabLoaded`, so the wait screen will loop forever waiting until his vehicleloading counter drops to 0

Best regards

looterz:
> 
> 
> Hi @looterz, thank you for your interest !
> 
> However I don't know if it's the right fix for that problem.
> Indeed this avoid silent failure and gives more logs about the failure, but at the end, loading screen will stay stuck on the VehicleSyncProcessor because we didn't reach the event broadcasting of VehicleCreated at the end of the `OnVehiclePrefabLoaded`, so the wait screen will loop forever waiting until his vehicleloading counter drops to 0
> 
> Best regards

This commit fixed the issue for our clients encountering infinite loading screen issue due to a SEAMOTH vehicle that was associated with a docking bay that didn't exist or couldn't be found at runtime during sync. What you are saying seems to be the case for `CyclopsInitialAsyncProcessor` but not for `VehicleInitialSyncProcessor`, only the Cyclops listens for `OnVehicleCreated` and adjusts its progress based on that hook, VehicleInitialSync simply waits for CreateVehicle to return and continues.

NitroxClient/GameLogic/Vehicles.cs
```diff
@@ -163,28 +213,44 @@ private void OnVehiclePrefabLoaded(TechType techType, GameObject prefab, NitroxI
...
+
+                    //Log.Debug($"Updated vehicle colors for {id}");
                 }
 
                 vehicle.GetComponent<LiveMixin>().health = health;
             }
             else if(techType == TechType.Cyclops)
             {
                 GameObject target = NitroxEntity.RequireObjectFrom(id);
+                if (target == null)
```
Measurity, line 242: Any object that inherits from `UnityEngine.Object` should not be compared against `== null` but instead just do `if (obj)` so that the implicit bool operator that Unity defines for these objects is triggered.

This is required because Unity objects can be non-null but dead. Which isn't good either.

NitroxClient/GameLogic/InitialSync/CyclopsInitialAsyncProcessor.cs
```diff
@@ -46,7 +52,7 @@ private void OnVehicleCreated(GameObject gameObject)
             cyclopsStillLoading--;
 
             // After all cyclops are created
-            if (cyclopsStillLoading == 0)
+            if (cyclopsStillLoading <= 0)
```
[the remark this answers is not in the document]
looterz: I did observe this going under 0, but it turned out to be unrelated to the vehicle sync issue I could reproduce, so I will revert this change.

NitroxClient/GameLogic/Vehicles.cs
```diff
@@ -137,23 +137,59 @@ private void OnVehiclePrefabLoaded(TechType techType, GameObject prefab, NitroxI
         {
             // Partially copied from SubConsoleCommand.OnSubPrefabLoaded
             GameObject gameObject = Utils.SpawnPrefabAt(prefab, null, spawnPosition);
+
+            if (gameObject == null)
```
Measurity, line 141:
```suggestion
            if (!gameObject)
```

NitroxClient/GameLogic/Vehicles.cs
```diff
@@ -137,23 +137,59 @@ private void OnVehiclePrefabLoaded(TechType techType, GameObject prefab, NitroxI
...
+                return;
+            }
+
             gameObject.transform.rotation = spawnRotation;
             gameObject.SetActive(true);
             gameObject.SendMessage("StartConstruction", SendMessageOptions.DontRequireReceiver);
             CrafterLogic.NotifyCraftEnd(gameObject, CraftData.GetTechType(gameObject));
+            
             Rigidbody rigidBody = gameObject.GetComponent<Rigidbody>();
+            if (rigidBody == null)
```
Measurity, line 153:
```suggestion
            if (!rigidBody)
```

NitroxClient/GameLogic/Vehicles.cs
```diff
@@ -200,7 +244,9 @@ private void OnVehiclePrefabLoaded(TechType techType, GameObject prefab, NitroxI
             }
 
             VehicleChildObjectIdentifierHelper.SetInteractiveChildrenIds(gameObject, interactiveChildIdentifiers); //Copy From ConstructorBeginCraftingProcessor
-                  
+            
+            Log.Debug($"Successfully created vehicle {id}");
```
Measurity, line 248:
```suggestion
```

Success debug logs don't add much value. Only add logs that tell us something is going wrong.

NitroxClient/GameLogic/Vehicles.cs
```diff
@@ -137,23 +137,59 @@ private void OnVehiclePrefabLoaded(TechType techType, GameObject prefab, NitroxI
...
+                if (!vehicle)
+                {
+                    Log.Error($"Failed to get vehicle component from prefab {prefab.ToString()} for {id}");
+                    return;
+                }
+
                 if (dockingBayId.HasValue)
                 {
                     GameObject dockingBayBase = NitroxEntity.RequireObjectFrom(dockingBayId.Value);
+                    if (!dockingBayBase)
```
Sunrunner37, line 175: Some of these are using RequireObjectFrom which will throw if the object doesn't exist.

NitroxClient/GameLogic/Vehicles.cs
```diff
@@ -182,9 +219,16 @@ private void OnVehiclePrefabLoaded(TechType techType, GameObject prefab, NitroxI
             else if(techType == TechType.Cyclops)
             {
                 GameObject target = NitroxEntity.RequireObjectFrom(id);
+                if (!target)
```
Sunrunner37, line 222: Same as above

acidscorch:
Hi @looterz @dartasen @Sunrunner37,

i'm not familiar with the vehicle classes. Could it be, that this bug is a follow-up problem of not correct saving and syncing Base IDs and Moonpool IDs between clients? If yes, this may be tested/reviewed again later. I'm currently working on Base-Building Sync. 

Best regards

Sunrunner37:
Yes Acid, I agree that the root cause is that the base id is not found.  Let's pause on this to see if updating the moonpool id resolves this issue.

dartasen:
I think we can close this PR, as long as it's not really fixing anything, just logging more informations about failures.
But indeed the bug is a follow-up problem of wrong moonpool ID (ID changed or wrong ID)

acidscorch:
Hi @looterz,

if you want, you can test my pr #1093.
This should fix loosing the ID of base objects including moonpool. So the vehicles should be properly assigned. 

Best regards

looterz:
> 
> 
> Hi @looterz,
> 
> if you want, you can test my pr #1093.
> This should fix loosing the ID of base objects including moonpool. So the vehicles should be properly assigned.
> 
> Best regards

Awesome, I'll test it as soon as I can. Thanks for the help.

PR #1088, opened 2020-05-09
Added exclusion to Optional<object> to always add all cecks because anything can be in it

Measurity:


NitroxServer/ConsoleCommands/BackCommand.cs
```diff
@@ -13,12 +14,13 @@ public BackCommand() : base("back", Perms.ADMIN, "Teleports you back on your las
         protected override void Execute(CallArgs args)
         {
             Validate.IsTrue(args.Sender.HasValue, "This command can't be used by CONSOLE");
-
             Player player = args.Sender.Value;
-
-            Validate.IsTrue(player.LastStoredPosition.HasValue, "No previous location...");
+            if (player.LastStoredPosition == null)
+            {
+                Log.InGame("No previous location...");
```
killzoms, line 20: As this is running on the server shouldn't it be
```csharp
SendMessage(args.Sender, "No previous location...";
```

dartasen: Yes we might be using SendMessage instead, or using Validate that will display an error message to the sender

Measurity: Preferable not `Validate` since this isn't an (exceptional) error.

PR #1089, opened 2020-05-10
VehicleColor logic update

dartasen:
- Fix a bug where _ExoSuit_ colors where not initialized causing **NRE** at initial syncing and also when trying to change colors inside a moonpool terminal _(Because exosuits SubName colors aren't defined yet for odd reasons)_
- Fix a bug where _Exosuit_ colors wheren't updated in live for players
- Removed Colours field for VehicleModel (Just redundant informations since colors are computed by Subnautica throught HSB field)

### DIFF
--- a/NitroxServer/ConsoleCommands/BackCommand.cs
+++ b/NitroxServer/ConsoleCommands/BackCommand.cs
@@ -15,11 +15,13 @@ namespace NitroxServer.ConsoleCommands
         {
             Validate.IsTrue(args.Sender.HasValue, "This command can't be used by CONSOLE");
             Player player = args.Sender.Value;
+
             if (player.LastStoredPosition == null)
             {
-                Log.InGame("No previous location...");
+                SendMessage(args.Sender, "No previous location...");
                 return;
             }
+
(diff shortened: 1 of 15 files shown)


NitroxServer/ConsoleCommands/BackCommand.cs
```diff
@@ -15,11 +15,13 @@ protected override void Execute(CallArgs args)
         {
             Validate.IsTrue(args.Sender.HasValue, "This command can't be used by CONSOLE");
             Player player = args.Sender.Value;
+
             if (player.LastStoredPosition == null)
             {
-                Log.InGame("No previous location...");
+                SendMessage(args.Sender, "No previous location...");
```
Coding-Hen, line 21: My personal view is this shouldn't be changed as part of this pr as it's irrelevant to the overall theme of the pr

Measurity: I think we can be an exception in this case since it's a one line fix and waiting around for another PR to touch this area is silly.

Issue #1051, opened 2020-04-21
Power management undsynchronised after disconnection reconnection


AshenShugar59:
Hello,

I am playing with 2 friends with you mod and we are catching some serious issue around structures.
Everytime we disconnect / reconnect, items are powered off, and we can't find a way to power on except destroy it and build it again.
The problem is that it is very specific to each user session.

Example :
- We are both on the server and build a fabricator, which works well
- One of us disconnect and reconnect, he will see the fabricator offline, as I will see it online and I can use it.
This happens systematically, not random.

It happens only on structure we build, not on the starting rescue cell.

Server version 1.2.1.0.

Thanks guys

acidscorch:
This issue adresses two themes:
1. Fabricator and other objects after reconnect have no power or are not working: This happens because these objects need to be linked to the underlying Base-Object which gets broken on reconnect (to be correct in detail, in most cases it is already broken while playing and not correctly saved on server). Work on this is already in progress, but this is a very complex theme.
2. General base energy syncing is not implemented yet. A first solution at reconnecting will be, that all the base energy objects (solar, reactor) will be initialized with 50% powerstate. This will result in having different energystates for every player but will make the mod playable and give an anchor point for future energy sync implementation.

Jannify:
Moved to https://github.com/SubnauticaNitrox/Nitrox/issues/1160

Issue #1030, opened 2020-04-10
[Base] Hull integrity bug


Logharaa:
Playing at 3, in the same base we didn't have the same hull integrity. As a result, one player saw the base sinking, while the others did not.

frgbf:
I have the same problem in my game, it looks like that deconstructing structures is not synchronizing properly, which may cause your problem

I found the following trace in the logs
```
[17:11:34 Client ERROR]: Error processing packet: [DeconstructionCompleted Id: a9656f85-984d-4dca-86ff-eb3aeb15c5a7]
NitroxModel.DataStructures.Util.OptionalEmptyException`1[UnityEngine.GameObject]: Optional <T> is empty:
	Game object required from id: a9656f85-984d-4dca-86ff-eb3aeb15c5a7
  at NitroxModel.Helper.Validate.IsPresent[T] (NitroxModel.DataStructures.Util.Optional`1[T] opt, System.String message) [0x0000f] in <b8ae7171f8254f8aaab8bc7d8e9c91e7>:0 
  at NitroxClient.MonoBehaviours.NitroxEntity.RequireObjectFrom (NitroxModel.DataStructures.NitroxId id) [0x00013] in <574153703ef24f60b801a84a2ebd5879>:0 
  at NitroxClient.Communication.Packets.Processors.DeconstructionCompletedProcessor.Process (NitroxModel.Packets.DeconstructionCompleted packet) [0x00006] in <574153703ef24f60b801a84a2ebd5879>:0 
  at NitroxClient.Communication.Packets.Processors.Abstract.ClientPacketProcessor`1[T].ProcessPacket (NitroxModel.Packets.Packet packet, NitroxModel.Packets.Processors.Abstract.IProcessorContext context) [0x00000] in <574153703ef24f60b801a84a2ebd5879>:0 
  at NitroxClient.MonoBehaviours.Multiplayer.ProcessPackets () [0x00040] in <574153703ef24f60b801a84a2ebd5879>:0 
```

tomi1206:
The same things happen for us (2 on a server). But more of that, when reconnection to the server, some base fragment aren't complete anymore, hull integrity can be negative and the base sinking

acidscorch:
This issue adresses 2 themes:
1. On reconnect bases are not restored correctly or restored in the wrong order which causes a funtional Base to break and sink. This is noticed for current rework and Objects will be loaded correctly and in the right order.
[...]

Issue #711, opened 2019-02-03
Inventory items being stolen


unlightednetting:
Version 1.0.3.1 from NexusMods. Fresh world, upgrade instructions of Uninstall Nitrox, Verify, and reinstall were taken.

This was originally going to be about items disappearing, but they are specifically being stolen.

Two player game, just started base building. The materials that are needed for any particular base piece are being taken from BOTH players' inventories.

As well, if the building player does not have every material needed but starts the building anyways, the materials that the builder is missing get taken from the other players' inventory, but the building does not actually finish, thereby sending the materials into the ether.

I have not tested if this is based on distance, and it does not seem to happen to happen to the lockers. As well ,this does not appear to happen reliably/every time for every material needed,, but testing does show it to be only when someone is building.

docquantum:
This is a known bug that's being worked on.

Thanks for the report!

0verdark:
@Measurity The issue has been resolved. You can download the latest **Nitrox Launcher** version which is: **1.2.2.0** from the **Official Website** or from **Nexus Mods.** You're Welcome!

PR #1093, opened 2020-05-11
Refactoring and fixing base building issues [review]

acidscorch:
Test required

❗ Info for testing this pull request: start with a new world, because old server save files are incorrect and cannot be fixed at client side during runtime.

Issue-Tracking:
#1120 (partially)
#1110 (Base related part)
#1051 (Issue key point 1)
#1030 (Issue key point 1)
#951
#691
#685
#711
#668
#675
#133
#132
#129
#127
#118

ToDo Major:

- [x] suppress hull calculation during initialsync
- [x] suppress item consumtion/granting during remote events
- [x] remove MultiplayerBuilder as it is a code duplicate and patch the nitroxneeds at the right place

ToDo Minor (in a later pr):
- sync bulkhead door state (currently implemented that they have state open on rejoin which is a nicer after load state)
- sync hull integrity
- sync energy states and consumption
- block simultanious constructing/deconstructing of same object by local and remote player or deconstructing an object that is currently used by a remote player (e.g. fabricator)


NitroxClient/GameLogic/Building.cs
```diff
@@ -1,213 +1,1312 @@
...
+            NitroxModel.Logger.Log.Debug("Constructable_Construct_Post - instance: " + instance + " tempId: " + tempId + " construced: " + instance._constructed + " amount: " + instance.constructedAmount + " remoteEventActive: " + remoteEventActive);
+#endif
+
+            //Check if we raised the event by using our own BuilderTool or if it came as post Event of a Remote-Action or Init-Action
+            if (lastHoveredConstructable != null && lastHoveredConstructable == instance && currentlyHandlingBuilderTool)
+            {
+
+                if (result && instance.constructedAmount < 1f)
+                {
+                    //Send every event, because recourses used for construction (or gained for deconstruction) are else not properly taken or credited,
```
acidscorch, line 607: to discuss

NitroxClient/GameLogic/Building.cs
```diff
@@ -1,213 +1,1312 @@
...
+            //
+            if (result && instance.constructedAmount == 1f && remoteEventActive)
+            {
+
+#if TRACE && BUILDING
+                NitroxModel.Logger.Log.Debug("Constructable_Construct_Post - finished construct remote");
+#endif
+
+
+                if (instance.gameObject.name.Contains("Solar") || instance.gameObject.name.Contains("Reactor"))
```
acidscorch, line 635: To be changed later

acidscorch:
> We do have some extensions method inside `NitroxModel.Helper.ReflectionHelper` that you can use for that 😀

Hi datarsan, can you tell me for which aspect do you suggest this? I don't get it :confused:

acidscorch:
Hi @all,

can someone tell me, why the github build check gets failed? I also have these UnityEngine warnings on my local environment, but the solution can be build. I have them only as warnings not as errors. 

Best regards

NitroxClient/GameLogic/Building.cs
```diff
@@ -1,213 +1,1326 @@
...
+                            }
+
+                            Transform camera = Camera.main.transform;
+                            Optional<RotationMetadata> rotationMetadata = rotationMetadataFactory.From(ghost);
+
+#if TRACE && BUILDING
+                            NitroxModel.Logger.Log.Debug("Constructable_NotifyConstructedChanged_Post - techType: " + instance.techType + " techType.Model(): " + instance.techType.Model());
+#endif
+
+                            //fix for wrong techType
```
Sunrunner37, line 929: wrong tech type?

NitroxClient/GameLogic/InitialSync/BuildingInitialSyncProcessor.cs
```diff
@@ -74,5 +92,97 @@ private void FinishedCompletedBuildings(object sender, EventArgs eventArgs)
             ThrottledBuilder.main.QueueDrained -= FinishedCompletedBuildings;
             completed = true;
         }
+
+        // Subnautica changes the Layout of Bases when new BasePieces are added or removed. A simple replay of all the saved BasePieces is not enough, because 
+        // deconstructig BasePieces (like Tubes) and building new BasePieces can cause the complete layout to change, so that following Pieces 
+        // (like Hatches) aren't fit right to the corresponding objects. Additionally we need to change the 
+        // The Method is not final. Some BasePiece Relations may be better validated by Face- or Cell-Connections, others still need fixed positons like 
+        // Solarpanels or Furniture. For now the BuildIndex should work. 
+        private List<BasePiece> GenerateBaseLayouts(List<BasePiece> basePieces)
```
Sunrunner37, line 101: Can we refactor this to have an ordered array of base piece types and cross reference that?  I think such an approach will remove the need to have so many loops.

NitroxClient/MonoBehaviours/ThrottledBuilder.cs
```diff
@@ -109,206 +98,29 @@ private void ActionBuildEvent(BuildEvent buildEvent)
...
-            if (opNewlyCreatedBase.HasValue)
-            {
-                GameObject newlyCreatedBase = (GameObject)opNewlyCreatedBase.Value;
-                NitroxEntity.SetNewId(newlyCreatedBase, newBaseId);
-            }
-            else
-            {
-                Log.Error("Could not assign new base id as no newly constructed base was found");
-            }
+            NitroxServiceLocator.LocateService<Building>().Constructable_ConstructionCompleted_Remote(constructionCompleted.PieceId);
```
Sunrunner37, line 108: Let's use dependency injection through the constructor instead for all of these methods.

NitroxClient/MonoBehaviours/ThrottledBuilder.cs
```diff
@@ -109,206 +98,29 @@ private void ActionBuildEvent(BuildEvent buildEvent)
...
-            /**
-             * Manually call start to initialize the object as we may need to interact with it within the same frame.
-             */
-            MethodInfo startCrafting = typeof(Constructable).GetMethod("Start", BindingFlags.NonPublic | BindingFlags.Instance);
-            Validate.NotNull(startCrafting);
-            startCrafting.Invoke(constructable, new object[] { });
+            NitroxServiceLocator.LocateService<Building>().Constructable_ConstructionBegin_Remote(constructionBegin.BasePiece);
         }
 
         private void ConstructionCompleted(ConstructionCompletedEvent constructionCompleted)
```
Sunrunner37, line 106: minor: we don't really need these anymore and can likely just move the one line up.

NitroxModel-Subnautica/DataStructures/GameLogic/Buildings/Rotation/Metadata/BaseFaceRotationMetadata.cs
```diff
@@ -0,0 +1,42 @@
...
+        // Base modules anchor based on a face.  This can be constructed via these two attributes.
+        [ProtoMember(1)]
+        public NitroxModel.DataStructures.Int3 AnchoredFaceCell { get; set; }
+
+        [ProtoMember(2)]
+        public int AnchoredFaceDirection { get; set; }
+
+        [ProtoMember(3)]
+        public int AnchoredFaceType { get; set; }
+
```
Sunrunner37, line 23: minor: if all faces are an anchored face then we may not need to prefix all with anchored.  Same in file below.

NitroxPatcher/Patches/Dynamic/CellManager_RegisterEntity_Patch.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
+using System.Reflection;
+using Harmony;
+using NitroxClient.GameLogic;
+using NitroxModel.Core;
+using UnityEngine;
+
+namespace NitroxPatcher.Patches.Dynamic
+{
+    public class CellManager_RegisterEntity_Patch : NitroxPatch, IDynamicPatch
```
Sunrunner37, line 10: Do we need this?

NitroxClient/GameLogic/Building.cs
```diff
@@ -1,213 +1,1326 @@
...
 using NitroxModel.Helper;
-using NitroxModel.Logger;
 using NitroxModel.Packets;
 using NitroxModel_Subnautica.Helper;
 using UnityEngine;
-using static NitroxClient.GameLogic.Helper.TransientLocalObjectManager;
 
 namespace NitroxClient.GameLogic
 {
     public class Building
```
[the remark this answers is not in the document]
acidscorch: Yes, that makes sense.

NitroxClient/GameLogic/Building.cs
```diff
@@ -1,213 +1,1326 @@
...
+        private string generateGameObjectTransformInfo(Transform transform)
+        {
+            return transform.name + " " + transform.parent.position.ToString() + " " + transform.GetSiblingIndex().ToString();
+        }
+
+        // Before the geometrychange we need to save the ids
+        public void Base_ClearGeometry_Pre(Base @base)
+        {
+            // Base_ClearGeometry_Pre is called twice. One time for the ghost layout and one time for the viewable layout. We are only interested in the viewable layout. 
+            {
```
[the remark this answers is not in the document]
acidscorch: I had another if check at this position which wasn't nessecary in the end. I will remove it.

NitroxClient/GameLogic/Building.cs
```diff
@@ -1,213 +1,1326 @@
...
+
+                foreach (Transform cellObject in cellObjects)
+                {
+                    if (cellObject != null)
+                    {
+                        for (int i = 0; i < cellObject.childCount; i++)
+                        {
+                            Transform surface = cellObject.GetChild(i);
+
+                            if (surface != null && surface.gameObject != null)
```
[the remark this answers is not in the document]
acidscorch: I have just overtaken the code from the current Base_ClearGeometry_Patch and renamed child with surface to make it a little bit more understandable what a child in this case stands for. Can you give me a hint or example what you mean with "implicit boolean operater" ? Does " if (surcafe) {... " work?

NitroxClient/GameLogic/Building.cs
```diff
@@ -1,213 +1,1326 @@
...
+            baseGhostIsFinishing = true;
+        }
+
+        private string generateGameObjectTransformInfo(Transform transform)
+        {
+            return transform.name + " " + transform.parent.position.ToString() + " " + transform.GetSiblingIndex().ToString();
+        }
+
+        // Before the geometrychange we need to save the ids
+        public void Base_ClearGeometry_Pre(Base @base)
```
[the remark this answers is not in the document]
acidscorch: 👍

NitroxClient/GameLogic/Building.cs
```diff
@@ -1,213 +1,1326 @@
...
+                            Transform surface = cellObject.GetChild(i);
+
+                            if (surface != null && surface.gameObject != null)
+                            {
+
+#if TRACE && BUILDING
+                                string info = generateGameObjectTransformInfo(surface);
+                                NitroxModel.Logger.Log.Debug("Base_ClearGeometry_Pre - current clear for : " + info);
+#endif
+                                if (!surface.name.Contains("CorridorConnector"))
```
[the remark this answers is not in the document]
acidscorch: CorridorConnectors have a special tech type and may be identified by these. But this will also enlarge this check, because there are different kinds of CorridorConnector for Hatches, Rooms, Moonpool, etc. The only common identification aspect i found was, that they all contain the string "CorridorConnector" in their names. Using a constant makes sense, i would put this as an internal enum in the BaseGeometry class, because it is just need there.

NitroxClient/GameLogic/Building.cs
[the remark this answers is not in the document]
Measurity: `UnityEngine.Object` has an implicit bool operator that checks if the object is alive. So you can either do `if (gameobjec) { // object is alive }` or `if (component) { // object is alive }`. If you do a null check then the object might be destroyed/dead but not null. Then the code crash when trying to use the dead object.

NitroxClient/GameLogic/InitialSync/BuildingInitialSyncProcessor.cs
acidscorch: Thinking more about this total method makes me unsure, if this can be eventually mostly dropped and only the buildindex and base sequences are needed. 
The first problem were unfinished peaces. These unfinished pieces should be now correctly handled, even if they are not loaded at the end of a base layout. 
The second problem is hull integrity. This will be overcome as soon as i found out how to suppress this on initial sync. My idea herefore is, to sort all items via parentbase and suppress hull integrity calculation except for the last finished piece of this base to let the game refresh the latest state. (For a more later PR: Thinking a little bit more further, the hull integrity sync could also be done in another separate InitialSync after the BuildingInitialsync which also can handle breaches and floot level of the base).
The third problem was energy. This should also be fixed, because alle energy producing and energy consuming objects are now correctly referenced by the corresponding base object and even if the energy generating object has been placed afterwards it should be just working as in normal singleplayer. 
The fourth problem were hatches on watertanks and on rooms for creating new corridors, this should also be no problem if the buildindex is correct. 

For now i would like to let the hullintegrity related sorts as they are and comment the code with the infos given here. 

For the use of an arry and cross referencing, can you give me a hint on that? I don't know exactly how you are meaning this.

NitroxClient/GameLogic/Building.cs
```diff
@@ -1,213 +1,1326 @@
...
+
+            baseGhostIsFinishing = true;
+        }
+
+        private string generateGameObjectTransformInfo(Transform transform)
+        {
+            return transform.name + " " + transform.parent.position.ToString() + " " + transform.GetSiblingIndex().ToString();
+        }
+
+        // Before the geometrychange we need to save the ids
```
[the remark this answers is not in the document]
acidscorch: I will take over the comment in the new BaseGeometry class.

NitroxClient/GameLogic/Building.cs
```diff
@@ -1,213 +1,1326 @@
...
+                                                if (surface.gameObject != null)
+                                                {
+                                                    parentCellContent = parentCellContent + surface.parent.GetChild(j).name;
+                                                }
+                                            }
+                                        }
+
+                                        GeometryInfo geometryInfo = new GeometryInfo() { Name = surface.name, Id = id, CellPosition = surface.parent.position, CellIndex = surface.GetSiblingIndex(), CellContent = parentCellContent };
+                                        idCacheForGeometryUpdate.Add(geometryInfo);
+                                        NitroxEntity.RemoveId(surface.gameObject);
```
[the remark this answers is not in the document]
acidscorch: I have noticed that Objects are not properly automaticaly destroyed by the game as long as there is a NitroxEntity applied to it. Destroying the object on our own also caused some wired effects. So i thought the best way is, to let the gameengine or unityengine do what it wants to do with the object and just revert the changes that nitrox did to the gameobject (adding a NitroxEntity) and so everything works fine. 
You can see this also if you apply a NitroxId to a ghost object. These objects are also not destroyed because the engine seems to perform a check if any component is still attached and so the NitroxEntity blocks this. I don't know if this is the same for all other entitys but for buildings i noticed this behaviour.

NitroxClient/GameLogic/Building.cs
```diff
@@ -1,213 +1,1326 @@
...
+                if (id == null && surface.GetComponent<BaseDeconstructable>() != null && !surface.name.Contains("CorridorConnector"))
+                {
+                    // In case a vertical connector exists and a ladder is build inside the base, then this vertical connector changes its shape and name
+                    if (surface.name.Contains("ConnectorTube"))
+                    {
+                        if (surface.name.Contains("Window"))
+                        {
+                            if (idCacheForGeometryUpdate.Exists(g => g.Name == surface.name.Replace("ConnectorTubeWindow", "ConnectorTube") && g.CellPosition == surface.parent.position && g.CellIndex == surface.GetSiblingIndex()))
+                            {
+                                id = idCacheForGeometryUpdate.Find(g => g.Name == surface.name.Replace("ConnectorTubeWindow", "ConnectorTube") && g.CellPosition == surface.parent.position && g.CellIndex == surface.GetSiblingIndex()).Id;
```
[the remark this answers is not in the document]
acidscorch: Okay

NitroxClient/GameLogic/Building.cs
```diff
@@ -1,213 +1,1326 @@
...
+
+                NitroxId id = NitroxEntity.GetIdNullable(surface.gameObject);
+                if (id == null && surface.GetComponent<BaseDeconstructable>() != null && !surface.name.Contains("CorridorConnector"))
+                {
+                    // In case a vertical connector exists and a ladder is build inside the base, then this vertical connector changes its shape and name
+                    if (surface.name.Contains("ConnectorTube"))
+                    {
+                        if (surface.name.Contains("Window"))
+                        {
+                            if (idCacheForGeometryUpdate.Exists(g => g.Name == surface.name.Replace("ConnectorTubeWindow", "ConnectorTube") && g.CellPosition == surface.parent.position && g.CellIndex == surface.GetSiblingIndex()))
```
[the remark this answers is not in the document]
acidscorch: If you have vertikal connector above a tube this has the name ConnectorTube. When you build a ladder through this connector it changes its name from ConnectorTube to WindowConnectorTube. So to find the gameobject again, we have to search for the new name.

NitroxClient/GameLogic/Building.cs
```diff
@@ -1,213 +1,1326 @@
...
+                        if (currentParentContent.Contains("RoomExteriorBottom") && !cachedInfo.CellContent.Contains("RoomExteriorBottom"))
+                        {
+                            moveFaceIndex = moveFaceIndex - 1;
+                        }
+                        if (currentParentContent.Contains("AdjustableSupport") && !cachedInfo.CellContent.Contains("AdjustableSupport"))
+                        {
+                            moveFaceIndex = moveFaceIndex - 1;
+                        }
+
+                        if (moveFaceIndex != 0)
```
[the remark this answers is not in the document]
acidscorch: I just wanted to separate the applying of a moveFaceIndex in a separate block for separate trace outputs. Though these are not needed anymore here and are not in this pr and only in my develop environment, i can refactor this and always use the same block as a little down lower with additional move index even if it is 0.

NitroxClient/GameLogic/Building.cs
```diff
@@ -1,213 +1,1326 @@
...
+                        {
+                            if (idCacheForGeometryUpdate.Exists(g => g.Name == surface.name && g.CellPosition == surface.parent.position && g.CellIndex == surface.GetSiblingIndex() + moveFaceIndex))
```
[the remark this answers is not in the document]
acidscorch: okay

NitroxClient/GameLogic/Building.cs
```diff
@@ -1,213 +1,1326 @@
...
+#endif
             }
+        }
 
-            if(parentBaseId == null)
+        // Reassign the cached id to the now constructable object
+        public bool ConstructableBase_SetState_Pre(ConstructableBase instance, bool value, bool setAmount)
+        {
+            // Cache the id of a constructableBase that gets 100% finished
+            if (value == true && setAmount == true && !instance._constructed)
```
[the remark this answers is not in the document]
acidscorch: Can you give me a hint what you mean with leaking implementation and abstraction in this case?

NitroxClient/GameLogic/Building.cs
```diff
@@ -1,213 +1,1326 @@
...
+                else
+                {
+                    return true;
+                }
             }
             else
             {
-                Base playerBase = gameObject.GetComponentInParent<Base>();
+                return true;
+            }*/
```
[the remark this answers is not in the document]
acidscorch: Some of them are approaches or hints for further implementation of features or other fixes. I can remove if not wanted.

NitroxClient/GameLogic/Building.cs
```diff
@@ -1,213 +1,1326 @@
...
 
-            NitroxId id = NitroxEntity.GetId(gameObject);
-            NitroxId parentId = null;
+#if TRACE && BUILDING && HOVER
+            NitroxModel.Logger.Log.Debug("BuilderTool_Pre_HandleInput");
+#endif
 
-            SubRoot sub = Player.main.currentSub;
+            currentlyHandlingBuilderTool = true;
+            return true;
```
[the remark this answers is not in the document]
acidscorch: The return true is currently in any case true, because there is no implementation yet wether the targetted object is in use by another player or not. When this is implemented, the return will be false in this case. This will suppress to execute the entire handleinput function of subnautica. An forbidden info will be given to the user as a visual (e.g. deconstruction not allowed in case of faces attached).

NitroxClient/GameLogic/Building.cs
```diff
@@ -1,213 +1,1326 @@
...
+        public void Constructable_Construct_Post(Constructable instance, bool result)
+        {
+
+#if TRACE && BUILDING
+            NitroxId tempId = NitroxEntity.GetIdNullable(instance.gameObject);            
+            NitroxModel.Logger.Log.Debug("Constructable_Construct_Post - instance: " + instance + " tempId: " + tempId + " construced: " + instance._constructed + " amount: " + instance.constructedAmount + " remoteEventActive: " + remoteEventActive);
+#endif
+
+            //Check if we raised the event by using our own BuilderTool or if it came as post Event of a Remote-Action or Init-Action
+            if (lastHoveredConstructable != null && lastHoveredConstructable == instance && currentlyHandlingBuilderTool)
```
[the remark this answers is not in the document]
acidscorch: Hmm, i think you are right. When i look at an object that is currently remote build. Should also check here if remoteEventActive.

NitroxClient/GameLogic/Building.cs
```diff
@@ -1,213 +1,1326 @@
...
+            // Construction finished of a remote is finsihed
+            if (result && instance.constructedAmount == 1f && remoteEventActive)
+            {
+
+#if TRACE && BUILDING
+                NitroxModel.Logger.Log.Debug("Constructable_Construct_Post - finished construct remote");
+#endif
+
+                // ##TODO BUILDING## This is a quick fix. Remove this block when energy-sync is implemented. 
+                if (instance.gameObject.name.Contains("Solar"))
```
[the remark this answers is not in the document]
acidscorch: Or maybe better remove in a separate Class. NitroxIds should be stable now and the Energy-Logic is similar on Cyclops and bases. Maybe first load all bases and vehicles and apply energy afterwards.

NitroxClient/GameLogic/Building.cs
acidscorch: I don't really know if this is a bug of subnautica or just intented to be so. If you want to build a Connector via Multiplayer-Builder you need the Enum Id = 5519. But when grabbing the TechType from a lokal builded Connector this has the Enum Id = 5508. Using this Id with the Multiplayer-Builder fails. So for saving or transfering to remote player changing tech type is needed. 
This is the only object i found, where the tech type changes happen.

NitroxClient/MonoBehaviours/ThrottledBuilder.cs
acidscorch: What do you mean with depedeny injection?

NitroxModel-Subnautica/DataStructures/GameLogic/Buildings/Rotation/Metadata/BaseFaceRotationMetadata.cs
acidscorch: I just wanted to make it a little bit clearer for what these members are needed in a RotationMetadata-Class.

NitroxPatcher/Patches/Dynamic/CellManager_RegisterEntity_Patch.cs
acidscorch: Currently not. It is/was for tracking, where and when a Base and/or a BaseRoot Object is generated from the game. The RegisterEntity patch may also be useful to analyse spawn issues, because this gets triggered for every entity in the game.

NitroxClient/GameLogic/Building.cs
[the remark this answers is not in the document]
acidscorch: Ah okay, thanks.

NitroxPower:
Hey acidscorch i would like to help you if i can with the testing. Let me know if you need my help or what i should do

acidscorch:
Hello @NitroxPower,

long time no see, very nice to read you again :) I would appreciate if you could help me with testing. 
(i also referrenced our old issue 691 here ;) )

The main problem with keeping NitroxIDs stable on base-parts are the geometry-updates which Subnautica performs when the layout of a base is updated. The problem is, that in these moments we have to reassign the NitroxIDs to the updated base-parts. This wouldn't be too complicated if not the order of the parts would change (e.g. stilts spawn, short additional corridors when placing a hatch on a room, putting multiple rooms above each other, ...)  or parts get renamed caused by shanging there surface (e.g. vertical connector with or without ladder inside). 

All constellations i've found are adressed, but maybe there are more that need to be considered. If you are able to compile my branch, it would be nice, if you could start over with a new world and just do some construction and deconstruction of every constellation you can get in mind, disconnect, reconnect and see if it is as you left. Most critical are all situations where objects change there visible surface or cause other objects to transform. 
If you can pin down a constellation that is reproducably breaking, please tell me what you did. This pr is only for base constrution/deconstruction and id stability, all the other things (hull sync, energy, and internal special state of module internals like reactorressources, filtered water, etc.) are not yet implemented. First we need to make sure everything has its right id. 

If you can find someone that also assits you in multiplayer-test this would be nice (my server is currently not available). 

I'm currently refactoring the code to implement the suggestions Sunrunner and Measurity made.

Best regards and thanks :)

NitroxPower:
Yeah its good to be back :) Alright first of all i hope i downloaded the branch correctly by downloading it from this way 
image

I was making tests about about every bases pieces and external stuff in general. Basically almost everything works fine and is sync correctly when you rejoin and stuff. Ok so here is the issues i adressed:

- There is some weird thing with the saving. Uppon the first time when i builded some stuff and i forced the server to save the current state by using save command and then rejoining non of the buildings was saved but a lot more time later after testing others stuff in the end when i saved and rejoined it did save them. (I am doing further testing for this)

- Power transmitter or Thermal plant break completly the building. Even removing them everything that i place disappears after i disconnect and reconnect. Nothing from the base pieces gets saved except the exterior modules that are placed on the ground like the Floodlight and growbed. Exterior modules that are placed on base pieces will disappear as well upon reconnecting.
Capture

- Solar Panels after reconnecting have a weird light issue 
image

-Never used Vertical connect before but i builded this and i think there should be a way from inside to climb up?
image
image

dartasen:
So I did some testing stuff as you asked
Sans titre

- Base fundations disappeared
- Some multirooms disappeared but the furnitures inside it are all present with correct positions (one with a bioreactor and leaf walls, other with radio, builder,... and a vertical tube connected with a ladder)
- Solar pannels seems better than before
- Solidifications for walls are way better than before

acidscorch:
Hi @NitroxPower, @dartasen,

thanks for the tests and the details given :)

For several mentioned aspects i also already found the reason:

* Problems after start with new world or starting a new base at another location: If the used geometrycache was empty, it failed.

* Problems with unfinished fortifications or bulkheads: They got loaded before the rest and block placement of finished items

* Light issue with Solarpanel: that is happening because the initial power value is set to 50% (i will remove this, energy comes later)

* Foundations i haven't analyzed yet

I also have found out how to supress Item stealing/granting for actions a remote Player does.

Currently i'm still refactoring and i give a ping if a new test would be nice.

(@NitroxPower: After building a vertical connector and having a Corridor on top and bottom, you should be able to build a ladder)

Thank you both and best regards

acidscorch:
Hi @dartasen, @Measurity, @Sunrunner37,

i could need a little help on patching. I'm trying to suppress the rotation hints while loading at initialsync or when a remote player places a new piece and rotationmetadata is applied. But it seems that returning false for the prefix doesn't lead to skipping the original method as it is intended to do. Anyone knows how to fix it or what i'm doing wrong here? Maybe it's just a special situation because the original class and the original method are both static. I've checked if the patch-code gots properly loaded and executed.

image

Best regards

acidscorch:
Hi @NitroxPower, @dartasen,

i have updated the pr to the current master and implemented a lot of fixes (see comments on 
a6a5eac). Would be nice if you could do another test with a fresh world.  

Unfortunately it seems that something of the rest of my local-branch has gotten broken beacuse i'm getting convert errors for color-convertions between Unity and Nitrox. So currently i myself am not able to test the changes on top of the current master, only in my dev-environment which is on the state before the latest release. 

Best regards

NitroxPower:
Hello @acidscorch  i want to make a quick question in the installation. In order to install the latest fixes you made we have to download the latest version of Nitrox and replace its files with the files of your latest release? Because there are some configuration files that do not exist in latest release and i believe i am doing something wrong
image

acidscorch:
Hi @NitroxPower,

you are right, there are different files now and it somehow seems to be incomplete because the NitroxLauchner is smaller (filesize) than it should be. Maybe this is related to the changes in library handling and assets in the latest commits. Let me first figure out, how this works and what has changed here. I give a ping, when i know how to test the pr without needing to compile it yourself. 

Best regards

Jannify:
@NitroxPower You have to delete the Managed folder an verify the game files through Steam/Epic. after that use the launcher normally. This is due changing which files are copied and which can be left in the launcher dir.

Measurity:
The color error is because it's using the old DLLs and so protobuf has issues matching the type. The new build script won't retroactively remove these old dll files. You have to clean it.
 
- Remove files in "Managed" dir of Subnautica and verify game with Steam/Epic
- Do a clean build of the Nitrox solution in your IDE
- Build again and test

Swiff4:
Hi @acidscorch,

could you please rebuild it so that I can test your modifications? I have no own IDE for it and would like to test it. Just an additionaly question: Is the version you build already merged with the master branch or is it an older version? Because with the old one which I could download yesterday from the pr we faced some issues that should be solved already like item exchange in stocks, etc.

Thanks and regards!

acidscorch:
Hi, 
thanks to @Jannify and @Measurity for the Infos for cleaning up subnautica installation. 👍 

@NitroxPower @Swiff4 @dartasen 
Currently it is not possible to take the latest artifacts from the check and combine it with the latest release to test the pr. I have uploaded a debug build on dropbox. ⚠️ This is less performant and may create larger log files then a release version.

To use it, you have to clean up the subnauctia game dir as @Jannify suggested, unzip the dropbox file in a separate folder and directly use the launcher in it. (Don't copy over any files to anywhere)

⚠️ This debug build is incompatible to the latest release 1.2.2.0 ! If you have switched back to using the release version, you'll have to clean up subnautica again to use the debug build again. 

https://www.dropbox.com/s/31rferevvcl3m7a/debug%20buidling-pr1%20acidscorch%202020-05-25.zip?dl=0

Greetings and very much thanks for your testing efforts and patience with me :)

NitroxPower:
@acidscorch Alright after the hole testing i made i am really suprised because its vastly improved congrats. I dont have a lot of things sto say i could barely find bugs in terms of the basic building.
Basic bugs:

- Observatory disappears after restarting the server 
Thats all everything else just works and saves fine after rejoining including the interior modules lighting and everything is just as how it should be after rejoining or restarting server. I am surprised that even the Reactors position was saved. There are the more specific related bugs like the state of things that doesn't get saved but overall its honestly amazing.

dartasen:
@acidscorch I also did some tests on my side, it really seems quite good since my last test !
I built a giant base with several floors and almost the entire base is rebuilt when I reconnect. There are however some bugs that I could see on the missing parts: 

- The flower pots on the walls never have the right orientation, (always offset by several degrees in the same direction)
- One of my corridors has disappeared (it was connected to an angle corridor and a multiroom)

We might need to find the pattern which causes a building to disappear, according to my tests it is often the same configuration which causes the disappearance

acidscorch:
@NitroxPower 
I've tried to reproduce the dissapearing of the observatory, but it didn't happen (tried with finished and unfinished objects). Maybe this was a follow up problem of something that happened before building the observatory. Could you check the log for errors and/or provide me your savegame, so i can check what has been saved and if it is a problem at loading or if it hasn't come to the save game at all. Or as alternative, try to find steps to exactly reproduce the disappearing starting from a new world. 

Best regards

acidscorch:
@dartasen 
Can you descripe the configuration which leads to disappearing or give a step by step guidance to come to the state from a new world? 

I must admit, i never used flower pots before :) I'll check this. 

Best regards

acidscorch:
@NitroxPower @dartasen 

I would like to clean up and optimize the code the next days, so that this pr can be merged and maybe someone else start on the dependend tasks (bulkhead door state, energy, and so on).
The points you mentioned i would like to better analyze later deeper to avoid that this here gets a never ending story to the last detail. Better let us open separated Issues for these minor things to also move the conversations to there and not here on the pr-section. 

For now can you both confirm that this pr solves the majority of base building issues and is stable enough for integration in Nitrox?

Thanks and best regards

dartasen:
@acidscorch I will try to figure out the steps to reproduce the bugs. 
[world.zip](https://github.com/SubnauticaNitrox/Nitrox/files/4681779/world.zip) (Try to rebuild angle corridor of the upper multiroom)

Indeed the PR seems to solve major issues after relog, however I can't really tell for now if it's ready for a merge, only did few tests and I didn't try with other players but I feel like you are on the right way.

Congrats for your job

NitroxPower:
I tracked down exactly how the observatory bug occured. I made this as you can see and placed a base pump in it. After that when i saved and restarted the server this happened. 
image

acidscorch:
@dartasen @NitroxPower 
Okay, i could reproduce both constellations. I give a ping when i found and fixed the code parts.

acidscorch:
@NitroxPower @dartasen 

New test build:
https://www.dropbox.com/s/9hz8wqttfkvypq2/debug%20buidling-pr1%20acidscorch%202020-05-26.zip?dl=0

* The waterpumps blocked the observatory because i missed observatory in the list of structure items that need to be loaded first.
* The corridor wasn't built because it couldn't be placed in the default rotation state. So i changed to aplying the rotationmeta data before the placement check is executed. This fixes all situations when you can not place an object in its default rotation (the visual buildertool ghost is red) but you can place it, when you rotate the piece (the buildertool ghost gets green).

Swiff4:
@acidscorch: Thanks for your work. I tested ist with a friend and ist seems to work! The issues we faced are already mentioned in other pr or reported bugs. I would agree to merge this as fast as possible as i think that 1.2.2.0 is currently not usable. At least for me and my friends the building issue is a big thing which stops us from using this Version. Meanwhile i will use your Dropbox Version. Can you please Update it with your last code corrections?

acidscorch:
@Swiff4  
Thanks for your tests and the positive feedback 👍 

But please be aware that this build is a debug build and not for release or broader use. You can use it for further testing, but don't share this or open any issues for this. As soon as this is merged and some other bugs are fixed or features implemented that is currently worked on, the Nitrox Team will provide a new release for community usage. 
And i will remove the dropbox-version as soon as tests are done. 

Best regards

Swiff4:
You are totally right. For sure this version just stays for me and my mates as we go on with testing. I need them as I couldn’t give you a Feedback according the Multiplayer syncs. 😉I just wanted to point out that your work is really needed. We started with the official versions and came up here as we faced the building issues.

Swiff4:
Hi @acidscorch,
I found a small thing that happend during my testing. During the base construction there seems to be an issue with spawing fishes as i had a fish swimming in my base after reconnect. Is this a topic for here or for a new issue? I picked it by mistake before taking a screenshot but he was definitly in the base as he swam against the walls and tries to get out ;)

Log attached.

[nitrox-2020-05-26.zip](https://github.com/SubnauticaNitrox/Nitrox/files/4683207/nitrox-2020-05-26.zip)

acidscorch:
Hi @Swiff4,
I believe that this not really is a nitrox related behaviour. I think i remember that this also happened while playing subnautica on playstation. Even with the bigger fishes that are not pickable. 
Googling for this will give you quite a lot hits for normal subnautica without nitrox.

acidscorch:
Hi, 

applying the rotation before the ghostmodel update seems to break reactors in base. I have to look into this again and maybe use a switch for inside and outside objects. 

Best regards

NitroxPower:
Is this a Nitrox related bug? I am pretty sure you can't do this in Subnautica 
image

acidscorch:
@NitroxPower,
yes this is also subnautica basic bug. Verified it in singleplayer in creative mode. Stack 3 rooms above each other and build a ladder from the lowest room and in the most upper room in the same corner. When you then look in the middle room you will see this :)

acidscorch:
https://www.dropbox.com/s/op8rlxg0jt0f9cl/debug%20buidling-pr1%20acidscorch%202020-05-26-2.zip?dl=0

* Reactors should be placed again now and rotated corridors also.

NitroxClient/MonoBehaviours/ThrottledBuilder.cs
```diff
@@ -83,14 +74,29 @@ private void ProcessBuildEventsUntilFrameBlocked()
...
+                    {
+                        prog = Convert.ToInt32(Count / 2 - prog * -1 / 2);
+                    }
+                    else
+                    {
+                        prog = Convert.ToInt32(prog / 2 + Count / 2);
+                    }
+
+                    WaitItem.SetProgress(prog, Count);
+                }
```
Sunrunner37, line 91: This is overloading throttled builder a bit as this use-case is only for loading.  Is there a way to expose the pending count as an attribute and keep all of the progress logic in the building's initial sync?

NitroxClient/MonoBehaviours/ThrottledBuilder.cs
Sunrunner37: We only want to call the locator once - similar to the other items in the Start method above.

NitroxModel-Subnautica/DataStructures/GameLogic/Buildings/Rotation/Metadata/BaseFaceRotationMetadata.cs
```diff
@@ -0,0 +1,42 @@
...
+namespace NitroxModel_Subnautica.DataStructures.GameLogic.Buildings.Rotation
+{
+    [Serializable]
+    [ProtoContract]
+    public class BaseFaceRotationMetadata : RotationMetadata
+    {
+        public const long VERSION = 1;
+
+        // Base modules anchor based on a face.  This can be constructed via these two attributes.
```
Sunrunner37, line 14: minor: two attributes?

NitroxServer/GameLogic/Bases/BaseManager.cs
```diff
@@ -9,153 +9,169 @@ namespace NitroxServer.GameLogic.Bases
...
-                basePiece = completedBasePieceHistory.Find(piece => piece.Id == id);
+                basePiece = allBasePieces.Find(piece => piece.Id == id);
 
                 if (basePiece != null)
                 {
                     basePiece.Metadata = Optional.OfNullable(metadata);
                 }
+                else
+                {
+                    NitroxModel.Logger.Log.Error("UpdateBasePieceMetadata - Received UpdateBasePieceMetadata for unknown NitroxID: " + id);
```
Sunrunner37, line 163: minor: no need to fully qualify these (here and everywhere)

NitroxServer/GameLogic/Bases/BaseManager.cs
```diff
@@ -9,153 +9,169 @@ namespace NitroxServer.GameLogic.Bases
...
+            NitroxModel.Logger.Log.Debug("BasePieceConstructionAmount - id: " + id + " - constructionAmount: " + constructionAmount);
+#endif
+
             BasePiece basePiece;
 
-            lock (partiallyConstructedPiecesById)
+            lock (allBasePieces)
             {
-                if (partiallyConstructedPiecesById.TryGetValue(id, out basePiece))
+                basePiece = allBasePieces.Find(piece => piece.Id == id);
```
Sunrunner37, line 62: It looks like almost all of the access patterns are looking up by id.  Should we index to a dictionary similar to how partiallyConstructedPiecesById used to be done in the constructor?

NitroxServer/GameLogic/Bases/BaseManager.cs
```diff
@@ -9,153 +9,169 @@ namespace NitroxServer.GameLogic.Bases
...
+#endif
+
+            lock (allBasePieces)
             {
-                partiallyConstructedPiecesById.Add(basePiece.Id, basePiece);
+                // Subnautica changes the order of basepieces if there are Layout-Changes on an existing base, e.g. by adding pieces or splitting a base apart
+                // in two bases by removing connector-tubes. This changes the order in which BasePieces are need to be generated by Nitrox on client side loading.
+                if (allBasePieces.Count > 0)
+                {
+                    int nextBuildIndex = allBasePieces.Max(piece => piece.BuildIndex) + 1;
```
Sunrunner37, line 43: Might be more friendly to compute upfront in the constructor once and use a member variable to increment.

NitroxClient/GameLogic/Building.cs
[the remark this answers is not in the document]
Sunrunner37: Sorry I don't think I was fully clear.  The patch should make the determination if it should suppress the code that is being mutated.  This is because the patch is close to the implementation and understands what the upcoming code will be doing.  The code in gamelogic is generally agnostic to who is calling it and sometimes has multiple invokers (such as through patches or packet processors).  With that being said, the patches can leverage gamelogic code to make an informed decision but we should tie it behind a descriptive name such as isRemotePlayerAlreadyInteracting().  I think any time that you are invoking a gamelogic method from a patch you should tie it to a descriptive name.

acidscorch:
Hello @Sunrunner, 
for some reason (i don't know why), i can't reply directly to your last comment on the Building.cs reffering the method naming and invoking gamelogic methods from patches, because there is no reply button on that post, so i do it here. 

(see screen below to assist in explanations following)

The naming i use has been deliberately chosen to explicitly show that these are the patch methods and where they are invoked from. These methods are not intended to be used by any other function (e.g. PacketProcessors) and only for the patches. If there would be a way to make them public only for a specific caller i would do it. Using a descriptive name could possibly prompt another developer to use it, because it sounds farmiliar and suiteable for what he wants to achieve. Then you have two different accesses to this method and in case you have to change something in the method (e.g. for updates of subnautica or bugfixes) it can be that you fix the code for the patch usage, but break it for the other usage . Same thoughts are for reusing the same method from a Patch call and a Packet Processor Call. Yes, there may be situations where the code thats need to be executed is identical or very similar, but crossusing should be avoided. Also i would avoid the way to then use private submethods that are called from patch methods and other methods, because this can also lead to the same situations. Especially in this case of a community project, where not every developer maybe fully aware of all dependencies a change may have. Better to have a copy of the code, which can be altered independently if needed. Or in cases of an override of an original method it is also better to have not used outsourcing to better compare the original (e.g. in dnspy) with the overridden variant in nitrox. 

I understand that these method names are not very speaking or intuitive in reading. Therefore i would better use more comments on the methods. This would also help other developers to understand why a particular patch is needed, what they are changing/doing and even could faster lead to the right place in case a bug is found and needs to be fixed. 

To make my thoughts a little bit clearer and to show which constellations from my point of view should be avoided please see the following overview. 

image

Best regards

NitroxServer/GameLogic/Bases/BaseManager.cs
acidscorch: Using a dictionary might also be possible. But this would lead to more code in handling the buildindex or other constellations where we may need to find a group of items based on a parameter (e.g. base splitting could cause a group of items to update their parents). List gives here more flexibility to not explicitly make use of for each loops.
In concerns of performance i'm with you, that the list operations are a little bit slower than the dictionarys, when the only use case is to get the value to a specific key. When you need to iterate through all key/value pairs of a dictionary the performance difference is not so big anymore. Also the difference is in magnitue of microseconds for the number of entrys that are handled here (maybe about 500 to 1000 items for multiple bases in full playthrough?)  and the methods are called only at specific situations in the game (maybe about 500 calls in a game session?). 
For updating movement infos that are need to be handled permanently i would switch, but here i would prefer to stay with the easier coding way and the more flexibility.

NitroxServer/GameLogic/Bases/BaseManager.cs
acidscorch: Yes this might be possible. This wouldn't decrement the buildindex when the last built basepiece gots deconstructed, but i see no trouble with missing numbers in the sequence. I will change it

NitroxModel-Subnautica/DataStructures/GameLogic/Buildings/Rotation/Metadata/BaseFaceRotationMetadata.cs
acidscorch: comment is wrong, i will remove

NitroxServer/GameLogic/Bases/BaseManager.cs
acidscorch: In some classes a normal Log output is not needed only for my trace outputs. If i make a using in these classes, this will lead to showing the using as not needed when the trace define is not set, so i prefer to use the full qualifier in the trace blocks. 

In this case here, you are right, this can be solved via a using.

NitroxClient/MonoBehaviours/ThrottledBuilder.cs
acidscorch: I think simply providing an attribute wouldn't work, because the InitialSync is waiting for completed == true. What i could do, is give the ThrottledBuilder an additional EventHandler that provides the progress. The listener can than be attached and deattached like the FinishedCompletedBuildings.

NitroxClient/MonoBehaviours/ThrottledBuilder.cs
acidscorch: Sorry i still don't understand what you are suggesting. The locator also gets called everytime an action is happening in a patch and a call is needed to a gamelogic class. I understand the locator as a dynamic reference to a static instance which has a permanent lifetime. I don't get the point why this would be harmful here to call it everytime a buildevent is processed.

Measurity:
FYI on replying to Sunrunner:
I had a similar issue of not being able to reply to Sunrunner's comments. The input box is only visible by going into the files tab and navigating to the comment. See: https://github.com/SubnauticaNitrox/Nitrox/pull/1093/files#r431616479

Sunrunner37:
Hello @acidscorch ,

Thank you for sharing your perspective.  To start, this is more of an exceptional circumstance as base pieces have a large amount of built-in game state that needs to be transitioned between many patches.  You'll find that the vast majority of patches are straight forward reactions that don't need complex interdependent operation.  With that being said, our guidance is to leverage existing game constructs to manage state when possible; this allows the game to effectively be a state machine instead of managing it ourselves.  When this is not possible, we definitely supplement this with data structures in game logic. 

In regards to the PR patch usage, the relationship between the patch and game logic is not appropriate.  Fundamentally, a method should not be cognizant about who is calling it or be tightly coupled to the state of the caller.  The language does not support whitelisting callers because that creates inappropriate coupling between classes. If a method invocation is ordered or stateful, then there should be an encapsulation around it that sets up the correct context beforehand or performs necessary validations on internal state. When a method's logic is truly an exclusive extension to a single caller, then the code should be moved to the caller.  I understand that this is both theoretical and idealistic (especially in unity); however, the feedback is to refactor into modularized blocks that are independently understandable while still articulating any stateful behavior, such as `stagePlacedBasePiece` as a fictitious example.  Lastly, I don't agree at all to code duplication over code re-use.  Code re-use is more applicable for a community project because duplicated code requires developers to know of the existence of the duplicated code to ensure parallel changes to both copies. Bugs and technical debt are both proportional to the amount of code you maintain. 

As for game logic, it should not know about the concept of patches or how to influence them. The reason for this is that patches have intimitate knowledge of how the target code works and has a responsibility to weave new code into it.  Moving a prefix's "return true" up a layer now misplaces the weaving responsibility and causes the game logic to inappropriate know about the calling code.  For stateful behavior (such as when multiple patches need to work together), we can always have a class that manages the state with public API methods; both SimulationOwnership and Entities are examples with clear naming and the ability to manage stateful data.  It is fine for a patch to invoke the layer above to collect data that informs a branching decision (note, this doesn't mean grabbing private state data - we should always use the public APIs). 

Your secondary question (and corresponding diagram) is to what extent these individual layers should interact; more specifically, when is it appropriate to short circuit the interaction with game logic.  Admittedly, we have historically had lax boundaries which have developed to reduce cognitive load.  For example, we have some code that constructs a packet directly in a patch or updates a game object in a processor.  Both cases can trivially be loaded into the domain of game logic with one additional layer of abstraction; however, if the item doesn't come with stateful side-effects then the gain is net neutral.  This leads to the question of patches and to what extent they should be cognizant about domain.  Generally, patches may understand domain that is localized to the code that they decorate.  It is not inherently bad for patches to have private methods or even minor state keeping (some of this is provided by harmony itself).  The goal is cognitive reduction and aiming to localize any actions.  Obviously there are judgement calls when you have multiple patch interactions but this is not the standard case in Nitrox. 

Lastly, we definitely will perform a refactoring pass on the client for BZ in the far future.  We've already executed this for both the model and server projects.  The major complexities lie with identifying common feature sets and creating interfaces (such as a vehicle factory).

acidscorch:
Hello @Sunrunner37,

thanks for your detailed reply and your point of view on this. 

I have implemented the variant i had in mind in the latest commit. This would be a solution to decouple also UnityEngine from the NitroxClient and also provides a speaking Interface to the NitroxClient. 

This way it is possible that a developer who has no deeper knowledge of Subnautica classes but knowledge of the Nitrox structure can preimplement the Nitrox related logics to the point of the interface. 

All the directly game related acting with objects happens in the PatchLogic class and an access to the patching related methods is not possible, because they are not exposed via the corresponding interface. This way, someone who is deep in subnautica mechanics and using dnspy can focus on game mechanic related things. As the PatchLogic now is directly attached to the patches, it should also be allowed to handle overrides and suppression of original game methods. 

Best Regards

NitroxClient/GameLogic/InitialSync/BuildingInitialSyncProcessor.cs
acidscorch: Implemented also sorting via ParentID in the initial sorting. 

The rest of the item adding i would let as it is, because this is only executed once in a game session and this code is easy to understand an readable.

Swiff4:
Should I test something again meanwhile you are reviewing it?

acidscorch:
Hi @Swiff4,

thanks for your interest, but currently no test is required. There has been a massive architectural change under the hood. The fundamental things i've checked and they work, but there are some minor constellations which i have to inspect again in detail. 
But first i would like to wait for the review. I give you a ping when it's again time for intense testing :)

Best regards

Sunrunner37:
Hello @acidscorch ,

Checking back in to this review as I have been offline for a bit.  I have reviewed the current implementation and here are my thoughts.

Overall, I have reservations about PatchLogic.  Patches in the PR are all one-line invocations into the PatchLogic layer.  The invoked methods, such as `BaseAddBulkheadGhost_UpdatePlacement_Pre`, are 1:1 with the patch itself.  These are so coupled with the caller that they are explicitly named after the caller.  As stated before, it isn't a good practice to have methods cognizant about their callers.  Furthermore, these methods are still making decisions on branching which is fundamentally the responsibility of the patch because it lives on top of the mutated code.  In my opinion, these two points are clear indicators that the logic should live within the caller.  

The reason that you are encapsulating all of this PatchLogic methods together is because of the state between them.  For example, `BaseAddBulkheadGhost_UpdatePlacement_Pre` checks if there is a `currentConstructedNewBasePiece`.  This type of state should be interacted with through a class that encapsulates the state keeping so it is decoupled from patch layer.  It isn't a good practice to encapsulate all of the building methods in a 1600 line file so they can access common member variables; this is more procedural and is functionally equivalent to keeping all of the logic in the patches and using statics to keep state (though, less readable). For full disclosure, the prior implementation using transient object manager to manage this state was inappropriate; it should have definitely been a well-defined class that understands how to keep track of the state.  We should use this as an opportunity to refactor and clean up these hacks. 

PatchLogic also raises another concern in regards to responsibility.  I don't think it is appropriate for NitroxPatcher to encapsulate stateful managers (such as `Building` in this PR).  The intention is to keep NitroxPatcher as a small layer that mutates existing UWE code.  As a general rule, we can inject code that would have reasonably been there if UWE had multiplayer functionality - no need to limit it to one line.  Some light state keeping is fine on a case-by-case basis (especially any provided by Harmony) but major stateful behavior should live in NitroxClient (GameLogic).  

In regards to decoupling Nitrox related logic from the underlying Subnautica code, I'm not certain to what extent this complexity adds value.  Generally speaking, Nitrox logic is simple state keeping and packet sending.  You'll already find that when it makes sense, we have decoupled these concepts (such as `SimulationOwnership` gamelogic specified above).  But I'm not seeing significant benefit by enforcing such a decoupling.  Your example about a developer implementing Nitrox logic without knowledge of Subnautica code seems suspect and may add unnecessary complexity.

One of your intentions was to facilitate interfacing for Below Zero.  Let's not worry too much about that right now.  It is likely that the proposed boundaries may not be appropriate.  The reason that I say that is because Subnautica and Below Zero share a substantial overlap in code.  We'll want to try to share logic as much as possible which will require creative refactoring - even in our game logic layer.  Artificially drawing a Subnautica bounding box around all building code may not be desirable as I don't want to maintain duplicate parallel implementations unless there is total divergence. 

Lastly, I agree that we should replace transpilers with prefix/postfix when possible.  However, I don't think that this is coupled with where the business logic resides.

acidscorch:
Hello @Sunrunner37,

thanks for your review and no problem, i know very well myself that time is a precious good. So thanks for your efforts. 

Yes, you're right, all the patches in this PR are 1:1 forwards to the Building class and yes they are directly coupled to the corresponding patch class. This forwarding was done, to get the logic to the place where the state informations are hold. This was formerly the GameLogic class Building residing in the NitroxClient. Through the change of moving the class to the Patch-Library and using an interface this is not needed anymore. Moving the code from all the forwarded methods to the Patch-Files seems fine if the needed state and flow informations also stay "near" the patches because they are also directly coupled to the caller and need fundamental information about what is the current flow state and are responsible for which behaviourchanges they cause in the patches that use them. 

Moving the patch methods out of the Building class would reduce this class to:
* Interface-Implementation to provide a service for the client to bring informations in or trigger activities
* Internal state informations (e.g. currentConstructedNewBasePiece) for informationholding/sharing between multiple patch classes
* Outward communication by using the IPaketSender service from the client and/or using other services provided by the client (e.g. SimulationOwnership)
Which seems fine for me as it creates an area of knowledge for this particular theme and all directly game dependent activities happen in the patch domain. 

As i have looked through the most of the other parts of the Nitrox code i couldn't find a part where state keeping of directly game related informations is actually explicit needed in the NitroxClient (but maybe i missed something here). So keeping patch relevant informations directly connected to the patches/game makes more sense to me than holding this in the NitroxClient. As you already mentioned yourself the state transfer via the harmony buildin feature is for very simple tasks and i think it is more ment for smaller projects or in cases where no further state transfer is needed than from before to after the method. But i don't think that the developers of harmony declared this as a preferably should use or must use feature. I think it's just a nice and fine little helper, if you don't need or want to use something more complex. 

For my impression the transfer line between game states & logic and the nitrox states & logic is not in the right place. ( I really don't want to upset anybode here and i have followed the progress on this project for a year now and also i know that there has been hours and hours of working, discussion and designing in this project from very much people. I just want to share my thoughts and knowledge and what i would suggest to change. I must admin i'm realy impressed by many of the parts, the dynamic loading and the launcher. )
Back to the transfer line. As i can see, there is gamelogic in every layer and that's why every layer has direct dependencies to subnautica. To overcome this, you have separated vertically. Which causes to have parts of classes and logic in the common libs and parts in the subnautica special libs. 
For my impression the transfer line lies somewhere between the most of the nitroxclient and the current gamelogic classes. If you would move the transfer line between the nitroxclient logic and the patchlogic, then everything above this layer could be dependency free and joined in generalized binaries. If there are no direct game dependencies also the model und structure classes could reside in the same library (e.g. Cylcops classes next to SeaTruck classes) because they don't interfere themselfes. As i can see from the BelowZero classes, some nitrox classes could also be reused (e.g. CorridorRotationMetadata) because i can't see a difference between SN and BZ at these objects. If  differences occur than this could also be handled via separate subnamespaces or even better with virtual classes with different overrides or at simple cases with if/else switching according to the main state if the client is currently running for SN or BZ. (e.g. like some patches switch for Multiplayer use).
For the serverside i can see that the spawning mechanism is implemented there. For my opinion this could also be done on the client side by using simulation ownership. Removing the dependencies from the server could also allow running a dedicated server without having to install entire SN or BZ on the server. (To admin, may be i oversee here something more.)

The significant benefit of this would be, that every nitrox element above the patch/core layer can be reused or extended without depencies to the other game or breaking the implementation for the other game caused by architecture breaks. If interfaces need to be extended or datastructures added for BZ this wouldn't interfere with the SN implementation, because this parts are just not implemented and not used in a SN environment and cause no harm. ViceVersa e.g. for Cyclops. 

The only thing that is directly game dependent is the patch/core dll which per se is not generally usable as it is dependend to the assembly and unity. The whole patcher and all logic in this dll need to be compiled against BZ. And here comes the point which i ment with copying over existing code to reuse instead of trying to outsource every bit that may be reusable. Even UWE didn't do this because as i can see e.g. all the cyclops classes are still there even if there is no cyclops in BZ and no "common" lib exists. And as for the SN lib is so stable and receiving an update per one or two years (?) after implementing one part bugfree in nitrox (e.g. building) there is no more need to maintain this parts of code often or maybe even never. Yes, there may be some classes that haven't differed between SN and BZ or parts of them may be identical. But i think, doing extraction here to layer above and or having code parts in the common and specialitys in some separated .dlls would clutter logic around the place which is more harder to maintain or harder to understand. It is only the patch/core layer that is effected by this, everything above can be so generalized that it fits for both SN and BZ without duplicates.

If the Bootloader and Launcher also are built on a generalized basis, the only thing the bootloader needs to do is load the right patch.dll if the user wants to play SN or BZ. 

If this architecture is used, we also come to the point where it is possible for someone who can use the "finished" nitrox solution (with all interfaces) and can (okay: more or less ) implement in the patch/core and apply all the changes that may be needed on this level/class for BZ (e.g. Builder.Begin SN vs. Builder.Begin BZ) without any knowledge of the rest of nitrox nor any need of change. 

As i assume, are the most challenges not residing in the nitrox classes or structures. As it seems there is quite a lot of knowledge for programming and also for some really hard nuts (e.g. Transpilers) or GUI work which i grant my most respect. That are things which i lack knowledge or haven't been my favorites. The most trouble seems to be the subnautica logic and workflows and findind the right spots to anchor in or out and to bring the things in a multiplayer environment without causing boomerang effects or unsynchronised actions. So i think separating these things that are directly attached to "game logic" may be better also for getting the full knowledge for the currently targeted aspect, as to maintain or maybe if necessary bugfix it later. As i can see are nearly all implementations of the current gamelogic classes on the nitrox side (my opinion) well done and are partly only unfinished or got stuck because something could not be implemented with subnautica. 

As you mentioned business logic as a keyword, so i hope i can assist my explanations with another powerpoint down below and maybe do a comparison to a database based software, which is one of the products i work on in my business. The assumtion is that every game instance of every player can be seen as a database (which gets manipulated by the player, timers and so on). The aim is to synchronize changes between the client-databases (which we also must achieve in our product for supporting distributed locations). The main difference is that there would be normally multiple clients or services above the business logic and not just one (or two if the player and the nitroxclient are seen as clients) like in this case. The point i want to mention is, that we use a core layer which directly sits above the database. This core layer is in our case mainly for access rights management and encrypting/decrypting database informations and for providing "public" access points. Everything that wants to manipulate anything in the database has to go through this. Even the business logic layer is ontop of this layer and cannot access through the core layer. Besides the security aspects this gives us also the ability to switch the underlying database (e.g. MSSQL vs Oracle) or allows us database updates (11.2 > 12.2) independent from the rest of our software by just exchanging the core implementation dlls. Without updating or changing any dependency on the layers above. 
In the screen i have tried to adopt the model on nitrox and show everything in green that i think can be reused and maybe only needed to be extended a little bit to fit both worlds. 


image


Lastly, to stay to my words, i told you on Discord that i would like to help you with the building theme. And this is what i still want to do. If the upper suggestion isn't your wish i would rollback my PR to before the outmoving Building from the NitroxClient, move the patchcodes and reimplement the former suggestions. This then should be the last tested state with the appropiate naming and code places. 

Best regards

Swiff4:
Hey guys,
how is it going? I know that you are all busy and I really appreciate your hard work on this project. I just wanted to ask if you can give me a short info when you expect to make further progress on this topic? My friends and me are waiting for this solution as we are facing many building issues and therefore stopped playing at the moment. Thank you very much for your time and effort!
Best regards,
Swiff4

acidscorch:
Hi @Swiff4,

i think you are right and i think you are not the only one waiting for the fixes to come online. I will create a new PR this week with the fixes in the former architecture style, so that it can be pulled independent from the discussion that is going on here. The PR will be online until saturday.

Greetings

dartasen:
This PR will never be updated, and we already have integrated some changes to the current master.
Thank's for the help, I'm closing this one.

PR #1097, opened 2020-05-11
Fix keypad doors not persisting state when reloading

Coding-Hen:
- Fixed keypad doors


Sunrunner37:
I'm not sure that this should be required as this is the same code in the metadata processor (which is executed when the entity spawns). Can you verify if the problem is that the door spawns after the keypad?  If so, we might want to do a small delay before executing the door open in the metadata processor.

Coding-Hen:
> 
> 
> I'm not sure that this should be required as this is the same code in the metadata processor (which is executed when the entity spawns). Can you verify if the problem is that the door spawns after the keypad? If so, we might want to do a small delay before executing the door open in the metadata processor.

Looks like the Starship doorLocked is not saved which is the actual issue for why the door is not reopening.

Sunrunner37:
> > I'm not sure that this should be required as this is the same code in the metadata processor (which is executed when the entity spawns). Can you verify if the problem is that the door spawns after the keypad? If so, we might want to do a small delay before executing the door open in the metadata processor.
> 
> @Sunrunner37 Looks like the Starship doorLocked is not saved which is the actual issue for why the door is not reopening.

Although the lock state may not be persisted, the replay of keypad should unlock it.  Do you know if StartshipDoors are always accompanied by keypads?

Coding-Hen:
As far as I'm aware the starshipdoors are not always accompanied by the keypad it can also be the weld doors and normal door for example

Sunrunner37:
I think we may need to apply StarshipDoorMetadata packet suppression in the KeypadMetadataProcessor as it can trigger a door unlock and cause packet echoing.

Coding-Hen:
> 
> 
> I think we may need to apply StarshipDoorMetadata packet suppression in the KeypadMetadataProcessor as it can trigger a door unlock and cause packet echoing. Let's also see if we need similar change for the precursor PR.

Yes it was something that I had noticed. Are you suggesting that we write a new packetsuppression or create use the existing packet supression?

Sunrunner37:
> > I think we may need to apply StarshipDoorMetadata packet suppression in the KeypadMetadataProcessor as it can trigger a door unlock and cause packet echoing. Let's also see if we need similar change for the precursor PR.
> 
> Yes it was something that I had noticed. Are you suggesting that we write a new packetsuppression or create use the existing packet supression?

I think we just need to wrap the open door invocation within keypadmetadataprocessor with packet suppression block for this new packet type.  Let's also look into the same for precursor door even though that has merged.

Coding-Hen:
Im not sure that it is needed for the precursor doorway as the metadata won't be constructed upon the processor processing, this should only be needed I believe when we call a function that has been patched such as toggledoor

NitroxClient/GameLogic/Spawning/Metadata/KeypadMetadataProcessor.cs
```diff
@@ -15,6 +19,7 @@ public override void ProcessMetadata(GameObject gameObject, KeypadMetadata metad
 
             if (metadata.Unlocked)
             {
+                PacketSuppressor<EntityMetadataUpdate> packetSuppressor =  NitroxServiceLocator.LocateService<IPacketSender>().Suppress<EntityMetadataUpdate>();
```
dartasen, line 22: Might be better to use a using statement to improve readability
```csharp
using (packetSender.Suppress<EntityMetadataUpdate>())
{
  //...
}
```

NitroxClient/GameLogic/Spawning/Metadata/StarshipDoorMetadataProcessor.cs
```diff
@@ -0,0 +1,28 @@
...
+using NitroxModel.Packets;
+using UnityEngine;
+
+namespace NitroxClient.GameLogic.Spawning.Metadata
+{
+    public class StarshipDoorMetadataProcessor : GenericEntityMetadataProcessor<StarshipDoorMetadata>
+    {
+        public override void ProcessMetadata(GameObject gameObject, StarshipDoorMetadata metadata)
+        {
+            Log.Debug($"Received starshipDoor metadata change for {gameObject.name} with data of {metadata}");
```
dartasen, line 15: Should we keep this log ?

NitroxClient/GameLogic/Spawning/Metadata/StarshipDoorMetadataProcessor.cs
```diff
@@ -0,0 +1,28 @@
...
+
+            StarshipDoor starshipDoor = gameObject.GetComponent<StarshipDoor>();
+            starshipDoor.doorLocked = metadata.DoorLocked;
+
+            if (!metadata.DoorLocked)
+            {
+                PacketSuppressor<EntityMetadataUpdate> packetSuppressor = NitroxServiceLocator.LocateService<IPacketSender>().Suppress<EntityMetadataUpdate>();
```
dartasen, line 22: Same for the using statement

NitroxClient/GameLogic/Spawning/Metadata/StarshipDoorMetadataProcessor.cs
Coding-Hen: Happy not to do this but as discussed before it was kept in to follow the SealedDoorMetadataProcessor Sunrunner wrote. Happy to remove as I agree possibly is just cluttering the logs

Jannify: Then maybe remove the log line from sun as well.

Coding-Hen:
Closing as changes this is now fixed as part of #1206

PR #1098, opened 2020-05-13
Cleaned up build scripts and used AssemblyResolve event instead of copying all dlls

Measurity:
- Changed NitroxModel and NitroxServer a lot to make it free of UnityEngine/UWE type references
 - Moved surrogates over from NitroxModel to NitroxModel-Subnautica to try keep NitroxModel UnityEngine.dll free.
 - Added `Nitrox.Bootloader` project that loads Nitrox based on where the launcher is located. The path is stored in "%AppData%\Roadming\Nitrox\launcherpath.txt" when Subnautica is started through the launcher. `Nitrox.Bootloader.dll` is the _only_ DLL that is copied to Subnautica managed directory.
 - Cleaned up the csproj files massively by using `Directory.Build.targets` and `Directory.Build.props` which MSBuild automatically includes in all projects.
 - Reverted LiteNetLib to upstream/0.8 because its csproj file is automatically 'changed' by the Directory.Build.* files.
 - Building solution and projects is now much faster since there is not copying going on beyond asset files like AssetBundles and static dll references.
 - Nitrox will now only do the patches if started through the launcher. Making it easier for people to switch between SP and MP.
 - Nitrox will now _not_ load when QModManager assembly has been loaded into Subnautica.


ZachTaMere:
Hello how i can download your version of nitrox to run my server with your version to test if the buildings works properly ?

dartasen:
> 
> 
> Hello how i can download your version of nitrox to run my server with your version to test if the buildings works properly ?

@ZachTaMere Can you please wait until the next nitrox release, It's a workplace here and If you have questions about Nitrox please join our [Discord](https://discord.gg/E8B4X9s) so our support team can answer to them.

Best regards

NitroxClient/GameLogic/Vehicles.cs
```diff
@@ -69,7 +70,7 @@ public VehicleModel BuildVehicleModelFrom(GameObject gameObject, TechType techTy
                         opvehicle.Value.ReflectionCall("RegenerateRenderInfo", false, false);
                     }
 
-                    HSB = opvehicle.Value.subName?.GetColors();
+                    hsb = opvehicle.Value.subName.GetColors().ToDto();
```
dartasen, line 73: We might need to keep all of those safe navigation operator because that could lead to NRE,
because as I commented on my hacky fix for exosuit, SubName objects aren't always initialised yet at prematured spawn

NitroxClient/GameLogic/Vehicles.cs
```diff
@@ -79,8 +80,8 @@ public VehicleModel BuildVehicleModelFrom(GameObject gameObject, TechType techTy
                         SubNameInput subNameInput = target.RequireComponentInChildren<SubNameInput>();
                         SubName subNameTarget = (SubName)subNameInput.ReflectionGet("target");
 
-                        name = subNameTarget?.GetName();
-                        HSB = subNameTarget?.GetColors();
+                        name = subNameTarget.GetName();
+                        hsb = subNameTarget.GetColors().ToDto();
```
dartasen, line 84: Same here, might be a good idea to play the safe card

Nitrox.Bootloader/Main.cs
```diff
@@ -0,0 +1,112 @@
...
+                Task.Factory.StartNew(() =>
+                {
+                    try
+                    {
+                        // Delete the path so that the launcher should be used to launch Nitrox
+                        File.Delete(nitroxLauncherPathFile);
+                    }
+                    catch (Exception)
+                    {
+                        // ignored
```
Jannify, line 36: Why not implement a exception just in case. Would be useful for devs in `DEBUG` if something is causing an error.

NitroxModel-Subnautica/DataStructures/DataExtensions.cs
```diff
@@ -0,0 +1,94 @@
...
+
+namespace NitroxModel_Subnautica.DataStructures
+{
+    public static class DataExtensions
+    {
+        public static NitroxVector3 ToDto(this Vector3 v)
+        {
+            return new NitroxVector3(v.x, v.y, v.z);
+        }
+
```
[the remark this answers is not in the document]
Measurity: I'd like to keep this ambiguous and not Nitrox specific in case we want to extend this to external services (outside of Nitrox).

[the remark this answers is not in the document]
Jannify: Ok, after your message in discord I know the meaning and I am fine with it :)

NitroxServer-Subnautica/SubnauticaServerAutoFacRegistrar.cs
```diff
@@ -38,11 +42,7 @@ public override void RegisterDependencies(ContainerBuilder containerBuilder)
             SubnauticaUwePrefabFactory prefabFactory = new SubnauticaUwePrefabFactory(resourceAssets.LootDistributionsJson);
             containerBuilder.Register(c => prefabFactory).As<UwePrefabFactory>().SingleInstance();
 
-            containerBuilder.Register(c => new Dictionary<TechTypeModel, IEntityBootstrapper>
-            {
-                [TechType.CrashHome.Model()] = new CrashFishBootstrapper(),
-                [TechType.Reefback.Model()] = new ReefbackBootstrapper()
-            }).SingleInstance();
+            containerBuilder.Register(c => new Dictionary<TechTypeModel, IEntityBootstrapper> { [TechType.CrashHome.ToDto()] = new CrashFishBootstrapper(), [TechType.Reefback.ToDto()] = new ReefbackBootstrapper() }).SingleInstance();
```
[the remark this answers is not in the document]
Measurity: Autoformat

NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs
```diff
@@ -352,7 +351,7 @@ private List<Entity> ConvertComponentPrefabsToEntities(List<PrefabAsset> prefabs
                 Entity prefabEntity = new Entity(transform.LocalPosition,
                              transform.LocalRotation,
                              transform.LocalScale,
-                             new TechType("None"),
+                             new TechType("None"),
```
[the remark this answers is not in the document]
Measurity: Oops.. really need to fix my autoformat so that I can use it on these large files. Because I don't use it, things like this happen.

Nitrox.Bootloader/Main.cs
```diff
@@ -0,0 +1,112 @@
...
+        
+        private static void BootstrapNitrox()
+        {
+            Assembly core = Assembly.Load(new AssemblyName("NitroxPatcher"));
+            Type mainType = core.GetType("NitroxPatcher.Main");
+            mainType.InvokeMember("Execute", BindingFlags.Public | BindingFlags.Static | BindingFlags.InvokeMethod, null, null, null);
+        }
+
+        
+        private static string NitroxLoadError
```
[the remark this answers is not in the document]
Measurity: I don't want to throw an error because it's 'ok' for it to not launch in some cases. But I can make it a method.

Nitrox.Bootloader/Main.cs
Measurity: It won't be useful. Nitrox won't run so reason enough to investigate.

NitroxLauncher/Patching/NitroxEntryPatch.cs
```diff
@@ -58,11 +60,35 @@ public void Apply()
...
+                throw error;
+            }
             File.Move(modifiedAssemblyCSharp, assemblyCSharp);
         }
 
+        private Exception RetryWait(Action action, int interval, int retries = 0)
+        {
+            Exception lastException = null;
+            retries++; // Always run at least once initially.
+            while (retries > 0)
```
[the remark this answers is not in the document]
Measurity: Doh :)

NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs
```diff
@@ -352,6 +351,7 @@ private List<Entity> ConvertComponentPrefabsToEntities(List<PrefabAsset> prefabs
                 Entity prefabEntity = new Entity(transform.LocalPosition,
                              transform.LocalRotation,
                              transform.LocalScale,
+
```
Jannify, line 354: A new line? :D

Nitrox.sln
```diff
@@ -26,124 +29,98 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "NitroxLauncher", "NitroxLau
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "LiteNetLib", "LiteNetLib\LiteNetLib\LiteNetLib.csproj", "{B0A014E4-0392-4E35-AA42-8FB5A40417E0}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Nitrox.Bootloader", "Nitrox.Bootloader\Nitrox.Bootloader.csproj", "{E4226522-9189-410B-93B2-792942FBD588}"
+EndProject
+Project("{D954291E-2A0B-460D-934E-DC6B0785DB48}") = "Nitrox.Subnautica.Assets", "Nitrox.Subnautica.Assets\Nitrox.Subnautica.Assets.shproj", "{0763D551-B1A8-4960-B88A-98833F957936}"
```
Jannify, line 34: I don't really know what this guid (`D954291E-2A0B-460D-934E-DC6B0785DB48`) stands for but it is different to all others (`FAE04EC0-301F-11D3-BF4B-00C04F79EFBC`), is this intended?

Nitrox.Bootloader/Main.cs
```diff
@@ -0,0 +1,109 @@
...
+            {
+                return "Nitrox launcher path not set in AppData. Nitrox will not start.";
+            }
+            if (AppDomain.CurrentDomain.GetAssemblies().FirstOrDefault(a => a.FullName.StartsWith("QMod", StringComparison.Ordinal)) != null)
+            {
+                return "Nitrox will not start because QModManager is active.";
+            }
+
+            return null;
+        }
```
Jannify, line 85: Imo this whole return message thing and checking for it is more complicated as it needs to be.
```suggestion
        private static bool ValidateNitroxSetup()
        {
            if (nitroxLauncherDir.Value == null)
            {
                Console.WriteLine("Nitrox launcher path not set in AppData. Nitrox will not start.");
                return true;
            }
            if (AppDomain.CurrentDomain.GetAssemblies().FirstOrDefault(a => a.FullName.StartsWith("QMod", StringComparison.Ordinal)) != null)
            {
                Console.WriteLine("Nitrox will not start because QModManager is active.");
                return true;
            }

            return false;
        }
```
That way you won't need to check for null

NitroxServer-Subnautica/Serialization/SubnauticaServerProtoBufSerializer.cs
```diff
@@ -2,8 +2,10 @@
 using NitroxModel.DataStructures.GameLogic.Buildings.Rotation;
 using NitroxModel_Subnautica.DataStructures.GameLogic;
 using NitroxModel_Subnautica.DataStructures.GameLogic.Buildings.Rotation;
+using NitroxModel_Subnautica.DataStructures.Surrogates;
 using NitroxServer.Serialization;
 using ProtoBufNet.Meta;
+using UnityEngine;
```
Sunrunner37, line 8: To what extent do we still need this dependency in the sever? Just a question for future reference.

NitroxServer/GameLogic/Unlockables/PDAStateData.cs
```diff
@@ -10,10 +10,10 @@ namespace NitroxServer.GameLogic.Unlockables
     public class PDAStateData
     {
         [ProtoMember(1)]
-        public ThreadSafeCollection<TechTypeModel> UnlockedTechTypes { get; } = new ThreadSafeCollection<TechType>();
+        public ThreadSafeCollection<TechTypeModel> UnlockedTechTypes { get; } = new ThreadSafeCollection<TechTypeModel>();
```
Sunrunner37, line 13: NitroxTechType?

NitroxModel-Subnautica/Helper/VehicleModelFactory.cs
```diff
@@ -4,6 +4,7 @@
 using NitroxModel.DataStructures.GameLogic;
 using NitroxModel.DataStructures.Util;
 using NitroxModel.Packets;
+using NitroxModel_Subnautica.DataStructures;
 using NitroxModel_Subnautica.DataStructures.GameLogic;
 using UnityEngine;
 using NitroxTechType = NitroxModel.DataStructures.TechType;
```
Sunrunner37, line 10: Should we rename the class to be consistent with the others?

NitroxServer/GameLogic/Players/PlayerData.cs
```diff
@@ -82,7 +82,7 @@ public class PersistedPlayerData
             public ushort Id { get; set; }
 
             [ProtoMember(5)]
-            public Vector3 SpawnPosition { get; set; }
+            public NitroxVector3 SpawnPosition { get; set; }
```
Sunrunner37, line 85: If we were able to completely decouple the server then do we need to rename all of the types?  I think the types are only really used on the client when constructing objects for the server or marshalling data off of packets (so they can simply use the toDto())

Nitrox.Bootloader/Main.cs
Measurity: Your way has more code in it and couples the console type to the method. Returning the error message is simple enough imo.

NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs
Measurity: Honestly, I don't even try anymore because autoformatting is annoying to setup in a way to Nitrox has its code style (and also some code style don't seem to work properly for me).

Nitrox.sln
Measurity: These things are generated by MSBuild. I can only guess what it means but it might be different because it's a shared project (not a library/exe).

NitroxServer-Subnautica/Serialization/SubnauticaServerProtoBufSerializer.cs
Measurity: Purely to help protobuf not crash on unknown types when it is reading the game's asset files. I don't know what the best solution is yet to remove the dependency on Unity/UWE code.

NitroxServer/GameLogic/Players/PlayerData.cs
Measurity: Not needed but the types are used in other projects where the distinction is useful to not have to use the `using` renaming.

Measurity:
I went with `ToDto()` instead of `ToNitrox()` because ToNitrox sounded too ambiguous as to what the conversion is used for.

NitroxModel-Subnautica/DataStructures/DataExtensions.cs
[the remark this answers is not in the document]
acidscorch: What does "Dto" mean? 
Maybe better use .ToUnityVector() and .ToNitroxVector() ? Then it should be clear enough what this method concrete does and not be too ambigous to confound with other "ToNitrox" conversions.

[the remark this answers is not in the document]
Jannify: @acidscorch `ToDto` means `To data transfer object` which is often used for such "data conversation". `ToNitroxVector` is (imo) too complicated because you already use it on a type like `NitroxVector`. I would just wait for @Sunrunner37 s opinion.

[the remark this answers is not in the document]
acidscorch: @Jannify Ah yes, i understand. (Sorry i'm not very familiar with extension methods.) 
This Method is for converting a UnityVector to a NitroxVector, correct? And you already have a .ToUnity() Method. Maybe then use .FromUnity() for this?

PR #1099, opened 2020-05-14
Add player banning

Coding-Hen:


Measurity:
This feature is nice to have but I have mixed feelings about banning players by name because it isn't identifiable to a person at all and easily changed. If the server was less dependant on the player name to know which player is what then this feature makes sense. As is, I think it's best to wait and improve the server first.

Sunrunner37:
Yeah banning by name is not really effective.  We may want to revisit this to do it by IP (similar can be done for hardcore banning).

PR #1105, opened 2020-05-17
Added precursor teleporer sync

Coding-Hen:
Sync currently isnt persistant as with other "doors"

### DIFF
--- a/NitroxClient/GameLogic/Spawning/Metadata/PrecursorTeleporterMetadataProcessor.cs
+++ b/NitroxClient/GameLogic/Spawning/Metadata/PrecursorTeleporterMetadataProcessor.cs
@@ -1,0 +1,19 @@
+﻿using NitroxModel.DataStructures.GameLogic.Entities.Metadata;
+using NitroxModel.Logger;
+using UnityEngine;
+
+namespace NitroxClient.GameLogic.Spawning.Metadata
+{
+    public class PrecursorTeleporterMetadataProcessor : GenericEntityMetadataProcessor<PrecursorTeleporterMetadata>
+    {
+        public override void ProcessMetadata(GameObject gameObject, PrecursorTeleporterMetadata metadata)
+        {
+            Log.Info($"Received precursor teleporter metadata change for {gameObject.name} with data of {metadata}");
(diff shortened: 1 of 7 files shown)


NitroxClient/GameLogic/Spawning/Metadata/PrecursorTeleporterMetadataProcessor.cs
```diff
@@ -0,0 +1,19 @@
...
+using NitroxModel.Logger;
+using UnityEngine;
+
+namespace NitroxClient.GameLogic.Spawning.Metadata
+{
+    public class PrecursorTeleporterMetadataProcessor : GenericEntityMetadataProcessor<PrecursorTeleporterMetadata>
+    {
+        public override void ProcessMetadata(GameObject gameObject, PrecursorTeleporterMetadata metadata)
+        {
+            Log.Info($"Received precursor teleporter metadata change for {gameObject.name} with data of {metadata}");
```
Jannify, line 11: Maybe Log.Debug? The `PrecursorTeleporterMetadata`-parameter seems more like debug but idk....

NitroxClient/GameLogic/Spawning/Metadata/PrecursorTeleporterMetadataProcessor.cs
```diff
@@ -0,0 +1,19 @@
...
+
+            PrecursorTeleporter precursorTeleporter = gameObject.GetComponent<PrecursorTeleporter>();
```
Jannify, line 13: Maybe check if the component is found, just to be save.

NitroxClient/GameLogic/Spawning/Metadata/PrecursorTeleporterMetadataProcessor.cs
Coding-Hen: This change can be made but its simular to the KeypadMetadataProcessor sunrunner created, so was just following that convention

Sunrunner37:
I think this is same behavior as keypad is to starship door? let's align on that one first.

PR #1118, opened 2020-05-26
Sync battery values for vehicles

Jannify:


NitroxServer/Communication/Packets/Processors/BatteryValueChangedPacketProcessor.cs
```diff
@@ -0,0 +1,27 @@
...
+
+        public BatteryValueChangedPacketProcessor(PlayerManager playerManager, InventoryManager inventoryManager)
+        {
+            this.playerManager = playerManager;
+            this.inventoryManager = inventoryManager;
+        }
+
+        public override void Process(BatteryValueChanged packet, Player player)
+        {
+            Log.Info(packet);
```
Measurity, line 22: This one is going to print a ton with more people. Probably best to remove it.

acidscorch:
Hi @Jannify,

with a first look over this i have two questions:

* May this also be usable for other batterys (e.g. in base chargers)?

* As it seems to me, this could cause endless loops or at least interfering packets, because every client updates the state of the battery permanently independent of other clients. I don't mean the pingback by suppressing the packetsend when processing an incoming packet. I mean that the battery state changes permanently in all clients and so this could cause creating packets from two clients at the same time? Maybe also a TransientObjectLock is needed to ensure that only one client is updating the state? I'm not sure if it is realy needed, just thinking about the flow between the clients. 

Best regards

Jannify:
I tried it with inventory tools and they battery value is synced. As long as it gets loaded with the `InventoryManager` it should be persistent.

Sunrunner37:
I think we already had vehicle power syncing. However, the update to the subnautica engine likely broke it.  Can you please check out the PowerMonitor class and PowerLevelChanged packet?

Jannify:
> I think we already had vehicle power syncing. However, the update to the subnautica engine likely broke it.  Can you please check out the PowerMonitor class and PowerLevelChanged packet?

I dug into the `PowerMonitor` code and as far as I see it is not really usable/obsolete. It patches the `EnergyInterface` class which is "one level above" `EnergyMixin` which represents one batter source. For the server to save these battery values we would needed to go through each of these sources. The `PowerLevelChangedProcessor` also has a copied version of the power calculation from SN which isn't really needed anymore.
I would like to switch to my battery/energy-source patching because it's much simpler and covers the save issue but of course I am open for discussion :D

PR #1121, opened 2020-05-28
Use Nlog logging V3.0

Jannify:


NitroxServer-Subnautica/Program.cs
```diff
@@ -44,6 +44,7 @@ private static void Main(string[] args)
             AppDomain.CurrentDomain.ReflectionOnlyAssemblyResolve += CurrentDomainOnAssemblyResolve;
             ConfigureConsoleWindow();
             ConfigureCultureInfo();
+            Log.Setup();
```
MarijnS95, line 47: @Jannify This is called way too late. I ran into it while updating the Linux PR (which is hell in and off itself, capitalization changes inconsistently with _every_ patch, and there have only been a few) and it crashes out when logging this warning I added: https://github.com/SubnauticaNitrox/Nitrox/pull/995/commits/377a33ca5513bf009f357fca045cb2d2543efcbb#diff-574d6978d6eaff031e9d30089aac9314R59

It's not purely that code though, look right above in argument handing: there are `Log.Info`s there.

What do you reckon, shall we reorder it like https://github.com/SubnauticaNitrox/Nitrox/pull/995/commits/08f5bc19bd3c906e8fcbb3348c8ff3fe4fff817d? Does the client set up logging early enough?

Jannify: Yeah that makes sense. It was very late as I made the pr 😅 and probably missed this.

PR #1125, opened 2020-06-01
Fixing chat hint icon obscuring SN-GUI

Jannify:
- Make chatKeyHint disappear after first use.
- Update NitroxUnity to function with newest changes to project structure.


NitroxClient/Unity/Helper/AssetBundleLoader.cs
```diff
@@ -42,10 +42,22 @@ public static IEnumerator LoadUIAsset(string name, string canvasObjectName, bool
...
+
+        public static void DisableAsset(string name)
+        {
+            string oldPath = Path.Combine(assetRootFolder, name);
+            if (File.Exists(oldPath))
             {
-                callback.Invoke(assetRoot.gameObject);
+                File.Move(oldPath, oldPath + ".disabled");
             }
+
```
Measurity, line 60: You can use Unity's playerprefs to store information on the client. Moving files around should be avoided unless absolutely necessary because IO can easily fail due to PC configuration.

NitroxClient/GameLogic/ChatUI/PlayerChatManager.cs
```diff
@@ -78,14 +89,32 @@ public void SendMessage()
...
+                }));
+            }
+        }
+
+        //TODO: Has to be reworked if the config API for NitroxClient is finished.
+        private void DisableChatKeyHint()
+        {
+            chatKeyHint.GetComponentInChildren<Text>().CrossFadeAlpha(0, 1, false);
+            chatKeyHint.GetComponentInChildren<Image>().CrossFadeAlpha(0, 1, false);
+            PlayerPrefs.SetInt("Nitrox.chatUsed", 1);
```
Measurity, line 116: Would be good to have this in a property on this class. So you can just set/get it.

Jannify: Do you mean that the string `Nitrox.chatUsed` should be defined as a const?

Measurity: No, like this:
```cs
private bool ChatUsed
{
    get => PlayerPrefs.GetInt("Nitrox.chatUsed") == 1;
    set => PlayerPrefs.SetInt("Nitrox.chatUsed", value ? 1 : 0);
}
```

PR #1141, opened 2020-06-16
add DontDestroyOnLoad to scene debugger list

killzoms:


NitroxClient/Debuggers/SceneDebugger.cs
```diff
@@ -237,31 +237,43 @@ private void RenderTabScenes()
...
+
+                    Log.Info($"Current Scene: {currentScene.name} current Index {i}");
+
                     bool isSelected = selectedScene.IsValid() && currentScene == selectedScene;
                     bool isLoaded = currentScene.isLoaded;
 
                     using (new GUILayout.HorizontalScope("Box"))
                     {
-                        if (GUILayout.Button($"{(isSelected ? ">> " : "")}{i}: {path.TruncateLeft(35)}", isLoaded ? "sceneLoaded" : "label"))
+                        if (GUILayout.Button($"{(isSelected ? ">> " : "")}{i}: {(currentScene.name == "DontDestroyOnLoad" ? currentScene.name : path.TruncateLeft(35))}", isLoaded ? "sceneLoaded" : "label"))
```
Measurity, line 261: Recommend storing the `currentScene.name == ` condition as a bool and reuse that for performance/readability reasons.

NitroxClient/Debuggers/SceneDebugger.cs
```diff
@@ -237,31 +237,43 @@ private void RenderTabScenes()
...
-                for (int i = 0; i < SceneManager.sceneCountInBuildSettings; i++)
+                for (int i = 0; i < SceneManager.sceneCountInBuildSettings+1; i++)
                 {
-                    Scene currentScene = SceneManager.GetSceneByBuildIndex(i);
-                    string path = SceneUtility.GetScenePathByBuildIndex(i);
+                    Scene currentScene;
+                    string path = "";
+                    if (i == SceneManager.sceneCountInBuildSettings)
+                    {
+                        currentScene = NitroxBootstrapper.Instance.gameObject.scene; // bit of a hack
```
Measurity, line 246: Could this be done instead with

 - https://docs.unity3d.com/ScriptReference/SceneManagement.Scene-name.html
OR
 - https://docs.unity3d.com/540/Documentation/ScriptReference/SceneManagement.SceneManager.GetActiveScene.html

killzoms: I used the i == SceneManager.sceneCountInBuildSettings instead of string equality for some performance reasons, also GetActiveScene will likely return the main scene and will not include the Scenes that are loaded as Additive, SceneManager.sceneCountInBuildSettings is always going to be the same number in a release build so simply adding 1 to it and then checking if the indexer is equal to that number is acceptable, I would've just simply moved it outside of the loop however that would require moving the Using further up the method(out side of the for loop) and duplicating logic

PR #1142, opened 2020-06-20
Moved GitHub files to .github directory.

Jannify:
Moved GitHub files to .github for a less bloated front page.


Coding-Hen:
I agree with @MarijnS95 with this doesn't seem to me that it bloats and as someone who spends time working on other platforms like GitLab I agree these aren't github specific.

Jannify:
Yeah makes sense. Just wanted to discuss this option.

PR #1154, opened 2020-06-24
Building pr2

acidscorch:
* Multiple changes for building - extract from PR 1093 for code compliance, contributing guidelines and architecture persistence

* No warranty for changes, all untested and not compileable caused by missing components like litenetlib and ionic, everything is coded "dry" and will not be further fixed or optimized by me, feel free to take over and change what is needed


dartasen:
Hi, just a quick message to say that all the preprocessors #if for debug purposes aren't useful since we switched to NLog and Log.Debug aren't here no more while in release mode

````csharp
        [Conditional("DEBUG")]
        public static void Debug(string message) => logger.Debug(message);
        [Conditional("DEBUG")]
        public static void Debug(object message) => Debug(message?.ToString());
````

Swiff4:
Hey guys, how is it going? Can I help you somehow with this building topic? Acidscorch invested a lot of time in two different approaches but both did not receive further comments or reviews. Please let me know of if I can help. Thanks and regards.

dartasen:
This PR will never be updated, and we already have integrated some changes to the current master.
Thank's for the help, I'm closing this one.

PR #1168, opened 2020-07-15
Fix for Inventory deletion

### LINKED ISSUE #1003: Items not saving in inventories of either base, player or storage
I have this issues that when ever I close the server and start the server all the item in the inventories even storage all disapear, for some reason only the repair tool are remain and other contrustion like mobile station contrustion are still remain

killzoms:
fixes #1003 
closes #1066


NitroxPatcher/Patches/Dynamic/uGUI_OnApplicationQuit_Patch.cs
```diff
@@ -0,0 +1,25 @@
...
+
+        public static void Prefix()
+        {
+            IMultiplayerSession multiplayerSession = NitroxServiceLocator.LocateService<IMultiplayerSession>();
+            multiplayerSession.Disconnect();
+        }
+
+        public override void Patch(HarmonyInstance harmony)
+        {
+            PatchPrefix(harmony, TYPE.GetMethod(TARGET_METHOD, BindingFlags.NonPublic | BindingFlags.Instance));
```
Jannify, line 22: Maybe change so it's consistent with other patches code style wise.

Packages/manifest.json
```diff
@@ -0,0 +1,39 @@
+{
+  "dependencies": {
```
Jannify, line 2: What's this file doing here?

killzoms: it appeared out of the blue so I left it there...

killzoms:
@Coding-Hen due to the overall "fix" being tiny felt it would be better to include the Autofac update with the patch instead of a new PR, also updating Autofac has some added benefits as well, besides matching the .Net framework version

Packages/manifest.json
Jannify: That file is usually generated inside unity projects but here it's in the root folder and not `/NitroxUnity`. So it's save to delete.

PR #1175, opened 2020-07-20
Fix gamemode transmission while joining a game

dartasen:
- Server gamemode wasn't transmitted to a player while joining a game
- Moved server log informations from WorldPersistence.cs to Server.cs

### DIFF
--- a/NitroxModel/Server/ServerConfig.cs
+++ b/NitroxModel/Server/ServerConfig.cs
@@ -7,9 +7,10 @@ namespace NitroxModel.Server
 {
     public class ServerConfig
     {
+        private readonly ServerConfigItem<ServerGameMode> gameModeSetting;
         private readonly ServerConfigItem<bool> disableConsoleSetting, disableAutoSaveSetting;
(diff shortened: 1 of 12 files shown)


NitroxModel/Server/ServerConfig.cs
```diff
@@ -200,9 +191,6 @@ public ServerSerializerMode SerializerModeEnum
         public PlayerStatsData DefaultPlayerStats => new PlayerStatsData(oxygenSetting.Value, maxOxygenSetting.Value, healthSetting.Value, foodSetting.Value, waterSetting.Value, infectionSetting.Value);
         #endregion
 
-        public bool IsGameMode(string gamemode)
-        {
-            return GameMode == gamemode;
-        }
+        public bool IsHardcore => GameModeEnum == ServerGameMode.HARDCORE;
```
Coding-Hen, line 194: I believe this was discussed before to have with the function as to allow for easy implementation of other gamemode checks. I would therforew suggest that this is left the same but checks the enum instead

PR #1178, opened 2020-07-22
MS Store Fix

killzoms:
```csharp
UnauthorizedAccessException: Access to the path "C:\WINDOWS\System32\servers" is denied.
  at NitroxClient.MonoBehaviours.Gui.MainMenu.MainMenuMultiplayerPanel.AddServer (System.String name, System.String ip)
```

### DIFF
--- a/NitroxClient/MonoBehaviours/Gui/MainMenu/MainMenuMultiplayerPanel.cs
+++ b/NitroxClient/MonoBehaviours/Gui/MainMenu/MainMenuMultiplayerPanel.cs
@@ -1,4 +1,5 @@
 ﻿using System.Collections.Generic;
+using System;
 using System.IO;
 using System.Net;
 using System.Net.Sockets;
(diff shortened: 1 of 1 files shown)


NitroxClient/MonoBehaviours/Gui/MainMenu/MainMenuMultiplayerPanel.cs
```diff
@@ -22,7 +23,8 @@ public class MainMenuMultiplayerPanel : MonoBehaviour
         private GameObject multiplayerButton;
         private Transform savedGameAreaContent;
         public GameObject SavedGamesRef;
-        public string SERVER_LIST_PATH = Path.Combine(".", "servers");
+        private static readonly Uri url = new Uri(System.Reflection.Assembly.GetAssembly(typeof(uGUI)).CodeBase);
```
Measurity, line 26: Use the `Location` property instead: https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly.codebase?view=netcore-3.1

PR #1179, opened 2020-07-26
Clean up warnings

killzoms:
Produced with the help of JR Alterra's Zoey


Dessix:
For the merger's information: This suppresses 253 warnings present in the repository upon clean checkout, and also removes a cyclic reference from the MSBuild results.

PR #1181, opened 2020-07-26
Improved QoL of /back and /warp

dartasen:


NitroxServer/ConsoleCommands/WarpCommand.cs
```diff
@@ -1,25 +1,24 @@
...
-            Validate.IsTrue(args.Sender.HasValue, "This command can't be used by CONSOLE");
-
-            Player otherPlayer = args.Get<Player>(0);
-            args.Sender.Value.Teleport(otherPlayer.Position);
+            Player player = args.Get<Player>(0);
+            Player destination = args.Get<Player>(1);
 
-            SendMessage(args.Sender, $"Teleported to {otherPlayer.Name}");
+            player.Teleport(destination.Position);
+            SendMessage(player, $"Teleported to {destination.Name}");
```
Jannify, line 21: Maybe we can implement it so that you can warp yourself with one parameter and somebody else with two?

NitroxServer/ConsoleCommands/Abstract/CallArgs.cs
```diff
@@ -5,7 +5,7 @@ namespace NitroxServer.ConsoleCommands.Abstract
 {
     public abstract partial class Command
     {
-        public class CallArgs
+        public sealed class CallArgs
```
Sunrunner37, line 8: minor: aren't classes sealed by default?

dartasen: Nop they aren't, sealed forbid inheritance which cause some minor speed improvements)

PR #1184, opened 2020-07-30
Added server password input when required

### LINKED ISSUE #1126: Implement server password UI on the client
Currently the server password does nothing besides rejecting join attempts if set to something. Implement an UI that ask for a password and proper feedback if valid or not.

Coding-Hen:
Im almost certian that changes will be needed as im not 100% happy with the code myself BUT it does work as an implementation.

Feedback would be most welcome, haven't done any unity UI's before.

Resolves #1126

### DIFF
--- a/NitroxClient/MonoBehaviours/Gui/MainMenu/JoinServer.cs
+++ b/NitroxClient/MonoBehaviours/Gui/MainMenu/JoinServer.cs
@@ -38,7 +38,7 @@ namespace NitroxClient.MonoBehaviours.Gui.MainMenu
         private Rect serverPasswordWindowRect = new Rect(Screen.width / 2 - 250, 200, 500, 200);
         private bool shouldFocus;
         private bool showingPasswordWindow;
-        private bool passwordEntered = false;
+        private bool passwordEntered;
         public static GameObject SaveGameMenuPrototype { get; set; }
 
         private static MainMenuRightSide RightSideMainMenu => MainMenuRightSide.main;
@@ -433,10 +433,11 @@ namespace NitroxClient.MonoBehaviours.Gui.MainMenu
                     break;
 
                 case MultiplayerSessionConnectionStage.AWAITING_RESERVATION_CREDENTIALS:
-                    if (passwordEntered)
+                    if (multiplayerSession.SessionPolicy.RequiresServerPassword)
                     {
-                        OnJoinClick();
-                        break;
+                        Log.InGame("Waiting for Server Password Input...");
+                        showingPasswordWindow = true;
+                        shouldFocus = true;
                     }
                     Log.InGame("Waiting for user input...");
                     RightSideMainMenu.OpenGroup("Join Server");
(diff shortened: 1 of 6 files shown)


NitroxClient/MonoBehaviours/Gui/MainMenu/JoinServer.cs
```diff
@@ -393,7 +398,15 @@ private void OnJoinClick()
             SetCurrentPreference(playerName, playerColor);
 
             PlayerSettings playerSettings = new PlayerSettings(playerColor.ToDto());
-            AuthenticationContext authenticationContext = new AuthenticationContext(playerName);
+            AuthenticationContext authenticationContext;
+            if (!passwordEntered)
```
dartasen, line 402: minor: invert boolean

Coding-Hen:
> 
> 
> Imo, we should ask for the password before even asking name/color. For that we have the SessionPolicy that can help you, because they are the first data sent to the potential player who wants to join
> 
> If we look at `NitroxModel.Packets.MultiplayerSessionPolicy` we can see that there is a an old stub for RequiresServerPassword that need to be replaced as well
> 
> ```cs
>         public MultiplayerSessionPolicy(string correlationId, bool disableConsole)
>             : base(correlationId)
>         {
>             // This is done intentionally. It is only a stub for future extension.
>             RequiresServerPassword = false;
>             AuthenticationAuthority = MultiplayerSessionAuthenticationAuthority.SERVER;
>             DisableConsole = disableConsole;
> ```

I agree with what you are saying I missed this when looking at the implementation, have changed to use this.

Measurity:
Agreed with Dartasen that the password should be shown before any other input/ui. Code looks good.

Coding-Hen:
> 
> 
> Agreed with Dartasen that the password should be shown before any other input/ui. Code looks good.

Code now displays the popup upon request of session policy rather than on regection from the server

Measurity:
I mean that it should validate the password before you have a chance to enter the other information.

NitroxModel/Server/ServerConfig.cs
```diff
@@ -192,5 +192,7 @@ public ServerSerializerMode SerializerModeEnum
         #endregion
 
         public bool IsHardcore => GameModeEnum == ServerGameMode.HARDCORE;
+
+        public bool IsPasswordRequired => ServerPassword != "";
```
dartasen, line 196: minor : string.Empty

NitroxClient/MonoBehaviours/Gui/MainMenu/JoinServer.cs
```diff
@@ -33,6 +33,11 @@ public class JoinServer : MonoBehaviour
         private PlayerPreferenceManager preferencesManager;
         public string ServerIp = "";
         public int ServerPort;
+        private string serverPassword = "";
+        private Rect serverPasswordWindowRect = new Rect(Screen.width / 2 - 250, 200, 500, 200);
+        private bool shouldFocus;
+        private bool showingPasswordWindow;
+        private bool passwordEntered;
```
dartasen, line 40: Minor: 
- Is it possible to cleanup all of these ? Regroup static, public, private, ...
- string.Empty also

NitroxClient/MonoBehaviours/Gui/MainMenu/JoinServer.cs
```diff
@@ -567,5 +585,110 @@ private void InitializeButtonElements(RectTransform joinServerBackground, GameOb
...
+            GUISkinUtils.RenderWithSkin(GetGUISkin(),
+                () =>
+                {
+                    using (new GUILayout.VerticalScope("Box"))
+                    {
+                        using (new GUILayout.HorizontalScope())
+                        {
+                            GUILayout.Label("Password:");
+                            GUI.SetNextControlName("serverPasswordField");
+                            // 120 so users can't go too crazy.
```
dartasen, line 646: What does that comment refers to ?

NitroxModel/Server/ServerConfig.cs
Measurity: Imo this isn't usually worth to bring up in review. `""` is just as clear as `string.Empty`. They are also both equivalent in terms of what IL is generated.

Measurity (approved):

PR #1196, opened 2020-08-06
Fix BatchParsing path

dartasen:


NitroxClient/Unity/Helper/GameObjectHelper.cs
```diff
@@ -104,13 +105,36 @@ public static string GetHierarchyPath(this Component component)
...
+
             return builder.ToString();
         }
+
+        public static string GetFullName(this GameObject obj)
+        {
+            Stack<string> stack = new Stack<string>();
+            Transform transform = obj.transform;
+
+            while (transform != null)
```
Jannify, line 124: Maybe use the unity way? => `while (transform)`

PR #1201, opened 2020-08-07
Fixed type exception due to RELEASE mode issue

Measurity:


NitroxModel-Subnautica/Helper/SubnauticaMap.cs
```diff
@@ -10,17 +10,20 @@ public class SubnauticaMap : Map
...
-            new NitroxTechType(TechType.Beacon.ToString())
+            new NitroxTechType("Pipe"),
+            new NitroxTechType("Constructor"),
+            new NitroxTechType("Flare"),
+            new NitroxTechType("Gravsphere"),
+            new NitroxTechType("PipeSurfaceFloater"),
+            new NitroxTechType("SmallStorage"),
+            new NitroxTechType("CyclopsDecoy"),
+            new NitroxTechType("LEDLight"),
+            new NitroxTechType("Beacon")
```
Dessix, line 26: This could use `nameof`, since the names exactly match. It's the same as the original code but at compile time instead of runtime, so the type information is still available. This, of course, assumes that the names for each item will exactly match those in the `enum`, but that appears to be true, and can be made policy.
```suggestion
            new NitroxTechType(nameof(TechType.Pipe)),
            new NitroxTechType(nameof(TechType.Constructor)),
            new NitroxTechType(nameof(TechType.Flare)),
            new NitroxTechType(nameof(TechType.Gravsphere)),
            new NitroxTechType(nameof(TechType.PipeSurfaceFloater)),
            new NitroxTechType(nameof(TechType.SmallStorage)),
            new NitroxTechType(nameof(TechType.CyclopsDecoy)),
            new NitroxTechType(nameof(TechType.LEDLight)),
            new NitroxTechType(nameof(TechType.Beacon))
```

NitroxModel-Subnautica/Helper/SubnauticaMap.cs
```diff
@@ -10,17 +10,20 @@ public class SubnauticaMap : Map
         private const int BATCH_SIZE = 160;
         private const int SKYBOX_METERS_ABOVE_WATER = 160;
 
-        public static readonly List<NitroxTechType> GLOBAL_ROOT_TECH_TYPES = new List<NitroxTechType>()
+        /// <summary>
+        ///     These types can't be statically typed because in RELEASE mode it cannot load the TechType type info.
```
Dessix, line 14:
```suggestion
        ///     TechType can't be introspected at runtime in RELEASE mode because its reflection info is elided.
```

PR #1203, opened 2020-08-07
Made versioning check more robust

dartasen:
This will allow us to bump the assembly version before releasing without having the popup

### DIFF
--- a/NitroxModel/Helper/WebHelper.cs
+++ b/NitroxModel/Helper/WebHelper.cs
@@ -26,21 +26,19 @@ namespace NitroxModel.Helper
             req.Method = "GET";
             req.UserAgent = "Nitrox";
             req.ContentType = "application/json";
-            req.Timeout = 1000;
-
-            Version version = new Version();
+            req.Timeout = 5000;
 
             try
             {
                 using (StreamReader sr = new StreamReader(req.GetResponse().GetResponseStream()))
                 {
                     string json = sr.ReadToEnd();
-                    Regex rx = new Regex("\\\"name\\\":\\\"(([^\\\"])*)\\\"");
+                    Regex rx = new Regex(@"""name"":""([^""]*)""");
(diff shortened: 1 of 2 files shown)


NitroxModel/Helper/WebHelper.cs
```diff
@@ -18,15 +20,15 @@ public static string GetPublicIP()
...
         }
 
-        public static string GetNitroxLatestVersion()
+        public static Version GetNitroxLatestVersion()
         {
             HttpWebRequest req = WebRequest.Create("https://api.github.com/repos/SubnauticaNitrox/Nitrox/releases/latest") as HttpWebRequest;
             req.Method = "GET";
             req.UserAgent = "Nitrox";
             req.ContentType = "application/json";
             req.Timeout = 1000;
```
Measurity, line 29: Only 1 second seems on the short side.
```suggestion
            req.Timeout = 5000;
```

NitroxModel/Helper/WebHelper.cs
```diff
@@ -18,15 +20,15 @@ public static string GetPublicIP()
...
 
-            string version = string.Empty;
+            Version version = new Version();
```
Measurity, line 31: It's possible that this instance won't be used so let's init to null and check for null later.
```suggestion
            Version version = null;
```

NitroxModel/Helper/WebHelper.cs
```diff
@@ -38,12 +40,13 @@ public static string GetNitroxLatestVersion()
...
                     }
                 }
             }
-            catch
+            catch (Exception ex)
             {
+                Log.Error(ex, "WebHelper : Error while fetching nitrox latest version on GitHub");
             }
 
             return version;
```
Measurity, line 52:
```suggestion
            return version ?? new Version();
```

PR #1215, opened 2020-08-13
Removed logging from metadata processors

Coding-Hen:


Jannify:
We discussed the inconsistent logging on discord. Here is the [link](https://discord.com/channels/525437013403631617/528915769107415040/743572798835785850) if you want to read it. Maybe you can bring this to a unified solution.

PR #1219, opened 2020-08-17
Added launcher path logging to help troubleshooting

Measurity:


Measurity:
Gonna try use a different loading tactic instead of this launcher path stuff.

PR #1221, opened 2020-08-18
Improved Nitrox loading logging

Measurity:
Also fixes issue where Launcher stays stuck when you try to close it due to trying to win32 postmessage on a dead process. So now just set it to `null` after disposing and check for that.


NitroxPatcher/Main.cs
```diff
@@ -28,16 +32,34 @@ public static void Execute()
...
+                Log.Error($"Failed to load one or more dependency types for Nitrox. Assembly: {ex.Types.FirstOrDefault()?.Assembly.FullName ?? "unknown"}");
+                foreach (Exception loaderEx in ex.LoaderExceptions)
+                {
+                    Log.Error(loaderEx);
+                }
+                throw;
+            }
+            catch (Exception)
+            {
+                Log.Error("Error initializing and loading dependencies.");
```
Jannify, line 60: Is there a "while" missing?

Jannify (approved):

NitroxPatcher/Main.cs
```diff
@@ -28,16 +32,34 @@ public static void Execute()
...
+                Log.Error($"Failed to load one or more dependency types for Nitrox. Assembly: {ex.Types.FirstOrDefault()?.Assembly.FullName ?? "unknown"}");
+                foreach (Exception loaderEx in ex.LoaderExceptions)
+                {
+                    Log.Error(loaderEx);
+                }
+                throw;
+            }
+            catch (Exception)
+            {
+                Log.Error("Error while initializing and loading dependencies.");
```
Dessix, line 60:
```suggestion
            catch (Exception ex)
            {
                Log.Error(ex, "Error while initializing and loading dependencies.");
```

Use `public static void Error(Exception ex, string message) => logger.Error(ex, message);` method variant.

NitroxPatcher/Main.cs
```diff
@@ -28,16 +32,34 @@ public static void Execute()
         {
             Log.Setup(true);
             Optional.ApplyHasValueCondition<Object>(o => (bool)o);
```
Dessix, line 34: Potentially explicitly use `UnityEngine.Object` and remove the `using Object = UnityEngine.Object` to avoid possible conflation with C#'s `object`?

Coding-Hen (approved):

dartasen (approved):

PR #1230, opened 2020-08-29
Added sync for incubator

Coding-Hen:
TODO: Make server aware of entity currently throws entity unknown to the server error when sending across the metadata

### DIFF
--- a/NitroxClient/GameLogic/Spawning/Metadata/IncubatorMetadataProcessor.cs
+++ b/NitroxClient/GameLogic/Spawning/Metadata/IncubatorMetadataProcessor.cs
@@ -1,0 +1,22 @@
+﻿using NitroxModel.DataStructures.GameLogic.Entities.Metadata;
+using NitroxModel.Helper;
+using UnityEngine;
+
+namespace NitroxClient.GameLogic.Spawning.Metadata
+{
+    public class IncubatorMetadataProcessor : GenericEntityMetadataProcessor<IncubatorMetadata>
+    {
+        public override void ProcessMetadata(GameObject gameObject, IncubatorMetadata metadata)
+        {
+            Incubator incubator = gameObject.GetComponent<Incubator>();
+            incubator.powered = metadata.Powered;
+            incubator.hatched = metadata.Hatched;
+
+            bool terminalActive = (bool)incubator.ReflectionGet("terminalActive");
+            incubator.computerTerminal.SetActive(terminalActive);
+            incubator.ReflectionSet("emissiveIntensity", (float)(terminalActive ? 1 : 0));
+            incubator.ReflectionCall("UpdateMaterialsEmissive");
+            incubator.terminalLight.intensity = (float)incubator.ReflectionGet("emissiveIntensity") * 1.85f;
+        }
+    }
+}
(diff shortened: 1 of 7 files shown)


NitroxClient/GameLogic/Spawning/Metadata/IncubatorMetadataProcessor.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        public override void ProcessMetadata(GameObject gameObject, IncubatorMetadata metadata)
+        {
+            Incubator incubator = gameObject.GetComponent<Incubator>();
+            incubator.powered = metadata.Powered;
+            incubator.hatched = metadata.Hatched;
+
+            bool terminalActive = (bool)incubator.ReflectionGet("terminalActive");
+            incubator.computerTerminal.SetActive(terminalActive);
+            incubator.ReflectionSet("emissiveIntensity", (float)(terminalActive ? 1 : 0));
```
Dessix, line 17:
```suggestion
            incubator.ReflectionSet("emissiveIntensity", (terminalActive ? 1f : 0f));
```

NitroxClient/GameLogic/Spawning/Metadata/IncubatorMetadataProcessor.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    public class IncubatorMetadataProcessor : GenericEntityMetadataProcessor<IncubatorMetadata>
+    {
+        public override void ProcessMetadata(GameObject gameObject, IncubatorMetadata metadata)
+        {
+            Incubator incubator = gameObject.GetComponent<Incubator>();
+            incubator.powered = metadata.Powered;
+            incubator.hatched = metadata.Hatched;
+
+            bool terminalActive = (bool)incubator.ReflectionGet("terminalActive");
```
Dessix, line 15:
```suggestion
            var terminalActive = incubator.ReflectionGet<bool>("terminalActive");
```

NitroxClient/GameLogic/Spawning/Metadata/IncubatorMetadataProcessor.cs
```diff
@@ -0,0 +1,22 @@
...
+            incubator.computerTerminal.SetActive(terminalActive);
+            incubator.ReflectionSet("emissiveIntensity", (float)(terminalActive ? 1 : 0));
+            incubator.ReflectionCall("UpdateMaterialsEmissive");
+            incubator.terminalLight.intensity = (float)incubator.ReflectionGet("emissiveIntensity") * 1.85f;
```
Dessix, line 19:
```suggestion
            incubator.terminalLight.intensity = incubator.ReflectionGet<float>("emissiveIntensity") * 1.85f;
```

NitroxClient/GameLogic/Spawning/Metadata/IncubatorMetadataProcessor.cs
Coding-Hen: Unless I am mistaken I do not believe this works due to how the ReflectionHelper returns the object type?

Dessix: My apologies, retracting this one and the other; I misinterpreted the intent of T in the `ReflectionHelper`s.

```
public static object ReflectionGet<T>(this T o, string fieldName, bool isPublic = false, bool isStatic = false)
//                  ^ expected T
```

Coding-Hen: Would be cool if we could do this though, I would have thought even with T though would it not have issues with say in this case passing in bool but the class being passed in as Incubator would that have an issue with the return type?

Sunrunner37:
Looks like this is still in draft.  What is remaining on this?

Coding-Hen:
Fixed with 7a949b7214d6df20838e8e28a60a373b36450a30

PR #1236, opened 2020-09-05
Rocket improvements

dartasen:
- [x] BUG : Buildbots are not sent when building the rocket (for synced clients only)
- [x] BUG : On reload, a finished rocket was rollback at his pre-final state
- [ ] ADDED : Sync for preflight checks switches
- [x] ADDED : Basic sync for rocket launch (Will need some brainstorms to improve the way we want to manage the end of the Nitrox gameplay)

### DIFF
--- a/NitroxClient/Communication/Packets/Processors/RocketEndLaunchProcessor.cs
+++ b/NitroxClient/Communication/Packets/Processors/RocketEndLaunchProcessor.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections;
+﻿using System;
+using System.Collections;
 using NitroxClient.Communication.Packets.Processors.Abstract;
 using NitroxClient.MonoBehaviours;
 using NitroxClient.Unity.Helper;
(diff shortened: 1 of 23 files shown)


NitroxModel-Subnautica/DataStructures/GameLogic/NeptuneRocketModel.cs
```diff
@@ -20,6 +20,9 @@ public class NeptuneRocketModel : VehicleModel
         [ProtoMember(3)]
         public bool ElevatorUp { get; set; }
 
+        [ProtoMember(4)]
+        public ThreadSafeCollection<PreflightCheck> PreflightChecks { get; set; }
```
Sunrunner37, line 24: ThreadSafeCollection?

NitroxClient/Communication/Packets/Processors/RocketEndLaunchProcessor.cs
```diff
@@ -0,0 +1,40 @@
...
+                launchRocket.StartCoroutine(endSequence);
+
+                HandReticle.main.RequestCrosshairHide();
+            }
+            catch (Exception ex)
+            {
+                Log.Error(ex, "An error occured while processing RocketEndLaunch packet");
+                Log.InGame("Error while processing a rocket launch packet :(");
+                throw;
+            }
```
Sunrunner37, line 37: I believe we automatically log at top level (same with below)

NitroxServer-Subnautica/Communication/Packets/Processors/RocketEndLaunchProcessor.cs
```diff
@@ -0,0 +1,23 @@
...
+
+        public RocketEndLaunchProcessor(PlayerManager playerManager)
+        {
+            this.playerManager = playerManager;
+        }
+
+        public override void Process(RocketEndLaunch packet, NitroxServer.Player player)
+        {
+            Log.Warn($"Starting end sequence with rocket {packet.Id}");
+            playerManager.SendPacketToOtherPlayers(packet, player);
```
Sunrunner37, line 20: Is this needed?  I think the default behavior is to echo packets to other players.  Is there some kind of state we should update?

NitroxModel-Subnautica/DataStructures/GameLogic/NeptuneRocketModel.cs
Measurity: Works better with protobuf because it's a custom type from us.

NitroxClient/Communication/Packets/Processors/RocketEndLaunchProcessor.cs
dartasen: Was just for testing purpose and will be removed in the final state, I needed to have a visual information when an error occurred, instead of having to check log files every game session

Issue #1235, opened 2020-09-03
Feature request: Pause game when no one is online


Aminity:
Hi and thank you for considering this request.

My request is for a config option to enable or disable when a server has been started to pauze progress/time if no one is online to play the game.

I'm no expert on coding, but I imagine the following things should be considered or done if this is implemented:

- Add a config option True or False for this feature(default being False if to keep inline with current implementation)
- Pause game after startup by default when above is True when server is started.

All suggestions below assume the feature is enabled, e.g. if admin paused the game, a player joining shouldn't trigger unpausing the game.
 - Unpause game when a player has loaded into the game.
- Pause game when:  last player connection drops, leaves.
- Check if pausing server impacts chat/commands system (e.g. does pausing game make communication impossible or the use of other commands, do certain commands need to be deactivated during pause because using them would break the game)
- Check if autosaves are impacted
- - Disable autosaves after the first succesful save while the game is paused
- - Re enable autosaves after unpausing game

I hope this contributes to the success of the project.

Kind regards,
Aminity

PR #1239, opened 2020-09-07
Added the ability to pause server when all players disconnect

Coding-Hen:
Added saving and reanable of events and server time when player disconnects/connects
Moved connection reanable to actually work as before would only work on second player connecting.
Resolves: #1235

### DIFF
--- a/NitroxServer/GameLogic/EventTriggerer.cs
+++ b/NitroxServer/GameLogic/EventTriggerer.cs
@@ -36,11 +36,11 @@ namespace NitroxServer.GameLogic
                 AuroraExplosionTime = RandomNumber(2.3d, 4d) * 1200d * 1000d; //Time.deltaTime returns seconds so we need to multiply 1000
             }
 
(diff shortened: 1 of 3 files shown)


NitroxServer/GameLogic/EventTriggerer.cs
```diff
@@ -55,12 +57,13 @@ private Timer CreateTimer(double time, StoryEventType eventType, string key)
             Timer timer = new Timer();
             timer.Elapsed += delegate
             {
-                Log.Info("Triggering event type " + eventType.ToString() + " at time " + time.ToString() + " with param " + key.ToString());
+                Log.Info("Triggering event type " + eventType + " at time " + time + " with param " + key);
```
Jannify, line 60: Should be `Log.Info($"Trigger event type {event Type}...`.

Jannify (changes requested):

NitroxServer/GameLogic/EventTriggerer.cs
Dessix:
```suggestion
                Log.Info($"Triggering event type {eventType} at time {time} with param {key}");
```

PR #1254, opened 2020-10-02
Implement Catch-all

killzoms:


Coding-Hen (approved):

NitroxPatcher/Main.cs
```diff
@@ -30,6 +30,19 @@ public static class Main
         public static void Execute()
         {
             Log.Setup(true);
+            Log.Info($"Using Nitrox version {Assembly.GetExecutingAssembly().GetName().Version} built on {File.GetCreationTimeUtc(Assembly.GetExecutingAssembly().Location)}");
```
Coding-Hen, line 33: Do we really need to log the time it was built, the version I would have thought would be enough?

Coding-Hen (approved):

Dessix (approved):

PR #1256, opened 2020-10-04
Fix SpawnAnyPendingChildren()

dartasen:
NitroxClient.GameLogic.Entities:
```csharp
private void SpawnAnyPendingChildren(Entity entity) {
           if (pendingParentEntitiesByParentId.TryGetValue(entity.Id, out  List<Entity> pendingEntities))
           {
                Optional<GameObject> parent = NitroxEntity.GetObjectFrom(entity.Id);

                foreach (Entity child in pendingEntities)
                {
                    Spawn(entity, parent); //<--- we shouldn't be trying to spawn parent entity n times
                }

                pendingParentEntitiesByParentId.Remove(entity.Id);
            }
}
```

PS: This doesn't fix structures loading, we need to investigate this bug by looking at the Pending entity system


Dessix (approved):
Looks good to me- I'm also a fan of the updates to inline out-var declaration and expression-bodied methods.
I am slightly concerned seeing how we track all entity IDs, and will have to look into whether or not we clean the list up elsewhere as entities are removed, but this PR doesn't introduce or substantially modify that behaviour, so: approved.

NitroxClient/GameLogic/Entities.cs
```diff
@@ -174,15 +161,10 @@ private void AddPendingParentEntity(Entity entity)
             pendingEntities.Add(entity);
         }
 
-        public bool WasSpawnedByServer(NitroxId id)
-        {
-            return alreadySpawnedIds.Contains(id);
-        }
+        private Int3 ToInt3(NitroxModel.DataStructures.Int3 int3) => new Int3(int3.X, int3.Y, int3.Z);
```
Jannify, line 164: This should be transferred to [DataExtensions](https://www.github.com/SubnauticaNitrox/Nitrox/tree/master/NitroxModel-Subnautica%2FDataStructures%2FDataExtensions.cs)

Jannify (changes requested):

NitroxModel-Subnautica/DataStructures/GameLogic/Buildings/Rotation/Metadata/BaseModuleRotationMetadata.cs
```diff
@@ -23,13 +23,13 @@ protected BaseModuleRotationMetadata() : base(typeof(BaseAddModuleGhost))
...
         {
-            Cell = cell.Model();
+            Cell = cell.ToDto();
             Direction = direction;
         }
 
         public override string ToString()
         {
-            return "[BaseModuleRotationMetadata cell: " + Cell + " direction: " + Direction + " ]";
+            return $"[BaseModuleRotationMetadata - Cell: {Cell}, Direction: {Direction}]";
```
Coding-Hen, line 32: I would suggest that this change is made inline with AnchoredFaceRotationMetadata
return $"[AnchoredFaceRotationMetadata Cell: {Cell} Direction: {Direction} FaceType: {FaceType}]";

Coding-Hen (changes requested):

NitroxModel/DataStructures/NitroxInt3.cs
```diff
@@ -0,0 +1,156 @@
...
+                   Y == int3.Y &&
+                   Z == int3.Z;
+        }
+
+        public override int GetHashCode()
+        {
+            int hashCode = -307843816;
+            hashCode = hashCode * -1521134295 + X.GetHashCode();
+            hashCode = hashCode * -1521134295 + Y.GetHashCode();
+            hashCode = hashCode * -1521134295 + Z.GetHashCode();
```
Dessix, line 56: This will throw with overflow errors under certain configurations if you do not mark it as [`unchecked`](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/unchecked), and will produce an overflow lint failure from SONAR.
```suggestion
            unchecked {
                hashCode = hashCode * -1521134295 + X.GetHashCode();
                hashCode = hashCode * -1521134295 + Y.GetHashCode();
                hashCode = hashCode * -1521134295 + Z.GetHashCode();
            }
```
Additionally, the factors on X, Y, and Z should use different component primes if possible, to avoid hash equivalency for axially orthogonal points.

NitroxModel/DataStructures/NitroxInt3.cs
```diff
@@ -0,0 +1,156 @@
...
+
+namespace NitroxModel.DataStructures
+{
+    /*
+     * NitroxInt3 allows NitroxModel to be decoupled from Assembly-csharp-firstpass
+     * as it is common code between subnautica and below zero
+     */
+    [Serializable]
+    [ProtoContract]
+    public class NitroxInt3
```
Dessix, line 13: A descriptive name like `Vec3I`may be preferable to smurf naming conventions?

NitroxModel/DataStructures/NitroxInt3.cs
```diff
@@ -0,0 +1,156 @@
...
+        {
+            return new NitroxInt3(u.X + s, u.Y + s, u.Z + s);
+        }
+
+        public static NitroxVector3 operator +(NitroxInt3 u, NitroxVector3 v)
+        {
+            return new NitroxVector3(u.X + v.X, u.Y + v.Y, u.Z + v.Z);
+        }
+        
+        public static implicit operator NitroxVector3(NitroxInt3 v)
```
Dessix, line 120: If `NitroxVector3` is implicitly convertible (via `implicit operator NitroxVector3`), consider adding equality operators against this type to avoid unnecessary conversion allocations where possible.

Additionally, consider adding comparisons and an implicit conversion to the non-DTO `Int3` type.

Dessix (changes requested):
A few concerns- the first of which is with an overflow in the hash function.

This PR has taken on too much work in one pass; The `Int3` work should be separate from the original PR, as they are orthogonal in purpose, and this work was only attached because someone saw ToInt3, not realizing that this PR just updated its syntax, and was not introducing the function or the type.

Jannify:
> A few concerns- the first of which is with an overflow in the hash function.
> 
> This PR has taken on too much work in one pass; The `Int3` work should be separate from the original PR, as they are orthogonal in purpose, and this work was only attached because someone saw ToInt3, not realizing that this PR just updated its syntax, and was not introducing the function or the type.

Yeah, I just thought of simple moving the method to DataExtension, not about all the name changing and so on.
And yes it should be split into two prs.

NitroxModel/DataStructures/NitroxInt3.cs
Jannify: Normally we just stick "Nitrox" in front of the SN name which would fit in this case ("Int3").

Dessix (approved):
Aside from ambiguity in why constructors are protected or not, I approve.

NitroxClient/GameLogic/Entities.cs
```diff
@@ -7,7 +7,6 @@
 using NitroxModel.DataStructures.GameLogic.Entities.Metadata;
 using NitroxModel.DataStructures.Util;
 using NitroxModel.Helper;
-using NitroxModel.Logger;
```
MarijnS95, line 10: This breaks the `Log.Error` inside `DEBUG` down below, at least on Linux:

```
  GameLogic/Entities.cs(144,17): error CS0103: The name 'Log' does not exist in the current context [/home/marijn/Code/C#/Nitrox/NitroxClient/NitroxClient.csproj]
```

Am I doing something stupid or does no-one ever build `DEBUG` anymore?

Coding-Hen: Nope this is a known problem fixed in #1260 Apparently this was due to the ide building in release doesnt see this as a build error unlike within the debug build due to the debug tags the rap the log in this file.

MarijnS95: Ah great, thanks for showing that!

PR #1257, opened 2020-10-05
Used helper to create vehicleMovementData to remove error message

Coding-Hen:
Fixes the "Got exosuit vehicle but no ExosuitMovementData" when exosuit is spawned in.

Really not sure if this is something that we want or the right way to approach it. In my mind this is semi not an issue anyway and if we even need to log this as it is an unlikely error and only stops the arm positions from being set.

### DIFF
--- a/NitroxClient/GameLogic/Vehicles.cs
+++ b/NitroxClient/GameLogic/Vehicles.cs
@@ -523,7 +523,18 @@ namespace NitroxClient.GameLogic
         {
             yield return new WaitForSeconds(cooldown);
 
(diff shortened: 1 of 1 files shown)


Sunrunner37:
This is kinda a hack as it sends exosuit for all vehicle types.  Not sure if we even need to send this packet.  If we do, it might be better to assign spawning player a transient lock and have them update the position instead.

Coding-Hen:
> 
> 
> This is kinda a hack as it sends exosuit for all vehicle types. Not sure if we even need to send this packet. If we do, it might be better to assign spawning player a transient lock and have them update the position instead.

I believe even with this hack that it will still send the correct type of movement data as it is still dependant on the type that is input. I do get your point though I'm not sure will do some testing, it was once upon a time to stop it getting stuck upon relog.

PR #1262, opened 2020-10-17
Update dependencies

Coding-Hen:
Updated all current dependencies to lastest working version
Unless I am mistake autofac is no longer supported above this version due to framework incompatibilities leading to loading errors


Jannify:
Is there something what is stopping this from getting merged?

PR #1265, opened 2020-10-19
Int3 refactor

dartasen:
* Since `Int3` is a "DTO" object, the name should be clear that it's a nitrox related object and not an `Int3 `from `Subnautica`
* Removed `.Global()` and `.Model()` extensions to match the `.ToDto()` `.ToUnity()` style


NitroxModel-Subnautica/Helper/VehicleModelFactory.cs
```diff
@@ -15,13 +15,13 @@ public static VehicleModel BuildFrom(NitroxTechType techType, NitroxId construct
...
                 case TechType.Exosuit:
-                    return new ExosuitModel(techType, constructedItemId, position, rotation, interactiveChildIdentifiers, Optional.Empty, name, hsb, health);
+                    return new ExosuitModel(techType, constructedItemId, position, rotation, interactiveChildIdentifiers, dockingBayId, name, hsb, health);
                 case TechType.Cyclops:
-                    return new CyclopsModel(techType, constructedItemId, position, rotation, interactiveChildIdentifiers, Optional.Empty, name, hsb, health);
+                    return new CyclopsModel(techType, constructedItemId, position, rotation, interactiveChildIdentifiers, dockingBayId, name, hsb, health);
                 case TechType.RocketBase:
-                    return new NeptuneRocketModel(techType, constructedItemId, position, rotation, interactiveChildIdentifiers, Optional.Empty, name, hsb, health);
+                    return new NeptuneRocketModel(techType, constructedItemId, position, rotation, interactiveChildIdentifiers, dockingBayId, name, hsb, health);
```
Jannify, line 24: Could you say/investigate if this change breaks something down the line?

NitroxModel/DataStructures/NitroxInt3.cs
```diff
@@ -0,0 +1,153 @@
...
+
+namespace NitroxModel.DataStructures
+{
+    /*
+     * NitroxInt3 model to allow NitroxModel to be decoupled from Assembly-csharp-firstpass
+     * as it is common between subnautica and below zero.
+     */
+    [Serializable]
+    [ProtoContract]
+    public class NitroxInt3
```
Jannify, line 13: What about making this a struct like the Sn version? 
grafik

Dessix: Agreed- this should only be merged once it's not going to cause huge amounts of heap allocations to use it.

NitroxModel-Subnautica/Helper/VehicleModelFactory.cs
dartasen: Sure!
My change doesn't do anything at all, since this factory is only used by `NitroxClient.GameLogic.Vehicles.BuildVehicleModelFrom(GameObject, TechType)`

```csharp
 return VehicleModelFactory.BuildFrom(
                    techType.ToDto(),
                    constructedObjectId,
                    gameObject.transform.position.ToDto(),
                    gameObject.transform.rotation.ToDto(),
                    childIdentifiers,
                    Optional.Empty, //<--------------
                    name,
                    hsb ?? VehicleHelper.GetDefaultColours(techType), //Shouldn't be null now, but just in case
                    health
                );
```

And we're only filling the "dockingBayId" with an Optional.Empty

Jannify (approved):

Dessix (approved):

PR #1272, opened 2020-10-31
Fixed connection via domain name

Coding-Hen:
Reworked ipv4/ipv6 and domain logic.
Changed format of server file

### DIFF
--- a/NitroxClient/MonoBehaviours/Multiplayer.cs
+++ b/NitroxClient/MonoBehaviours/Multiplayer.cs
@@ -19,7 +19,6 @@ using NitroxModel.Logger;
 using NitroxModel.Packets;
 using NitroxModel.Packets.Processors.Abstract;
 using NitroxModel_Subnautica.Helper;
-using NitroxModel_Subnautica.Logger;
 using UnityEngine;
 using UnityEngine.SceneManagement;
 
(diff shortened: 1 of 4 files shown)


Measurity (changes requested):

Measurity (changes requested):
As discussed in Discord.

Add extra field to "Add server" prompt for port (fill out port as 11000 by default).
Change server save format to store ip and port separately.

NitroxClient/MonoBehaviours/Multiplayer.cs
```diff
@@ -77,7 +76,6 @@ public static IEnumerator LoadAsync()
 
         public void Awake()
         {
-            Log.InGameLogger = new SubnauticaInGameLogger();
```
dartasen, line 80: Not sure to understand why you did remove this logger init ?

Measurity (approved):
Works nicely.

Only minor thing that Log.InGameLogger init is better suited in `NitroxPatcher.Main.Execute` or close to it than hidden away in MainMenuMultiplayerPanel: https://github.com/SubnauticaNitrox/Nitrox/pull/1272/files#diff-8b008cc9cc06ae159843c594eb48d03feea944f2b508d1c57b39cc7326f41b75R37

Measurity (approved):

dartasen (approved):

PR #1283, opened 2020-11-17
Gun fixed

killzoms:
made so that progress isn't lost if I disappear/get distracted

Current Status:

- [x] Gun fully Spawning
- [ ] FPS stabilized
    - [ ] locate key source of FPS issues


killzoms:
Average FPS around the Gun = 2-3 FPS

PR #1286, opened 2020-11-21
Initial Infection not set to Cured by default.

MrPurple6411:
Set the default infection level to 0.1f to match the value set by vanilla game on first launch.

End result:

New players now start the game as NOT Infected but also Not Cured as in the vanilla game.

### DIFF
--- a/NitroxClient/GameLogic/InitialSync/PlayerInitialSyncProcessor.cs
+++ b/NitroxClient/GameLogic/InitialSync/PlayerInitialSyncProcessor.cs
@@ -66,17 +66,14 @@ namespace NitroxClient.GameLogic.InitialSync
         {
             if (statsData != null)
             {
-                using (packetSender.Suppress<PlayerStats>())
+                Player.main.oxygenMgr.AddOxygen(statsData.Oxygen);
+                Player.main.liveMixin.health = statsData.Health;
+                Player.main.GetComponent<Survival>().food = statsData.Food;
+                Player.main.GetComponent<Survival>().water = statsData.Water;
+                Player.main.infectedMixin.SetInfectedAmount(statsData.InfectionAmount);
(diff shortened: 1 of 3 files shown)


NitroxServer-Subnautica/App.config
```diff
@@ -16,7 +16,7 @@
         <add key="StartHealth" value="80" />
         <add key="StartFood" value="50.5" />
         <add key="StartWater" value="90.5" />
-        <add key="StartInfection" value="0" />
+        <add key="StartInfection" value="0.1" />
```
Sunrunner37, line 19: Let's add a comment to both of these specifying why 0.1 is required.

MrPurple6411: Left a comment at the second location as the App.config cannot have comments in it.

NitroxClient/GameLogic/InitialSync/PlayerInitialSyncProcessor.cs
```diff
@@ -66,17 +66,14 @@ private void SetPlayerStats(PlayerStatsData statsData)
...
-                    Player.main.oxygenMgr.AddOxygen(statsData.Oxygen);
-                    Player.main.liveMixin.health = statsData.Health;
-                    Player.main.GetComponent<Survival>().food = statsData.Food;
-                    Player.main.GetComponent<Survival>().water = statsData.Water;
-                    Player.main.infectedMixin.SetInfectedAmount(statsData.InfectionAmount);
-                    if (statsData.InfectionAmount > 0f)
-                    {
-                        Player.main.infectionRevealed = true;
-                    }
+                    Player.main.infectionRevealed = true;
```
Jannify, line 76: If you look into Subnautica code this can probably lead to the "Infection-Reveal-Sequence" not being played now that we changed the initial value.
grafik

PR #1289, opened 2020-11-21
Add April Fools logic

killzoms:
Note: Only replaces initially spawned Reefbacks does not touch ones already in the world

### DIFF
--- a/NitroxClient/GameLogic/Spawning/ReefbackEntitySpawner.cs
+++ b/NitroxClient/GameLogic/Spawning/ReefbackEntitySpawner.cs
@@ -1,4 +1,5 @@
-﻿using NitroxModel.DataStructures.GameLogic;
+﻿using System;
+using NitroxModel.DataStructures.GameLogic;
 using NitroxModel.DataStructures.Util;
 using NitroxModel.Helper;
 using UnityEngine;
@@ -8,7 +9,7 @@ namespace NitroxClient.GameLogic.Spawning
     public class ReefbackEntitySpawner : IEntitySpawner
     {
         private readonly DefaultEntitySpawner defaultSpawner;
-
+        private readonly bool aprilFools = DateTime.Today is { Month: 4, Day: 1 };
         public ReefbackEntitySpawner(DefaultEntitySpawner defaultSpawner)
         {
             this.defaultSpawner = defaultSpawner;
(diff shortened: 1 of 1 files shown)


NitroxServer-Subnautica/GameLogic/Entities/Spawning/ReefbackBootstrapper.cs
```diff
@@ -10,6 +11,7 @@ namespace NitroxServer_Subnautica.GameLogic.Entities.Spawning.EntityBootstrapper
     class ReefbackBootstrapper : IEntityBootstrapper
     {
         private float creatureProbabiltySum = 0;
+        private readonly bool aprilFools = DateTime.Today.Month == new DateTime(1, 4, 1).Month && DateTime.Today.Day == new DateTime(1, 4, 1).Day;
```
dartasen, line 14: I think this can be simplified with something like that
```
DateTime.Now.Month == 4 && DateTime.Now.Day == 1
```

Measurity: Or `DateTime.Today == new DateTime(DateTime.Today.Year, 4, 1)`

Dessix: `DateTime.Today is {} today && today.Month == 4 && today.Day == 1`

NitroxServer-Subnautica/GameLogic/Entities/Spawning/ReefbackBootstrapper.cs
```diff
@@ -38,7 +40,10 @@ public void Prepare(Entity parentEntity, DeterministicBatchGenerator determinist
                         for(int i = 0; i < totalToSpawn; i++)
                         {
                             NitroxId id = deterministicBatchGenerator.NextId();
-                            Entity child = new Entity(localSpawnPosition, new NitroxQuaternion(0, 0, 0, 1), new NitroxVector3(1, 1, 1), creature.techType.ToDto(), parentEntity.Level, creature.classId, true, id, null, parentEntity);
+                            NitroxTechType techType = aprilFools ? parentEntity.TechType : creature.techType.ToDto();
```
Dessix, line 43: Won't this result in infinite recursion?

killzoms: Nope, since the Reefback bootstrapper handles its own spawning it doesnt use the regular spawning code for spawning its children, thus no infinite recursion

killzoms:
Shall I make it a client side only "bug" with an April fools tag above them? @Sunrunner37

killzoms:
Update: now it is client side and effects already spawned Reefbacks as well

Measurity:
@killzoms Think we can close this one for now. Can reopen at a later time but I think this april fools is spoiled :P

PR #1290, opened 2020-11-22
Update Harmony to v2.0.4

Coding-Hen:
Updating harmony to 2.0.4 
Added new finalizer patching method
Changed Fabricator to GhostCrafter to also support syncing of Workbench


Sunrunner37:
Will only consider after the next release.

NitroxClient/Communication/Packets/Processors/GhostCrafterBeginCraftingProcessor.cs
```diff
@@ -5,29 +5,28 @@
 using NitroxClient.Unity.Helper;
 using NitroxModel.Packets;
 using NitroxModel_Subnautica.DataStructures;
-using NitroxModel_Subnautica.Helper;
 using UnityEngine;
 
 namespace NitroxClient.Communication.Packets.Processors
 {
-    public class FabricatorBeginCraftingProcessor : ClientPacketProcessor<FabricatorBeginCrafting>
+    public class GhostCrafterBeginCraftingProcessor : ClientPacketProcessor<GhostCrafterBeginCrafting>
```
Sunrunner37, line 12: For my education, what is considered to be a "Ghost"?  I believe the initial abstraction was QOL for developers so they didn't need to understand internal UWE verbiage.

Coding-Hen: If I understand your question correctly. The GhostCrafter is just a parent class of both the workbench and the fabricator. This allows us to therefore patch this class rather than the fabricator and support both without any additional changes being needed to the code. Does that answer your question?

Sunrunner37: Sounds good to me, can we update with latest hintpath change and I will merge?

PR #1291, opened 2020-11-23
Added support to sync sound

### LINKED ISSUE #37: [Audio] Play audio effects
Currently behavior and usually animations are synced for actions, but sound is left out. Since sound is positional, this gives a much more immersive experience than if the other players are virtually silent.

#### TODO:
- Go over all the synced actions, and implement sounds on the receiver side if they play a sound on the sending side.
- Light toggles.
- Vehicle enters/exists.
- Base building and interaction.
- Swimming/walking.
- Tool usage.

Jannify:
This pr does not cover all sounds, that's not the goal. It's a framework with a few sounds synced. The rest of the is for jrs or who want to work on an easy task.

Closes #37


Jannify:
Technical ready but will investigate some time to check if 2D audio can be synced too.  If that is not ready till release you can merge it.

Sunrunner37:
Looks good.  Let me know when investigation is complete and I'll merge.

Jannify:
So I think everything is working now but didn't had the time to test every synced sound from the list (it's completely guest by the name). So we have to test this with multiple people. It can happen that the sounds come from other patches and were already "synced" (intended or not). These then need to suppress the new packets to prevent double playing.

Jannify:
There is still a bug where the sound is much silenter than it should be. Cyclops horn is a good example.

Jannify:
I think it is working now. It's 4 am. I will do testing on Monday.

romgerman:
How is it going?

Coding-Hen:
Just did a quick test of this on the whole seems much better now. I have a feeling that this will probably evolve overtime as we decide that some sounds are needed/not and distances might change. A few things stuck out to me:
- Fire extinguisher can get stuck on for one player and endlessly loop
- Seamoth explosions arent global radius when close
- Seamoth lights on/off are not global radius when close (This was global before, not sure if intended to not have radius now?)

I will continue to investigate and will add to it if I find anymore

Jannify:
Will look into fixing these bug and then make this pr ready to review.

Jannify:
This pr does not cover all the sounds, that's not the goal. It should lay the ground work and syncs a few, the rest is for the jrs or so who want to work on an easy task.

Coding-Hen (approved):
Looks good for a first implementation as you say this can be refined and improved over time with minor changes where needed

Jannify:
Closes #37

Issue #1223, opened 2020-08-19
Seamoth randomly disappear or explode


MysticCoss:
With 1 vehicle bay we can not create more than one seamoth. We made 2 vehicle bay, then each person must launch their own seamoth. when enter other's seamoth it will sometimes disappear or explode

Coding-Hen:
Please verify this is still an issue. Are you able to provide the exact way you produced this? I am unable to reproduce.

EnzoTools:
> 请确认这仍然是一个问题。您是否能够提供产生此效果的确切方法？我无法复制。

Mine didn't explode, but it got stuck and couldn't move, so I had to use the booster to get home  :( 

After I restarted the server, he was back to normal, unable to move again a few minutes later

EnzoTools:
> Please verify this is still an issue. Are you able to provide the exact way you produced this? I am unable to reproduce.

Mine didn't explode, but it got stuck and couldn't move, so I had to use the booster to get home :(

After I restarted the server, he was back to normal, unable to move again a few minutes later

MysticCoss:
@Coding-Hen all things happened: i had to created 2 mobile vehicle bays, each person must create their own seamoth on their own MVB. After that we went to life pod under 200m. Because of the limitation we had to  park there and swam deeper. When we came back, my friend's seamoth disappeared. Maybe it was destroyed because of depth. We came back to base for him to build a seamoth again. We left those alone to do some stuffs. When we came back to base again, i entered my friend's seamoth and it exploded instantly.
It's kind of random event. I have tried reproduce but fail

AzraeltheWicked:
I can confirm the event, though not the cause.

I have gone into wrecks on multiple occasions and in multiple locations and when I come back out my seamoth will have vanished.  There's no sound or debris indicating explosion and in one case I was only on the other side of a wall (maybe 5m away).  Seamoth just vanishes.

Ohimark12:
yea this happened to me yesterday and it seems that it thinks your hitting something at a much faster speed than you are and does a ton of damage i just accidentally bumped into my friend and his seamoth blew up buuut we left the server and rejoined and it was back at low health.. but back thats all i have about this matter

EinLuigi:
it is true i can confirm it

dartasen:
New version of Nitrox is available, we shipped some fix for this. I'm closing it !

Xavego0n:
[...]

Issue #753, opened 2019-03-18
[Vehicle] Docking/Undocking animation and sound issues


Ghaarg:
There are minor problems with docking (for cyclops):
- Docking into a cyclops will play the docking sounds for all player regardless of position. Super hard to find where the sounds come from
- Undocking will just release the vehicle after the animation are played for the person in the vehicle. Looks terrible but works just fine

Copied from PR [#748](https://github.com/SubnauticaNitrox/Nitrox/pull/748)

acidscorch:
Hi @Ghaarg,

For the sounds, maybe this helps:

![image](https://user-images.githubusercontent.com/48223659/83644986-3595e880-a5b2-11ea-9641-f812fe070421.png)


And for the undocking i would suggest: 

![image](https://user-images.githubusercontent.com/48223659/83647821-a094ee80-a5b5-11ea-9c75-48a4bafe9238.png)

The exact values may need to be found by testing. This gives a smooth movement downwards. Maybe on the sending side the position updates should be suppressed as long as the cinematic is running. After the cinematic, the position on the remote client (evaluate how much speed is needed and how far away the position from the cyclops should be) needs to be very close to the position on the undocking client to avoid a reposition jump. 

Best regards

Jannify:
With https://github.com/SubnauticaNitrox/Nitrox/pull/1291 a sound suppressor can be used in the docking processor.

Issue #118, opened 2018-02-04
[Vehicle] Seamoth docking sync


Sunrunner37:
- [ ] Docking inside cyclops
- [ ] Docking inside moon pool

hamsteyr:
Docking seems to work, but undocking still seems to have an issue with syncing.

Replicable.

Agent77326:
In my case neither did work, tested with 2 players and 1.0.0.9 from today.

Aocxiez:
Case close thank you for your hardwork.

kinvaris:
TLDR: we reproduced it twice on a different server with same versions. 

Same issue occured during 2 players with Nitrox version `Alpha 1.1.0.0`. We were playing on september build of 2018. 

During the first time, we thought we could not fix it. The bug starts when you try to dock the seamoth into the moonpool or cyclops. 

Every player is immediately desynced. The desync can only be fixed by destroying the moonpool/cyclops and performing a re-login into the server. (This we noticed during our second playthrough.)

The details of the Subnautica build to download via steam console: `download_depot 264710 264712 5196577974721678848 :`

App id: 264710
Depot id: 264712
Manifest id (version of build): 5196577974721678848

Reference Subnautica build in steamDB: https://steamdb.info/depot/264712/manifests/

acidscorch:
Please retest the moonpool related part after upcoming fix for sync and store moonpool Ids.

PR #1292, opened 2020-11-23
Seamoth kill fix and clientside simulation extended

Ghaarg:
This includes:
- ~~Client side now keeps track of other player's simulation ownerships~~
- ~~Client can now override simulation behaviour if needed. Needs explicit lookup if override exists. Should not change previous behavior~~
- Initial simulation ownership: server now sends initial ownerships for global entities and vehicles (bases should follow at a later date)
- Fixed a bug that resulted in only one simulation ownership change when a player disconnects
- Extended simulation ownership change on disconnected player for all simulated entries instead of only entites.
- LiveMixin for vehicles + cyclops will only be simulated by one player. This should stop random explosions
- LiveMixin will be synchronized by simulating player for vehicles + cyclops
- All LiveMixin related behavior will be controled by LiveMixinManager
- Weld action will be synchronized for vehicles
- Docking is improved. Less/no docking explosions and cyclops should not bounce around anymore #118, #1223, #753
- From previous fix: Players in waterparks will have the right swimming animation for players joining (this should also be corrected for players in a vehicle)


NitroxClient/GameLogic/InitialSync/RemotePlayerInitialSyncProcessor.cs
```diff
@@ -62,15 +79,18 @@ public override IEnumerator Process(InitialPlayerSync packet, WaitScreen.ManualW
...
+                            swimming = false;
                         }
                     }
                     else
                     {
                         Log.Error("Could not spawn remote player into subroot/escape pod with id: " + playerData.SubRootId.Value);
                     }
                 }
-
+                if (!swimming)
```
Coding-Hen, line 90: Does this also count for the player being above water on land or a base/lifeboat roof?

Ghaarg: I don't think we currently check that but as far as i see it, it could be easily added via
```C#
float oceanLevel = Ocean.main.GetOceanLevel();
swimming = position.y < oceanLevel;
```

NitroxClient/Communication/Packets/Processors/VehicleDockingProcessor.cs
```diff
@@ -28,10 +35,15 @@ public override void Process(VehicleDocking packet)
...
+
             using (packetSender.Suppress<VehicleDocking>())
             {
-                vehicleDockingBay.SetVehicleDocked(vehicle);
+                Log.Debug($"Set vehicle docked for {vehicleDockingBay.gameObject.name}");
+                vehicle.GetComponent<MultiplayerVehicleControl>().Exit();
+                Log.Debug($"before dock isKinematic: {vehicle.useRigidbody.isKinematic}");
+                vehicleDockingBay.DockVehicle(vehicle);
+                vehicle.useRigidbody.isKinematic = false;
+                Log.Debug($"after dock isKinematic: {vehicle.useRigidbody.isKinematic}");
```
Sunrunner37, line 46: We can probably remove these three logs.  especially this one (since we just set the field).

NitroxClient/Communication/Packets/Processors/VehicleDockingProcessor.cs
```diff
@@ -28,10 +35,15 @@ public override void Process(VehicleDocking packet)
...
-            
+
             using (packetSender.Suppress<VehicleDocking>())
             {
-                vehicleDockingBay.SetVehicleDocked(vehicle);
+                Log.Debug($"Set vehicle docked for {vehicleDockingBay.gameObject.name}");
+                vehicle.GetComponent<MultiplayerVehicleControl>().Exit();
+                Log.Debug($"before dock isKinematic: {vehicle.useRigidbody.isKinematic}");
+                vehicleDockingBay.DockVehicle(vehicle);
+                vehicle.useRigidbody.isKinematic = false;
```
Sunrunner37, line 45: Let's add a comment saying that the ordering is important because the function above changed isKinematic.

NitroxClient/Communication/Packets/Processors/VehicleDockingProcessor.cs
```diff
@@ -41,6 +53,12 @@ IEnumerator DisablePilotingAfterAnimation(NitroxId vehicleId, ushort playerId)
         {
             yield return new WaitForSeconds(3.0f);
             vehicles.SetOnPilotMode(vehicleId, playerId, false);
+            GameObject vehicleGo = NitroxEntity.RequireObjectFrom(vehicleId);
+            Vehicle vehicle = vehicleGo.RequireComponent<Vehicle>();
+            if (!vehicle.docked)
+            {
+                Log.Error($"Vehicle not docked after docking process");
```
Sunrunner37, line 60: minor: let's add the vehicle id.

NitroxClient/Communication/Packets/Processors/VehicleUndockingProcessor.cs
```diff
@@ -31,26 +35,35 @@ public override void Process(VehicleUndocking packet)
...
+                        playerInstance.AnimationController.UpdatePlayerAnimations = false;
+                    }
+                }
+                else
+                {
+                    vehicleDockingBay.SetVehicleUndocked();
+                    vehicleDockingBay.ReflectionSet("vehicle_docked_param", false);
+                    vehicleDockingBay.ReflectionSet("_dockedVehicle", null);
+                    vehicle.docked = false;
+                    //vehicle.useRigidbody.AddForce(Vector3.down * 5f, ForceMode.VelocityChange);
```
Sunrunner37, line 58: minor: dead code

NitroxClient/Communication/Packets/Processors/VehicleUndockingProcessor.cs
```diff
@@ -31,26 +35,35 @@ public override void Process(VehicleUndocking packet)
...
-                vehicles.SetOnPilotMode(packet.VehicleId, packet.PlayerId, true);
-                vehicleDockingBay.subRoot.BroadcastMessage("OnLaunchBayOpening", SendMessageOptions.DontRequireReceiver);
-                SkyEnvironmentChanged.Broadcast(vehicleGo, (GameObject)null);
-
-                
-
-                vehicle.StartCoroutine(WaitBeforePushDown(vehicle, vehicleDockingBay));
-                
+                Optional<RemotePlayer> player = remotePlayerManager.Find(packet.PlayerId);
+                if (packet.UndockingStart)
```
Sunrunner37, line 39: minor: can we break this into two functions (undocking/docking) related logic?

NitroxClient/Communication/Packets/Processors/WeldActionProcessor.cs
```diff
@@ -0,0 +1,52 @@
...
+using NitroxClient.GameLogic;
+using NitroxClient.MonoBehaviours;
+using NitroxModel.DataStructures.Util;
+using NitroxModel.Logger;
+using NitroxModel.Packets;
+using UnityEngine;
+
+namespace NitroxClient.Communication.Packets.Processors
+{
+    class WeldActionProcessor : ClientPacketProcessor<WeldAction>
```
Sunrunner37, line 12: can this be made more generic for LiveMixing?  For instance, we may want a system were some knifes a fish and wants the same behavior (packet sent out for simulator to update health).

NitroxClient/Communication/Packets/Processors/WeldActionProcessor.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        public WeldActionProcessor(IMultiplayerSession multiplayerSession, SimulationOwnership simulationOwnership)
+        {
+            this.multiplayerSession = multiplayerSession;
+            this.simulationOwnership = simulationOwnership;
+        }
+
+        public override void Process(WeldAction packet)
+        {
+            Optional<GameObject> opGameObject = NitroxEntity.GetObjectFrom(packet.Id);
```
Sunrunner37, line 25: Alternatively, I think there is a RequireObjectfFrom which will internally throw.  That will stop only this packet processor and is safe to do.

NitroxClient/Communication/Packets/Processors/WeldActionProcessor.cs
```diff
@@ -0,0 +1,52 @@
...
+            if (opGameObject.HasValue)
+            {
+                if (simulationOwnership.HasAnyLockType(packet.Id))
+                {
+                    GameObject gameObject = opGameObject.Value;
+                    LiveMixin liveMixin = gameObject.GetComponent<LiveMixin>();
```
Sunrunner37, line 31: Similarly RequireComponent (unity helper) here.

NitroxClient/GameLogic/InitialSync/RemotePlayerInitialSyncProcessor.cs
```diff
@@ -51,9 +51,26 @@ public override IEnumerator Process(InitialPlayerSync packet, WaitScreen.ManualW
...
                             player.SetSubRoot(subroot);
                             // Set the animation for the remote player to standing instead of swimming if player is not in a flooded subroot
+                            // or in a waterpark
+                            
                             if (!subroot.IsUnderwater(player.Body.transform.position))
                             {
-                                player.UpdateAnimation(AnimChangeType.UNDERWATER, AnimChangeState.OFF);
+                                swimming = false;
+                                // We know that we are in a subroot. But we can also be in a waterpark in a subroot, where we would swim
+                                BaseRoot baseRoot = subroot.GetComponentInParent<BaseRoot>();
```
Sunrunner37, line 60: minor: let's extract into a private method to reduce nesting.

NitroxClient/GameLogic/SimulationOwnership.cs
```diff
@@ -53,6 +77,10 @@ public void ReceivedSimulationLockResponse(NitroxId id, bool lockAquired, Simula
             {
                 SimulateEntity(id, lockType);
             }
+            else if (!OtherPlayerHasAnyLock(id))
+            {
+                AddSimulationLockForUnknown(id, lockType);
+            }
```
Sunrunner37, line 83: If this happens, then it looks like a result of a race condition; i.e., lock acquired before us but broadcasted later.  Should we just ignore this case and allow future lock packets to update the list?

NitroxClient/GameLogic/SimulationOwnership.cs
```diff
@@ -67,14 +95,71 @@ public void ReceivedSimulationLockResponse(NitroxId id, bool lockAquired, Simula
...
+            else
+            {
+                Log.Warn($"Tried to remove non existing simulation override for NitroxId: {id}");
+            }
+        }
+
+        public bool SimulationLockOverrideActive(NitroxId id)
+        {
+            return simulationOverride.Contains(id);
+        }
```
Sunrunner37, line 163: Can you describe the behavior here?

NitroxClient/GameLogic/Vehicles.cs
```diff
@@ -212,10 +214,14 @@ public void UpdateVehiclePosition(VehicleMovementData vehicleModel, Optional<Rem
                 if (subRoot)
                 {
                     mvc = subRoot.gameObject.EnsureComponent<MultiplayerCyclops>();
-                    subRoot.GetComponent<LiveMixin>().health = vehicleModel.Health;
                 }
                 else if (vehicle)
                 {
+                    if (vehicle.docked)
+                    {
+                        Log.Debug($"For vehicle {vehicleModel.Id} position update while docked, will not execute");
```
Sunrunner37, line 222: I think this might be very noisy... maybe combine the two conditons into just the else if.

NitroxModel/Packets/LiveMixinHealthChanged.cs
```diff
@@ -0,0 +1,63 @@
...
+using ProtoBufNet;
+
+namespace NitroxModel.Packets
+{
+    [Serializable]
+    [ProtoContract]
+    public class LiveMixinHealthChanged : Packet
+    {
+        [Serializable]
+        class DamageTakenData
```
Sunrunner37, line 18: minor: Let's put in a new class in model.

NitroxModel/Packets/LiveMixinHealthChanged.cs
```diff
@@ -0,0 +1,63 @@
...
+        {
+            public NitroxVector3 Position { get; set; }
+
+            public ushort Damagetype { get; set; }
+
+            public Optional<NitroxId> DealerId { get; set; }
```
Sunrunner37, line 24: Is this metadata relevant?

NitroxModel/Packets/LiveMixinHealthChanged.cs
```diff
@@ -0,0 +1,63 @@
...
+        }
+        Optional<DamageTakenData> damageTakenData;
```
Sunrunner37, line 26: minor: missing accessor

NitroxModel/Packets/LiveMixinHealthChanged.cs
```diff
@@ -0,0 +1,63 @@
...
+        public NitroxId Id { get; set; }
+
+        public float LifeChanged { get; set; }
+
+        
+        public NitroxVector3 Position { get { return damageTakenData.HasValue ? damageTakenData.Value.Position : default; } }
+
+        public ushort Damagetype { get { return damageTakenData.HasValue ? damageTakenData.Value.Damagetype : default; } }
+
+        public Optional<NitroxId> DealerId { get { return damageTakenData.HasValue ? damageTakenData.Value.DealerId : default; } }
```
Sunrunner37, line 38: minor: Is this model also intended to be used on its own?  Can it just be fields here?

NitroxModel/Packets/LiveMixinHealthChanged.cs
```diff
@@ -0,0 +1,63 @@
...
+                {
+                    Position = position,
+                    Damagetype = damageType,
+                    DealerId = dealerId
+                };
+            }
+            else
+            {
+                damageTakenData = Optional.Empty;
+            }
```
Sunrunner37, line 60: This seems like business logic that should be moved out from the constructor.

NitroxPatcher/Patches/Dynamic/LiveMixin_AddHealth_Patch.cs
```diff
@@ -0,0 +1,98 @@
...
+using NitroxModel.DataStructures.GameLogic;
+using NitroxModel.DataStructures.Util;
+using NitroxModel.Logger;
+using NitroxModel.Packets;
+using NitroxModel_Subnautica.DataStructures;
+using UnityEngine;
+
+namespace NitroxPatcher.Patches.Dynamic
+{
+    class LiveMixin_AddHealth_Patch : NitroxPatch, IDynamicPatch
```
Sunrunner37, line 20: If possible, let's see if we can federate out some of the state keeping and logic into a new class that we invoke.

NitroxPatcher/Patches/Dynamic/Vehicle_OnKill_Patch.cs
```diff
@@ -12,8 +15,13 @@ public class Vehicle_OnKill_Patch : NitroxPatch, IDynamicPatch
         public static readonly MethodInfo TARGET_METHOD = TARGET_CLASS.GetMethod("OnKill", BindingFlags.NonPublic | BindingFlags.Instance);
 
         public static void Prefix(Vehicle __instance)
-        {
-            NitroxServiceLocator.LocateService<Vehicles>().BroadcastDestroyedVehicle(__instance);
+        {            
+            NitroxId id = NitroxEntity.GetId(__instance.gameObject);
+            SimulationOwnership simulationOwnership = NitroxServiceLocator.LocateService<SimulationOwnership>();
+            if (simulationOwnership.HasAnyLockType(id) || simulationOwnership.OtherPlayerHasAnyLock(id))
```
Sunrunner37, line 21: Why even check for a lock? in both cases you stop simulating.

NitroxModel/Packets/WeldAction.cs
```diff
@@ -0,0 +1,22 @@
...
+    public class WeldAction : Packet
+    {
+        public NitroxId Id { get; }
+        public float HealthAdded { get; }
+
+        public WeldAction(NitroxId id, float healthAdded)
+        {
+            Id = id;
+            HealthAdded = healthAdded;
+        }
```
dartasen, line 20: minor: Add a toString to models
```csharp
{
    return $"[WeldAction - Id: {Id}, HealthAdded: {HealthAdded}]";
}
```

NitroxClient/Communication/Packets/Processors/VehicleUndockingProcessor.cs
Ghaarg: Maybe i should reword the boolean "UndockingStart". Both parts are undocking related logic. The first part starts the undocking process (when player enters vehicle) and second part (UndockingStart=false) when the undocking is finished (the real undocking)
I can separate both anyway.

NitroxClient/Communication/Packets/Processors/WeldActionProcessor.cs
Ghaarg: We can do a generic LiveMixinChangeRequest. The idea behind a destinct WeldAction packet and processor was that for other actions (knifing a fish) it may not be sufficient to only change the health.
For example hitting with a knife: If we then want to also synchronize slashing sounds, only requesting health change will not cut it.

NitroxModel/Packets/LiveMixinHealthChanged.cs
Ghaarg: Had them just as fields before but NitroxVector3 and ushort cannot be used with Optional. But we don't need those fields for Health healed. Therefore, i separated them into an internal class that can be set empty to (hopefully) reduce package size.

NitroxPatcher/Patches/Dynamic/Vehicle_OnKill_Patch.cs
Ghaarg: Could happen that the client does not have notice of a lock. Removing the simulation would then raise an error.

NitroxPatcher/Patches/Dynamic/LiveMixin_AddHealth_Patch.cs
Ghaarg: You mean something like an LiveMixin manager? I hesitated because i was not sure if we will use that for other entities. If wanted, i will refactor this into it's own management class.

NitroxClient/Communication/Packets/Processors/LifeMixinChangedProcessor.cs
```diff
@@ -0,0 +1,71 @@
...
+                        {
+                            liveMixin.AddHealth(packet.LifeChanged);
+                        }
+                    }
+                    
+                    // Check if the health calculated by the game is the same as the calculated damage from the simulator
+                    if (liveMixin.health != packet.TotalHealth)
+                    {
+                        Log.Warn($"Calculated health and send health for {packet.Id} do not align (Calculated: {liveMixin.health}, send:{packet.TotalHealth}). This will be correted but should be investigated");
+                        liveMixin.health = packet.TotalHealth;
```
Sunrunner37, line 65: What is the advantage of syncing the damage taking and not just health amount directly?

NitroxClient/Communication/Packets/Processors/VehicleUndockingProcessor.cs
```diff
@@ -31,26 +35,66 @@ public override void Process(VehicleUndocking packet)
...
             vehicle.docked = false;
-            vehicle.useRigidbody.AddForce(Vector3.down * 5f, ForceMode.VelocityChange);
+            Optional<RemotePlayer> player = remotePlayerManager.Find(packet.PlayerId);
+            if (player.HasValue)
+            {
+                // Sometimes the player is not set accordingly which stretches the player's model instead of putting them in place
+                // after undocking. This fixes it (the player rigid body seems to not be set right sometimes)
+                player.Value.SetSubRoot(null);
+                player.Value.SetVehicle(null);
+                player.Value.SetVehicle(vehicle);
```
Sunrunner37, line 94: ?

NitroxServer/GameLogic/Vehicles/VehicleManager.cs
```diff
@@ -30,7 +30,21 @@ public void UpdateVehicle(VehicleMovementData vehicleMovement)
...
+            }
+        }
+
+        public void UpdateVehicleHealth(NitroxId vehicleId, float newHealth)
+        {
+            if (vehiclesById.ContainsKey(vehicleId))
+            {
+                if (newHealth > 0)
+                {
+                    vehiclesById[vehicleId].Health = newHealth;
```
Measurity, line 42: Use dictionary `.TryGetValue(key, out ..)` instead to reduce lookup cost by 1.

NitroxServer/GameLogic/SimulationOwnership.cs
```diff
@@ -105,5 +105,17 @@ public bool RevokeOwnerOfId(NitroxId id)
...
             return false;
         }
+
+        public Player GetPlayerForLock(NitroxId id)
+        {
+            lock (playerLocksById)
+            {
+                if(playerLocksById.ContainsKey(id))
+                {
+                    return playerLocksById[id].Player;
```
Measurity, line 115: Use dictionary `.TryGetValue(key, out ..)` instead to reduce lookup cost by 1.

NitroxPatcher/Patches/Dynamic/LiveMixin_AddHealth_Patch.cs
```diff
@@ -0,0 +1,98 @@
...
+                Tuple<LiveMixin, float> parameter = null;
+                PARAMETER_PER_ENTITY.TryGetValue(id, out parameter);
+                if (parameter != null)
+                {
+                    parameter.Item1.AddHealth(parameter.Item2);
+                    PARAMETER_PER_ENTITY.Remove(id);
+                }
+                else
+                {
+                    Log.Warn($"Got a simulation lock response for entity {id} but did not find the corresponding parameter. This should not happen!");
```
Measurity, line 93: Should be error if it shouldn't happen.

NitroxClient/GameLogic/SimulationOwnership.cs
```diff
@@ -67,14 +111,76 @@ public void ReceivedSimulationLockResponse(NitroxId id, bool lockAquired, Simula
...
+        public void RemoveSimulationOverride(NitroxId id)
+        {
+            if (simulationOverride.Contains(id))
+            {
+                simulationOverride.Remove(id);
+            }
+            else
+            {
+                Log.Warn($"Tried to remove non existing simulation override for NitroxId: {id}");
+            }
```
Measurity, line 178: Shorten to:
```suggestion
            if (!simulationOverride.Remove(id))
            {
                Log.Warn($"Tried to remove non existing simulation override for NitroxId: {id}");
            }
```

NitroxClient/GameLogic/SimulationOwnership.cs
```diff
@@ -67,14 +111,76 @@ public void ReceivedSimulationLockResponse(NitroxId id, bool lockAquired, Simula
...
+        }
+
+        public void AddSimulationLockForUnknown(NitroxId id, SimulationLockType lockType)
+        {
+            simulatedIdsByLockType[id] = new PlayerLock(null, lockType);
+        }
+
+        public void AddSimulationOverride(NitroxId id)
+        {
+            if (simulatedIdsByLockType.ContainsKey(id) && simulatedIdsByLockType[id].PlayerId == muliplayerSession.Reservation.PlayerId)
```
Measurity, line 154: Use `simulatedIdsByLockType.TryGetValue(id, out ...)` to reduce amount of lookups. Then just keep the ` && simulatedIdsByLockType[id].PlayerId == muliplayerSession.Reservation.PlayerId` as is.

NitroxClient/Communication/Packets/Processors/WeldActionProcessor.cs
```diff
@@ -0,0 +1,52 @@
...
+                {
+                    Log.Error($"Got WeldAction packet for {packet.Id} but did not find the lock corresponding to it");
+                }
+            }
+            else
+            {
+                Log.Error($"Did not find GameObject {packet.Id} for WeldAction even though this player should have the simulation lock");
+            }
+        }
+    }
```
Measurity, line 51: I recommend inverting the if checks here to log & `return` to reduce code indentation and make it easier to read.

Measurity (commented):
Some code style feedback

NitroxServer/GameLogic/Entities/EntitySimulator.cs
```diff
@@ -78,7 +89,26 @@ private void AssignEntitiesToNewPlayers(Player oldPlayer, List<Entity> entities,
...
                         Log.Info("Player " + player.Name + " has taken over simulating " + entity.Id);
                         ownershipChanges.Add(new SimulatedEntity(entity.Id, player.Id, true, DEFAULT_ENTITY_SIMULATION_LOCKTYPE));
-                        return;
+                        break;
+                    }
+                }
+            }
+        }
+
+        private void AssignEntitiesToNewPlayers(Player oldPlayer, List<NitroxId> entities, List<SimulatedEntity> ownershipChanges)
```
Sunrunner37, line 98: Looks like the diff is messed up.  this function is duplicated.

NitroxServer/Communication/Packets/Processors/WeldActionProcessor.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        public override void Process(WeldAction packet, Player player)
+        {
+            Player simulatingPlayer = simulationOwnershipData.GetPlayerForLock(packet.Id);
+
+            if (simulatingPlayer != null)
+            {
+                Log.Debug($"Send WeldAction to simulating player {simulatingPlayer.Name} for entity {packet.Id}");
+                simulatingPlayer.SendPacket(packet);
+            }
```
Sunrunner37, line 30: else, should we assign simulation the requester and send them back the packet?

NitroxServer/GameLogic/Vehicles/VehicleManager.cs
```diff
@@ -25,52 +25,65 @@ public IEnumerable<VehicleModel> GetVehicles()
...
+        {
+            if (vehiclesById.TryGetValue(vehicleId, out VehicleModel vehicleModel))
+            {
+                if (newHealth > 0)
+                {
+                    vehicleModel.Health = newHealth;
+                }
+                else
+                {
+                    RemoveVehicle(vehicleId);
```
Sunrunner37, line 45: Should we rely on the VehicleRemoved packet to do this?  In case something weird happens, the server might lose track of this instance.

NitroxPatcher/Patches/Dynamic/LiveMixin_AddHealth_Patch.cs
```diff
@@ -0,0 +1,70 @@
...
+using NitroxModel.DataStructures.GameLogic;
+using NitroxModel.DataStructures.Util;
+using NitroxModel.Logger;
+using NitroxModel.Packets;
+using NitroxModel_Subnautica.DataStructures;
+using UnityEngine;
+
+namespace NitroxPatcher.Patches.Dynamic
+{
+    class LiveMixin_AddHealth_Patch : NitroxPatch, IDynamicPatch
```
Sunrunner37, line 20: Let's see if we can refactor into its own class.  I think if you do this, then it would allow us to remove the need for simulation override (because the class can manage this stateful suppression).

NitroxPatcher/Patches/Dynamic/LiveMixin_AddHealth_Patch.cs
```diff
@@ -0,0 +1,70 @@
...
+    {
+        public static readonly Type TARGET_CLASS = typeof(LiveMixin);
+        public static readonly MethodInfo TARGET_METHOD = TARGET_CLASS.GetMethod("AddHealth", BindingFlags.Public | BindingFlags.Instance);
+        public static readonly Dictionary<NitroxId, float> ORIGINAL_HEALTH_PER_ENTITY = new Dictionary<NitroxId, float>();
+        public static readonly Dictionary<NitroxId, Tuple<LiveMixin, float>> PARAMETER_PER_ENTITY = new Dictionary<NitroxId, Tuple<LiveMixin, float>>();
```
Sunrunner37, line 25: is this even used?  (same with class below)

NitroxPatcher/Patches/Dynamic/LiveMixin_AddHealth_Patch.cs
```diff
@@ -0,0 +1,70 @@
...
+using NitroxModel_Subnautica.DataStructures;
+using UnityEngine;
+
+namespace NitroxPatcher.Patches.Dynamic
+{
+    class LiveMixin_AddHealth_Patch : NitroxPatch, IDynamicPatch
+    {
+        public static readonly Type TARGET_CLASS = typeof(LiveMixin);
+        public static readonly MethodInfo TARGET_METHOD = TARGET_CLASS.GetMethod("AddHealth", BindingFlags.Public | BindingFlags.Instance);
+        public static readonly Dictionary<NitroxId, float> ORIGINAL_HEALTH_PER_ENTITY = new Dictionary<NitroxId, float>();
```
Sunrunner37, line 24: in harmony, you can use a __state variable as a parameter to pass from prefix to postfix.

NitroxPatcher/Patches/Dynamic/Player_SetCurrentSub_Patch.cs
```diff
@@ -14,16 +14,20 @@ public class Player_SetCurrentSub_Patch : NitroxPatch, IDynamicPatch
...
 
             if (sub != null)
             {
                 subId = NitroxEntity.GetId(sub.gameObject);
             }
-
-            NitroxServiceLocator.LocateService<LocalPlayer>().BroadcastSubrootChange(Optional.OfNullable(subId));
+            // When in the water of the moonpool, it can happen that you hammer change requests
+            // while the sub is not changed. This will prevent that
+            if (__instance.GetCurrentSub() != sub)
```
Sunrunner37, line 27: Is this the fix?  I thought in the moonpool, the player is being forced out of the sub with waves?  Per this, the subroot is actually staying the same?

NitroxPatcher/Patches/Dynamic/VehicleDockingBay_OnTriggerEnter.cs
```diff
@@ -0,0 +1,45 @@
...
+using NitroxModel.Logger;
+using UnityEngine;
+
+namespace NitroxPatcher.Patches.Dynamic
+{
+    class VehicleDockingBay_OnTriggerEnter : NitroxPatch, IDynamicPatch
+    {
+        public static readonly Type TARGET_CLASS = typeof(VehicleDockingBay);
+        public static readonly MethodInfo TARGET_METHOD = TARGET_CLASS.GetMethod("OnTriggerEnter", BindingFlags.NonPublic | BindingFlags.Instance);
+        private static Vehicle prevInterpolatingVehicle = null;
```
Sunrunner37, line 18: __state here as well.

NitroxPatcher/Patches/Dynamic/LiveMixin_AddHealth_Patch.cs
Ghaarg: Consolidated the code into the LiveMixinManager and removed the simulationOverride.

NitroxPatcher/Patches/Dynamic/LiveMixin_AddHealth_Patch.cs
Ghaarg: Done

NitroxPatcher/Patches/Dynamic/Player_SetCurrentSub_Patch.cs
Ghaarg: Yes the subroot actually does not change. But the client calls the subrootchanged method anyway (like a machinegun)

NitroxPatcher/Patches/Dynamic/VehicleDockingBay_OnTriggerEnter.cs
Ghaarg: Done

NitroxServer/Communication/Packets/Processors/WeldActionProcessor.cs
Ghaarg: Due to the initial simulation ownership and changed simulation ownership when a player disconnects, it should not be possible that no player has the simulation of an vehicle. Therefore, no assign in retrospect is needed.

NitroxServer/GameLogic/Vehicles/VehicleManager.cs
Ghaarg: I can't think of an instance where it could happen that a vehicle has no health and is not destroyed.
Just did this to reduce the packet load (1 packet instead of 2 for the health change and vehicle kill)
There were edge cases where the non-simulating player destroyed the vehicle on his client but that's it.

NitroxServer/GameLogic/Entities/EntitySimulator.cs
Ghaarg: Changed the name of the method to show that it is not the same as the other.

NitroxClient/GameLogic/InitialSync/RemotePlayerInitialSyncProcessor.cs
```diff
@@ -36,7 +36,7 @@ public override IEnumerator Process(InitialPlayerSync packet, WaitScreen.ManualW
 
                 List<TechType> equippedTechTypes = playerData.EquippedTechTypes.Select(techType => techType.ToUnity()).ToList();
                 RemotePlayer player = remotePlayerManager.Create(playerData.PlayerContext, equippedTechTypes);
-
+                bool swimming = true;
```
Sunrunner37, line 39: Looks like you only call `IsSwimming` when they are in a subroot.  What about when they are not in a subroot but above elevation? (such as on top of the escape pod).  Should we always call the below function?

NitroxClient/GameLogic/InitialSync/SimulationOwnershipInitialSyncProcessor.cs
```diff
@@ -0,0 +1,42 @@
...
+
+            DependentProcessors.Add(typeof(BuildingInitialSyncProcessor));
+            DependentProcessors.Add(typeof(CyclopsInitialAsyncProcessor));
+            DependentProcessors.Add(typeof(VehicleInitialSyncProcessor));
+        }
+
+        public override IEnumerator Process(InitialPlayerSync packet, WaitScreen.ManualWaitItem waitScreenItem)
+        {
+            int idsSynced = 0;
+            foreach (NitroxId ownerId in packet.InitialSimulationOwnerships)
```
Sunrunner37, line 32: minor: this is an entityId.

NitroxClient/GameLogic/InitialSync/SimulationOwnershipInitialSyncProcessor.cs
```diff
@@ -0,0 +1,42 @@
...
+            {
+                waitScreenItem.SetProgress(idsSynced++, packet.InitialSimulationOwnerships.Count);
+                // Initial locks are transient
+                simulationOwnership.SimulateEntity(ownerId, SimulationLockType.TRANSIENT);
+                Log.Debug($"Transient simulation ownership for {ownerId} from initial sync");
+                yield return null;
```
Sunrunner37, line 38: do you need a frame in between all of these?  I think you might just want a yield break at the very end.

NitroxClient/GameLogic/LiveMixinManager.cs
```diff
@@ -0,0 +1,133 @@
...
+using NitroxModel_Subnautica.DataStructures;
+using UnityEngine;
+
+namespace NitroxClient.GameLogic
+{
+    public class LiveMixinManager
+    {
+        private readonly IMultiplayerSession multiplayerSession;
+        private readonly SimulationOwnership simulationOwnership;
+        private readonly Dictionary<NitroxId, Tuple<float, float>> outstandingChangeHealth = new Dictionary<NitroxId, Tuple<float, float>>();
```
Sunrunner37, line 23: Can we make a small struct to encapsulate tuple members?

NitroxClient/GameLogic/Vehicles.cs
```diff
@@ -493,7 +499,15 @@ public void BroadcastVehicleUndocking(VehicleDockingBay dockingBay, Vehicle vehi
...
             ushort playerId = multiplayerSession.Reservation.PlayerId;
 
-            VehicleUndocking packet = new VehicleUndocking(vehicleId, dockId, playerId);
+            PacketSuppressor<Movement> movementSuppressor = packetSender.Suppress<Movement>();
+            PacketSuppressor<VehicleMovement> vehicleMovementSuppressor = packetSender.Suppress<VehicleMovement>();
+            if (!undockingStart)
+            {
+                movementSuppressor.Dispose();
+                vehicleMovementSuppressor.Dispose();
+            }
```
Sunrunner37, line 508: These are only disposed when not we are not starting to undock?

NitroxClient/GameLogic/Vehicles.cs
```diff
@@ -470,10 +475,11 @@ public void BroadcastVehicleDocking(VehicleDockingBay dockingBay, Vehicle vehicl
             packetSender.Send(packet);
 
             PacketSuppressor<Movement> movementSuppressor = packetSender.Suppress<Movement>();
-            vehicle.StartCoroutine(AllowMovementPacketsAfterDockingAnimation(movementSuppressor));
+            PacketSuppressor<VehicleMovement> vehicleMovementSuppressor = packetSender.Suppress<VehicleMovement>();
```
Sunrunner37, line 478: Can we move to the start of the coroutine?

NitroxModel/Packets/LiveMixinHealthChanged.cs
```diff
@@ -0,0 +1,62 @@
...
+            TechType = techType;
+            Id = id;
+            TotalHealth = totalHealth;
+            LifeChanged = lifeChanged;
+            DamageTakenData = Optional.Empty;
+        }
+
+        public override string ToString()
+        {
+            if (DamageTakenData.HasValue)
```
Sunrunner37, line 52: I think you can add a toString() to DamageTakenData and it will be automatically be invoked if present from optional's toString(). (although I may be wrong).

NitroxServer/Communication/Packets/Processors/WeldActionProcessor.cs
Sunrunner37: Let's replace with a Validate.NotNull() instead of null check then.

NitroxServer/GameLogic/Entities/EntitySimulator.cs
```diff
@@ -78,7 +89,26 @@ private void AssignEntitiesToNewPlayers(Player oldPlayer, List<Entity> entities,
...
                         Log.Info("Player " + player.Name + " has taken over simulating " + entity.Id);
                         ownershipChanges.Add(new SimulatedEntity(entity.Id, player.Id, true, DEFAULT_ENTITY_SIMULATION_LOCKTYPE));
-                        return;
+                        break;
+                    }
+                }
+            }
+        }
+
+        private void AssignSimulationsToNewPlayers(Player oldPlayer, List<NitroxId> entities, List<SimulatedEntity> ownershipChanges)
```
Sunrunner37, line 98: Is the old method `AssignEntitiesToNewPlayers` still used?  I don't see any references (though github is hard to navigate code in the review tool)

NitroxServer/GameLogic/Vehicles/VehicleManager.cs
Sunrunner37: Sounds good! I didn't realize destroy packet was removed.

PR #1300, opened 2020-12-04
Set default culture for all threads in server

Measurity:


NitroxServer-Subnautica/Program.cs
```diff
@@ -153,8 +153,8 @@ private static void ConfigureCultureInfo()
             cultureInfo.NumberFormat.NumberDecimalSeparator = ".";
             cultureInfo.NumberFormat.NumberGroupSeparator = ",";
 
-            Thread.CurrentThread.CurrentCulture = cultureInfo;
-            Thread.CurrentThread.CurrentUICulture = cultureInfo;
+            CultureInfo.DefaultThreadCurrentCulture = cultureInfo;
+            CultureInfo.DefaultThreadCurrentUICulture = cultureInfo;
```
Sunrunner37, line 157: Does this include the current thread or only newly spawned threads?

Measurity: > If you have not explicitly set the culture of any existing threads executing in an application domain, setting the DefaultThreadCurrentCulture property also changes the culture of these threads. However, if these threads execute in another application domain, their culture is defined by the DefaultThreadCurrentCulture property in that application domain or, if no default value is defined, by the default system culture. Because of this, we recommend that you always explicitly set the culture of your main application thread, and not rely on the DefaultThreadCurrentCulture property to define the culture of the main application thread.

https://docs.microsoft.com/en-us/dotnet/api/system.globalization.cultureinfo.defaultthreadcurrentculture?view=net-5.0

PR #1301, opened 2020-12-04
Added config and directory commands to server console

Measurity:
- Config command opens the configuration file in notepad
 - Directory command opens the directory of the main program (ie. Nitrox Launcher)

### DIFF
--- a/NitroxServer/ConsoleCommands/ConfigCommand.cs
+++ b/NitroxServer/ConsoleCommands/ConfigCommand.cs
@@ -1,0 +1,52 @@
+﻿using System.Diagnostics;
+using System.IO;
+using System.Reflection;
+using System.Threading.Tasks;
+using NitroxModel.Core;
+using NitroxModel.DataStructures.GameLogic;
+using NitroxModel.Logger;
+using NitroxModel.Serialization;
+using NitroxServer.ConsoleCommands.Abstract;
+using NitroxServer.Serialization;
+
+namespace NitroxServer.ConsoleCommands
+{
+    internal class ConfigCommand : Command
+    {
+        public ConfigCommand() : base("config", Perms.CONSOLE, "Opens the server configuration file")
+        {
+        }
+
+        protected override void Execute(CallArgs args)
+        {
+            ServerConfig currentActiveConfig = NitroxServiceLocator.LocateService<ServerConfig>();
+            string configFile = Path.Combine(Path.GetDirectoryName(Assembly.GetEntryAssembly()?.Location) ?? "", currentActiveConfig.FileName);
+            if (!File.Exists(configFile))
+            {
+                Log.Error($"Could not find config file at: {configFile}");
+                return;
+            }
+
+            Task.Run(async () =>
+            {
+                await StartProcessAsync("notepad", configFile);
+                ServerConfig newConfigFromFile = PropertiesWriter.Deserialize<ServerConfig>();
+                if (!ServerConfig.ServerConfigComparer.Equals(currentActiveConfig, newConfigFromFile))
+                {
+                    Log.Info("Config file has changed. Restart the server to make the changes take effect!");
+                }
+            });
+        }
+
+        private Task StartProcessAsync(string fileName, string arguments)
+        {
+            ProcessStartInfo info = new ProcessStartInfo();
+            info.FileName = fileName;
+            info.Arguments = arguments;
+            info.UseShellExecute = false;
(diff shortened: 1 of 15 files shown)


NitroxServer/ConsoleCommands/ConfigCommand.cs
```diff
@@ -0,0 +1,70 @@
...
+        }
+
+        private async Task StartProcessAsync(string fileName, string arguments)
+        {
+            ProcessStartInfo info = new ProcessStartInfo();
+            info.FileName = fileName;
+            info.Arguments = arguments;
+            info.UseShellExecute = false;
+            using Process process = Process.Start(info);
+            while (process?.HasExited == false)
```
Jannify, line 64:
```suggestion
            while (!process?.HasExited)
```

NitroxServer/ConsoleCommands/ConfigCommand.cs
```diff
@@ -0,0 +1,70 @@
...
+            {
+                Log.Error($"Could not find config file at: {configFile}");
+                return;
+            }
+
+            Task.Run(async () =>
+            {
+                try
+                {
+                    await StartProcessAsync("notepad", configFile);
```
Jannify, line 43: Is it possible to open the standard programm linked to the file type instead of notepad? Thinking about linux compatibility...

NitroxServer/ConsoleCommands/DirCommand.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Diagnostics;
+using System.IO;
+using System.Reflection;
+using NitroxModel.DataStructures.GameLogic;
+using NitroxModel.Logger;
+using NitroxServer.ConsoleCommands.Abstract;
+
+namespace NitroxServer.ConsoleCommands
+{
+    internal class DirectoryCommand : Command
```
Jannify, line 11: Class and file name are not the same.

NitroxServer/ConsoleCommands/ConfigCommand.cs
```diff
@@ -0,0 +1,70 @@
...
+                }
+                finally
+                {
+                    configOpenLock.Release();
+                }
+                ServerConfig newConfigFromFile = PropertiesWriter.Deserialize<ServerConfig>();
+                if (!ServerConfig.ServerConfigComparer.Equals(currentActiveConfig, newConfigFromFile))
```
Jannify, line 50: What about comparing the "last time edited"-time? Fells lighter then the `ServerConfigComparer` class.

Jannify (changes requested):

NitroxServer/ConsoleCommands/ConfigCommand.cs
Measurity: Definitely lighter but I want to only do this if the values are different from before. Not if it has extra whitespace or saved twice which will change file edit time. I think this is the best way to do that.

NitroxServer/ConsoleCommands/ConfigCommand.cs
Measurity: image

NitroxServer/ConsoleCommands/ConfigCommand.cs
Sunrunner37: Is this equality comparator only for showing the log message?  Maybe we should simply always show the log message, i.e.: "if you made changes, restart the server for them to take effect"

Measurity: Changed the message and removed the comparer class.

NitroxServer/ConsoleCommands/ConfigCommand.cs
Measurity: Had to use pinvoke to get the standard program. As a fallback it iterates windows registry for the possible programs. If that can't be found it defaults to notepad. I added the code in a way that should make it easy to add cross platform support. Will throw `NotSupportedException` if left unimplemented so we'll notice it.

Coding-Hen:
Would this not be better to have as a button on the launcher to edit the config rather than once it has already launched?

Measurity:
Yes it would. But can't hurt to have this in server as well. Could add a restart server command later or add support for hot-reloading the config without restarting the server by broadcasting the necessary changes to all connected players. That's probably far future stuff though.

Jannify (approved):

PR #1309, opened 2020-12-10
Fixed testing errors and added more tests

Jannify:
- Fixed testing errors because of accessing runtime and unity values
- Updated MSTest to v2
- Added NitroxInt3Test
- Added WorldPersistenceTest
- Added GitHub actions support but it's disabled because it's not downloading "resources.assets" which needs to be changed by the creator of the depotDownloader.


NitroxClient/Communication/NetworkingLayer/LiteNetLib/LiteNetLibClient.cs
```diff
@@ -22,10 +23,14 @@ public class LiteNetLibClient : IClient
...
 
-        public LiteNetLibClient(PacketReceiver packetReceiver, NetworkDebugger networkDebugger)
+        public LiteNetLibClient(PacketReceiver packetReceiver)
         {
             this.packetReceiver = packetReceiver;
-            this.networkDebugger = networkDebugger;
+
+            if (NitroxEnvironment.IsNormal) //Testing would fail because NetworkDebugger inherits from Unity.
+            {
+                networkDebugger = NitroxServiceLocator.LocateService<NetworkDebugger>();
```
Sunrunner37, line 32: Is it possible to wire up mocks instead?

.github/workflows/buildingRelease.yml
```diff
@@ -8,13 +8,47 @@ jobs:
...
+      # uses: microsoft/setup-msbuild@v1.0.2
+
+    # - name: Setup VSTest and add to PATH
+      # uses: darenm/Setup-VSTest@v1
+
+    # - name: Build Debug with MSBuild
+      # run: msbuild Nitrox.sln -p:Configuration=Debug
+
+    # - name: Run VSTest
+      # run: vstest.console.exe ./NitroxTest/bin/Debug/NitroxTest.dll
```
Sunrunner37, line 54: Is this supposed to be commented out?

Jannify: Yeah, like written in the pr description GitHub is failing currently because the depot downloader isn't downloading resources.asset. So I comment it out until this is changed.

Jannify:
Waiting for #1325 to switch from classes to records

NitroxModel/DataStructures/GameLogic/Entity.cs
```diff
@@ -118,5 +119,58 @@ private void JsonAfterDeserialization(StreamingContext context)
...
+                   ExistingGameObjectChildIndex == other.ExistingGameObjectChildIndex &&
+                   ChildEntities.SequenceEqual(other.ChildEntities);
+        }
+
+        public override bool Equals(object obj)
+        {
+            return ReferenceEquals(this, obj) || obj is Entity other && Equals(other);
+        }
+
+        public override int GetHashCode()
```
Sunrunner37, line 155: Hmm this might be dangerous as we are indexing on entity a lot of places and using the default reference equals implementation.  Also, not sure what the performance characteristics of these are as they run on child entities.  Is there a way we can forgot messing with these?

Jannify: Yes I can use a custom equality checker in the EntityManager from NitroxServer.

Jannify:
It's based on #1337 because target declaration is different when using `PackageReference`, so merge that one first.

NitroxServer/GameLogic/Entities/EntityData.cs
```diff
@@ -45,5 +45,38 @@ public static EntityData From(List<Entity> entities)
         {
             return new EntityData { Entities = entities };
         }
+
+        public bool Equals(EntityData other)
```
Sunrunner37, line 49: I have mixed feelings about generating equals/hashcode for so many classes to support one unit test. These data models are used/indexed in strange ways.  Any type of missed field (now or in the future) will result in strange behavior that is very difficult to diagnose.  They are also very bad on performance as many of them are recursively calling children equality; when in actual production code we only need reference equality. The equals/hashcode methods are also not unit tested, so net coverage is going down - we are essentially adding what is functionally an integration test.  Instead, I would be more comfortable adding individual tests to verify each model can be serialized/deserialized properly.  That way a missed member would only be impacting a test and not actual production code.

NitroxServer/GameLogic/Bases/BaseData.cs
```diff
@@ -1,12 +1,14 @@
...
 using System.Collections.Generic;
+using System.Linq;
 using Newtonsoft.Json;
 using ProtoBufNet;
 
 namespace NitroxServer.GameLogic.Bases
 {
     [ProtoContract, JsonObject(MemberSerialization.OptIn)]
-    public class BaseData
+    public sealed record BaseData
```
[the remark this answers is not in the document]
Jannify: Do you mean why `record` instead of `class`? It automatic implements equality checks based on values and not reference. So huge chunks of repeating code (like `Equals(object obj)` and `GetHashCode()`) can be avoided. Only if List/Dictionaries are in the `record` we have to override with `Equals(T other)`.

NitroxServer/GameLogic/Entities/EntityData.cs
Jannify: - I know it's not the nicest way but the only other solution I can think of is a huge chunk of equality checks in `WorldPersistenceTest` which nobody will update either.
- I don't see why there will be a performance impact. The equality checks of the models are just used in the test (and shouldn't be somewhere else because they just store data for serialization).
- Net coverage isn't going down on my side: 
grafik
- I am fine with a test class for each model but that would result in 12+ test classes. Just tell me if I should do it that way, I am really eager to implement this because in the past serialization broke a few times and a test like this could indicate an error immediately.

Sunrunner37: The behavior and performance changes can occur when indexing any of these datastructures in a dictionary or hashset.  Whenever you insert into these datastructures, it will execute these methods; this happens recursively for children if they are included in the equals/hashcode.  I don't trust our codebase enough to keep all of these updated and think the possibility of inadvertent bugs is likely.  Would rather punt those to the unit tests.

As for the net coverage, all of these lines will show as covered because they get executed from a unit test.  However, there is nothing doing actual assertions that is saying that the equality/hashcode operations are performing as expected.  So they are showing as covered because they are used to check equality as a side-effect of another unit test for serialization.

 I think we can have a middle ground with a single serialization test class with different methods for each object type. If they all collectively serialize/deserialize, then it would be a significant increase in confidence.

NitroxTest/Serialization/WorldPersistenceTest.cs
```diff
@@ -0,0 +1,426 @@
...
+        private PersistedWorldData worldData;
+        private PersistedWorldData[] worldsDataAfter;
+
+        [TestInitialize]
+        public void Setup()
+        {
+            NitroxServiceLocator.InitializeDependencyContainer(new SubnauticaServerAutoFacRegistrar(), new TestAutoFacRegistrar());
+            NitroxServiceLocator.BeginNewLifetimeScope();
+
+            tempSaveFilePath = Path.Combine(Path.GetTempPath(), "NitroxTestTempDir");
```
Coding-Hen, line 42: Just a suggestion to maybe use a mock file system so that we don't actually have to save to disk?

NitroxTest/Serialization/WorldPersistenceTest.cs
```diff
@@ -0,0 +1,426 @@
...
+                }
+
+                Assert.IsTrue(worldData.WorldData.GameData.StoryGoals.CompletedGoals.SequenceEqual(worldDataAfter.WorldData.GameData.StoryGoals.CompletedGoals), "WorldData.GameData.StoryGoals.CompletedGoals is not equal");
+                Assert.IsTrue(worldData.WorldData.GameData.StoryGoals.RadioQueue.SequenceEqual(worldDataAfter.WorldData.GameData.StoryGoals.RadioQueue), "WorldData.GameData.StoryGoals.RadioQueue is not equal");
+                Assert.IsTrue(worldData.WorldData.GameData.StoryGoals.GoalUnlocks.SequenceEqual(worldDataAfter.WorldData.GameData.StoryGoals.GoalUnlocks), "WorldData.GameData.StoryGoals.GoalUnlocks is not equal");
+
+                Assert.AreEqual(worldData.WorldData.GameData.StoryTiming.ElapsedTime, worldData.WorldData.GameData.StoryTiming.ElapsedTime, "WorldData.GameData.StoryTiming.ElapsedTime is not equal");
+                Assert.AreEqual(worldData.WorldData.GameData.StoryTiming.AuroraExplosionTime, worldData.WorldData.GameData.StoryTiming.AuroraExplosionTime, "WorldData.GameData.StoryTiming.AuroraExplosionTime is not equal");
```
Coding-Hen, line 175:
```suggestion
                Assert.AreEqual(worldData.WorldData.GameData.StoryTiming.ElapsedTime, worldDataAfter.WorldData.GameData.StoryTiming.ElapsedTime, "WorldData.GameData.StoryTiming.ElapsedTime is not equal");
                Assert.AreEqual(worldData.WorldData.GameData.StoryTiming.AuroraExplosionTime, worldDataAfter.WorldData.GameData.StoryTiming.AuroraExplosionTime, "WorldData.GameData.StoryTiming.AuroraExplosionTime is not equal");
```

NitroxTest/Serialization/WorldPersistenceTest.cs
```diff
@@ -0,0 +1,426 @@
...
+                            Id = 2,
+                            Name = "Test2",
+                            IsPermaDeath = false,
+                            Permissions = Perms.PLAYER,
+                            SpawnPosition = NitroxVector3.One,
+                            SubRootId = new NitroxId(),
+                            CurrentStats = new PlayerStatsData(40, 40, 30, 29, 28, 0),
+                            EquippedItems = new List<EquippedItemData>(0),
+                            Modules = new List<EquippedItemData>(0)
+                        }
```
Coding-Hen, line 353: Is it worth adding another with permadeath as true and maybe some modules/equipteditems to make sure that they save correctly

Coding-Hen (changes requested):
Looks good, does look like some of the other tests are failing still.

Jannify:
> Looks good, does look like some of the other tests are failing still. 

Are your sure? At my end it runs all test smoothly. On the other side I do it with ReSharper. Can you try it a few times in a row in the vs window? Last time this helped me.

NitroxTest/Serialization/WorldPersistenceTest.cs
Jannify: I can changed it to a stream.

Coding-Hen:
> 
> 
> > Looks good, does look like some of the other tests are failing still.
> 
> Are your sure? At my end it runs all test smoothly. On the other side I do it with ReSharper. Can you try it a few times in a row in the vs window? Last time this helped me. If not can you say which tests are failing?

Sometimes I love visual studio... Works with both resharper and normal now

NitroxTest/Serialization/WorldPersistenceTest.cs
Jannify: Now that I think about it to test the save and load function of the `WorldPersistence.cs` we can't use bytestream or mock filesystem.

Measurity: Looks like all the required APIs used by `LoadFromFile` do accept `Stream` so with a tiny bit of refactoring (moving code into a new `LoadFromStream`) you can make it testable by using `MemoryStream`.

Jannify: For coverage and "don't edit the code to suit the test" I would leave it as it is.

PR #1313, opened 2020-12-13
Fix sensitive IP and password

Coding-Hen:
Fixed server IP and passwords being sensitive


Sunrunner37:
The purpose of log sensitive was so that the information did not make it into nitrox log files.  I think we should figure out why log sensitive is not functioning as intended.

Measurity (changes requested):
This is not the proper way to fix it. Should fix `Log` class to use the sensitivity logic for console again.

Coding-Hen:
Closing in favour of Measurity's

PR #1319, opened 2020-12-13
Added more metadata to console logging if not a console app

Measurity:


NitroxModel/Logger/Log.cs
```diff
@@ -33,7 +33,12 @@ public static void Setup(bool asyncConsoleWriter = false, InGameLogger inGameLog
                      .MinimumLevel.Debug()
                      .WriteTo.Logger(cnf =>
                      {
-                         string consoleTemplate = "{Timestamp:HH:mm:ss.fff} {Message}{NewLine}{Exception}";
+                         string consoleTemplate = isConsoleApp switch
+                         {
+                             false => $"{{Timestamp:HH:mm:ss.fff}} [{GetLoggerName()}{{{nameof(PlayerName)}}}][{{Level:u3}}] {{Message}}{{NewLine}}{{Exception}}",
+                             _ => "{Timestamp:HH:mm:ss.fff} {Message}{NewLine}{Exception}"
```
Jannify, line 39: Can we add `[]` around the time too?

Measurity: Ok

PR #1323, opened 2020-12-14
Changed log file names to include server, client or launcher

Measurity:
Previously one log file was shared across all programs but that made the log file harder to read.

### DIFF
--- a/NitroxModel/Logger/Log.cs
+++ b/NitroxModel/Logger/Log.cs
@@ -29,13 +29,14 @@ namespace NitroxModel.Logger
             {
                 throw new Exception($"{nameof(Log)} setup should only be executed once.");
             }
+            PlayerName = "";
             logger = new LoggerConfiguration()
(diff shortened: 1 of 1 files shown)


NitroxModel/Logger/Log.cs
```diff
@@ -193,6 +198,28 @@ private static string GetLoggerName()
...
+                return false;
+            }
+
+            return GetLoggerName() switch
+            {
+                { } s when IndexOfAny(s, "server") => "server",
+                { } s when IndexOfAny(s, "launcher") => "launcher",
+                _ => "game"
+            };
+        }
```
Sunrunner37, line 221: Too much cognitive complexity - why not a simple if/elseif?

Measurity: `foreach` was unnecessary. Just liked expression syntax more myself. Used if else now.

PR #1325, opened 2020-12-14
Upgraded projects to C# 9

Measurity:
Added dependencies and dummy classes where needed to support as many C# 9 features as possible.


Dessix (approved):
I love the cleanup from the lack of hintpaths etc. As long as it resolves to the right versions to mesh well with Unity, let's get 9'd up :+1:

PR #1327, opened 2020-12-17
Added automatic UPNP port forwarding on server start

Coding-Hen:


NitroxServer/Communication/NetworkingLayer/NitroxServer.cs
```diff
@@ -29,12 +29,13 @@ public NitroxServer(PacketHandler packetHandler, PlayerManager playerManager, En
 
             portNumber = serverConfig.ServerPort;
             maxConn = serverConfig.MaxConnections;
+            SetupUPNP();
```
Measurity, line 32: IO should not be done in constructors. Constructors are expected to be fast and not cause side effects. Consider putting it in start by making it virtual or require implementors of this class to call it if needed.

NitroxServer/Communication/NetworkingLayer/NitroxServer.cs
```diff
@@ -67,5 +68,24 @@ protected void ProcessIncomingData(NitroxConnection connection, Packet packet)
...
+            NatUtility.DeviceFound += DeviceFound;
+            NatUtility.StartDiscovery();
+        }
+
+        private void DeviceFound(object sender, DeviceEventArgs args)
+        {
+            try
+            {
+                INatDevice device = args.Device;
+                device.CreatePortMapAsync(new Mapping(Protocol.Udp, portNumber, portNumber, 0, "Nitrox Server - Subnautica"));
```
Measurity, line 83: Does not say anything if succeeded. Probably best to let this throw an exception that should be handled so that the error message has more detail.

Measurity (changes requested):

NitroxServer/Communication/NetworkingLayer/NitroxServer.cs
Measurity: Especially in debug mode I expect it to tell me if the port is already forwarded.

Measurity:
https://github.com/OpenRA/OpenRA uses https://github.com/lontivero/Open.NAT and the API is better designed even though it's older. Also the wiki says it has improved discovery performance because searches only for devices that can open ports. Could we use that instead of Mono.NAT?

NitroxServer/Communication/NetworkingLayer/NitroxServer.cs
Measurity: Let's also auto expire so that the port doesn't remain open forever when they stop using Nitrox. A week or month sounds good to me.

Coding-Hen: Added logs and auto expire of 2 hours, do we think that is long enough?
Also added debug logging when in debug mode and a more user friendly message for release build

Coding-Hen:
> 
> 
> https://github.com/OpenRA/OpenRA uses https://github.com/lontivero/Open.NAT and the API is better designed even though it's older. Also the wiki says it has improved discovery performance because searches only for devices that can open ports. Could we use that instead of Mono.NAT?
Looking at the wiki and while attempting to implement it many of the benifits/features did not seem to work such as the auto renew and mapping exceptions (This may be down to my router perhaps), so for now sticking with Mono.Nat as it seems to offer at baseline the same features

Measurity (approved):
Not sure if `packages.json` should be used or if we can just use `ProjectReference` tag in .csproj file. But it's minor and can be fixed later.

NitroxServer/Communication/NetworkingLayer/NitroxServer.cs
```diff
@@ -67,5 +67,32 @@ protected void ProcessIncomingData(NitroxConnection connection, Packet packet)
...
+            NatUtility.DeviceFound += DeviceFound;
+            NatUtility.StartDiscovery();
+        }
+
+        private async void DeviceFound(object sender, DeviceEventArgs args)
+        {
+            try
+            {
+                INatDevice device = args.Device;
+                await device.CreatePortMapAsync(new Mapping(Protocol.Udp, 11000, 11000, 86400, "Nitrox Server - Subnautica"));
```
Measurity, line 82:
```suggestion
               await device.CreatePortMapAsync(new Mapping(Protocol.Udp, 11000, 11000, (int)TimeSpan.FromDays(1).TotalSeconds, "Nitrox Server - Subnautica"));
```

image

Measurity (approved):

PR #1331, opened 2020-12-20
Initial Implementation of P2P tunneling

killzoms:
### **Currently Implemented**

- [x] Server side implementation of the tunneling system
- [ ] Client side implementation of the tunneling system
- [ ] Steam P2P
- [ ] Epic Games P2P
- [ ] Discord P2P
- [ ] ALL THE P2P


Exso12Ecesis:
P2P tunneling would help users be able to use a VPN, thus mitigate the potential for being attacked on the internet by exposing a (otherwise worthless) port on your home router to the world. Coupled with the changes I am planning to make with using the UUID as a unique identifier for the client/server paradigm, coupled with P2P tunneling, I think that both would greatly improve security. It is never a good idea to expose your home IP to the internet, and then advertise you are running a piece of code on a machine from that IP and port. There are people who automatically scan the internet for vulnerable devices and try to hack in to them with bots, and their scan rate varies wildly, but a single modern gaming PC with a 3080 and a Ryzen 5950X can iterate the entire IPv4 space in a matter of hours. If a vulnerability were discovered for Nitrox that allowed them to hack in to the server, it would only take a couple of hours before they could potentially add it to the bot. However, the bad guys are shit at their job, and are slow to adopt. The exploit usually leaks to the darknet during this period from either insiders or people like me who capture malware in the wild and figure out how they work and then try to actively exploit the information gathered by the binary to either stop or thwart the C2 machine controlling it. So someone like me would probably find it, report it and you would know about it long before it becomes actively used in a campaign. If you would like to know more about these people, you can learn about the people who fight them, like http://greynoise.io/. I use them all the time at work to determine who is actively attacking my clients, and who is just scanning the internet. The name is a play on the concept of signal to noise in SIGINT, where you are actively trying to filter nonsense from what you actually care about receiving. It has helped the intelligence community immensely.

Measurity:
Closing PR due to inactivity

PR #1337, opened 2020-12-21
Switching from package.config to PackageReference

Jannify:
If you wondering why I changed the `PackageReference`-format, vs did this as I ran the option to automatic migrate to it.


Coding-Hen:
I have no issue with this but is there any benefit over what we have now? I was under the impression package is the standard

Jannify:
> 
> 
> I have no issue with this but is there any benefit over what we have now? I was under the impression package is the standard

Here are a few benefits listed by Microsoft: https://docs.microsoft.com/en-us/nuget/consume-packages/migrate-packages-config-to-package-reference
Overall `packages.config` is no longer under active development.

PR #1338, opened 2020-12-21
Improved Sync for FireDouse events

### LINKED ISSUE #1173: [World] Fire extinguisher only removes fire for the player using it.
**Version:**
1.2.2.0

**Bug:**
When you are on the Aurora and one player uses the fire extinguisher the fire will only be removed for that player. The other player still sees it and gets damage of it. It simply isn't synced. 

**Expected:**
When one player uses a fire extinguisher, the fire will go out for both users.

s-reich:
Fixes #1173 
* fixed broken rate limiter
* using position-based fire detection in case the fire's NitroxId is not known on receiving end

### DIFF
--- a/NitroxClient/Communication/Packets/Processors/FireDousedProcessor.cs
+++ b/NitroxClient/Communication/Packets/Processors/FireDousedProcessor.cs
@@ -1,8 +1,10 @@
 ﻿using NitroxClient.Communication.Abstract;
 using NitroxClient.Communication.Packets.Processors.Abstract;
 using NitroxClient.MonoBehaviours;
+using NitroxClient.Unity.Helper;
 using NitroxModel.DataStructures;
 using NitroxModel.DataStructures.Util;
+using NitroxModel_Subnautica.DataStructures;
 using NitroxModel.Packets;
 using NitroxModel.Logger;
 using NitroxModel.Helper;
@@ -25,16 +27,16 @@ namespace NitroxClient.Communication.Packets.Processors
         /// </summary>
         public override void Process(FireDoused packet)
         {
-            Optional<GameObject> gameObject = NitroxEntity.GetObjectFrom(packet.Id);
+            Optional<GameObject> gameObject = NitroxEntity.GetObjectFrom(packet.FireId);
             if (!gameObject.HasValue)
             {
                 // cannot find that ID. try to douse something at the same location.
 
                 Optional<GameObject> parentObject = NitroxEntity.GetObjectFrom(packet.ParentId);
-                Validate.IsPresent(parentObject, $"Found neither fire {packet.Id} nor parent {packet.ParentId}");
+                Validate.IsPresent(parentObject, $"Found neither fire {packet.FireId} nor parent {packet.ParentId}");
 
                 // look for closest fire
-                Vector3 firePos = new Vector3(packet.xpos, packet.ypos, packet.zpos);
+                Vector3 firePos = packet.Position.ToUnity();
                 float maxDist = float.MaxValue;
 
                 foreach (Fire candidate in parentObject.Value.GetComponentsInChildren<Fire>())
@@ -50,12 +52,12 @@ namespace NitroxClient.Communication.Packets.Processors
                 Validate.IsPresent(gameObject, $"Did not find any fires in parent {packet.ParentId}");
 
                 NitroxId replacedId = NitroxEntity.GetId(gameObject.Value);
-                Log.Info($"FireDousedProcessor: Using substitute fire {replacedId} instead of nonexistant {packet.Id}");
+                Log.Info($"FireDousedProcessor: Using substitute fire {replacedId} instead of nonexistant {packet.FireId}");
             }
 
             using (packetSender.Suppress<FireDoused>())
             {
-                gameObject.Value.GetComponent<Fire>().Douse(packet.DouseAmount);
+                gameObject.Value.RequireComponent<Fire>().Douse(packet.DouseAmount);
             }
         }
     }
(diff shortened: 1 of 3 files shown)


NitroxModel/Packets/FireDoused.cs
```diff
@@ -11,14 +11,22 @@ public class FireDoused : Packet
     {
         public NitroxId Id { get; }
         public float DouseAmount { get; }
+        public NitroxId ParentId { get; }
+        public float xpos { get; }
+        public float ypos { get; }
+        public float zpos { get; }
```
Sunrunner37, line 18: You can use NitroxVector3

NitroxClient/Communication/Packets/Processors/FireDousedProcessor.cs
```diff
@@ -22,11 +25,37 @@ public FireDousedProcessor(IPacketSender packetSender)
...
+                Validate.IsPresent(gameObject, $"Did not find any fires in parent {packet.ParentId}");
+
+                NitroxId replacedId = NitroxEntity.GetId(gameObject.Value);
+                Log.Info($"FireDousedProcessor: Using substitute fire {replacedId} instead of nonexistant {packet.Id}");
+            }
 
             using (packetSender.Suppress<FireDoused>())
             {
-                fireGameObject.RequireComponent<Fire>().Douse(packet.DouseAmount);
+                gameObject.Value.GetComponent<Fire>().Douse(packet.DouseAmount);
```
Sunrunner37, line 58: I think we still want RequireComponent so this doesn't null pointer on missing (hard to debug vs Required)

NitroxClient/Communication/Packets/Processors/FireDousedProcessor.cs
```diff
@@ -22,11 +25,37 @@ public FireDousedProcessor(IPacketSender packetSender)
         /// </summary>
         public override void Process(FireDoused packet)
         {
-            GameObject fireGameObject = NitroxEntity.RequireObjectFrom(packet.Id);
+            Optional<GameObject> gameObject = NitroxEntity.GetObjectFrom(packet.Id);
+            if (!gameObject.HasValue)
+            {
+                // cannot find that ID. try to douse something at the same location.
```
Sunrunner37, line 31: Is it possible that we can deterministically know these IDs ahead of time.  For example, if there are N number of fire spawn points, we can make them deterministic based on parent cyclops id (NitroxId.nextId()).

NitroxModel/Packets/FireDoused.cs
s-reich: Found it, thanks.

NitroxClient/Communication/Packets/Processors/FireDousedProcessor.cs
s-reich: Ack.
Hard to navigate when IntelliSense does not show up the out-of-scope extensions...

NitroxClient/Communication/Packets/Processors/FireDousedProcessor.cs
s-reich: That goes way beyond my knowledge of Unity, Nitrox or Subnautica :)
I'm just glad the ParentID was known in case of Aurora ... will try to find out how to start fires on Cyclops this evening and see if my patch handles those well.

Are there other places where a Fire might show up?

NitroxModel/Packets/FireDoused.cs
```diff
@@ -9,21 +10,27 @@ namespace NitroxModel.Packets
...
+            FireId = fireId;
             DouseAmount = douseAmount;
+            ParentId = parentId;
+            Position = position;
         }
 
         public override string ToString()
         {
-            return "[FireDoused Id: " + Id + "Douse Amount: " + DouseAmount + "]";
+            return "[FireDoused Id: " + FireId + "Douse Amount: " + DouseAmount + "]";
```
Coding-Hen, line 33:
```suggestion
            return $"[FireDoused Id: {FireId} Douse Amount: {DouseAmount}]";
```
To be inline with other packet tostring methods

NitroxClient/Communication/Packets/Processors/FireDousedProcessor.cs
Sunrunner37: Sorry for the slow reply, it is the holidays.

After examining the code, it looks like cyclops fires will always lookup a valid gameobject due to being tagged with an id on spawn.  So you don't need to be concerned with that use-case.

However, I think this implementation may only work for local players that are in the immediate area.  If one player extinguishes the fire, while the other one is far away, then the other player will still see the fire.  Can you verify that this is the case?

I think the right long-term fix is for the server to know about the fires so that it can appropriately allocate an id on spawn.  That would allow you to use Nitrox's EntityMetadata system to store the douse amount; that way, when players enter the area, it will automatically apply the correct douse and even persist it on server reset.  

Example metadata: https://github.com/SubnauticaNitrox/Nitrox/blob/master/NitroxClient/GameLogic/Spawning/Metadata/KeypadMetadataProcessor.cs

Area where we spawn (see if the parent is there by adding breakpoint on class id): https://github.com/SubnauticaNitrox/Nitrox/blob/8815fb0a8fc80ff030bc12c46dce10fe380febde/NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs#L265

Sunrunner37: Just to update this.  The server knows about the parent objects in the aurora.  Thus, if the fire does not have a nitrox id, you can go one level up to get it.  Then you can use the entity metadata system to persist this data.  That will remove the need for the location checks and will work for out of range players.

Sunrunner37:
Just checking in on this.

Jannify:
Closed due to inactivity. If someone likes to pick up the code and finish it they are free to do.

PR #1341, opened 2020-12-26
Persistent Storage and Sync of Plant Growth

### LINKED ISSUE #810: Plant pot syncing
When I plant plants in a plant pot after a while I can see different state, that my brother on same LAN. After relogging the plants are like just planted. Tested only large indoors plant pot with three different large edible plants from the island without gun

s-reich:
Fixes #810

- Client records current time and adds it to the item metadata
- Server stores that time and sends it to other clients
- Clients receiving such item data will "fix" plant growth according to the given time.

This will **not**
- remember or populate fruit on plants (Seaweed / Lanterns)
- automatically fix existing plants -- these need to be harvested and replanted.


s-reich:
Putting up the PR for review, needs some testing with an actual second player (although relogging is a pretty good simulator for this).

Coding-Hen:
Nice to have somebody new contribute welcome (Have you joined our discord?). Did some quick testing seems that the player who plants gets the correct growing but the other players get set to fully grown straight away in most circumstances, however I did get some sucsess with acid mushrooms in the outside grownbed occasionally. Reconnection gives the same growth as a other players and doesnt seem to sync across progress though I imagine this is linked to the former issue above
image
image
Will do some further testing see if I can issolate the reason for this

s-reich:
@Coding-Hen : I checked again and I see two problems:
1) I'm calling `GrowingPlant.SetProgress` with a percentage (0..100); it probably has to be a fraction (0..1)
2) `DayNightCycle.main.timePassedAsDouble` might not be what I need. Needs more checking, but it might still progress while the server is paused.

s-reich:
Works pretty well with the latest commit.
Just found a new problem with Alien Containment unit, which needs extra code to create common sub-element IDs for all clients (Planter inside the facility)

NitroxClient/GameLogic/Bases/Spawning/BaseRoomWaterParkProcessor.cs
```diff
@@ -0,0 +1,51 @@
...
+using NitroxClient.GameLogic.Helper;
+
+namespace NitroxClient.GameLogic.Bases.Spawning
+{
+    /*
+     * When a bio reactor is created, two objects are spawned: the main world object (BaseBioReactorGeometry) and
+     * the core power logic as a separate game object (BaseBioReactor, also known as a 'module').  The BaseBioReactor 
+     * resides as a direct child of the base object (probably so UWE could iterate them easy).  When the object spawns, 
+     * we use this class to set a deterministic id seeded by the parent id.  This keeps inventory actions in sync.
+     */
```
Sunrunner37, line 15: We need to update the comment.

NitroxClient/GameLogic/Bases/Spawning/BaseRoomWaterParkProcessor.cs
```diff
@@ -0,0 +1,51 @@
...
+        {
+            TechType.BaseWaterPark
+        };
+
+        public override void SpawnPostProcess(Base latestBase, Int3 latestCell, GameObject finishedPiece)
+        {
+            NitroxId pieceId = NitroxEntity.GetId(finishedPiece);
+
+            WaterParkPiece waterParkPiece = finishedPiece.GetComponent<WaterParkPiece>();
+            if(null==waterParkPiece)
```
Sunrunner37, line 28: use the implicit boolean operator for unity objects.  Also, no yoda conditions.

NitroxClient/GameLogic/Entities.cs
```diff
@@ -66,9 +66,16 @@ public void Spawn(List<Entity> entities)
...
+                    try
+                    {
+                        Optional<GameObject> parent = NitroxEntity.GetObjectFrom(entity.ParentId);
+                        Spawn(entity, parent);
+                        SpawnAnyPendingChildren(entity);
+                    }
+                    catch(OptionalEmptyException<GameObject> e)
+                    {
+                        Log.Error($"Failed to spawn Entity {entity.Id}, a {entity.TechType}: {e.Message}");
+                    }
```
Sunrunner37, line 78: ?

NitroxClient/GameLogic/Helper/PlantableHelper.cs
```diff
@@ -0,0 +1,206 @@
+﻿using System;
+using UnityEngine;
+using NitroxModel.DataStructures.GameLogic;
+using NitroxModel.Logger;
+using NitroxClient.MonoBehaviours;
+
+namespace NitroxClient.GameLogic.Helper
+{
+    public static class GameObjectExtensions
```
Sunrunner37, line 9: We reserve extensions for only very common reusable logic.  Since items like 'FixPlantGrowth' are specific pieces of logic, it should go into a file specific to it.

NitroxClient/GameLogic/Helper/PlantableHelper.cs
```diff
@@ -0,0 +1,206 @@
...
+        /// <summary>
+        /// Pereform a recursive dump of the objectr into the clients log file.
+        /// The dump will contain a parent trace (object names and types)
+        /// and a tree of all children (defined by Transform).
+        /// For each object shown, the attached Componet types are listed.
+        /// </summary>
+        /// <param name="gameObject"></param>
+        /// <param name="currentDepth"></param>
+        /// <param name="maxDepth"></param>
+        public static void DumpRecursive(this GameObject gameObject, int maxDepth = 10, int currentDepth=0)
```
Sunrunner37, line 58: We have similar functionality already in DebugUtils

NitroxClient/Communication/Packets/Processors/ItemContainerAddProcessor.cs
```diff
@@ -27,6 +27,12 @@ public override void Process(ItemContainerAdd packet)
             NitroxEntity.SetNewId(item, itemData.ItemId);
             
             itemContainer.AddItem(item, itemData.ContainerId);
+
+            // special Planting helper
+            if( itemData is PlantableItemData plantableData)
+            {
+                item.FixPlantGrowth(plantableData);
+            }
```
Sunrunner37, line 35: Would prefer a ItemContainerAdd post processor.  Similar to pattern used for other things (BasePieceMetadata, DroppedItems, EntityMetadata).

NitroxClient/GameLogic/ItemContainers.cs
```diff
@@ -26,8 +26,19 @@ public void BroadcastItemAdd(Pickupable pickupable, Transform containerTransform
...
             byte[] bytes = SerializationHelper.GetBytes(pickupable.gameObject);
+            NitroxId ownerId = GetOwner(containerTransform);
 
-            ItemData itemData = new ItemData(GetOwner(containerTransform), itemId, bytes);
+            ItemData itemData;
+            Plantable plant = pickupable.GetComponent<Plantable>();
+            if (plant && plant.currentPlanter)
+            {
+                // special case: we want to remember the time when the plant was added, so we can simulate growth
+                itemData = new PlantableItemData(ownerId, itemId, bytes, DayNightCycle.main.timePassedAsDouble);
```
Sunrunner37, line 36: Is timePassedAsDouble not serialized in bytes?  Is there a way to add it?

NitroxClient/GameLogic/ItemContainers.cs
```diff
@@ -26,8 +26,19 @@ public void BroadcastItemAdd(Pickupable pickupable, Transform containerTransform
         {
             NitroxId itemId = NitroxEntity.GetId(pickupable.gameObject);
             byte[] bytes = SerializationHelper.GetBytes(pickupable.gameObject);
+            NitroxId ownerId = GetOwner(containerTransform);
 
-            ItemData itemData = new ItemData(GetOwner(containerTransform), itemId, bytes);
+            ItemData itemData;
+            Plantable plant = pickupable.GetComponent<Plantable>();
+            if (plant && plant.currentPlanter)
```
Sunrunner37, line 33: Same as above, prefer some factory.

NitroxClient/GameLogic/Entities.cs
s-reich: Unrelated change. Problem here was that when spawning of a single entity failed, the whole entity spawner failed and would end up as stalling login.

NitroxClient/GameLogic/ItemContainers.cs
s-reich: `class Plantable` does contain `public float plantAge`, but storing that won't help without any reference time.

PR #1349, opened 2020-12-30
Updated Debug seed

killzoms:


killzoms:
This Seed positions the EscapePod at -111.881, 0, -319.9703 which is 2.6489 units away from our original debug spawn location -112.2, 0, -322.6

PR #1351, opened 2020-12-30
Added seed to test and a little bit of styling

Jannify:


NitroxServer/Serialization/World/WorldData.cs
```diff
@@ -33,16 +33,16 @@ public class WorldData
         [JsonProperty, ProtoMember(6)]
         public EscapePodData EscapePodData { get; set; }
 
-
         [JsonProperty, ProtoMember(8)]
```
dartasen, line 36: I think the ProtoMember should be at 7

PR #1353, opened 2021-01-02
Updated MainMenu-UI

Jannify:
- Reduced load/reload time and freeze frame on button presses.
- Added caching for frequently used GameObjects, Transforms, etc.
- Fixed a potential bug in `NitroxServer.SetupUPNP()`

### DIFF
--- a/NitroxClient/MonoBehaviours/Gui/MainMenu/MainMenuMultiplayerPanel.cs
+++ b/NitroxClient/MonoBehaviours/Gui/MainMenu/MainMenuMultiplayerPanel.cs
@@ -5,7 +5,6 @@ using System.Net;
 using System.Net.Sockets;
 using System.Text.RegularExpressions;
 using NitroxClient.Unity.Helper;
-using NitroxModel.Core;
 using NitroxModel.Logger;
 using UnityEngine;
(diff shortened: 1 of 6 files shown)


NitroxClient/MonoBehaviours/Gui/MainMenu/MainMenuMultiplayerPanel.cs
```diff
@@ -78,55 +84,48 @@ public void CreateServerButton(string text, string joinIp, string joinPort)
...
             serverLines.RemoveAt(index);
             File.WriteAllLines(SERVER_LIST_PATH, serverLines.ToArray());
         }
 
-        public void OpenJoinServerMenu(string serverIp, string serverPort)
+        public static void OpenJoinServerMenu(string serverIp, string serverPort)
         {
+            if (main == null)
+            {
+                return;
```
Measurity, line 116: Should throw exception because this is an important function. If it fails, program doesn't work right.

Measurity (approved):
Totally overlooked that 11000 static port before 👍

Coding-Hen:
Nice catch on the port I clearly forgot to reverse this change once I had finished testing

PR #1360, opened 2021-01-04
Fixed server restart not waiting long enough to save

Measurity:


NitroxServer/ConsoleCommands/RestartCommand.cs
```diff
@@ -28,8 +28,9 @@ protected override void Execute(CallArgs args)
                 return;
             }
 
+            server.Save();
             using Process proc = Process.Start(program);
-            server.Stop();
+            server.Stop(false);
```
Sunrunner37, line 33: Can you move this line above the process start?

PR #1370, opened 2021-01-06
Added save file upgradability

Jannify:
- Added two new server commands: `swapSerializer` and `upgrade`
  - It's now possible to change the serializer (Protobuf/Json) while the server is running
  - Save files can now be upgraded to new versions by editing the raw json data.
- Fixed a bug where the server commands changing config values were unable to save the config

Example:
The field `Players` in `PlayerData` was renamed to `Persons` in 1.4.0.1. Adding this class
```cs
public class Upgrade_1_4_0_1 : SaveDataUpgrade
{
    public override Version TargetVersion { get; } = new Version(1, 4, 0, 1);

    protected override void UpgradePlayerData(JObject data)
    {
        data["Players"].Rename("Persons");
    }
}
```
will rename the field in the save file so that it's usable again.

Important: the `upgrade` command needs to be send before the server overwrites the files (normally after 12 sec)


NitroxServer/Server.cs
```diff
@@ -66,12 +66,8 @@ public void Save()
                 return;
             }
 
-            // Don't overwrite config changes that users made to file
-            if (!File.Exists(serverConfig.FileName))
-            {
-                NitroxConfig.Serialize(serverConfig);
-            }
             IsSaving = true;
+            NitroxConfig.Serialize(serverConfig); // This is overwriting the config file => server has to be closed before making changes to it
```
Measurity, line 70: I guess the comment before wasn't clear that this introduces a bug and overrides the player's own config setting on close. Any command that changes settings should read the existing settings, set the value, then write it back immediately.

Jannify: "_Any command that changes settings should read the existing settings, set the value, then write it back immediately._" Unfortunately that is not the case. I will revert my changes and write a config save function.

Measurity: A config save function is not necessary I think. The `swapSerializer` can resave the config file when called. It should be transparent to the user to avoid confusion on when things are saved (should always be saved unless logged/printed otherwise).

NitroxServer/Serialization/SaveDataUpgrades/SaveDataUpgrade.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System;
+using System.IO;
+
+namespace NitroxServer.Serialization.Upgrade
+{
+    public abstract class SaveDataUpgrade
+    {
+        public abstract Version TargetVersion { get; }
```
Measurity, line 8: This should probably be a range instead of 1 target version so that really old Nitrox versions aren't erroneously getting "upgraded" to an invalid state.

Measurity (changes requested):

NitroxServer/Serialization/SaveDataUpgrades/SaveDataUpgrade.cs
Jannify: It was intended like this so that for each version there is an upgrade file to it from the last one. I think that a range would complicated thinks with this feature carrying on.

Measurity:
```cs
                foreach (SaveDataUpgrade upgrade in upgrades)
                {
                    if (upgrade.TargetVersion > saveFileVersion.Version)
                    {
                        SendMessage(args.Sender, $"Applying {upgrade.GetType().Name}");
                        upgrade.UpgradeData(saveDir, fileEnding);
                    }
                }
```
This code applies all upgrades if TargetVersion is higher so if you tried to upgrade from 0.0.1 to 1.5 you can get issues with order of upgrades and just incompatibilities in general. Unless I'm misunderstanding the code.

Jannify: As far as my testing goes Autofac is sorting them alphabetically . I tested it with `Upgrade_1_4_0_0` and `Upgrade_1_3_5_0` while 1.3.5.0 was executed first.

NitroxServer/Serialization/SaveDataUpgrades/SaveDataUpgrade.cs
```diff
@@ -0,0 +1,44 @@
...
+            File.WriteAllText(baseDataPath, UpgradeBaseData(File.ReadAllText(baseDataPath)));
+            File.WriteAllText(playerDataPath, UpgradePlayerData(File.ReadAllText(playerDataPath)));
+            File.WriteAllText(worldDataPath, UpgradeWorldData(File.ReadAllText(worldDataPath)));
+            File.WriteAllText(entityDataPath, UpgradeEntityData(File.ReadAllText(entityDataPath)));
+        }
+
+        protected virtual string UpgradeBaseData(string data)
+        {
+            return data;
+        }
```
Sunrunner37, line 26: Wouldn't it be better to version the data models themselves instead of mutating raw string data?  IMO, upgrade base data should take a BaseData_1_3 and output a BaseData_1_4.

Jannify: Yeah that would be the better approach but then we would need to save each iteration of the save file structure with all changing classes which would probably be too much after a handful of version.

killzoms: I mean we can remove those changes if the version is 5 versions prior or more to help keep bloat down

NitroxServer/ConsoleCommands/UpgradeCommand.cs
```diff
@@ -0,0 +1,60 @@
...
+                    if (upgrade.TargetVersion > saveFileVersion.Version)
+                    {
+                        SendMessage(args.Sender, $"Applying {upgrade.GetType().Name}");
+                        upgrade.UpgradeData(saveDir, fileEnding);
+                    }
+                }
+                worldPersistence.SaveDataSerializer.Serialize(Path.Combine(saveDir, "Version" + fileEnding), new SaveFileVersion());
+                SendMessage(args.Sender, $"Save file was successfully upgraded to {NitroxEnvironment.Version}");
+                server.Stop(false);
+                Thread.Sleep(2000);
```
Measurity, line 56: Why sleep?

Jannify: That way the last message is readable. Without it the server is closed immediately which gives the impression of a fatal error.

Measurity: Make sense. Though I recommend adding a feature to make the server stop with a message to display which will:
1. Show message in console.
2. Wait for user to enter any key to stop.

killzoms: Wouldnt that be implementable via adding some method to NitroxLogger
```csharp
void LogAndWait()
{
  Info("Press any key to continue");
  Console.Read();
}
```

Measurity: I don't agree with that because you don't always have a console when logging. Would need to do something similar to `InGameLogger` abstraction to noop if no console is available.

Jannify: I would add the method to NitroxServer/Program

NitroxServer/Serialization/SaveDataUpgrades/SaveDataUpgrade.cs
Jannify: I counted and we have around 25 files/classes in NitroxServer used for saving, not including gamelogic stored in NitroxModel. Even if we just support 5 versions it will bloat up the program and the exported dll.

Sunrunner37: To be clear, I only think we should master a new version when the model changes, not a duplicate of every model for every version.  However, we should likely only start that when in beta as the changes right now still aren't backwards compatible even if you update the model (missing stuff in the world).  However, we can define the framework for it now.

Jannify: Yeah, editing raw json is not the ideal way but as you said it lays a foundation (like saving the assembly version in the save files). I will look into making the json editing easier/more useful. 

The current solution should just be for things easy to fix and something like "adding new values to models" will break it, but if we just remove a field it is fixable by this early approach.

Jannify: I switched from string based editing to `JObject` which should improve editing a lot.

PR #1373, opened 2021-01-08
Fixed loading of missing translation

MatrixDJ96:
This commit aims to fix blank menu in Subnautica

immagine

### DIFF
--- a/NitroxPatcher/Patches/Persistent/Language_LoadLanguageFile_Patch.cs
+++ b/NitroxPatcher/Patches/Persistent/Language_LoadLanguageFile_Patch.cs
@@ -17,7 +17,7 @@ namespace NitroxPatcher.Patches.Persistent
         public static void Postfix(string language, Dictionary<string, string> ___strings)
         {
             string[] files = {
-                Path.Combine(NitroxAppData.Instance.LauncherPath, "LanguageFiles", "English.json"), //Using English as fallback.
+                Path.Combine(NitroxAppData.Instance.LauncherPath, "LanguageFiles", "English.json"), // Using English as fallback.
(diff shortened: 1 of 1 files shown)


NitroxPatcher/Patches/Persistent/Language_LoadLanguageFile_Patch.cs
```diff
@@ -26,6 +26,7 @@ public static void Postfix(string language, Dictionary<string, string> ___string
                 if (!File.Exists(file))
                 {
                     Log.Warn($"No language file was found for at {file}. Using English as fallback");
+                    break;
```
Jannify, line 29: `continue` fits here more because you want to terminate the failing language file not all other in the array.

MatrixDJ96: I want to stop because no more languages are loading after the first (English)

MatrixDJ96: I think that in this case using `break` or `continue` is the same but if you prefer i could change it

Jannify: You are right the outcome is the same but if (for whatever reason) we add more language files to load or just English fails the other file wont be loaded.

MatrixDJ96 (commented):
Changed `break` with `continue`

Jannify (approved):

Coding-Hen (approved):

dartasen (approved):

PR #1374, opened 2021-01-08
Added EntityData.nitrox to zip backup

ProjectPatatoe:
EntityData was not being added in backup zip

### DIFF
--- a/NitroxServer/Serialization/World/WorldPersistence.cs
+++ b/NitroxServer/Serialization/World/WorldPersistence.cs
@@ -128,6 +128,7 @@ namespace NitroxServer.Serialization.World
                     zipFile.AddFile(Path.Combine(saveDir, "BaseData" + fileEnding));
                     zipFile.AddFile(Path.Combine(saveDir, "PlayerData" + fileEnding));
                     zipFile.AddFile(Path.Combine(saveDir, "WorldData" + fileEnding));
+                    zipFile.AddFile(Path.Combine(saveDir, "EntityData" + fileEnding));
                     zipFile.Save(Path.Combine(saveDir, "worldBackup.zip"));
                 }
 #if DEBUG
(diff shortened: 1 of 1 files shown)


dartasen (approved):

dartasen:
Maybe It should be better to add every file inside the save dir, so we don't need to maintain this area everytime the save format is changed

ProjectPatatoe:
yea thats a good idea. I'll change that.

ProjectPatatoe:
I made it for any .nitrox because an existing zip would be included. If we want to change where the zip is saved then it would make sense to grab the entire directory.

Coding-Hen (approved):

NitroxServer/Serialization/World/WorldPersistence.cs
```diff
@@ -124,10 +124,8 @@ internal Optional<World> LoadFromFile(string saveDir)
                 //Backup world if loading fails
                 using (ZipFile zipFile = new ZipFile())
                 {
-                    zipFile.AddFile(Path.Combine(saveDir, "Version" + fileEnding));
-                    zipFile.AddFile(Path.Combine(saveDir, "BaseData" + fileEnding));
-                    zipFile.AddFile(Path.Combine(saveDir, "PlayerData" + fileEnding));
-                    zipFile.AddFile(Path.Combine(saveDir, "WorldData" + fileEnding));
+                    string[] nitroxFiles = Directory.GetFiles(saveDir, "*.nitrox");
```
Jannify, line 127: That doesn't include json files if used (see `fileEnding`).

Jannify (changes requested):

Jannify (approved):

dartasen (approved):

Coding-Hen (approved):

Measurity (approved):

PR #1384, opened 2021-01-11
Launcher Tweaks

diabolical17:
# Fade between pages.

Feel free to modify the speed if it doesn't feel correct.

gif

# Debug Mode

Can be built onto, currently it shows a border and label when the application is ran in the DEBUG configuration.

Screenshot

Currently, the variable stating that debug mode is enabled can be found in `NitroxLauncher.MainWindow.DebugMode`, however, in the future it may be saved as a setting so it can be used in RELEASE configuration to reproduce bugs. That is an entirely different PR though.

### DIFF
--- a/NitroxLauncher/Controls/FaderFrame.cs
+++ b/NitroxLauncher/Controls/FaderFrame.cs
@@ -1,0 +1,117 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Windows.Controls;
+using System.Diagnostics;
+using System.Windows.Navigation;
+using System.Windows.Media.Animation;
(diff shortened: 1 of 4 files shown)


diabolical17:
This also adds a new namespace:

`NitroxLauncher.Controls`

It currently only contains one control, the FaderFrame. But it can be populated when the need arises.

diabolical17:
My next PR will allow the user to enable debug mode in the settings to access a developer tab.

This developer tab will contain features such as log analysis and launching Subnautica Nitrox with an external console window.

diabolical17:
Closing - Electron Launcher makes this redundant.

PR #1386, opened 2021-01-13
Nitrox Public API v0

diabolical17:
This puts down the infrastructure of the Nitrox Public API.

## Infomation

The Nitrox Public API is a public API that runs on `http://{NitroxServerIP}:3000/` and includes various endpoints to collect data or interact with your Nitrox Server.

## Spec

The NPAPI is built using Grapevine, a not very well known library that uses Attributes to register methods.

Example:

```cs
[RestRoute("Delete", "/server/stop")]
public async Task stopServer(IHttpContext ctx) {
    Server srv = Server.Instance;
    if (ctx.Request.Headers.Get("Auth") != srv.GetServerConfig().AdminPassword) {
        ctx.Response.StatusCode = 401;
        await ctx.Response.SendResponseAsync(JsonConvert.SerializeObject(new UnauthorizedResponse("Please pass your ADMIN PASSWORD through the Auth header to access this endpoint."), Formatting.Indented));
    } else {
        await ctx.Response.SendResponseAsync(JsonConvert.SerializeObject(new ServerStoppedResponse(), Formatting.Indented));
        srv.Stop();
    }
}
```

These Attributes allow easy addition of more methods, allowing the API to be updated easily.

## Usages

Although you may not be able to think of many usages, here is a wide range:

- Discord Bots
- Website Integrations 
	+ Show currently online players on your website or etc.
- Future Nitrox Usages
	+ Public serverlist?
	+ Paid Hosting?
	+ Web Admin Panels?

## Documentation and Access

The server can be automatically ran when the server is started using the command `/api enable/disable`. It can be stopped using the `/api stop` and restarted using the `/api start` commands.

Documentation: https://documenter.getpostman.com/view/9952152/TVzSmJ8j

### DIFF
--- a/NitroxServer/Communication/NetworkingLayer/NitroxServer.cs
+++ b/NitroxServer/Communication/NetworkingLayer/NitroxServer.cs
@@ -74,6 +74,15 @@ namespace NitroxServer.Communication.NetworkingLayer
             NatUtility.StartDiscovery();
         }
 
+        public PlayerManager GetPlayerManager()
+        {
+            return playerManager;
+        }
+        public int GetPlayerCount()
+        {
+            return playerManager.GetConnectedPlayers().Count;
+        }
+
         private async void DeviceFound(object sender, DeviceEventArgs args)
         {
             try
--- a/NitroxServer/ConsoleCommands/APICommand.cs
+++ b/NitroxServer/ConsoleCommands/APICommand.cs
@@ -1,0 +1,88 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Net;
+using System.Text;
+using System.Threading.Tasks;
+using NitroxModel.DataStructures.GameLogic;
+using NitroxServer.ConsoleCommands.Abstract;
+using NitroxServer.ConsoleCommands.Abstract.Type;
+
+namespace NitroxServer.ConsoleCommands
+{
+    public class APICommand : Command
+    {
+        private string GetIPAddress()
+        {
+            string address = "";
+            WebRequest request = WebRequest.Create("http://checkip.dyndns.org/");
+            using (WebResponse response = request.GetResponse())
+            using (StreamReader stream = new StreamReader(response.GetResponseStream()))
+            {
+                address = stream.ReadToEnd();
+            }
+
+            int first = address.IndexOf("Address: ") + 9;
+            int last = address.LastIndexOf("</body>");
+            address = address.Substring(first, last - first);
+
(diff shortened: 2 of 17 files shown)


diabolical17:
Discussion from Discord:

NitroxServer/Communication/NetworkingLayer/NitroxServer.cs
```diff
@@ -74,6 +74,15 @@ protected void SetupUPNP()
...
 
+        public PlayerManager GetPlayerManager()
+        {
+            return playerManager;
+        }
+        public int GetPlayerCount()
+        {
+            return playerManager.GetConnectedPlayers().Count;
+        }
+
```
Jannify, line 85: We use Autofac to get these objects.
Eg. `NitroxServiceLocator.LocateService<PlayerManager>();` or with parameters and initializing it with Autofac (see [here](https://github.com/SubnauticaNitrox/Nitrox/blob/c7dc535fc8658edb220363ba68fb5fab2675a3d6/NitroxServer/ServerAutoFacRegistrar.cs#L33)).

So pls remove any unnecessary getter/setter and use one of the options above.

NitroxServer/ConsoleCommands/APICommand.cs
```diff
@@ -0,0 +1,88 @@
...
+            {
+                address = stream.ReadToEnd();
+            }
+
+            int first = address.IndexOf("Address: ") + 9;
+            int last = address.LastIndexOf("</body>");
+            address = address.Substring(first, last - first);
+
+            return address;
+        }
```
Jannify, line 31: Duplicate: See [WebHelper.cs](https://github.com/SubnauticaNitrox/Nitrox/blob/master/NitroxModel/Helper/WebHelper.cs)

NitroxServer/Server.cs
```diff
@@ -54,11 +61,20 @@ public string SaveSummary
...
                 builder.AppendLine($" - Known tech: {world.GameData.PDAState.KnownTechTypes.Count}");
                 builder.AppendLine($" - Vehicles: {world.VehicleManager.GetVehicles().Count()}");
+                builder.AppendLine($" - NPAPI Enabled: {serverConfig.EnablePublicAPI}");
 
                 return builder.ToString();
             }
         }
 
+        public APIServer API { get; }
+        public bool Paused { get; private set; }
```
Jannify, line 71: Minor: Move up to the other fields/properties.

Jannify (changes requested):

PR #1402, opened 2021-01-19
Fixed StoryTimingData and Seed from resetting

Jannify:
grafik


Coding-Hen (approved):
Tested and seems to work as expected

Exso12Ecesis:
I have also tested and confirmed the behavior described in the snapshot snippet occurs now, there also seems to be no regression issues with the PDA Data.

image

PR #1407, opened 2021-01-21
Added unique user identification token usage to Nitrox.

Exso12Ecesis:
I need the team's help to address the FIXME and regression issue in ShowAddServerWindow(). The regression issue is that users who have the 3 piped or 2 piped format will not store the token subsequently.

### DIFF
--- a/NitroxClient/MonoBehaviours/Gui/MainMenu/MainMenuMultiplayerPanel.cs
+++ b/NitroxClient/MonoBehaviours/Gui/MainMenu/MainMenuMultiplayerPanel.cs
@@ -9,14 +9,15 @@ using NitroxModel.Logger;
 using UnityEngine;
 using UnityEngine.Events;
 using UnityEngine.UI;
+using NitroxClient.Persistence;
+using NitroxClient.Persistence.Model;
+using LitJson;
 
 namespace NitroxClient.MonoBehaviours.Gui.MainMenu
 {
     public class MainMenuMultiplayerPanel : MonoBehaviour
     {
-        private const string SERVER_LIST_PATH = ".\\servers";
         private static MainMenuMultiplayerPanel main;
-
 
         private Rect addServerWindowRect = new Rect(Screen.width / 2 - 250, 200, 500, 200);
         private GameObject loadedMultiplayerRef;
@@ -48,10 +49,7 @@ namespace NitroxClient.MonoBehaviours.Gui.MainMenu
             savedGameAreaContent = loadedMultiplayerRef.RequireTransform("Scroll View/Viewport/SavedGameAreaContent");
             deleteButtonRef = savedGamesRef.GetComponent<MainMenuLoadPanel>().saveInstance.GetComponent<MainMenuLoadButton>().deleteButton;
 
-            if (!File.Exists(SERVER_LIST_PATH))
-            {
-                AddServer("local server", "127.0.0.1", "11000", Guid.NewGuid());
-            }
+            PersistedClientData.InitalizeServerList();
 
             CreateButton(Language.main.Get("Nitrox_AddServer"), ShowAddServerWindow);
             LoadSavedServers();
@@ -96,17 +94,12 @@ namespace NitroxClient.MonoBehaviours.Gui.MainMenu
 
         private void AddServer(string name, string ip, string port, Guid token)
         {
-            using (StreamWriter sw = new StreamWriter(SERVER_LIST_PATH, true))
-            {
-                sw.WriteLine($"{name}|{ip}|{port}|{token}");
-            }
+            PersistedClientData.EmplaceServer(name, ip, port, token);
         }
 
         private void RemoveServer(int index)
         {
-            List<string> serverLines = new List<string>(File.ReadAllLines(SERVER_LIST_PATH));
-            serverLines.RemoveAt(index);
-            File.WriteAllLines(SERVER_LIST_PATH, serverLines.ToArray());
+            PersistedClientData.RemoveServerByIndex(index);
         }
 
         public static void OpenJoinServerMenu(string serverIp, string serverPort, Guid token)
@@ -151,34 +144,9 @@ namespace NitroxClient.MonoBehaviours.Gui.MainMenu
 
         private void LoadSavedServers()
         {
-            using (StreamReader sr = new StreamReader(SERVER_LIST_PATH))
+            foreach(SavedServer server in PersistedClientData.GetServers())
             {
-                string line;
-                while ((line = sr.ReadLine()) != null)
-                {
-                    string[] lineData = line.Split('|');
-                    string serverName = lineData[0];
-                    string serverIp = lineData[1];
-                    string serverPort;
-                    Guid token = Guid.NewGuid();
-                    if (lineData.Length == 3)
-                    {
-                        serverPort = lineData[2];
-                        //FIXME: Update the at rest server file with the new token.
-                    }
-                    else if (lineData.Length == 4)
-                    {
-                        serverPort = lineData[2];
-                        token = new Guid(lineData[3]);
-                    }
-                    else
-                    {
-                        Match match = Regex.Match(serverIp, @"^(.*?)(?::(\d{3,5}))?$");
-                        serverIp = match.Groups[1].Value;
-                        serverPort = match.Groups[2].Success ? match.Groups[2].Value : "11000";
-                    }
-                    CreateServerButton($"{Language.main.Get("Nitrox_ConnectTo")} <b>{serverName}</b>\n{serverIp}:{serverPort}", serverIp, serverPort, token);
-                }
+                CreateServerButton($"{Language.main.Get("Nitrox_ConnectTo")} <b>{server.Name}</b>\n{server.Ip}:{server.Port}", server.Ip, server.Port, new Guid(server.Token));
             }
         }
 
--- a/NitroxClient/MonoBehaviours/Gui/MainMenu/MainMenuMultiplayerPanel.cs
+++ b/NitroxClient/MonoBehaviours/Gui/MainMenu/MainMenuMultiplayerPanel.cs
@@ -50,7 +50,7 @@ namespace NitroxClient.MonoBehaviours.Gui.MainMenu
 
             if (!File.Exists(SERVER_LIST_PATH))
             {
-                AddServer("local server", "127.0.0.1", "11000");
+                AddServer("local server", "127.0.0.1", "11000", Guid.NewGuid());
             }
 
             CreateButton(Language.main.Get("Nitrox_AddServer"), ShowAddServerWindow);
@@ -68,7 +68,7 @@ namespace NitroxClient.MonoBehaviours.Gui.MainMenu
             multiplayerButtonButton.onClick.AddListener(clickEvent);
         }
 
-        private void CreateServerButton(string text, string joinIp, string joinPort)
+        private void CreateServerButton(string text, string joinIp, string joinPort, Guid token)
         {
             GameObject multiplayerButtonInst = Instantiate(multiplayerButton, savedGameAreaContent, false);
             multiplayerButtonInst.name = (savedGameAreaContent.childCount - 1).ToString();
@@ -81,7 +81,7 @@ namespace NitroxClient.MonoBehaviours.Gui.MainMenu
             multiplayerButtonButton.onClick.AddListener(() =>
             {
                 txt.GetComponent<Text>().color = prevTextColor; // Visual fix for black text after click (hover state still active)
-                OpenJoinServerMenu(joinIp, joinPort);
+                OpenJoinServerMenu(joinIp, joinPort, token);
             });
 
             GameObject delete = Instantiate(deleteButtonRef, multiplayerButtonInst.transform, false);
@@ -94,11 +94,11 @@ namespace NitroxClient.MonoBehaviours.Gui.MainMenu
             });
         }
 
-        private void AddServer(string name, string ip, string port)
+        private void AddServer(string name, string ip, string port, Guid token)
         {
             using (StreamWriter sw = new StreamWriter(SERVER_LIST_PATH, true))
             {
-                sw.WriteLine($"{name}|{ip}|{port}");
+                sw.WriteLine($"{name}|{ip}|{port}|{token}");
             }
         }
 
@@ -109,7 +109,7 @@ namespace NitroxClient.MonoBehaviours.Gui.MainMenu
             File.WriteAllLines(SERVER_LIST_PATH, serverLines.ToArray());
         }
 
-        public static void OpenJoinServerMenu(string serverIp, string serverPort)
+        public static void OpenJoinServerMenu(string serverIp, string serverPort, Guid token)
         {
             if (main == null)
             {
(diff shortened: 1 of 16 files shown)


Exso12Ecesis:
One of the ideas I had was we use a class similar to PersistedPlayerData but for client stuff. This server flat file nonsense isn't going to scale with whatever we need to store in the client at rest. We should also have functions exposed from this class that will help implementers out, like say for example we give it an IP and port, and it returns us a token or something. We should have a helper function for each setting, or we dynamically just expose something like what nlohmann JSON does, where you can let the implementer specify the key/value in the pair, then we just handle that as a C# object type. Although I'm not sure if there is something like that for C#, considering Microsoft wants you to stay in their ecosystem or whatever.

NitroxServer/GameLogic/PlayerManager.cs
```diff
@@ -72,6 +72,15 @@ public IEnumerable<Player> GetAllPlayers()
                 return new MultiplayerSessionReservation(correlationId, rejectedState);
             }
 
+            bool hasSeenPlayerBefore = player != null;
+//#if RELEASE
+            if (hasSeenPlayerBefore && player.Token != authenticationContext.Token)
```
Sunrunner37, line 77: Let's make this check configurable, defaulting to false to start.

NitroxClient/MonoBehaviours/Gui/MainMenu/MainMenuMultiplayerPanel.cs
```diff
@@ -218,8 +225,9 @@ private void OnAddServerButtonClicked()
             serverNameInput = serverNameInput.Trim();
             serverHostInput = serverHostInput.Trim();
             serverPortInput = serverPortInput.Trim();
-            AddServer(serverNameInput, serverHostInput, serverPortInput);
-            CreateServerButton($"{Language.main.Get("Nitrox_ConnectTo")} <b>{serverNameInput}</b>\n{serverHostInput}:{serverPortInput}", serverHostInput, serverPortInput);
+            Guid token = Guid.NewGuid();
```
Sunrunner37, line 228: Let's generate server side and send back.

NitroxClient/MonoBehaviours/Gui/MainMenu/MainMenuMultiplayerPanel.cs
```diff
@@ -160,17 +160,24 @@ private void LoadSavedServers()
                     string serverName = lineData[0];
```
Sunrunner37, line 160: Let's use the json serializer.

NitroxServer/GameLogic/Players/PersistedPlayerData.cs
```diff
@@ -40,6 +41,9 @@ public class PersistedPlayerData
         [JsonProperty, ProtoMember(10)]
         public bool IsPermaDeath { get; set; }
 
+        [JsonProperty, ProtoMember(11)]
+        public Guid Token { get; set; }
```
Sunrunner37, line 45: is this type appropriately handled by both serializers?

Exso12Ecesis: I did not test the PROTOBUF serializer scenario, only the JSON one.

NitroxClient/MonoBehaviours/Gui/MainMenu/MainMenuMultiplayerPanel.cs
Exso12Ecesis: I agree, this is nonsense. We will be playing patty cake with this stupid flat file forever if we keep this up.

NitroxClient/MonoBehaviours/Gui/MainMenu/MainMenuMultiplayerPanel.cs
Exso12Ecesis: If the attacker wishes to send corrupted data to the server as their tracking protocol, it would only be useful to tracking hackers, and while we have done that (thank you idiot bad guys) by tracing their bad implementation of RSA keys, I do not see the reason to care, since we not trying to solve the problem but move it, security through obscurity I like to call it.

NitroxServer/GameLogic/PlayerManager.cs
Exso12Ecesis: I agree, I much prefer that, I just did that for developmental and testing purposes which is why it was commented out in the commit.

NitroxServer/GameLogic/Players/PersistedPlayerData.cs
Jannify: Pls add this field to the `WorldPersistentTest` to check if it is working with ProtoBuf.

Exso12Ecesis:
I seem to be stuck trying to get the JSON configuration class working. I don't know enough about C# as to why it wouldn't work, all the code seems like it should. I tried using dnSpy to debug the C# running inside the client, but it's not able to find Subnautica.exe.

NitroxClient/Persistence/PersistedClientData.cs
```diff
@@ -0,0 +1,76 @@
...
+            CheckSingleton();
+            return model.SavedServers;
+        }
+
+        public static void InitalizeServerList()
+        {
+            if (!File.Exists(FILE_NAME))
+            {
+                model = new PersistedClientDataModel();
+                model.SavedServers.Add(new SavedServer() { Name = "local server", Ip = "127.0.0.1", Port = "1200", Token = Guid.NewGuid() });
```
Exso12Ecesis, line 61: Change to port 11000.

NitroxClient/MonoBehaviours/Gui/MainMenu/MainMenuMultiplayerPanel.cs
```diff
@@ -151,34 +143,9 @@ private void OnGUI()
 
         private void LoadSavedServers()
         {
-            using (StreamReader sr = new StreamReader(SERVER_LIST_PATH))
+            foreach(SavedServer server in PersistedClientData.GetServers())
             {
-                string line;
-                while ((line = sr.ReadLine()) != null)
```
Exso12Ecesis, line 157: Don't forget to handle the regression scenario later when the JSON is working.

NitroxTest/Serialization/WorldPersistenceTest.cs
```diff
@@ -301,6 +301,7 @@ public void PlayerDataTest()
                     Assert.AreEqual(playerData.Permissions, playerDataAfter.Permissions, "PlayerData.Players.Permissions is not equal");
                     Assert.AreEqual(playerData.NitroxId, playerDataAfter.NitroxId, "PlayerData.Players.NitroxId is not equal");
                     Assert.AreEqual(playerData.IsPermaDeath, playerDataAfter.IsPermaDeath, "PlayerData.Players.IsFurniture is not equal");
+                    Assert.AreEqual(playerData.Token, playerDataAfter.Token, "PlayerData.Players.Token is not equal");
```
Exso12Ecesis, line 304: Could there be more things to test for? Check the other tests for what they are doing an duplicate behaviour appropriately. Maybe also make a test for invalid tokens.

Exso12Ecesis (commented):
Notes.

NitroxClient/Persistence/Model/PersistedClientDataModel.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace NitroxClient.Persistence.Model
+{
+    [Serializable]
```
Exso12Ecesis, line 6: I tried making this a ProtobufContract/JsonObject like on the server, but the client stopped working after I did, so I had to remove it.

Exso12Ecesis:
I have finished conducting a pentest (giggidy) against my code. I found there are two vectors to steal the token. One, is you compromise the client and just straight up steal the token at rest, very easy to do previously mentioned. But if you have Wireshark running in the network and ARP poison the target, you can sniff the traffic and steal the token on handshake. While there are two vectors for attack against this, I still feel it fits with the security through obscurity model, as we're not trying to solve the problem, only move it. Once the JSON persistence class is finished, I recommend we move forward with merging.

NitroxClient/Persistence/Model/PersistedClientDataModel.cs
Jannify: Json (at least the version from the server) can't be used in Client and Model because Oculus is using a custom (and old) version which is interfering.

Jannify: You could use LitJson if Newtonsoft isn't working for your needs.

Exso12Ecesis: I do not like the idea of pulling in yet another JSON library to the project. That sounds reductive to future maintainers as they will have to support 2 JSON implementations (asinine). I instead suggest we get Newtonsoft  working somehow, I just need to figure out why my current solution is not working on the client... I feel it has something to do with the File class API on Windows not working right injected in to the client...

Jannify: LitJson is already used in Subnautica and Nitrox for Asset handling

Exso12Ecesis: 🤔 I will try it and see what I come up with, but I feel that if the (de)SerializeObject functions weren't working, there would just be a blank file on the client's disk, where there is not... I feel the point of failure is at the File class function calls...

Exso12Ecesis: Through some digging on the documentation, I found there isn't similar behaviour to Newtonsoft that just simply takes a C# object and returns a string and vice versa. https://litjson.net/api/LitJson/JsonWriter/ I would have to write a custom implementation of JsonWriter for each and every function call, and expecting the implementer in the future to do that too might be asking too much...

Exso12Ecesis: I confirmed my theory, if you just replace the Desearlize calls with a blank string, the file still does not get created... It's clearly a problem with the client not working with files correctly, I just have 0 idea what's causing it...

NitroxClient/Persistence/PersistedClientData.cs
```diff
@@ -0,0 +1,76 @@
...
+        public static void EmplaceServer(string name, string ip, string port, Guid token)
+        {
+            CheckSingleton();
+            SavedServer savedServer = new SavedServer() { Name = name, Ip = ip, Port = port, Token = token };
+            if(GetToken(savedServer.Ip, savedServer.Port) != Guid.Empty) //Some validation is better than no validation.
+            {
+                throw new Exception($"Failed to add '{ip}':'{port}' because there is already an entry for it.");
+            }
+            model.SavedServers.Add(savedServer);
+            File.WriteAllText(JsonConvert.SerializeObject(model), FILE_NAME);
```
Exso12Ecesis, line 47: The parameters are backwards.

Exso12Ecesis: I have tried 

`File.WriteAllText(".\\client.json", "Test");`

But I still can't get the client to write a dang file to the disk! This is beyond frustrating!

Exso12Ecesis: Fuck my life, yo. Just fuck it. So @Jannify was right, there seems to be a problem with Newtonsoft running inside the client. As soon as you comment out the Newtonsoft code from my singleton class, the client is able to write a file to the disk. Can someone please explain that behavior. That makes no sense to me AT ALL!

Exso12Ecesis: https://youtu.be/yf_17TYAPbI

I seriously need a break.

Exso12Ecesis: https://litjson.net/docs/quickstart/mapping-json-to-objects

I found the same behaviour occurs when you swap out Newtonsoft JSON for LitJson. The client isn't able to write files to the disk anymore. I am giving up for now.

Exso12Ecesis: This this needs to be (kinda) encrypted anyway on release, we should just instead of relying on JSON, maybe use ProtoBuf exclusively, but this will make testing and development an absolute nightmare.

Exso12Ecesis: I tried protbuf, same behavior. I give up.

Exso12Ecesis:
For S&G because I was bored and stuck on this issue, I implemented the Steam API checks to the server with constant data in the code, and while this addresses the issue with the original ticket, I was able to hack the system and steal the token just like the GUID with MITM attacks, so unless we start encrypting the traffic, the core issue will remain unsolved...

Exso12Ecesis:
So I'm at the end of my rope, at this point I would just attach a debugger and see what nonsense it's doing, but I can't. I looked at the projects [wiki](https://github.com/micaww/nitrox-wiki/blob/master/Setting-up-a-development-environment-for-Nitrox.md) but the instructions are deprecated. So, I looked on the dnSpy's readme, and it says to go here: https://github.com/dnSpy/dnSpy/releases/unity for the binaries, but that link is also invalid. So, now I'm trying to figure out how  in the world to download the unity mono.dll project and compile it. The barrier for entry for this seems way too damn high. Why is it like this? Did Unity sue them or something?

Exso12Ecesis:
I have found a temporary solution to debugging. I can surround the offending code with a try/catch and dump the exception to file. I found that the cause of the issue is this: Max allowed object depth reached while trying to export from type System.Guid.

Exso12Ecesis:
I know I just pushed the change for the JSON class that needs to be reviewed first, but I went ahead and did the encryption anyway because I was bored. https://github.com/Exso12Ecesis/Nitrox/tree/JSONEncrypt We should work out a better way of storing the key, preferably one that is generated on the client side then saved somewhere, that will make it less likely for someone to weaponize the code, since they will also have to incorporate loading the key in to their attack platform.

Exso12Ecesis:
I have implemented my idea in the [JSONEncrypt](https://github.com/Exso12Ecesis/Nitrox/tree/JSONEncrypt) branch. I think it is a sufficient system that will defeat most attackers. Even if they know the system is running AES-128 and they also know we're using a weak PRNG algorithm to generate the initialization vector and key, they would still need to compromise the user token somehow, which is probably above the skill level of most of the attackers this project would be dealing with. I think my branch is sufficient to solve the ticket's core issue.

Exso12Ecesis:
@Sunrunner37 @Jannify @Measurity Is there something on my end that's holding this up?

NitroxTest/Serialization/WorldPersistenceTest.cs
Jannify, line 304: In addition to that you need generate a new token in `GeneratePersistedWorldData()`

Jannify (commented):
Code wise it seems fine to me (except `WorldPersistenceTest.cs`). 
In a few places `Guid token` looks a bit to generic (maybe something like "authToken" would be better). 
I will do additional testing if I have the time (and a volunteer).

NitroxTest/Serialization/WorldPersistenceTest.cs
Exso12Ecesis: I have added the requested changes in a subsequent commit. Please confirm.

Jannify (commented):
Okay I tested the pr with @Coding-Hen and we found a few problems with the discord integration.
You can't connect (with the same user name) to a server you have previously connect to though the main menu. We tried it with a domain as IP in main menu, if you have saved a real IP you would probably get a (fatal) error.

You could prevent that (IP main-menu <-> IP DiscordRP) with checking for servers with the same IP in DiscordRPController before creating a new one. To also support domains they would needed to be resolved before that.

Measurity:
Closing PR due to inactivity

PR #1417, opened 2021-02-08
Update LauncherLogic.cs

ethelin:


Coding-Hen (commented):
What is the reason for this change?

Dessix (changes requested):
Without an explanation, this isn't a pull request.

PR #1433, opened 2021-03-19
Enable loading multiplayer with QMod installed

Coding-Hen:
Stop qmod initialisation on load and restore on exit

### DIFF
--- a/NitroxLauncher/Patching/QModHelper.cs
+++ b/NitroxLauncher/Patching/QModHelper.cs
@@ -1,4 +1,5 @@
-﻿using System.IO;
+﻿using System;
+using System.IO;
 using NitroxModel.Logger;
 
 namespace NitroxLauncher.Patching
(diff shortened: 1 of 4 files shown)


Jannify (commented):
The solution seems fine to me. Would add info logging on qmod restore/remove and error logging if the file is not found.

Coding-Hen:
> 
> 
> The solution seems fine to me. Would add info logging on qmod restore/remove and error logging if the file is not found.

Yes good idea will add that in. Will also remove the existing checks for qmod being installed so that our code is cleaned up. If the patching fails for whatever reason and they are still able to launch into the game qmods will still catch Nitrox being installed and throw an error.

Jannify (approved):

NitroxLauncher/Patching/QModHelper.cs
```diff
@@ -0,0 +1,52 @@
+﻿using System.IO;
+using NitroxModel.Logger;
+
+namespace NitroxLauncher.Patching
+{
+    public static class QModHelper
+    {
+        private static readonly string originalFileName = "winhttp.dll";
```
dartasen, line 8: minor: const

NitroxLauncher/Patching/QModHelper.cs
```diff
@@ -0,0 +1,52 @@
...
+            string fileToRenamePath = Path.Combine(subnauticaBasePath, fileToRename);
+            if (!File.Exists(fileToRenamePath))
+            {
+                Log.Error($"QMod entry cannot be found, please uninstall QMod");
+                return;
+            }
+
+            QModPatched = !QModPatched;
+            string newFilePath = Path.Combine(subnauticaBasePath, newFileName);
+            File.Move(fileToRenamePath, newFilePath);
```
dartasen, line 42: Might be good to handle exceptions, especially with IO actions.

NitroxLauncher/Patching/QModHelper.cs
Coding-Hen: Const cant be used here as static and readonly cant be used in conjunction with const

Measurity: A const is by default a static so you can remove both `static` and `readonly` to make this work.

Coding-Hen: Good to know, will make the change

PR #1436, opened 2021-03-27
Refactor PlatformDetection inside Launcher

dartasen:
* Moved the detection to NitroxModel
* Introduced the "Platform" Enum in order to figure out which platform we are dealing with
* Cleaned up LauncherLogic.cs  a bit so it can be more readable

### DIFF
--- a/NitroxLauncher/LauncherLogic.cs
+++ b/NitroxLauncher/LauncherLogic.cs
@@ -13,6 +13,7 @@ using NitroxLauncher.Events;
 using NitroxLauncher.Pages;
 using NitroxLauncher.Patching;
 using NitroxModel;
+using NitroxModel.Discovery;
 using NitroxModel.Helper;
 using NitroxModel.Logger;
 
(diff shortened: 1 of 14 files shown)


NitroxModel/NitroxModel.csproj
```diff
@@ -237,10 +239,7 @@
       <Name>LiteNetLib</Name>
     </ProjectReference>
   </ItemGroup>
-  <ItemGroup>
-    <Folder Include="OS\MacOS\" />
-    <Folder Include="OS\Unix\" />
-  </ItemGroup>
+  <ItemGroup />
```
Jannify, line 242: Probably unintended

Jannify (approved):
Besides `NitroxModel.csproj` it seems fine to me

NitroxModel/NitroxModel.csproj
dartasen: Visual doing some weird things as always 😪

Measurity:
Personally not a fan of using #region because I don't see much use in them and they add extra clutter.

NitroxLauncher/LauncherLogic.cs
```diff
@@ -340,31 +331,39 @@ private async Task<Process> WaitForProcessAsync()
             }
 
             return await Task.Run(async () =>
+            {
+                for (int i = 0; i < 1000; i++)
                 {
-                    int waitTimeInMill = 1000;
-                    for (int i = 0; i < 1000; i++)
+                    Process[] processes = Process.GetProcessesByName("Subnautica");
```
Measurity, line 337: We need to dispose all `Process` objects otherwise we leak memory here (was already an issue before changes you made)

Measurity (approved):

Coding-Hen (approved):

NitroxLauncher/LauncherLogic.cs
dartasen: Should be good now

PR #1441, opened 2021-03-30
Server PostSaveCommand

grandoth:
Add a config property for a Server PostSaveCommand
The property can be empty or be set to a path of an executable, bat, PowerShell script, etc
Following a save completing, the command will attempt to be called (non-blocking)
This could allow a server admin to perform automatic backups of the world

Also cleanup Server.cs usings to be alphabetical


NitroxServer/Server.cs
```diff
@@ -69,7 +70,25 @@ public void Save()
...
+
+                    // Call external tool for backups, etc
+                    if (!string.IsNullOrEmpty(postSaveCommandPathTrimmed) ||
+                        File.Exists(postSaveCommandPathTrimmed))
+                    {
+                        Process.Start(postSaveCommandPathTrimmed);
+                        Log.Info($"Post-save command completed successfully: {postSaveCommandPathTrimmed}");
+                    }
+                }
+                catch (Exception Ex)
```
Measurity, line 87:
```suggestion
                catch (Exception ex)
```
Lowercase var names.

NitroxServer/Server.cs
```diff
@@ -69,7 +70,25 @@ public void Save()
...
             NitroxConfig.Serialize(serverConfig); // This is overwriting the config file => server has to be closed before making changes to it
-            worldPersistence.Save(world, serverConfig.SaveName);
+            if (worldPersistence.Save(world, serverConfig.SaveName))
+            {
+                try
+                {
+                    string postSaveCommandPathTrimmed = serverConfig.PostSaveCommandPath.Trim('"');
+
+                    // Call external tool for backups, etc
+                    if (!string.IsNullOrEmpty(postSaveCommandPathTrimmed) ||
```
Measurity, line 80: Don't need to check for empty, File.Exists will do it.

NitroxServer/Serialization/ServerConfig.cs
```diff
@@ -43,6 +43,9 @@ public int SaveInterval
             }
         }
 
+        [PropertyDescription("Command to run following a successful word save (e.g. world backup .exe, .bat, or PowerShell script)")]
+        public string PostSaveCommandPath { get; set; } = string.Empty;
```
Measurity, line 47: Could override this setter to trim the given value so that other parts of code don't need to do it.

NitroxServer/Server.cs
```diff
@@ -69,7 +70,25 @@ public void Save()
...
+                        File.Exists(postSaveCommandPathTrimmed))
+                    {
+                        Process.Start(postSaveCommandPathTrimmed);
```
Measurity, line 83: Recommend using the `FileSystem.OpenOrExecuteFile` function instead of process.start so that files are opened with default editor. Make sure to dispose of the Process object, as to not leak memory.

Jannify (commented):
Seems fine to me. 
Maybe add the command into the error log and use "IsNullOrWhiteSpace" instead of "IsNullOrEmpty.

Measurity (changes requested):

grandoth:
Thanks for the comments. My intent is to address them in a few days. An exceptionally busy weekend. Cheers.

NitroxServer/Server.cs
grandoth: FileSystem.OpenOrExecuteFile doesn't work with .bat on my system. I did dispose of the process now. Good catch.

grandoth:
Changes made. Think it's ready for final review.

Measurity (approved):

NitroxServer/Serialization/ServerConfig.cs
```diff
@@ -43,6 +45,13 @@ public int SaveInterval
             }
         }
 
+        [PropertyDescription("Command to run following a successful word save (e.g. .exe, .bat, or PowerShell script). ")]
```
Coding-Hen, line 48: Minor: typo
```suggestion
        [PropertyDescription("Command to run following a successful world save (e.g. .exe, .bat, or PowerShell script). ")]
```

Coding-Hen (changes requested):

NitroxServer/Server.cs
```diff
@@ -72,7 +74,27 @@ public void Save()
             }
 
             IsSaving = true;
-            worldPersistence.Save(world, serverConfig.SaveName);
+
+            if (worldPersistence.Save(world, serverConfig.SaveName) && !string.IsNullOrWhiteSpace(serverConfig.PostSaveCommandPath))
```
Jannify, line 78:
```suggestion
            bool savedSuccessfully = worldPersistence.Save(world, serverConfig.SaveName);
            if (savedSuccessfully && !string.IsNullOrWhiteSpace(serverConfig.PostSaveCommandPath))
```
Seeing [sunrunners opinion](https://github.com/SubnauticaNitrox/Nitrox/pull/1519#discussion_r667355638) on logic inside ifs, this is probably fitting better as calling `worldPersistence.Save()` has not really something to do with checking if the post save command can be executed.

PR #1477, opened 2021-05-15
Update German.json

Frostie314159:
Just a bit of spell checking.


Nitrox.Subnautica.Assets/LanguageFiles/German.json
```diff
@@ -25,7 +25,7 @@
   "Nitrox_LostConnection": "Die Verbindung zum Server wurde unterbrochen",
   "Nitrox_PlayerKicked": "Du bist vom Server geflogen",
   // In game logging
-  "Nitrox_UnableToConnect": "Es ist unmöglich folgenen Server zu erreichen:",
+  "Nitrox_UnableToConnect": "Der Server kann nicht erreicht werden.",
```
Jannify, line 28: This change is sounding better but you have to keep in mind that after it the severs IP address is added. So at least the "." should be a ":" like before.

PR #1478, opened 2021-05-16
Fix spelling pda scanner entry progress

Coding-Hen:
Looking through logs and found this small spelling error so cleaned it up. Also added a little more information to the log so its a little more meaningful.

### DIFF
--- a/NitroxClient/Communication/Packets/Processors/PDAScannerEntryProgressProcessor.cs
+++ b/NitroxClient/Communication/Packets/Processors/PDAScannerEntryProgressProcessor.cs
@@ -5,7 +5,6 @@ using NitroxClient.Communication.Packets.Processors.Abstract;
 using NitroxModel.Logger;
 using NitroxModel.Packets;
 using NitroxModel_Subnautica.DataStructures;
-using NitroxModel_Subnautica.Helper;
 
 namespace NitroxClient.Communication.Packets.Processors
 {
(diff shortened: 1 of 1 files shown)


Jannify (approved):

NitroxClient/Communication/Packets/Processors/PDAScannerEntryProgressProcessor.cs
```diff
@@ -24,22 +23,21 @@ public override void Process(PDAEntryProgress packet)
...
-                        Log.Info("PDAEntryProgress Upldate Old:" + entry.unlocked + " New" + packet.Unlocked);
+                        Log.Info($"PDAEntryProgress Update For TechType:{techType} Old:{entry.unlocked} New:{packet.Unlocked}");
                         entry.unlocked = packet.Unlocked;
                     }
                 }
                 else
                 {
-                    Log.Info("PDAEntryProgress New TechType:" + techType + " Unlocked:" + packet.Unlocked);
+                    Log.Info($"PDAEntryProgress New TechType:{techType} Unlocked:{packet.Unlocked}");
                     MethodInfo methodAdd = typeof(PDAScanner).GetMethod("Add", BindingFlags.NonPublic | BindingFlags.Static, null, new Type[] { typeof(TechType), typeof(int) }, null);
```
dartasen, line 39: We might wanna use the Reflection helper class we have

NitroxClient/Communication/Packets/Processors/PDAScannerEntryProgressProcessor.cs
```diff
@@ -24,22 +23,21 @@ public override void Process(PDAEntryProgress packet)
             using (packetSender.Suppress<PDAEntryProgress>())
             {
                 TechType techType = packet.TechType.ToUnity();
-                PDAScanner.EntryData entryData = PDAScanner.GetEntryData(techType);
 
                 PDAScanner.Entry entry;
                 if (PDAScanner.GetPartialEntryByKey(techType, out entry))
```
dartasen, line 28:
```suggestion
                if (PDAScanner.GetPartialEntryByKey(techType, out PDAScanner.Entry entry))
```

NitroxClient/Communication/Packets/Processors/PDAScannerEntryProgressProcessor.cs
Coding-Hen: Doesn't look like we are able to do that as c# doesnt allow static types (which PDAScanner is) to be used as type arguments which is what our helper uses.

dartasen (approved):

NitroxClient/Communication/Packets/Processors/PDAScannerEntryProgressProcessor.cs
Jannify: But you could put the MethodInfo outside the method as a readonly field. Would improve performance a bit.

PR #1486, opened 2021-05-24
Improvements to commands workflow

dartasen:
* Rephrase sentences from the server to make it clearer
* Fixed a regression causing the need to restart on every config change (even through commands)
* Some refactor and improvements
* Introduced command flags `NO_CONSOLE` and `DEBUG_ONLY` to add more flexibility on command behaviours
* Introduced a new command `/promote {player} {rank}`

Current ranks : 
- `NONE ` : Can't use any command  
- `PLAYER` : Basic command (help, list, summary, whois, msg)
- `MODERATOR` : Basic moderating (teleport, warp, kick, promote, ..)
- `ADMIN` : Almost everything
- `CONSOLE` : Everything

* Added a server config to set the default perm to a new player


NitroxServer-Subnautica/Program.cs
```diff
@@ -73,7 +76,10 @@ private static async Task Main(string[] args)
                 {
                     throw new Exception("Unable to start server.");
                 }
-                Log.Info("Server is waiting for players!");
+                watch.Stop();
+
+                Log.Info($"Server started ({Math.Round(watch.Elapsed.TotalSeconds, 1)}s)");
```
Coding-Hen, line 81: I agree with some text changes here as to actually saying the server has started but I'm not sure of the value of having a timer and telling them how long it took, will this be useful?

NitroxServer/Communication/NetworkingLayer/NitroxServer.cs
```diff
@@ -64,7 +64,7 @@ protected void ProcessIncomingData(NitroxConnection connection, Packet packet)
             }
             catch (Exception ex)
             {
-                Log.Error("Exception while processing packet: " + packet + " " + ex);
+                Log.Error(ex, $"Exception while processing packet: {packet}");
```
Coding-Hen, line 67: Do we not want to keep this for debugging if we get logs sent to us?

NitroxServer/Server.cs
```diff
@@ -137,7 +142,7 @@ public void PauseServer()
             DisablePeriodicSaving();
             world.EventTriggerer.PauseWorldTime();
             world.EventTriggerer.PauseEventTimers();
-            Log.Info("Server has paused");
+            Log.Info("Server has paused, waiting for players");
```
Coding-Hen, line 145: Might be worth slightly editing this just o make it even more clear
```suggestion
            Log.Info("Server has paused, waiting for players to connect");
```

Coding-Hen (commented):
Looks much cleaner without having to load the config file each time

NitroxServer/ConsoleCommands/DirectoryCommand.cs
```diff
@@ -8,24 +9,25 @@
...
+
             if (!Directory.Exists(dir))
             {
-                Log.Error($"Unable to open Nitrox directory '{dir}' because it does not exist.");
-                return;
+                throw new ArgumentException($"Unable to open Nitrox directory '{dir}' because it does not exist.");
             }
 
             Process.Start(dir);
+            Log.Info($"Opening directory '{dir}'");
```
Jannify, line 30: Maybe we should add a `dir` parameter to `Log.sensitiveLogParameters`. I thinking about some users who have their full name as their pc account and depending on the path it could be saved to the log file.

NitroxServer/Serialization/World/WorldPersistence.cs
```diff
@@ -28,16 +28,23 @@ public class WorldPersistence
...
             this.config = config;
 
-            Serializer = config.SerializerMode == ServerSerializerMode.PROTOBUF ? protoBufSerializer : jsonSerializer;
+            UpdateSerializer(config.SerializerMode);
+        }
+
+        internal void UpdateSerializer(ServerSerializerMode mode)
+        {
+            Serializer = (mode == ServerSerializerMode.PROTOBUF) ? protoBufSerializer : jsonSerializer;
         }
```
Jannify, line 48: This should error out `WorldPersistenceTest.Setup()`. Are your sure all tests are still working?

Jannify (changes requested):

NitroxServer/Communication/NetworkingLayer/NitroxServer.cs
Coding-Hen: Ignore my comment missed the exception still being logged, thought we had removed the logging for it.

NitroxServer/ConsoleCommands/DirectoryCommand.cs
Measurity: Yes, would recommend rename to `path` and then check if the PC user name is in the path and redact it away.

NitroxServer-Subnautica/Program.cs
dartasen: Imo, time metrics are important when you're dealing with servers, always seen those kind of informations when launching game servers, but I can remove it, if no one else find it useful

NitroxServer/ConsoleCommands/DebugStartMapCommand.cs
```diff
@@ -11,11 +9,11 @@ namespace NitroxServer.ConsoleCommands
 {
     internal class DebugStartMapCommand : Command
     {
-        private readonly PlayerManager playerManager;
         private readonly RandomStartGenerator nitroxRandomStart;
+        private readonly PlayerManager playerManager;
         private readonly World world;
 
-        public DebugStartMapCommand(PlayerManager playerManager, RandomStartGenerator nitroxRandomStart, World world) : base("debugstartmap", Perms.CONSOLE, "warning: spawns blocks")
+        public DebugStartMapCommand(PlayerManager playerManager, RandomStartGenerator nitroxRandomStart, World world) : base("debugstartmap", Perms.CONSOLE | Perms.DEBUG, "Spawns blocks at spawn positions")
```
Jannify, line 16: You added `Perms.Any`, isn't this usable here?

NitroxServer/Server.cs
Jannify, line 145: Re Hen: Might be worth slightly editing this just to make it even more clear
```suggestion
            Log.Info("Server has paused, waiting for players to connect");
```

Jannify (changes requested):

NitroxServer/ConsoleCommands/OpCommand.cs
```diff
@@ -6,18 +6,17 @@ namespace NitroxServer.ConsoleCommands
 {
     internal class OpCommand : Command
     {
-        public OpCommand() : base("op", Perms.ADMIN, "Sets an user as admin")
+        public OpCommand() : base("op", Perms.ADMIN, "Sets a user as admin")
```
Jannify, line 9:
```suggestion
        public OpCommand() : base("op", Perms.ADMIN, "Sets an user as admin")
```

Jannify (approved):
Code wise it seems fine to me.

NitroxServer/ConsoleCommands/OpCommand.cs
Measurity: `a` is correct here https://english.stackexchange.com/questions/105116/is-it-a-user-or-an-user

Jannify: Ok sry :D

dartasen: I discovered that weird rule few days ago too

NitroxModel/DataStructures/GameLogic/Perms.cs
```diff
@@ -1,10 +1,21 @@
...
         CONSOLE
     }
+
+    [Flags]
+    public enum PermsFlag : sbyte
+    {
+        NONE = 0x0,
+        DEBUG_ONLY = 0x1,
+        NO_CONSOLE = 0x2
+    }
```
Measurity, line 20: Why use `sbyte` here? Do we need negative values?

NitroxServer/ConsoleCommands/Abstract/Command.cs
```diff
@@ -65,9 +73,18 @@ public void TryExecute(Optional<Player> sender, string[] args)
...
         }
 
+        public bool CanExecute(Perms treshold)
+        {
+#if RELEASE
+            return RequiredPermLevel <= treshold && !Flags.HasFlag(PermsFlag.DEBUG_ONLY);
+#else
+            return RequiredPermLevel <= treshold;
+#endif
+        }
```
Measurity, line 83: Omit preprocessor directives here, omit "debug" commands instead if in release mode.

```suggestion
        public bool CanExecute(Perms threshold)
        {
            return RequiredPermLevel <= threshold;
        }
```

NitroxServer/ConsoleCommands/AutosaveCommand.cs
```diff
@@ -9,15 +8,19 @@ namespace NitroxServer.ConsoleCommands
...
             AddParameter(new TypeBoolean("on/off", true));
+
+            this.serverConfig = serverConfig;
         }
 
         protected override void Execute(CallArgs args)
         {
             bool toggle = args.Get<bool>(0);
-            ServerConfig serverConfig = NitroxConfig.Deserialize<ServerConfig>();
+
```
Measurity, line 23: Removing this will cause existing changes in config file to be overwritten. I think this is suprising behaviour for users so I recommend adding it back.

NitroxServer/ConsoleCommands/ChangeServerGamemodeCommand.cs
```diff
@@ -10,24 +11,33 @@ namespace NitroxServer.ConsoleCommands
...
         protected override void Execute(CallArgs args)
         {
             ServerGameMode sgm = args.Get<ServerGameMode>(0);
 
-            serverConfig.GameMode = sgm;
-            playerManager.SendPacketToAllPlayers(new GameModeChanged(sgm));
+            if (serverConfig.GameMode != sgm)
+            {
+                serverConfig.GameMode = sgm;
+                NitroxConfig.Serialize(serverConfig);
```
Measurity, line 32: This causes existing changes to config file to be overwritten by whatever the settings were in memory. It should deserialize, make the changes, and then serialize again (maybe turn this into a method).

NitroxServer/ConsoleCommands/ChangeServerPasswordCommand.cs
```diff
@@ -10,21 +9,24 @@ namespace NitroxServer.ConsoleCommands
...
             AddParameter(new TypeString("password", false));
+
+            this.serverConfig = serverConfig;
         }
 
         protected override void Execute(CallArgs args)
         {
             string password = args.Get(0) ?? string.Empty;
 
-            ServerConfig serverConfig = NitroxConfig.Deserialize<ServerConfig>();
```
Measurity, line 22: Same issue here

NitroxServer/ConsoleCommands/KickCommand.cs
```diff
@@ -14,31 +14,46 @@ internal class KickCommand : Command
...
+                    playerManager.SendPacketToAllPlayers(ownershipChange);
+                }
+
+                playerManager.SendPacketToOtherPlayers(new Disconnect(playerToKick.Id), playerToKick);
+                SendMessage(args.Sender, $"The player {playerToKick.Name} has been disconnected");
+            }
+            else
+            {
+                SendMessage(args.Sender, $"You're not allowed to kick {playerToKick.Name}");
+            }
```
Measurity, line 56: Minor: return early (flip the `if` for "not allowed to kick") to reduce indentation of the code here.

NitroxServer/ConsoleCommands/SwapSerializerCommand.cs
```diff
@@ -2,32 +2,41 @@
...
+
             this.worldPersistence = worldPersistence;
-            this.protoBufSerializer = protoBufSerializer;
-            this.jsonSerializer = jsonSerializer;
+            this.serverConfig = serverConfig;
         }
 
         protected override void Execute(CallArgs args)
         {
-            ServerConfig serverConfig = NitroxConfig.Deserialize<ServerConfig>();
```
Measurity, line 25: Same issue here with not deserializing (anymore).

NitroxServer/ConsoleCommands/TimeCommand.cs
```diff
@@ -9,15 +9,16 @@ internal class TimeCommand : Command
...
-            this.timeKeeper = timeKeeper;
             AddParameter(new TypeString("day/night", false));
+
+            this.timeKeeper = timeKeeper;
         }
 
         protected override void Execute(CallArgs args)
         {
-            string time = args.Get(0);
+            string time = args.Get<string>(0);
```
Measurity, line 21: Not sure why this change is done. If we want to require specifying the type we should remove the method `public string Get()` in CallCargs.cs that makes this possible.

Measurity (changes requested):

Measurity:
I'll make a new PR to fix the serialization after this one is merged in.

Measurity (approved):

NitroxServer/ConsoleCommands/PromoteCommand.cs
```diff
@@ -0,0 +1,39 @@
+﻿using NitroxModel.DataStructures.GameLogic;
+using NitroxServer.ConsoleCommands.Abstract;
+using NitroxServer.ConsoleCommands.Abstract.Type;
+
+namespace NitroxServer.ConsoleCommands
+{
+    internal class PromoteCommand : Command
+    {
+        public PromoteCommand() : base("promote", Perms.MODERATOR, "Sets specific permissions to a user")
```
Coding-Hen, line 9: I feel like promote is the wrong name here as it could be a demotion if you set their permissions lower. Would giverole or something be more descriptive?

NitroxServer/ConsoleCommands/StopCommand.cs
```diff
@@ -4,17 +4,18 @@
 
 namespace NitroxServer.ConsoleCommands
 {
-    internal class ExitCommand : Command
+    internal class StopCommand : Command
     {
         public override IEnumerable<string> Aliases { get; } = new[] { "exit", "halt", "quit" };
 
-        public ExitCommand() : base("stop", Perms.ADMIN, "Stops the server")
+        public StopCommand() : base("stop", Perms.ADMIN, "Stops the server")
```
Coding-Hen, line 11: Again maybe a rename to shutdown or something as stop isn't too descriptive of what it does?

Coding-Hen (commented):
Just my personal option but looks good overall

PR #1496, opened 2021-06-01
Create Turkish.json

ltibm90:
Here is Turkish translation.


Jannify:
We need to verify the translation so please be patient until we find a native speaker in the team (or friends).

Issue #1359, opened 2021-01-04
Glitches with the Prawnsuit, Seamoth, and Cyclops 


ItzKyros:
### Version

 1.4.0.0
### _Seamoth:_

**1. Random Explosions**

- At random times the Seamoth will explode for no reason.
- This can happen when someone is in the vehicle or while they aren't present.

**2. Turning Invisible**

- This seems to also happen at random times however it only occurs for one player.
- There are some instances where someone enters the Seamoth and both the Seamoth and the player become invisible to other players.
- After this glitch occurs, my friend and I leave the game and join back which normally fixes the issue. However, when joining back, one of us or both of us spawn at a random location on the map. This can range from outside the map to 100 meters away from where the player left.
- Once the player leaves the vehicle, both the Seamoth and player become visible.

**3. Modules only appear and work for one player**

- Putting an upgrade module in the Seamoth will only work for the player who did it. Anyone else who tries to use the Seamoth won't see or be able to use the upgrades.

**4. Entering the vehicle facing sideways**

- Sometimes when you enter the Seamoth, you face a random direction at an awkward angle. 
- In order to fix it you have to leave the vehicle and re-enter.

**5. Seamoth gets stuck in place**

- Sometimes when entering a Seamoth and trying to drive forward the vehicle gets stuck in place. This only occurs for one player.
- In order to fix this glitch someone else has to get in and drive the Seamoth and then get back out. Then the other player can move the Seamoth again.

### _Prawnsuit:_

**1. Upgrades only appear and work for 1 player at a time.**

- This is the same bug listed in number 3 for the Seamoth.

**2. Does not charge**

 - When loaded into Cyclops, the Prawnsuit will continue to take power from the Cyclops trying to charge itself. Power is drained from the Cyclops but the Prawnsuit never receives the power.
- Essentially, the Prawnsuit constantly draws power from the Cyclops to charge itself but never actually charges. This results in the Cyclops losing its power for no reason at a rate of about 1% per second.

**3. Frozen player in the moonpool**

- Players may randomly freeze when they enter the Prawnsuit in the moonpool
- When a player enters the Prawnsuit while it's parked in the moonpool, sometimes they appear frozen on another player's screen. For them, it appears that they are moving around in the Prawnsuit doing their own business. However, for another player, the person in the Prawnsuit appears frozen hanging inside the moonpool. 
- When this happens, the frozen player seems to be greyed/blacked out,
[...]

Issue #1293, opened 2020-11-23
[Cyclops] Battery drain is simulated by each player


Ghaarg:
### Version
Current master branch (24.11.2020)

### Bug
It seems that with more than one player, a cyclops battery drain is higher than usual. This can lead to high drain overall and more packages send.
Cause: All players will simulate the battery drain and send the data to the server, instead of one player sending the information.

### Expected
Less drain, fewer packages send

### Steps to reproduce
Build cyclops and let more than one player join.

Shirkie01:
This may be fixed with #1659, needs verification.

PR #1499, opened 2021-06-13
Fixed non fully charged vehicle in cyclops bay would drain all energy without getting charged

S0nix12:
This is a fix for the bug that a Prawn Suit that is not fully charged is docked into the cyclops it would drain all power from the cyclops trying to charge itself but never getting any power.
The bug is also mentioned in this issue: #1359 

It did happen because the patched ModifyCharge method checked if the modification did a change to a different natural number before sending a sync packet. The problem was that this check did not compare the old charge with the new charge value but the new charge value with a value applying the modification again.
The Prawn Suit was draining 1 energy per second (normal power cells) and distribute it to it's 2 power cells. Each got 0.5 energy. The patched method now checked newCharge against newCharge + 0.5 and detected a natural number change although no change occurred. After that the current charge was floored removing all added power again and sending this to the other clients. As all clients simulate the recharging inside the cyclops they would override each other with the wrong floored values never being able to add any energy to the PrawnSuit but still consuming from the Cyclops. 

This change does not solve the problem that energy consumption inside the cyclops is simulated by all player also mentioned in this issue #1293 
This is also noticeable as recharging a vehicle inside the cyclops takes more energy from the cyclops than what is added to the vehicle but with this change the vehicle will still properly charge and the cyclops won't leak all energy.

### DIFF
--- a/NitroxPatcher/Patches/Dynamic/EnergyMixin_ModifyCharge_Patch.cs
+++ b/NitroxPatcher/Patches/Dynamic/EnergyMixin_ModifyCharge_Patch.cs
@@ -16,7 +16,7 @@ namespace NitroxPatcher.Patches.Dynamic
         public static readonly Type TARGET_CLASS = typeof(EnergyMixin);
         public static readonly MethodInfo TARGET_METHOD = TARGET_CLASS.GetMethod("ModifyCharge", BindingFlags.Public | BindingFlags.Instance);
 
-        public static void Postfix(float __result, EnergyMixin __instance)
+        public static void Postfix(EnergyMixin __instance, float __result)
         {
             GameObject battery = __instance.GetBattery();
             if (battery)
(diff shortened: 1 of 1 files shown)


NitroxPatcher/Patches/Dynamic/EnergyMixin_ModifyCharge_Patch.cs
```diff
@@ -16,12 +16,12 @@ public class EnergyMixin_ModifyCharge_Patch : NitroxPatch, IDynamicPatch
         public static readonly Type TARGET_CLASS = typeof(EnergyMixin);
         public static readonly MethodInfo TARGET_METHOD = TARGET_CLASS.GetMethod("ModifyCharge", BindingFlags.Public | BindingFlags.Instance);
 
-        public static void Postfix(EnergyMixin __instance, float amount)
+        public static void Postfix(float __result, EnergyMixin __instance)
```
Coding-Hen, line 19:
```suggestion
        public static void Postfix(EnergyMixin __instance, float __result)
```
Having these the other way round to be more inline with other patches

Coding-Hen (changes requested):
Need to test code is working as described. Just a small note against parameter ordering.

Good find on the cause of this issue

Jannify (approved):
Verified the bug (on master) and the fix in game.

PR #1508, opened 2021-06-30
Refactored ServerConfig to serialize and deserialize the same instance

Measurity:
- [x] Verify all cases where config serialization happens are handled.

### DIFF
--- a/NitroxServer-Subnautica/Program.cs
+++ b/NitroxServer-Subnautica/Program.cs
@@ -33,7 +33,7 @@ namespace NitroxServer_Subnautica
         private static async Task Main(string[] args)
         {
             // The thread that writers to console is paused while selecting text in console. So console writer needs to be async.
-            Log.Setup(true, isConsoleApp: true);
+            Log.Setup(asyncConsoleWriter: true, isConsoleApp: true);
             AppDomain.CurrentDomain.UnhandledException += CurrentDomainOnUnhandledException;
 
             ConfigureCultureInfo();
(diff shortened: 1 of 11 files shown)


NitroxServer-Subnautica/Program.cs
```diff
@@ -262,7 +269,6 @@ private static void StopAndExitServer()
         {
             Log.Info("Exiting ...");
             Server.Instance.Stop();
-            Environment.Exit(0);
```
Jannify, line 265: Wouldn't this cause the cmd window (if you choose this option) to stay and not automatically close?

NitroxServer-Subnautica/Program.cs
```diff
@@ -32,7 +33,7 @@ public class Program
         private static async Task Main(string[] args)
         {
             // The thread that writers to console is paused while selecting text in console. So console writer needs to be async.
-            Log.Setup(asyncConsoleWriter: true, isConsoleApp: true);
+            Log.Setup(true, isConsoleApp: true);
```
Jannify, line 36: Why removing just one side? All or nothing :P

Measurity: It formatted this automatically.

NitroxServer-Subnautica/Program.cs
Measurity: It still automatically closes when clicking exit button on console window or CTRL+C. Though naming could be changed, will do.

PR #1528, opened 2021-07-17
Fix for fabricator looping bug

dartasen:
* Some basic refactor to lighten the code
* Time will now be resync for everyone on player rejoin (We might need to implement a cron resync system for those kind of data from the server)
* Refactored TimeKeeper to match hardcoded value from subnautica (also switched from float to double, since we can deal with big numbers, and subnautica is internaly using double instead of float)
* Fix for the fabricator/workbench looping bug
* Fix for single use mobile vehicle bay bug
* Fix for pda journal starting at -5 causing multiple bugs

### DIFF
--- a/NitroxClient/Communication/Packets/Processors/GhostCrafterItemPickupProcessor.cs
+++ b/NitroxClient/Communication/Packets/Processors/GhostCrafterItemPickupProcessor.cs
@@ -1,5 +1,4 @@
-﻿using NitroxClient.Communication.Abstract;
-using NitroxClient.Communication.Packets.Processors.Abstract;
+﻿using NitroxClient.Communication.Packets.Processors.Abstract;
 using NitroxClient.MonoBehaviours;
 using NitroxClient.Unity.Helper;
 using NitroxModel.Packets;
@@ -9,27 +8,12 @@ namespace NitroxClient.Communication.Packets.Processors
 {
     public class GhostCrafterItemPickupProcessor : ClientPacketProcessor<GhostCrafterItemPickup>
     {
-        private readonly IPacketSender packetSender;
-
-        public GhostCrafterItemPickupProcessor(IPacketSender packetSender)
-        {
-            this.packetSender = packetSender;
-        }
-
         public override void Process(GhostCrafterItemPickup packet)
         {
             GameObject gameObject = NitroxEntity.RequireObjectFrom(packet.GhostCrafterId);
             CrafterLogic crafterLogic = gameObject.RequireComponentInChildren<CrafterLogic>(true);
 
-            if (crafterLogic.numCrafted > 0)
-            {
-                crafterLogic.numCrafted--;
-
-                if (crafterLogic.numCrafted == 0)
-                {
-                    crafterLogic.Reset();
-                }
-            }
+            crafterLogic.Reset();
         }
     }
 }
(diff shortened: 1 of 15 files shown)


NitroxClient/Communication/Packets/Processors/GhostCrafterItemPickupProcessor.cs
```diff
@@ -9,26 +8,14 @@ namespace NitroxClient.Communication.Packets.Processors
...
 
             if (crafterLogic.numCrafted > 0)
             {
-                crafterLogic.numCrafted--;
-
-                if (crafterLogic.numCrafted == 0)
-                {
-                    crafterLogic.Reset();
-                }
+                crafterLogic.Reset();
```
Jannify, line 18: Wouldn't this cause an issues if multiple items are crafted but only a few could be picked up (because of inventory space)?

NitroxServer/GameLogic/EventTriggerer.cs
```diff
@@ -4,19 +4,22 @@
...
         private Stopwatch stopWatch;
-        private Dictionary<String, Timer> eventTimers = new Dictionary<string, Timer>();
+        
+
         public double ElapsedTime;
         public double AuroraExplosionTime;
+
         public EventTriggerer(PlayerManager playerManager, double elapsedTime, double? auroraExplosionTime)
         {
+            eventTimers = new();
```
Jannify, line 22: Minor: I personally would move this back to the field as it has no need to be in the constructor, but do as you think.

NitroxServer/GameLogic/TimeKeeper.cs
```diff
@@ -6,58 +6,53 @@ namespace NitroxServer.GameLogic
...
-            if (player != null)
-            {
-                player.SendPacket(new TimeChange(CurrentTime));
-            }
-            else
-            {
-                PlayerManager playerManager = NitroxServiceLocator.LocateService<PlayerManager>();
-                playerManager.SendPacketToAllPlayers(new TimeChange(CurrentTime));
-            }
+            PlayerManager playerManager = NitroxServiceLocator.LocateService<PlayerManager>();
```
Jannify, line 46: Should be moved to the constructor

NitroxServer/Communication/Packets/Processors/PlayerJoiningMultiplayerSessionProcessor.cs
```diff
@@ -18,31 +18,29 @@ public class PlayerJoiningMultiplayerSessionProcessor : UnauthenticatedPacketPro
         private readonly TimeKeeper timeKeeper;
         private readonly World world;
 
-        public PlayerJoiningMultiplayerSessionProcessor(TimeKeeper timeKeeper,
-            PlayerManager playerManager, World world)
+        public PlayerJoiningMultiplayerSessionProcessor(TimeKeeper timeKeeper, PlayerManager playerManager, World world)
         {
-            this.timeKeeper = timeKeeper;
             this.playerManager = playerManager;
+            this.timeKeeper = timeKeeper;
```
Coding-Hen, line 24: Should these be the other way around as the parameters are still in the original order?

NitroxServer/GameLogic/TimeKeeper.cs
```diff
@@ -1,63 +1,59 @@
...
-                                                                      SUBNAUTICA_FUTURE_START_DATE.Second);
+        public DateTime ServerStartTime { get; set; }
 
-        // Discrepancy value for player based time modifications
-        private float correctionValue;
+        private readonly PlayerManager playerManager;
+        private double correctionValue;
 
-        public DateTime ServerStartTime { get; set; } = DateTime.UtcNow;
+        public TimeKeeper(PlayerManager playerManager)
```
Coding-Hen, line 16: Does this change now work, where does playermanager get set?

dartasen: Dependency injection

Jannify (approved):

PR #1529, opened 2021-07-18
Automatic savefile upgrade

Coding-Hen:
Includes the changes from #1517 and upgrades the file to the latest version upon startup to prevent wiping of original save and creation of a new world


NitroxServer/Serialization/World/WorldPersistence.cs
```diff
@@ -214,5 +213,49 @@ public World CreateWorld(PersistedWorldData pWorldData, ServerGameMode gameMode)
 
             return world;
         }
+
+        private void UpgradeSave()
+        {
+            string saveDir = config.SaveName;
```
Jannify, line 219: Would make `saveDir` a parameter just to be safe

NitroxServer/ConsoleCommands/UpgradeCommand.cs
```diff
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
```
Jannify, line 1: I don't see the reason to keep this command when we do it on load automatically

NitroxServer/Serialization/World/WorldPersistence.cs
```diff
@@ -214,5 +213,49 @@ public World CreateWorld(PersistedWorldData pWorldData, ServerGameMode gameMode)
...
+            // SaveFileVersion structure was updated in V1.5.0.0
+            // This can be removed with V1.6.0.0 or later
+            if (File.ReadAllText(Path.Combine(saveDir, $"Version{FileEnding}")).Contains("BaseDataVersion"))
+            {
+                saveFileVersion = new SaveFileVersion(new Version(1, 4, 0, 0));
+            }
+
+            if (saveFileVersion.Version == NitroxEnvironment.Version)
+            {
+                Log.Info("Save files are already at the newest version");
```
Jannify, line 232: Don't know if we should keep this Log. Would be executed on every server start...

Jannify (changes requested):

NitroxServer-Subnautica/Program.cs
```diff
@@ -37,6 +37,7 @@ private static async Task Main(string[] args)
             AppDomain.CurrentDomain.UnhandledException += CurrentDomainOnUnhandledException;
 
             ConfigureCultureInfo();
+            Log.Info($"Starting NitroxServer V{NitroxEnvironment.Version} for Subnautica");
```
dartasen, line 40:
```suggestion
            Log.Info($"Starting NitroxServer V{NitroxEnvironment.Version}");
```

NitroxServer/ConsoleCommands/UpgradeCommand.cs
```diff
@@ -34,6 +34,13 @@ protected override void Execute(CallArgs args)
 
             SaveFileVersion saveFileVersion = worldPersistence.Serializer.Deserialize<SaveFileVersion>(Path.Combine(saveDir, $"Version{fileEnding}"));
 
+            // SaveFileVersion structure was updated in V1.5.0.0
+            // This can be removed with V1.6.0.0 or later
```
dartasen, line 38: Is this something worth keeping ? I do think it's not worth it maintaining save files at this stage of the developement, imo upgrades would fit better for future version, once the core mod is stabilized

NitroxServer/Serialization/World/WorldPersistence.cs
```diff
@@ -214,5 +213,49 @@ public World CreateWorld(PersistedWorldData pWorldData, ServerGameMode gameMode)
...
         }
+
+        private void UpgradeSave()
+        {
+            string saveDir = config.SaveName;
+
+            SaveFileVersion saveFileVersion = Serializer.Deserialize<SaveFileVersion>(Path.Combine(saveDir, $"Version{FileEnding}"));
+
+            // SaveFileVersion structure was updated in V1.5.0.0
+            // This can be removed with V1.6.0.0 or later
```
dartasen, line 224: same

NitroxServer/Serialization/World/WorldPersistence.cs
```diff
@@ -91,12 +94,8 @@ internal Optional<World> LoadFromFile(string saveDir)
             try
             {
                 PersistedWorldData persistedData = new();
-                SaveFileVersion saveFileVersion = Serializer.Deserialize<SaveFileVersion>(Path.Combine(saveDir, $"Version{FileEnding}"));
 
-                if (saveFileVersion == null || saveFileVersion.Version != NitroxEnvironment.Version)
-                {
-                    throw new InvalidDataException("Version file is empty or save data files are too old");
-                }
```
dartasen, line 99: I think those kind of data are good to keep, maybe just a log

NitroxServer-Subnautica/Program.cs
Jannify: As this is in `NitroxServer-Subnautica` I saw this fitting. Especially with BZ in mind.

NitroxServer/Serialization/World/WorldPersistence.cs
Jannify: Yeah the `saveFileVersion == null` check should be kept imo.

NitroxServer/ConsoleCommands/UpgradeCommand.cs
Jannify: I would leave it in for this version. It's just a couple lines of extra code and the upgrade script is already written. On top of that we could use it as a small testing/beta phase on the users as only a few of them were using JSON as save-file-format.

NitroxServer/Serialization/World/WorldPersistence.cs
Coding-Hen: Yes I agree, it was something I was thinking about removing as you are right does show each startup.

NitroxServer/Serialization/World/WorldPersistence.cs
Coding-Hen: Wont this get caught by the check of is the file version.json exists? I believe the removed saveFileVersion with deserialization defaults to version 1.5.0.0 when it cannot deserialise the content so should never be null

Jannify: Even if the file does not exist or is corrupted?

Coding-Hen: If it doesnt exist then the [check](https://github.com/SubnauticaNitrox/Nitrox/blob/master/NitroxServer/Serialization/World/WorldPersistence.cs#L85) for a previous file should catch it unless I am mistaken. If it is currupted like for example a 1.4 file with the BaseDataVersion then it defaults to 1.5.0.0

Jannify (approved):

PR #1531, opened 2021-07-19
Moved server list handling to separate class and moved file out of Subnautica folder

Measurity:
Code for moving servers file:
https://github.com/SubnauticaNitrox/Nitrox/pull/1531/files#diff-3e79ff7d448e24c0bcab15fec82aa7cd292b05f94f7c5937c8d81e554d1f97c1R33-R37

### DIFF
--- a/NitroxModel/Helper/Validate.cs
+++ b/NitroxModel/Helper/Validate.cs
@@ -68,6 +68,21 @@ namespace NitroxModel.Helper
             }
         }
 
+        public static void String(string s)
+        {
+            if (!string.IsNullOrWhiteSpace(s))
+            {
+                return;
+            }
+            
+            Optional<string> paramName = GetParameterName<string>();
+            if (paramName.HasValue)
+            {
+                throw new ArgumentException("String must not be null of whitespace", paramName.Value);
+            }
+            throw new ArgumentException("String must not be null of whitespace");
+        }
+
         public static void IsFalse(bool b, string message)
         {
             if (b)
--- a/NitroxModel/Helper/Validate.cs
+++ b/NitroxModel/Helper/Validate.cs
@@ -76,11 +76,7 @@ namespace NitroxModel.Helper
             }
             
             Optional<string> paramName = GetParameterName<string>();
-            if (paramName.HasValue)
-            {
-                throw new ArgumentException("String must not be null of whitespace", paramName.Value);
-            }
-            throw new ArgumentException("String must not be null of whitespace");
+            throw new ArgumentException("String must not be null of whitespace", paramName.OrElse(null));
         }
 
         public static void IsFalse(bool b, string message)
--- a/NitroxModel/Serialization/ServerList.cs
+++ b/NitroxModel/Serialization/ServerList.cs
@@ -1,0 +1,134 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Net;
+using NitroxModel.Helper;
+using NitroxModel.Logger;
+
+namespace NitroxModel.Serialization
+{
+    public class ServerList
+    {
+        private const string SERVERS_FILE_NAME = "servers";
+        private static ServerList instance;
+        private readonly List<Entry> entries = new();
+        public static ServerList Instance => instance ??= From(DefaultFile);
+
+        private static ServerList Default
+        {
+            get
+            {
+                ServerList list = new();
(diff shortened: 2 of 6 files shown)


NitroxModel/Serialization/ServerList.cs
```diff
@@ -0,0 +1,134 @@
...
+        public static ServerList From(string file = null)
+        {
+            // TODO: Remove backward compatibility after 1.5.0.0 release (no need to move old server file).
+            if (File.Exists(SERVERS_FILE_NAME))
+            {
+                File.Move(SERVERS_FILE_NAME, DefaultFile);
+            }
+
+            // Create file if it doesn't exist yet.
+            file ??= DefaultFile;
```
Coding-Hen, line 40: While I am not against using this, I'm wondering if we should be using this until c# 8 becomes more widely known. Is there a possibility this might be a little unclear as it's not a widely known operator?

Measurity: It's a limited operator for null checks only. Similar to the limited features of `+=`, `/=` etc. If you know `??` you likely also know what `??=` does. I like using newer syntax because it keeps the code modern and concise and it's a good learning experience if you're unfamiliar.

Feedback like this really aught to come from someone actually having trouble understand the syntax so we can discuss the way forward. I think that, as long as the good is understood, that there is no issue.

Coding-Hen: Yes I agree it's good to use new things, happy to continue use of it. Just wanted to throw a thought for people who may not have been familiar with the operator.

NitroxModel/Helper/Validate.cs
```diff
@@ -68,6 +68,21 @@ public static void IsFalse(bool b)
             }
         }
 
+        public static void String(string s)
```
dartasen, line 71: This method name sounds not really clear to me

NitroxModel/Helper/Validate.cs
```diff
@@ -68,6 +68,21 @@ public static void IsFalse(bool b)
...
+                return;
+            }
+            
+            Optional<string> paramName = GetParameterName<string>();
+            if (paramName.HasValue)
+            {
+                throw new ArgumentException("String must not be null of whitespace", paramName.Value);
+            }
+            throw new ArgumentException("String must not be null of whitespace");
+        }
```
dartasen, line 84: Is this whole extra logic needed ? I think this can be simplified

NitroxModel/Serialization/ServerList.cs
```diff
@@ -0,0 +1,134 @@
...
+
+        public IEnumerable<Entry> Entries => entries;
+
+        public static ServerList From(string file = null)
+        {
+            // TODO: Remove backward compatibility after 1.5.0.0 release (no need to move old server file).
+            if (File.Exists(SERVERS_FILE_NAME))
+            {
+                File.Move(SERVERS_FILE_NAME, DefaultFile);
+            }
```
dartasen, line 37: Even if it's not a nominal scenario, we should probably go the safe way

NitroxModel/OS/FileSystem.cs
```suggestion
            // TODO: Remove backward compatibility after 1.5.0.0 release (no need to move old server file).
            FileSystem.instance.ReplaceFile(SERVERS_FILE_NAME, DefaultFile);
```

NitroxModel/Serialization/ServerList.cs
```diff
@@ -0,0 +1,134 @@
...
+        {
+            public string Name { get; }
+            public IPAddress Address { get; }
+            public int Port { get; }
+
+            public Entry(string name, IPAddress address, int port)
+            {
+                Validate.String(name);
+                Validate.NotNull(address);
+                Validate.IsTrue(port is > ushort.MinValue and <= ushort.MaxValue);
```
dartasen, line 98:
```suggestion
                Validate.IsTrue(port is > 1024 and <= ushort.MaxValue);
```

NitroxModel/Helper/Validate.cs
Measurity: Yup, ArgumentException checks for null so can just pass null into it to remove if statement.

NitroxModel/Serialization/ServerList.cs
Measurity: Fixed ReplaceFile to work across drives as well.

dartasen (approved):

PR #1534, opened 2021-07-21
WIP: Fix for multiple events related to Aurora

dartasen:
Since we're emulating events based on client events, we need to do something about events like aurora that we're emulating serverside.
* Introduced a whitelist for story events (since)
* Changes to StoryEventProcessor to avoid looping with the patch with the new whitelist

This should fix wrong aurora warnings order, and also multiple explosions


dartasen:
Should be ready

NitroxClient/Communication/Packets/Processors/StoryEventProcessor.cs
```diff
@@ -0,0 +1,67 @@
...
+    {
+        public override void Process(StoryEventSend packet)
+        {
+            switch (packet.StoryEventType)
+            {
+                case StoryEventType.PDA:
+                case StoryEventType.RADIO:
+                case StoryEventType.ENCYCLOPEDIA:
+                case StoryEventType.STORY:
+                    using (NitroxServiceLocator.LocateService<IPacketSender>().Suppress<StoryEventSend>())
```
Measurity, line 20: Add constructor to set `IPacketSender` as local field. `NitroxServiceLocator` is slow and should not be used if possible (because it's an anti pattern).

PR #1537, opened 2021-07-24
Fixed FMOD volume sync going max volume if source distance is out of range

Measurity:


NitroxClient/GameLogic/FMOD/FMODSystem.cs
```diff
@@ -11,19 +12,16 @@ namespace NitroxClient.GameLogic.FMOD
...
         {
             this.packetSender = packetSender;
-
-            string soundsWhitelist = Properties.Resources.soundsWhitelist;
-
+            string soundsWhitelist = Resources.soundsWhitelist;
             if (string.IsNullOrWhiteSpace(soundsWhitelist))
             {
-                Log.Error(new NullReferenceException(), "[FMODSystem]: soundsWhitelist.csv is null or whitespace");
+                Log.Error(new Exception(), "[FMODSystem]: soundsWhitelist.csv is null or whitespace");
```
Jannify, line 24: Why changing it to a generic exception? If NRE isn't fitting I would remove it completely and only log the message.

Measurity: Because it's null or whitespace, not just null. But still wanted the stack trace.

Measurity: The stack trace is useless, with this `new` on same line. Will remove it.

NitroxClient/MonoBehaviours/MultiplayerSeaMoth.cs
```diff
@@ -24,9 +24,10 @@ protected override void Awake()
 
         protected void Update()
         {
+            // Clamp volume between 0 and 1 (nothing or max). Going below 0 turns up volume to max.
             float distance = Vector3.Distance(Player.main.transform.position, transform.position);
-            rpmSound.GetEventInstance().setVolume(1 - distance / radiusRpmSound);
-            revSound.GetEventInstance().setVolume(1 - distance / radiusRevSound);
+            rpmSound.GetEventInstance().setVolume(Mathf.Clamp01(1 - distance / radiusRpmSound));
```
Jannify, line 29: Would do the same in `NitroxServer/Communication/Packets/Processors/PlayFMODAssetProcessor.cs`.

Jannify (changes requested):

Jannify (approved):

PR #1543, opened 2021-07-25
Fixed error loading server list when missing port on older server lists

Coding-Hen:


Measurity:
#1544 which do you like best?

Jannify:
Would say #1544 as it doesn't require regex and allocation of a list

Coding-Hen:
Closed in favour of #1544

PR #1545, opened 2021-07-25
Allows cfg to be updated with any changes rather than always using default properties for those missing

Coding-Hen:
Removed serialize from save due to it being created on load.
Switched out deserialize for Update as it calls deserialize for us


Measurity:
Should update the `Update` function so you can pass null to it:

```cs
        public void Update(Action<T> config)
        {
            try
            {
                Deserialize();
                config?.Invoke(this as T);
            }
            finally
            {
                Serialize();
            }
        }
```

Measurity (approved):

PR #1548, opened 2021-07-26
Switched version check to website rather than github

Coding-Hen:
Prevents a github tag causing an upgrade notification before we have released on website


Measurity:
We should rate limit update check to minimum once a day, maybe more.

dartasen:
Those changes will be integrated with the new launcher

PR #1551, opened 2021-07-31
Fixed modules not saving correctly due to using containerid

Coding-Hen:
Switches from using container id to item id as to not overide the whole container and just modify the item


Measurity (approved):
Confirmed working between server restarts, still synced for multiple players.

PR #1552, opened 2021-07-31
Moving ToString() from all packets to Packet.cs

Jannify:


NitroxModel/Packets/StoryEventSend.cs
```diff
@@ -5,28 +5,23 @@ namespace NitroxModel.Packets
...
-        public StoryEventSend(StoryEventType storyEventType, string key = "") : base()
+        public StoryEventSend(StoryEventType eventType, string key = "")
         {
-            StoryEventType = storyEventType;
+            EventType = eventType;
             Key = key;
         }
 
-        public override string ToString()
+        public enum StoryEventType
```
Measurity, line 17: Name could be shorter. `EventType` or just `Type`.

NitroxModel/Packets/Packet.cs
```diff
@@ -60,40 +63,58 @@ public enum UdpChannelId
...
         public WrapperPacket ToWrapperPacket()
         {
             return new WrapperPacket(Serialize());
         }
+
+        public override string ToString()
+        {
+            StringBuilder builder = new($"[{GetType().Name}: ");
+
+            foreach (PropertyInfo propertyInfo in GetType().GetProperties(BindingFlags.Public | BindingFlags.Instance))
```
Measurity, line 96: Should be cached per type (i.e. using a Dictionary) to improve performance when ToString() is called a lot.

Measurity (changes requested):

NitroxModel/Packets/Packet.cs
```diff
@@ -60,40 +65,64 @@ public enum UdpChannelId
...
+                cachedPropertiesByType.Add(packetType, properties);
+            }
+
+            StringBuilder builder = new($"[{packetType.Name}: ");
+            foreach (PropertyInfo property in properties)
+            {
+                if (property.Name is nameof(DeliveryMethod) or nameof(UdpChannel))
+                {
+                    continue;
+                }
```
Measurity, line 110: These properties should be excluded in the cache to reduce impact when ToString() is called a lot.

Measurity (approved):

PR #1555, opened 2021-08-01
Force Destroy Nearby Ghosts

GreenMachine2o3:
Not sure if this is done the best way, but maybe it will inspire you guys to write something similar or suggest improvements. I've been using this to deal with the occasional appearance of red ghosts. Doesn't solve the root problem but is a decent workaround so I figured I'd share.

**Testing:**
Walk up to red ghost -> press U -> boom


dartasen:
Thanks for your contribution, I don't think binding this to an hotkey is the best way to handle this, but could be a great idea for additionnal server check on startup or new debug command.

Will still working on bases to reduce those kind of bugs

PR #1556, opened 2021-08-02
Added reset argument to restart command

Measurity:


NitroxServer/ConsoleCommands/RestartCommand.cs
```diff
@@ -30,8 +31,15 @@ protected override void Execute(CallArgs args)
...
-
             server.Stop();
+            // If reset, delete save.
+            if (args.Get<bool>(0))
+            {
+                if (Directory.Exists(serverConfig.SaveName))
+                {
+                    Directory.Delete(serverConfig.SaveName, true);
+                }
+            }
```
dartasen, line 42: I don't think it's the responsability of the server restart command to reset a world ?

Measurity: Initially I had it as separate command but I needed the same code so I merged it with this.

Jannify: This also deletes any backup. That's a very risky command addition

NitroxServer/ConsoleCommands/RestartCommand.cs
```diff
@@ -30,9 +33,44 @@ protected override void Execute(CallArgs args)
...
+            // Ensure save folder is inside Nitrox folder, as to not accidentally delete files in the wrong location.
+            if (savePath.IndexOf(NitroxAppData.Instance.LauncherPath, StringComparison.OrdinalIgnoreCase) != 0)
+            {
+                return;
+            }
+            
+            foreach (string file in Directory.EnumerateFiles(savePath, "*", SearchOption.AllDirectories))
+            {
+                if (".zip".Equals(Path.GetExtension(file), StringComparison.OrdinalIgnoreCase))
```
Jannify, line 60: Minor: Would swap the sides of the `.Equals`

Jannify (commented):
I am still thinking that the reset command shouldn't be nested/executed in the restart even if it shares common code. As it's very destructive it should have its own `reset` command and a confirmation requirement (like you need to use `reset true` or `reset <admin password>`).
Code wise it seems fine.

Coding-Hen:
I agree with Jannify about a separate command. I think having them have to type in the admin password makes good sense as it can't be accidentally typed,  even if we changed the command to be server console only it would cause the user to go out of their way to reset which adds that confirmation. Alternatively we add a confirmation check but I don't think the permission system is really up for that as it stands

Issue #1368, opened 2021-01-06
[Bug] Radio displays it has a messages but nothing plays.


ProjectPatatoe:
### Version
1.4.0.0

### Bug
Radio shows message to be played, but i hear nothing

### Expected
Some message to be played

### Steps to reproduce
Enter game or wait a long time.

### Other details:
No one in our group (4 players) ever received the sunbeam radio events

TeamDoodz:
When this happened to me, there was an IndexOutOfBounds exception I believe.

Coding-Hen:
Do you have a playerlog for those IndexOutOfBounds exceptions? @TeamDoodz

TeamDoodz:
[20201227.log](https://github.com/SubnauticaNitrox/Nitrox/files/5810830/20201227.log)

Here it is. Sorry for the late reply and the immature usernames...

diabolical17:
Here, found the cause:

```log
17:01:55.212 [Server][ERR] Exception while processing packet: NitroxModel.Packets.RadioPlayPendingMessage System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection.
Parameter name: index
   at System.ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument argument, ExceptionResource resource)
   at System.Collections.Generic.List`1.RemoveAt(Int32 index)
   at NitroxModel.DataStructures.ThreadSafeCollection`1.RemoveAt(Int32 index) in C:\Users\Sunrunner\Desktop\Code\Nitrox\NitroxModel\DataStructures\ThreadSafeCollection.cs:line 188
   at NitroxServer.RadioPlayPendingMessageProcessor.Process(RadioPlayPendingMessage packet, Player player) in C:\Users\Sunrunner\Desktop\Code\Nitrox\NitroxServer\Communication\Packets\Processors\RadioPlayPendingMessageProcessor.cs:line 21
   at NitroxServer.Communication.Packets.Processors.Abstract.AuthenticatedPacketProcessor`1.ProcessPacket(Packet packet, IProcessorContext player) in C:\Users\Sunrunner\Desktop\Code\Nitrox\NitroxServer\Communication\Packets\Processors\Abstract\AuthenticatedPacketProcessor.cs:line 10
[...]

PR #1560, opened 2021-08-04
Added logging for Radio bug ArgumentOutOfRangeException

tornac1234:
Concerning [Issue #1368](https://github.com/SubnauticaNitrox/Nitrox/issues/1368)

### DIFF
--- a/NitroxServer/GameLogic/Unlockables/StoryGoalData.cs
+++ b/NitroxServer/GameLogic/Unlockables/StoryGoalData.cs
@@ -2,7 +2,6 @@
 using Newtonsoft.Json;
 using NitroxModel.DataStructures;
 using NitroxModel.DataStructures.GameLogic;
-using NitroxModel.Logger;
 using ProtoBufNet;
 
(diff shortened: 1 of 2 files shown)


NitroxServer/GameLogic/Unlockables/StoryGoalData.cs
```diff
@@ -18,9 +19,18 @@ public class StoryGoalData
...
         {
-            RadioQueue.RemoveAt(0);
+            if (RadioQueue.Count > 0)
+            {
+                RadioQueue.RemoveAt(0);
+                return true;
+            }
+            else
+            {
+                Log.Error("Couldn't remove the latest radio message: the radio queue is empty");
```
Measurity, line 31: You're logging twice for same issue so this log can be moved to the RadioPlay.Process instead (and adding the packet info to it as well).

tornac1234:
Ok I just changed it

Measurity (approved):

PR #1561, opened 2021-08-05
Fixed seed loading from config file in production environments

Yoshie2000:
The seed from the server config file wasn't actually being used anywhere, so I implemented it into the world generation for production environments.

### DIFF
--- a/NitroxServer/Serialization/World/WorldPersistence.cs
+++ b/NitroxServer/Serialization/World/WorldPersistence.cs
@@ -159,6 +159,9 @@ namespace NitroxServer.Serialization.World
 #if DEBUG
                     Seed = "TCCBIBZXAB"
 #endif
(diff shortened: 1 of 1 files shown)


NitroxServer/Serialization/World/WorldPersistence.cs
```diff
@@ -158,6 +158,9 @@ private World CreateFreshWorld()
                     ServerStartTime = DateTime.UtcNow,
 #if DEBUG
                     Seed = "TCCBIBZXAB"
+#endif
+#if (!DEBUG)
```
Measurity, line 162:
```suggestion
#else
```

Measurity (commented):
Thanks for fixing this.

Measurity (approved):

PR #1562, opened 2021-08-06
Additional help information for command options

Coding-Hen:
Console:
Help without command
image
Help with command
image
Invalid Command
image

### DIFF
--- a/NitroxServer/ConsoleCommands/Abstract/Command.cs
+++ b/NitroxServer/ConsoleCommands/Abstract/Command.cs
@@ -88,7 +88,11 @@ namespace NitroxServer.ConsoleCommands.Abstract
             }
 
             cmd.AppendFormat(" {0}", string.Join(" ", Parameters));
-            return cropText ? $"{cmd}" : $"{cmd,-32} - {Description}";
+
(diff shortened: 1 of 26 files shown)


dartasen:
I was more thinking of extending the help command with something similar to bukkit help command : 

i.e: 
`/help {command}` : 
```txt
=== Showing command list ===
     /changeservergamemode [gamemode]     -   Allows to change server gamemode
     ....
```

```txt
`/help changeservergamemode` :
=== Showing help for changeservergamemode === 
Allows to change server gamemode

[gamemode] => gamemode of the world (values: SURVIVAL, FREEDOM, CREATIVE, HARDCORE)
```

dartasen:
Related to #1574

dartasen (commented):
Maybe the UI can be improved a bit to be more sexy, imo the pattern ` {name}            -   {description}` does not really fit parameters and is repetitive if we look at the previous line which is the command description line (might suggest to the user that they're in the same category)

Code wise, might be great to add some desc generating behaviours, especially for Enums, so we don't need to edit descriptions everytime we do some edit in models, what do you think ?

dartasen:
Closed for inactivity

PR #1563, opened 2021-08-07
Added scan progress sync

tornac1234:
Concerning Issue#1467

### DIFF
--- a/NitroxClient/GameLogic/InitialSync/PdaInitialSyncProcessor.cs
+++ b/NitroxClient/GameLogic/InitialSync/PdaInitialSyncProcessor.cs
@@ -72,12 +72,17 @@ namespace NitroxClient.GameLogic.InitialSync
         private void SetPDAEntryPartial(List<PDAEntry> entries)
         {
             List<PDAScanner.Entry> partial = (List<PDAScanner.Entry>)(typeof(PDAScanner).GetField("partial", BindingFlags.NonPublic | BindingFlags.Static).GetValue(null));
-
+            
             foreach (PDAEntry entry in entries)
             {
-                partial.Add(new PDAScanner.Entry { progress = entry.Progress, techType = entry.TechType.ToUnity(), unlocked = entry.Unlocked });
+                PDAScanner.Entry newEntry = new PDAScanner.Entry { progress = entry.Progress, techType = entry.TechType.ToUnity(), unlocked = entry.Unlocked };
+                // If, for some reason this happens, at least the client will be able to ignore it, in the other case, he wouldn't be able to connect
+                if (newEntry.progress != 0f)
+                {
+                    partial.Add(newEntry);
+                    Log.Debug($"New partial entry: {newEntry}");
+                }
             }
-
             Log.Debug($"PDAEntryPartial: New added: {entries.Count}, Total: {partial.Count}");
         }
 
--- a/NitroxPatcher/Patches/Dynamic/PDAScanner_Scan_Patch.cs
+++ b/NitroxPatcher/Patches/Dynamic/PDAScanner_Scan_Patch.cs
@@ -2,10 +2,12 @@
 using System.Collections.Generic;
 using System.Reflection;
 using System.Reflection.Emit;
+using System.Threading;
 using HarmonyLib;
 using NitroxClient.GameLogic;
 using NitroxModel.Core;
 using NitroxModel.Helper;
+using NitroxModel.Logger;
 
 namespace NitroxPatcher.Patches.Dynamic
 {
@@ -17,9 +19,13 @@ namespace NitroxPatcher.Patches.Dynamic
         public static readonly OpCode INJECTION_OPCODE = OpCodes.Call;
         public static readonly object INJECTION_OPERAND = typeof(ResourceTracker).GetMethod("UpdateFragments", BindingFlags.Public | BindingFlags.Static);
 
+        public static readonly OpCode INJECTION_OPCODE_2 = OpCodes.Ldsfld;
+        public static readonly object INJECTION_OPERAND_2 = typeof(PDAScanner).GetField("cachedProgress", BindingFlags.NonPublic | BindingFlags.Static);
+
         public static IEnumerable<CodeInstruction> Transpiler(MethodBase original, IEnumerable<CodeInstruction> instructions)
         {
             Validate.NotNull(INJECTION_OPERAND);
+            Validate.NotNull(INJECTION_OPERAND_2);
 
             foreach (CodeInstruction instruction in instructions)
             {
@@ -32,6 +38,10 @@ namespace NitroxPatcher.Patches.Dynamic
                      * >> PDAScanner_Scan_Patch.Callback();
                      */
                     yield return new CodeInstruction(OpCodes.Call, typeof(PDAScanner_Scan_Patch).GetMethod("Callback", BindingFlags.Static | BindingFlags.Public));
+                }
+                else if(instruction.opcode.Equals(INJECTION_OPCODE_2) && instruction.operand.Equals(INJECTION_OPERAND_2))
+                {
+                    yield return new CodeInstruction(OpCodes.Call, typeof(PDAScanner_Scan_Patch).GetMethod("ProgressCallback", BindingFlags.Static | BindingFlags.Public));
                 }
             }
         }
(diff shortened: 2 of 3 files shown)


NitroxPatcher/Patches/Dynamic/PDAScanner_Scan_Patch.cs
```diff
@@ -53,6 +63,77 @@ public static void Callback()
...
+
+                // Check if the entry is already in the partial list, and then, change the progress
+                if (PDAScanner.GetPartialEntryByKey(scanTarget.techType, out PDAScanner.Entry partialEntry) && partialEntry.progress > entry.progress)
+                {
+                    entry.progress = partialEntry.progress;
+                    PDAScanner.scanTarget.progress = partialEntry.progress;
+                }
+
+                // Updating the progress throttling
+                if (ThrottlingEntries.ContainsKey(scanTarget.techType))
```
Jannify, line 122: Maybe use `TryGetValue`

NitroxServer/Server.cs
```diff
@@ -57,6 +57,7 @@ public string SaveSummary
                 builder.AppendLine($" - Encyclopedia entries: {world.GameData.PDAState.EncyclopediaEntries.Count}");
                 builder.AppendLine($" - Storage slot items: {world.InventoryManager.GetAllStorageSlotItems().Count}");
                 builder.AppendLine($" - Inventory items: {world.InventoryManager.GetAllInventoryItems().Count}");
+                builder.AppendLine($" - Partially unlocked: {world.GameData.PDAState.PartiallyUnlockedByTechType.Count}");
```
Jannify, line 60: As it's has no "Tech" in the message there is no clear understanding what is really partial unlocked.

NitroxClient/GameLogic/InitialSync/PdaInitialSyncProcessor.cs
```diff
@@ -72,12 +72,17 @@ private void SetPDAEntryComplete(List<NitroxTechType> pdaEntryComplete)
...
         {
             List<PDAScanner.Entry> partial = (List<PDAScanner.Entry>)(typeof(PDAScanner).GetField("partial", BindingFlags.NonPublic | BindingFlags.Static).GetValue(null));
-
+            
             foreach (PDAEntry entry in entries)
             {
-                partial.Add(new PDAScanner.Entry { progress = entry.Progress, techType = entry.TechType.ToUnity(), unlocked = entry.Unlocked });
+                PDAScanner.Entry newEntry = new PDAScanner.Entry { progress = entry.Progress, techType = entry.TechType.ToUnity(), unlocked = entry.Unlocked };
+                // If, for some reason this happens, at least the client will be able to ignore it, in the other case, he wouldn't be able to connect
+                if (newEntry.progress != 0f)
```
Jannify, line 80: I think `newEntry.progress` is a float. Won't there be problems with equality checks?

NitroxPatcher/Patches/Dynamic/PDAScanner_Scan_Patch.cs
```diff
@@ -53,6 +63,77 @@ public static void Callback()
...
+                Thread.Sleep(2000);
+                Log.Debug($"Throttled {techType} waiting");
+            }
+            while (DateTimeOffset.UtcNow.ToUnixTimeMilliseconds() < ThrottlingEntries[techType].Item1.ToUnixTimeMilliseconds() + 2000);
+            Log.Debug($"Throttled {techType} will now be sent");
+
+            // Don't send the packet if the scan is already finished
+            if (!PDAScanner.ContainsCompleteEntry(techType))
+            {
+                NitroxServiceLocator.LocateService<PDAManagerEntry>().Progress(ThrottlingEntries[techType].Item3);
```
Jannify, line 87: As this will be executed fairly often `PDAManagerEntry` should be cached.

NitroxPatcher/Patches/Dynamic/PDAScanner_Scan_Patch.cs
```diff
@@ -53,6 +63,77 @@ public static void Callback()
             }
         }
 
+        public static readonly int PACKET_SENDING_RATE = 500; // in milliseconds
+
+        // Tuple<LatestProgressTime, LatestPacketSendTime, Entry>
+        public static Dictionary<TechType, Tuple<DateTimeOffset, DateTimeOffset, PDAScanner.Entry>> ThrottlingEntries = new Dictionary<TechType, Tuple<DateTimeOffset, DateTimeOffset, PDAScanner.Entry>>();
+        // Store the thread concerning a certain type
+        public static Dictionary<TechType, Thread> ThrottlingThreads = new Dictionary<TechType, Thread>();
```
Jannify, line 71: Won't it be possible to combine the tuple and the thread in one class so you can have only one dictionary?

NitroxPatcher/Patches/Dynamic/PDAScanner_Scan_Patch.cs
```diff
@@ -53,6 +63,77 @@ public static void Callback()
...
+
+        // After 2 seconds without scanning, a packet will be sent with the latest progress
+        public static void Throttle(TechType techType)
+        {
+            do
+            {
+                Thread.Sleep(2000);
```
Jannify, line 78: I don't know for sure but in Unity environments coroutines or the Update function of MonoBehaviors are usually used for things like this. @Measurity do you have more knowledge in this field?

Jannify (changes requested):
These are only changes I spotted while looking at it. I am probably bound to my phone for the next to weeks so another dev has to do the in-game checking.

NitroxPatcher/Patches/Dynamic/PDAScanner_Scan_Patch.cs
Measurity: Coroutines or tasks are the way to go for short workloads. Thread objects are quite heavy because they need to allocate their own stack and introduce more load on the Windows thread scheduler.

NitroxClient/GameLogic/InitialSync/PdaInitialSyncProcessor.cs
tornac1234: I'll test it again asap but I'm quite sure it solved the problem

NitroxPatcher/Patches/Dynamic/PDAScanner_Scan_Patch.cs
tornac1234: Sounds good to me, I'll make this change

NitroxPatcher/Patches/Dynamic/PDAScanner_Scan_Patch.cs
```diff
@@ -53,6 +63,77 @@ public static void Callback()
...
+                Log.Debug($"Throttled {techType} waiting");
+            }
+            while (DateTimeOffset.UtcNow.ToUnixTimeMilliseconds() < ThrottlingEntries[techType].Item1.ToUnixTimeMilliseconds() + 2000);
+            Log.Debug($"Throttled {techType} will now be sent");
```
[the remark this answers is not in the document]
tornac1234: Oh yes that's my bad I forgot to remove these debugging logs (not necessary)

NitroxPatcher/Patches/Dynamic/PDAScanner_Scan_Patch.cs
tornac1234: I'm not really sure to understand what you mean for this one

Measurity: `NitroxServiceLocator.LocateService<PDAManagerEntry>()` takes a bit to execute so it should be cached in a field on this class.

tornac1234: Oh ok

tornac1234:
As I won't be available before next week's week end, just consider that this issue is paused. I'll get back to it as soon as I'm able to write code again :D

tornac1234:
I thought of something about this PR, (to modify later when we'll look more into it), currently the scan progress concerns all members of one species and not only one entity as it is in the current game. If entities IDs are always the same over different sessions, then I'll make the necessary change that will save the progress for each entity only.

PR #1578, opened 2021-08-30
BuilderTool_HandleInput_Patch Transpiler fix

MrPurple6411:
DeconstructionBegin needs a NitroxId not a gameobject. 

Every time I would try to deconstruct something it would throw an exception as this method was being provided with the wrong type.


dartasen:
Maybe it would be better to add a callback method in the transpiler instead of writing the whole code in IL, so our IDE can check for dumb errors like it (see Rocket transpilers for example)

Jannify (approved):

PR #1579, opened 2021-08-30
Add OnEnable and OnDisable to NitroxEntity

killzoms:
This theoretically should help with some pointless log spam

### DIFF
--- a/NitroxClient/MonoBehaviours/NitroxEntity.cs
+++ b/NitroxClient/MonoBehaviours/NitroxEntity.cs
@@ -115,7 +115,7 @@ namespace NitroxClient.MonoBehaviours
             }
         }
 
-        public void OnDisable() // Called with Destroy as well
+        public void OnDestroy()
         {
             if (Id == null)
             {
--- a/NitroxClient/MonoBehaviours/NitroxEntity.cs
+++ b/NitroxClient/MonoBehaviours/NitroxEntity.cs
@@ -115,17 +115,6 @@ namespace NitroxClient.MonoBehaviours
             }
         }
 
-        public void OnDestroy()
-        {
-            if (Id == null)
-            {
-                Log.Error($"Nitrox Id was null on {gameObject.name} in OnDisable");
-                return;
-            }
-
-            gameObjectsById.Remove(Id);
-        }
-
         public void OnEnable()
         {
             if (Id == null)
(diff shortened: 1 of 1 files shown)


NitroxClient/MonoBehaviours/NitroxEntity.cs
```diff
@@ -115,6 +115,22 @@ public void Start()
             }
         }
 
+        public void OnDisable() // Called with Destroy as well
+        {
+            if (Id != null) // Theoretically shouldnt happen but Unity sucks
+            {
+                gameObjectsById.Remove(Id);
+            }
```
dartasen, line 123: Maybe we should log this happy case if it shouldn't happen ?

NitroxClient/MonoBehaviours/NitroxEntity.cs
```diff
@@ -115,6 +115,22 @@ public void Start()
...
+        }
+
+        public void OnEnable()
+        {
+            if (Id != null)
+            {
+                gameObjectsById[Id] = gameObject;
+            }
```
dartasen, line 131: same here

dartasen (changes requested):

Measurity:
Does this have to do with serialization that skips this ID? If we know what's causing it then it should have a comment in code.

Measurity:
TODO: Verify that this won't prematurely unassign the Nitrox ID (i.e. object is still needed somewhere by our code) when said object is far away.

Measurity:
Can we move this code to OnDestroy instead?

killzoms:
@Measurity done

NitroxClient/MonoBehaviours/NitroxEntity.cs
```diff
@@ -115,6 +115,28 @@ public void Start()
...
+        {
+            if (Id == null)
+            {
+                Log.Error($"Nitrox Id was null on {gameObject.name} in OnEnable");
+                return;
+            }
+
+            gameObjectsById[Id] = gameObject;
+        }
+
```
Jannify, line 139:
```suggestion
        public void OnEnable()
        {
            if (Id == null)
            {
                Log.Error($"Nitrox Id was null on {gameObject.name} in OnEnable");
                return;
            }

            gameObjectsById[Id] = gameObject;
        }

        public void OnDestroy()
        {
            if (Id == null)
            {
                Log.Error($"Nitrox Id was null on {gameObject.name} in OnDestroy");
                return;
            }

            gameObjectsById.Remove(Id);
        }
```

Jannify (approved):

Issue #1109, opened 2020-05-22
[Server] Load Batches Up-Front


Sunrunner37:
When the server starts for the first time, it will not have any entities loaded; all entity loading happens on-demand when a client loads a particular batch for the first time.  Loading these entities on demand causes problems for clients with low-spec computers as it is resource intensive and can cause frame lag.  

We want to add the functionality for the server to load batches the first time the server starts.  This will loop through all possible batches and try to load them into the entity cache.  Ideally, this functionality is gated behind a new configuration.  Below are the relevant code areas:

Batch spawning: https://github.com/SubnauticaNitrox/Nitrox/blob/master/NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs#L69

Map Sizes: 
https://github.com/SubnauticaNitrox/Nitrox/blob/master/NitroxModel-Subnautica/Helper/SubnauticaMap.cs

PR #1580, opened 2021-08-30
Load Entity Batches Up-Front and Cache

MrPurple6411:
https://github.com/SubnauticaNitrox/Nitrox/issues/1109

Add functionality for the server to Cache all entity batches on the NEXT server start. This will loop through all possible batches and try to load them into the entity cache. 

Gated behind a new configuration option:
`# Set to true to Cache entities for the whole map on next run. `
`# WARNING! Will make server load take longer on the cache run but players will gain a performance boost when entering new areas.`
`CreateFullEntityCache=True`

### DIFF
--- a/NitroxServer-Subnautica/Program.cs
+++ b/NitroxServer-Subnautica/Program.cs
@@ -41,8 +41,8 @@ namespace NitroxServer_Subnautica
 
             AppMutex.Hold(() =>
             {
-                Log.Info("Waiting for 30 seconds on other Nitrox servers to initialize before starting..");
-            }, 30000);
+                Log.Info("Waiting on other Nitrox servers to initialize before starting..");
+            }, 120000);
             Server server;
             Task listenForCommands;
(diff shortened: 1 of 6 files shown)


Jannify:
Fixes #1109

Jannify:
If the batches are loaded p front I would disable the log message with "Spawning X entities..." because it's quit spam intensive. In exchange I would add a log message before and after the loading telling the user that this process can take up to 2 minutes,

MrPurple6411:
> If the batches are loaded p front I would disable the log message with "Spawning X entities..." because it's quit spam intensive. In exchange I would add a log message before and after the loading telling the user that this process can take up to 2 minutes,

yea I left the spam because otherwise it just feels like the server load stalled...   maybe I could change it to just post a % of total cells completed every 5% or so calculated by cells completed vs total cells?

Measurity (approved):

PR #1581, opened 2021-08-30
Ion Cube Parenting Fix

MrPurple6411:
Correctly Parent ion cubes to the pedestals so they are now accessible.

Subnautica_VCBarPw5vU


Jannify:
While the fix is fine maybe we should implement a processor system like in `NitroxClient.GameLogic.Bases.Spawning` for future changes and adaptability.

MrPurple6411:
> While the fix is fine maybe we should implement a processor system like in `NitroxClient.GameLogic.Bases.Spawning` for future changes and adaptability.

ok so I got it semi working using client side processor BUT it only seems to work sometimes and will break if a player teleports to the objects......... also if a second player comes near it breaks...........   I REALLY think the only way to do this reliably is on the server side and the only place I can think of to correct this is at this spot where I have access to the identifiers needed to do it. 


That being said..... I will also take a peek into WHY the prefab has the ion cube set to the wrong parent in the first place as that ion cube should not even exist in the prefab so SOMEWHERE we have put it there and attached it to the wrong thing...  now if I can fix that then this whole code won't be needed at all.

MrPurple6411:
Should be ready for merge after yall test it out. :-)

NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs
```diff
@@ -359,8 +359,51 @@ private List<Entity> ConvertComponentPrefabsToEntities(List<PrefabAsset> prefabs
                              counter++,
                              parent);
 
-                prefabEntity.ChildEntities = ConvertComponentPrefabsToEntities(prefab.Children, prefabEntity, deterministicBatchGenerator);
+                if (prefab.Name.Contains("(Placeholder)"))
```
dartasen, line 362: Maybe leave a comment for this little buddy ?
 I personnaly have no clue about it

NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs
```diff
@@ -359,8 +359,51 @@ private List<Entity> ConvertComponentPrefabsToEntities(List<PrefabAsset> prefabs
...
+                {
+                    PrefabAsset spawnablePrefab = spawnablePrefabs.Find((x) => x.TransformAsset == transform);
+                    if (spawnablePrefab != null)
+                    {
+                        Optional<UweWorldEntity> opWorldEntity = worldEntityFactory.From(spawnablePrefab.ClassId);
 
+                        if (!opWorldEntity.HasValue)
+                        {
+                            Log.Debug("Unexpected Empty WorldEntity! " + spawnablePrefab.ClassId);
```
dartasen, line 371: String concat is bad, we should always use string interpolation

NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs
```diff
@@ -330,15 +332,13 @@ private void CreatePrefabPlaceholdersWithChildren(Entity entity, string classId,
...
-
-                entity.ChildEntities.AddRange(ConvertComponentPrefabsToEntities(group.ExistingPrefabs, entity, deterministicBatchGenerator));
             }
         }
 
         // Entities that have been spawned by a parent prefab (child game objects baked into the prefab).
         // created separately as we don't actually want to spawn these but instead just update the id.
         // will refactor this piece a bit later to split these into a new data structure.
-        private List<Entity> ConvertComponentPrefabsToEntities(List<PrefabAsset> prefabs, Entity parent, DeterministicBatchGenerator deterministicBatchGenerator)
+        private List<Entity> ConvertComponentPrefabsToEntities(List<PrefabAsset> prefabs, Entity parent, ref DeterministicBatchGenerator deterministicBatchGenerator, ref List<PrefabAsset> spawnablePrefabs)
```
dartasen, line 341: ref ?

NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs
MrPurple6411: lol  yea that part is actually copy pasted from further up in the code so I will fix both.

NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs
MrPurple6411: that List is used and altered throughout the recursive spawning.

NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs
MrPurple6411: Sorry what about this part?  That line that is removed here is actually just altered and moved down below.

NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs
MrPurple6411: unless you mean for the deterministic and that would be a mis commit of a troubleshooting step I was doing when it all of a sudden told me I had duplicate entities and I thought somehow it was not making new id's lol   yea that ref is not needed and I will fix it asap

NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs
MrPurple6411: 532514d

dartasen (approved):

NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs
```diff
@@ -359,8 +359,58 @@ private List<Entity> ConvertComponentPrefabsToEntities(List<PrefabAsset> prefabs
                              counter++,
                              parent);
 
-                prefabEntity.ChildEntities = ConvertComponentPrefabsToEntities(prefab.Children, prefabEntity, deterministicBatchGenerator);
+                // Checkes if the current object being setup is a Placeholder object.
+                if (prefab.Name.Contains("(Placeholder)"))
```
Jannify, line 363: I would add a comment here saying that this is a safe method as you checked that the placeholders are all named like this.

MrPurple6411: 2cd9536

Issue #1468, opened 2021-05-12
wreck's not spawning, or partially spawning


pkTux:
### Version

1.4.0.0
### Bug
wrecks sometimes fail to spawn  & can be swam thru, other times partially spawn

### Expected
wreck exists

### Steps to reproduce
[Player.log](https://github.com/SubnauticaNitrox/Nitrox/files/6462611/Player.log)

sometimes happens when you travel great distances, from reconnect point.

### Other details:
windows 10.

1Kr0noZ1:
Can confirm this one. Sometimes wrecks are stretched out vertically over significant depths. Where it should be on the sea floor, some parts may actually be there with the rest of the wreck being in pieces and floating above in a disconnected mess (meaning if it's one of the wrecks you are supposed to get into, sometimes the doors are just free floating and the actual interior bits are present in the vicinity but blocked by invisible walls, loot also doesn't spawn in properly). It mostly seems to happen to wrecks far away from the initial spawn location (making it hard to notice at first) but can persist through server restarts regardless of players being in the area on spawn or not.

Mephisto090:
I can confirm that too. Makes it almost impossible to play as it is difficult to unlock new things

AwesomenessZ:
I've found that using the cheat freecam, moving a little, and typing freecam again to deactivate fixes the position of the wrecks

PR #1584, opened 2021-09-01
Spawning code race condition fix.

MrPurple6411:
Seems to fix https://github.com/SubnauticaNitrox/Nitrox/issues/1468

### DIFF
--- a/NitroxClient/GameLogic/Spawning/DefaultEntitySpawner.cs
+++ b/NitroxClient/GameLogic/Spawning/DefaultEntitySpawner.cs
@@ -53,11 +53,15 @@ namespace NitroxClient.GameLogic.Spawning
 
         private GameObject CreateGameObject(TechType techType, string classId)
         {
-            IPrefabRequest prefabRequest = PrefabDatabase.GetPrefabAsync(classId);
-            if (!prefabRequest.TryGetPrefab(out GameObject prefab))
+            GameObject prefab = null;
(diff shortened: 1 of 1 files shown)


NitroxClient/GameLogic/Spawning/DefaultEntitySpawner.cs
```diff
@@ -53,11 +53,15 @@ public Optional<GameObject> Spawn(Entity entity, Optional<GameObject> parent, En
...
         private GameObject CreateGameObject(TechType techType, string classId)
         {
-            IPrefabRequest prefabRequest = PrefabDatabase.GetPrefabAsync(classId);
-            if (!prefabRequest.TryGetPrefab(out GameObject prefab))
+            GameObject prefab = null;
+            if (PrefabDatabase.TryGetPrefabFilename(classId, out string filename))
             {
-                prefab = CraftData.GetPrefabForTechType(techType, false);
+                prefab = Resources.Load<GameObject>(filename);
+            }
```
dartasen, line 60: Maybe we should implement both CreateGameObject and CreateGameObjectAsync, so we can switch easily between them

MrPurple6411: This PRis just to fix the sync method. 

I will be making a full overhaul PR for the spawning system separately.

Measurity (approved):

dartasen (approved):

PR #1586, opened 2021-09-04
Add null check to avoid Null ref exception when no vehicle docked.

MrPurple6411:


Coding-Hen (approved):

NitroxPatcher/Patches/Dynamic/VehicleDockingBay_OnTriggerEnter.cs
```diff
@@ -21,7 +21,7 @@ public static bool Prefix(VehicleDockingBay __instance, Collider other)
         {
             Vehicle vehicle = other.GetComponentInParent<Vehicle>();
             prevInterpolatingVehicle = (Vehicle)__instance.ReflectionGet("interpolatingVehicle");
-            return NitroxServiceLocator.LocateService<SimulationOwnership>().HasAnyLockType(NitroxEntity.GetId(vehicle.gameObject));
+            return vehicle is null || NitroxServiceLocator.LocateService<SimulationOwnership>().HasAnyLockType(NitroxEntity.GetId(vehicle.gameObject));
```
Measurity, line 24: Should be `== null` instead.

`is null` is not compatible with Unity because Unity equality to do internal "object is alive" checks which `is null` skips. Leading to objects that are non-null but dead inside Unity.

See: https://blog.unity.com/technology/custom-operator-should-we-keep-it

Measurity (changes requested):

NitroxPatcher/Patches/Dynamic/VehicleDockingBay_OnTriggerEnter.cs
MrPurple6411: 6b3818b

PR #1591, opened 2021-09-08
Improving Transform logic on server

Jannify:


NitroxModel/Helper/Matrix4x4Extension.cs
```diff
@@ -10,7 +10,7 @@ public static Matrix4x4 Compose(NitroxVector3 localPosition, NitroxQuaternion lo
             Matrix4x4 translationMatrix = Matrix4x4.CreateTranslation(localPosition.X, localPosition.Y, localPosition.Z);
             Matrix4x4 rotationMatrix = Matrix4x4.CreateFromQuaternion((Quaternion)localRotation);
             Matrix4x4 scaleMatrix = Matrix4x4.CreateScale(localScale.X, localScale.Y, localScale.Z);
-            return rotationMatrix * scaleMatrix * translationMatrix;
+            return scaleMatrix *rotationMatrix *  translationMatrix;
```
killzoms, line 13: add space before ``rotationMatrix``

killzoms (changes requested):

NitroxModel/Helper/Matrix4x4Extension.cs
Coding-Hen: I assume here the change in order is due to arithmetic operations. If so should we not encapsulate the first calculation with brackets to ensure correct ordering when multiplication is performed?

killzoms: well technically the first calculation is already performed first(order of operations) its just when it comes to Matrix multiplication the order of the variables matters

killzoms: easy way to write it

Matrix ***A*** * Matrix ***B*** = Matrix ***C***
Matrix ***B*** * Matrix ***A*** = Matrix ***D***

Matrix ***D*** != Matrix ***C***

PR #1597, opened 2021-09-15
Please add simplified Chinese (I have compiled it)

2293745032:
I have compiled a simplified Chinese translation file.Please add it to the module file
Translation preview here:(in Simplified Chinese)
xg
效果

Here is the file of the language:
[Chinese (Simplified).zip](https://github.com/SubnauticaNitrox/Nitrox/files/7169553/Chinese.Simplified.zip)


Coding-Hen (changes requested):
Glad you got it working. You seem to have changed many more files than just adding the Chinese language file. Might be easier to just start a fresh pr off master then add the language file on its own as this has many conflicts and unrelated changes.

Measurity:
Make sure you don't have diverging branches with same work already on master.

Fix it and create a new PR, please.

PR #1609, opened 2021-09-18
Refactored platforms

Measurity:
This PR adds a lot of Windows native structs to better inspect processes which we can use to detect piracy later or extend modding capabilities of Nitrox by hooking native functions.

 - Fixed switching to different subnautica installation not being used for the NitroxEntryPatch logic (would require launcher restart before).
 - Fixed steam process getting closed if Nitrox Launcher closes. (it captures the steam process instead if steam wasn't started before starting game).
 - Passed Steam app ID to game process so that Steam does not need to restart the game. We can then pass the nitrox launcher path as environment variable to the game process and it will stick.

TODO:
 - [x] Implement EGS platform support just like Steam is now
 - [x] Implement Discord platform support (confirmed working by Cookay)

MS Store is broken anyway and hopefully changes soon to be less finicky to mod so I left it unimplemented (throws errors if users try to).


NitroxModel/Discovery/InstallationFinders/DiscordGameFinder.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.IO;
+
+namespace NitroxModel.Discovery.InstallationFinders
+{
+    public class DiscordGameFinder : IFindGameInstallation
+    {
+        /// <summary>
+        ///     Subnautica Discord is either in appdata or in C:. So for now we just check these 2 paths until we have a better way.
+        /// </summary>
+        /// <returns></returns>
```
dartasen, line 12:
```suggestion

```

NitroxModel/Discovery/InstallationFinders/DiscordGameFinder.cs
```diff
@@ -0,0 +1,34 @@
...
+        public string FindGame(IList<string> errors = null)
+        {
+            bool HasGameInDir(string path)
+            {
+                return File.Exists(Path.Combine(path, GameInfo.Subnautica.ExeName));
+            }
```
dartasen, line 18: Maybe extract it to a proper private method ?
Since it's not accessing any sort of data from the function itself, I don't see nested func being useful here

NitroxModel/Discovery/InstallationFinders/DiscordGameFinder.cs
```diff
@@ -0,0 +1,34 @@
...
+
+            string path = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), "DiscordGames", "Subnautica", "content");
+            if (HasGameInDir(path))
+            {
+                return path;
+            }
+            path = @"C:\Games\Subnautica\content";
```
dartasen, line 25: Maybe leave a comment to explain where it's coming from, is it something standard ?

NitroxModel/Platforms/Store/Egs.cs
```diff
@@ -0,0 +1,45 @@
...
+            throw new NotImplementedException();
+        }
+
+        public async Task<ProcessEx> StartGameAsync(string pathToGameExe)
+        {
+            // Normally should call StartPlatformAsync first. But Subnautica will start without EGS.
+            return await Task.FromResult(ProcessEx.Start(pathToGameExe,
+                                                         new[] { ("NITROX_LAUNCHER_PATH", Path.GetDirectoryName(Assembly.GetEntryAssembly().Location)) },
+                                                         Path.GetDirectoryName(pathToGameExe),
+                                                         "-EpicPortal -vrmode"));
```
dartasen, line 42:
```suggestion
                                                         "-EpicPortal -vrmode none"));
```
I think we should leave game arguments as parameters, since we might be interested in adding vrmode options inside the launcher or provide a custom launch argument ?

dartasen (commented):
Despite the fact that native stuff are almost impossible to review, PR looks good

PR #1614, opened 2021-10-04
Constructable mp inventory 'sharing' fix

codefaux:
Bug: When building, if player (Builder) attempts to construct an object but does not have all required items in inventory, other players' inventory will have said item deleted.

Cause: Constructable class, Construct function - run on all clients regardless of wether they're building or not.

Proposed fix: It's a hack, but it works - simply check if the client is targetting an object, if that object is constructable, and if they're holding the construct buttons. Better fix would be to verify player is weilding BuilderTool, is targetting same constructable, is holding construct buttons -- but this is my first attempt at using Harmony, Visual Studio 2019, interfacing with the Unity engine, etc etc etc etc, so it's a hack....but to my surprise it works.

FLAWS: Currently the only flaw we've noticed to this fix is buildables don't seem to update their texture mid-construction for other players but I've tested this for a few hours with a few players and it seems to not introduce unexpected behaviors.

Will be testing if this cooperates with PR#1587 (https://github.com/SubnauticaNitrox/Nitrox/pull/1587) after stability is verified.

- This is my first pull request, be gentle lol

### DIFF
--- a/NitroxPatcher/Patches/Dynamic/Constructable_Construct_Patch.cs
+++ b/NitroxPatcher/Patches/Dynamic/Constructable_Construct_Patch.cs
@@ -22,10 +22,6 @@ namespace NitroxPatcher.Patches.Dynamic
                 NitroxServiceLocator.LocateService<Building>().ChangeConstructionAmount(__instance.gameObject, __instance.constructedAmount);
             }
 
-            // If we are constructing a base piece then we'll want to store all of the BaseGhost information
-            // as it will not be available when the construction hits 100%
-            BaseGhost baseGhost = __instance.gameObject.GetComponentInChildren<BaseGhost>();
-
             UnityEngine.GameObject gameObject;
             float dist;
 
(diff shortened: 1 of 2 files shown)


codefaux:
Also apparently as a bonus, you get my fix for world persistence A) not providing a stack trace and B) overwriting the previous world zip blindly. Please at least implement the zip date.

Measurity:
Thanks for making this PR. 

I'd rather not merge hacks into Nitrox if they have a more technically correct solution, because they can lead to more bugs that can get much harder to debug later on, especially if we start depending on this hack. You're free to use it of course.

As for the stack trace, most users won't know what to do with it and it likely won't be very useful either because it's corrupted save that will cause the error. I.e. it should not be printed to the console. If you could put the changes for datetime-zip in a separate PR I can merge it while we think about this PR some more.

codefaux:
Roger, fair and understood. Will revise.

While I have your attention - do we have a way to identify the player's held tool, by either reference or type?

I can piece the rest together myself, but I couldn't find that. (Is there documentation I've missed somewhere? Lol)

Measurity:
We have a sync for the player held item that should help you. Class is `QuickSlots_SelectInternal_Patch`.

codefaux:
Scratch this one - will submit proper PR without accidental patches

PR #1615, opened 2021-10-04
MP Building Inventory 'sharing' fixup

codefaux:
Second attempt, this time without extra accidental patches, and a proper non-hack solution (so far as I can see it)

Bug: When building, if player (Builder) attempts to construct an object but does not have all required items in inventory, other players' inventory will have said item deleted.

Cause: Constructable class, Construct function - run on all clients regardless of wether they're building or not.

Proposed fix: Proper checks throughout, to ensure player is holding the builder, using the tool, aiming at the right thing, etc etc.

FLAWS: Currently the only flaw we've noticed to this fix is buildables don't seem to update their texture mid-construction for other players but I've tested this with a few players and it seems to not introduce unexpected behaviors.

Will be testing if this cooperates with PR#1587 (#1587) after stability is verified.
(EDIT: Played all day w/ #1587 applied, no issues so far.)

This is my second pull request, be gentle lol

### DIFF
--- a/NitroxPatcher/Patches/Dynamic/Constructable_Construct_Patch.cs
+++ b/NitroxPatcher/Patches/Dynamic/Constructable_Construct_Patch.cs
@@ -25,6 +25,21 @@ namespace NitroxPatcher.Patches.Dynamic
             UnityEngine.GameObject gameObject;
             float dist;
 
+            // If we are constructing a base piece then we'll want to store all of the BaseGhost information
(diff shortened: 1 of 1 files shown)


Measurity:
If you update this PR from changes on master you can access private fields directly and no longer need reflection (because DLLs are publicized now).

Make sure to run the Nitrox.BuildTool first and then rebuild the solution.

dartasen:
Closed for inactivity

PR #1626, opened 2021-10-15
Cleans publicised assemblies on clean of project

Coding-Hen:
Cleans the bin folder and removes the publicised assemblies meaning manual removal is not needed. Especially when switching between games


Measurity:
I noticed that when I rebuild the solution it will also run the "clean" targets of the projects so it will always fail to build the solution correctly. I think it's possible to only clean if you select only the Nitrox.BuildTool for cleaning by adding a condition that checks which projects are requested to build.

PR #1627, opened 2021-10-15
Below zero

Coding-Hen:
Initial works and by no means complete, includes lots of changes that will probably be in the upcoming giant update.


AntonIXO:
Did you try to build it? Why all changes comented?

Jannify:
We probably should make the same split in client as in `NitroxServer` and `NitroxServer-Subnautica`.

Coding-Hen:
Going to close this for now just from a point of view of the drastic changes we need to make, will probably reopen after the modpocalypse as code bases should be closer together

Issue #1602, opened 2021-09-17
Scanner room not syncing


lucafont2:
### Version
Alpha 1.5.0.1

### Bug
In the scanner room, the status of the modules and whether the search for an item is in progress is on, is not synchronized between the players.

### Other details:
[LOGFILE](https://cdn.lucaservers.com/SUB/09-17-02.log)
[LOGFILE_PRE](https://cdn.lucaservers.com/SUB/09-17-02-prev.log)

Measurity:
Duplicate of #1425

PR #1644, opened 2021-11-07
Scanner room sync

tornac1234:
Fixes [Issue#1602](https://github.com/SubnauticaNitrox/Nitrox/issues/1602)

The goal is to sync every action of the scanner room (also called Map Room)
The first commit includes these changes:
- Camera's movement sync (with a lock system on every MapRoomScreen: the screens to take control over a camera and over each camera)
- Scanner room's states are now synced and saved (the docking points states are synced and also the object being scanned with the map scanner)
- Some stuff around Metadata saving/updating/processing has been modified/fixed
- There's a try of making a "Packet Saver" in the PlayerMovement, the objective is to not send movement packets when the player is not moving anymore but it lacks some tests and tweaks to work

### DIFF
--- a/NitroxClient/GameLogic/Bases/Metadata/MapRoomMetadataProcessor.cs
+++ b/NitroxClient/GameLogic/Bases/Metadata/MapRoomMetadataProcessor.cs
@@ -1,0 +1,80 @@
+﻿using NitroxClient.MonoBehaviours;
+using NitroxClient.Unity.Helper;
+using NitroxModel.DataStructures;
+using NitroxModel.DataStructures.GameLogic.Buildings.Metadata;
+using NitroxModel.Logger;
+using NitroxModel_Subnautica.DataStructures;
+using UnityEngine;
+
+namespace NitroxClient.GameLogic.Bases.Metadata
+{
+    public class MapRoomMetadataProcessor : GenericBasePieceMetadataProcessor<MapRoomMetadata>
+    {
+        public override void UpdateMetadata(NitroxId id, MapRoomMetadata metadata, bool initialSync)
+        {
+            if (!initialSync)
+            {
+                // Log.Debug("UpdateMetadata() but not initial sync");
+                return;
+            }
+            // Log.Debug($"UpdateMetada({id}, {metadata})");
+
+            // MRF_GO is the MapRoomFunctionality's GameObject which has the NitroxId provided in the metadata
+            GameObject MRF_GO;
+            if (!NitroxEntity.TryGetObjectFrom(metadata.MapRoomFunctionalityId, out MRF_GO))
+            {
+                Log.Error($"There was an error while processing MapRoomMetadata for [NitroxId: {metadata.MapRoomFunctionalityId}], the corresponding game object couldn't be found");
+                return;
+            }
+            
+            MapRoomCameraDocking cameraDocking1 = MRF_GO.FindChild("dockingPoint1").GetComponent<MapRoomCameraDocking>();
+            MapRoomCameraDocking cameraDocking2 = MRF_GO.FindChild("dockingPoint2").GetComponent<MapRoomCameraDocking>();
+            if (cameraDocking1.AliveOrNull() != null && cameraDocking2.AliveOrNull() != null)
+            {
+                Log.Debug($"Processing CameraDockingMetadata [{metadata}]");
+                // Simulate Undock but don't do it else Postfix will happen
+                if (!metadata.CameraDocked1)
+                {
+                    if (cameraDocking1.camera != null)
+                    {
+                        NitroxEntity.RemoveFrom(cameraDocking1.camera.gameObject);
+                        UnityEngine.Object.Destroy(cameraDocking1.camera.gameObject);
+                        cameraDocking1.camera = null;
+                        cameraDocking1.cameraDocked = false;
+                    }
+                }
+                else if (metadata.Camera1NitroxId != null)
+                {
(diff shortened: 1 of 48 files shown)


NitroxPatcher/Patches/Dynamic/Player_EnterLockedMode_Patch.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+    public class Player_EnterLockedMode_Patch : NitroxPatch, IDynamicPatch
+    {
+        private static readonly MethodInfo TARGET_METHOD = Reflect.Method((Player t) => t.EnterLockedMode(default(Transform), default(bool)));
+        private static LocalPlayer localPlayer;
+
+        public static void Postfix(Transform parent, bool teleport)
+        {
+            localPlayer ??= Resolve<LocalPlayer>();
+            if (parent == null && teleport == false)
```
Jannify, line 18:
```suggestion
            if (!parent && !teleport)
```

NitroxModel/Packets/BasePieceMetadataChanged.cs
```diff
@@ -7,11 +7,13 @@ namespace NitroxModel.Packets
...
     public class BasePieceMetadataChanged : Packet
     {
+        public NitroxId BaseParentId { get; }
         public NitroxId PieceId { get; }
         public BasePieceMetadata Metadata { get; }
 
-        public BasePieceMetadataChanged(NitroxId pieceId, BasePieceMetadata metadata)
+        public BasePieceMetadataChanged(NitroxId baseParentId, NitroxId pieceId, BasePieceMetadata metadata)
         {
+            BaseParentId = baseParentId;
```
Jannify, line 16: Is this really required? Can't you just use the parent or search for BasePiece where this id is needed?

NitroxServer/Communication/Packets/Processors/CameraMovementPacketProcessor.cs
```diff
@@ -0,0 +1,38 @@
...
+            Optional<Entity> entity = entityManager.GetEntityById(packet.CameraMovementData.Id);
+            if (entity.HasValue)
+            {
+                entity.Value.Transform.Position = packet.Position;
+                entity.Value.Transform.Rotation = packet.BodyRotation;
+            }
+
+            if (player.Id == packet.PlayerId)
+            {
+                player.Position = packet.Position;
```
Jannify, line 32: This will probably cause the player to spawn on the camera position if the connection gets cut of or even disconnect.

NitroxServer/Communication/Packets/Processors/CameraMovementPacketProcessor.cs
```diff
@@ -0,0 +1,38 @@
+﻿using NitroxModel.DataStructures.GameLogic;
+using NitroxModel.DataStructures.Util;
+using NitroxModel.Packets;
+using NitroxServer.Communication.Packets.Processors.Abstract;
+using NitroxServer.GameLogic;
+using NitroxServer.GameLogic.Entities;
+
+namespace NitroxServer.Communication.Packets.Processors
+{
+    class CameraMovementPacketProcessor : AuthenticatedPacketProcessor<CameraMovement>
```
Jannify, line 10: Here in NitroxServer a few public are missing

NitroxServer/Communication/Packets/Processors/DockingStateChangeProcessor.cs
```diff
@@ -0,0 +1,21 @@
...
+        public DockingStateChangeProcessor(PlayerManager playerManager)
+        {
+            this.playerManager = playerManager;
+        }
+
+        public override void Process(DockingStateChange packet, Player player)
+        {
+            playerManager.SendPacketToOtherPlayers(packet, player);
+        }
+    }
```
Jannify, line 20: Is this really needed? Isn't this the default behavior of the server  if no processor is there? There a few instances of this in your files.

Jannify (changes requested):

NitroxClient/GameLogic/Bases/Metadata/MapRoomMetadataProcessor.cs
```diff
@@ -0,0 +1,80 @@
...
+using NitroxModel_Subnautica.DataStructures;
+using UnityEngine;
+
+namespace NitroxClient.GameLogic.Bases.Metadata
+{
+    public class MapRoomMetadataProcessor : GenericBasePieceMetadataProcessor<MapRoomMetadata>
+    {
+        public override void UpdateMetadata(NitroxId id, MapRoomMetadata metadata, bool initialSync)
+        {
+            if (!initialSync)
```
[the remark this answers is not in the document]
tornac1234: The players may receive update metadata packets during playtime but we don't want the local state of the maproom dockings to be updated after the initial sync so it's why I restricted it

NitroxModel/Packets/BasePieceMetadataChanged.cs
tornac1234: The client processor for basepiecemetadata still needs the piece id

Jannify: I mean the `baseParentId`, sry. Is this Id needed in many Metadata processor or only a few?

tornac1234: baseParentId is necessary for the server's processor in the BaseManager

dartasen:
Inactive, i'll close it for now

PR #1651, opened 2021-11-15
Fixes the issue of entities not respawning at each server restart

tornac1234:
Local position should never be changed


killzoms:
this change is not the correct approach for this, it is likely better to dig deeper as the position should be updated for moving objects

Measurity (approved):

Issue #1457, opened 2021-05-10
[Server] Add joining queue


Jannify:
When multiple users are joining at the same time inventory items can desync. 
To prevent that and future issues the server should establish a joining queue so that `InitialSyncProcessors` of different players are not executed at the same time.
Also the server should "kick" the player if he disconnects during that process.

TwinBuilderOne:
Can I be listed as the assignee for this?

PR #1655, opened 2021-11-25
Join Queue

TwinBuilderOne:
If a player attempts to join the server while another player is joining or loading into the game, they will be added to a queue and can complete the initial connection once the first player has finished.


TwinBuilderOne:
There is currently a bug where the server does not receive the packet that the client uses to let the server know when it is finished loading in. I've tried everything I can think of to debug this several times and I have been unable to find the issue. I doubt my WiFi is the problem, but at this point, that's becoming more and more likely. I would greatly appreciate it if someone else tested this themselves.

NitroxServer/Serialization/ServerConfig.cs
```diff
@@ -66,6 +68,17 @@ public int MaxConnections
             }
         }
 
+        public int InitialSyncTimeout
+        {
+            get => initialSyncTimeoutSetting;
+
+            set
+            {
+                Validate.IsTrue(value > 0, "InitialSyncTimeout must be greater than 0");
```
dartasen, line 77: I think we shouldn't be allowing small values (cause 0ms of timeout is quite strange like it is for 50, ...)

NitroxClient/Communication/MultiplayerSession/MultiplayerSessionManager.cs
```diff
@@ -78,6 +78,13 @@ public void RequestSessionReservation(PlayerSettings playerSettings, Authenticat
 
         public void ProcessReservationResponsePacket(MultiplayerSessionReservation reservation)
         {
+            if (reservation.ReservationState == MultiplayerSessionReservationState.ENQUEUED_IN_JOIN_QUEUE)
+            {
+                // TODO: Add this message in other languages
+                Log.InGame("Waiting in join queue...");
```
dartasen, line 84:
```suggestion
              Log.InGame(Language.main.Get("Nitrox_Waiting"))
```
And you can simply add an entry `Nitrox_Waiting` to the english json in `Nitrox.Subnautica.Assets/LanguageFiles/English.json`

NitroxServer/Serialization/ServerConfig.cs
dartasen, line 77: I think we shouldn't be allowing small values (cause 0ms of timeout is quite strange like it is for 50, ...)

NitroxModel/MultiplayerSession/MultiplayerSessionReservationState.cs
```diff
@@ -22,6 +22,9 @@ public enum MultiplayerSessionReservationState
 
         [Description("The server is using hardcore gamemode, player is dead.")]
         HARDCORE_PLAYER_DEAD = 1 << 4,
+
+        [Description("Another user is currently joining the server.")]
```
dartasen, line 26: We can't join the server is someone is already joining ? This player shouldn't be just queued ?

NitroxServer/GameLogic/PlayerManager.cs
```diff
@@ -171,6 +202,31 @@ public void PlayerDisconnected(NitroxConnection connection)
             }
         }
 
+        public void FinishProcessingReservation()
+        {
+            PlayerCurrentlyJoining = false;
+
+            NitroxModel.Logger.Log.Info($"Finished processing reservation. Remaining requests: {JoinQueue.Count}");
```
dartasen, line 209:
```suggestion
            Log.Info($"Finished processing reservation. Remaining requests: {JoinQueue.Count}");
```

NitroxModel/MultiplayerSession/MultiplayerSessionReservationState.cs
TwinBuilderOne: This reservation state lets the client know that they have been enqueued in the join queue. They are then dequeued by another MultiplayerSessionReservation that lets them know if they have been reserved.

TwinBuilderOne:
Fixes #1457

NitroxServer/GameLogic/PlayerManager.cs
```diff
@@ -171,6 +222,27 @@ public void PlayerDisconnected(NitroxConnection connection)
...
+
+            if (JoinQueue.Count > 0)
+            {
+                var keyValuePair = JoinQueue.Dequeue();
+                NitroxConnection requestConnection = keyValuePair.Key;
+                MultiplayerSessionReservationRequest reservationRequest = keyValuePair.Value;
+                MultiplayerSessionReservation reservation = ReservePlayerContext(requestConnection,
+                    reservationRequest.PlayerSettings,
+                    reservationRequest.AuthenticationContext,
+                    reservationRequest.CorrelationId);
```
Jannify, line 240: A few things.

- I know you used var because of the length of the type but we don't use var if it's not 100% needed.
- To remove the clutter I added a new line between the "variable defining" part and the "data processing"
- Imo it's better if the parameter are align with the others and not just indent by 1 tab.

```suggestion
                KeyValuePair<NitroxConnection, MultiplayerSessionReservationRequest> keyValuePair = JoinQueue.Dequeue();
                NitroxConnection requestConnection = keyValuePair.Key;
                MultiplayerSessionReservationRequest reservationRequest = keyValuePair.Value;

                MultiplayerSessionReservation reservation = ReservePlayerContext(requestConnection,
                                                                                 reservationRequest.PlayerSettings,
                                                                                 reservationRequest.AuthenticationContext,
                                                                                 reservationRequest.CorrelationId);
```

NitroxServer/Serialization/ServerConfig.cs
```diff
@@ -9,6 +9,8 @@ namespace NitroxServer.Serialization
     public class ServerConfig : NitroxConfig<ServerConfig>
     {
         private int maxConnectionsSetting = 100;
+
+        private int initialSyncTimeoutSetting = 120000;
```
Jannify, line 13: I would double this value if not triple. We know we have a few players with shitty internet/PCs and there will be a bunch of them in the help channel "WHY NOT WORKING?????"

Jannify (changes requested):

NitroxServer/GameLogic/PlayerManager.cs
```diff
@@ -143,6 +181,19 @@ public Player PlayerConnected(NitroxConnection connection, string reservationKey
...
+            JoinQueue.Clear();
+            while (copy.Any())
+            {
+                var item = copy.Dequeue();
+
+                if (item.Key != connection)
+                {
+                    JoinQueue.Enqueue(item);
+                }
+            }
```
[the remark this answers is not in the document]
TwinBuilderOne: Yeah, that would be better. I think I was under the impression that JoinQueue was readonly at the time.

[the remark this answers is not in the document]
Jannify: It is but you can change that, can't you?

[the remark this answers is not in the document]
Jannify: Btw you can make the two variables at the start of the class private iirc.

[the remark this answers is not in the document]
TwinBuilderOne: The problem with that is JoinQueue is read-only.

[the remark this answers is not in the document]
Jannify: What about adding the `set;` in L26 and using this: `JoinQueue = (ThreadSafeQueue<KeyValuePair<NitroxConnection, MultiplayerSessionReservationRequest>>)JoinQueue.Where(item => !Equals(item.Key, connection));`?

TwinBuilderOne:
Ok, the changes have been made.

Measurity:
Please add unit tests for thread safe queue

Measurity (approved):
Looks good, see previous comment. Will merge after that.

PR #1656, opened 2021-11-27
Minor launcher fixes

Measurity:


NitroxModel/Platforms/Store/Steam.cs
```diff
@@ -75,9 +74,10 @@ public async Task<ProcessEx> StartGameAsync(string pathToGameExe, int steamAppId
             }
 
             return ProcessEx.Start(pathToGameExe,
-                                   new[] { ("SteamGameId", steamAppId.ToString()), ("SteamAppID", steamAppId.ToString()), (NitroxUser.LAUNCHER_PATH_ENV_KEY, NitroxUser.LauncherPath), },
-                                   Path.GetDirectoryName(pathToGameExe)
+                                   new[] { ("SteamGameId", steamAppId.ToString()), ("SteamAppID", steamAppId.ToString()), (NitroxUser.LAUNCHER_PATH_ENV_KEY, NitroxUser.LauncherPath) },
+                                   Path.GetDirectoryName(pathToGameExe),
+                                   "-vrmode none"
```
dartasen, line 79: Maybe we should add this as a default parameter of StartPlatform or something similar ? So we can override those settings later from the options page

dartasen (approved):

Issue #1471, opened 2021-05-13
Rocket stages can easily desync


dartasen:
### Bug
Server doesn't check for invalid stages order, for whatever reason if a player desync from others, they can send invalid data to the server that will be persisted. So it's possible to rollback some stages of the rocket on reload.

dartasen:
Partially fix with #1503

PR #1658, opened 2021-11-29
Fix rocket stages and preflight checks bugs

tornac1234:
- Check for null PreflightChecks
- Prevent a client from sending back the event of completing a preflight check

### DIFF
--- a/NitroxClient/Communication/Packets/Processors/RocketPreflightCompleteProcessor.cs
+++ b/NitroxClient/Communication/Packets/Processors/RocketPreflightCompleteProcessor.cs
@@ -1,7 +1,6 @@
 ﻿using NitroxClient.Communication.Abstract;
 using NitroxClient.Communication.Packets.Processors.Abstract;
 using NitroxClient.MonoBehaviours;
-using NitroxModel.Core;
 using NitroxModel_Subnautica.Packets;
 using UnityEngine;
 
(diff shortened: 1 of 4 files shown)


tornac1234:
Does it finally fix [Issue#1471](https://github.com/SubnauticaNitrox/Nitrox/issues/1471) ?

NitroxClient/Communication/Packets/Processors/RocketPreflightCompleteProcessor.cs
```diff
@@ -59,5 +62,14 @@ public override void Process(RocketPreflightComplete packet)
             }
 
         }
+
+        private void CompletePreflightCheck(PreflightCheckSwitch preflightCheckSwitch)
+        {
+            packetSender ??= NitroxServiceLocator.LocateService<IPacketSender>();
```
Jannify, line 68: You can get assign the packetSender in a constructor in this class.

NitroxClient/Communication/Packets/Processors/RocketPreflightCompleteProcessor.cs
```diff
@@ -1,12 +1,15 @@
...
 using NitroxClient.MonoBehaviours;
+using NitroxModel.Core;
 using NitroxModel_Subnautica.Packets;
 using UnityEngine;
 
 namespace NitroxClient.Communication.Packets.Processors
 {
     public class RocketPreflightCompleteProcessor : ClientPacketProcessor<RocketPreflightComplete>
     {
+        private IPacketSender packetSender;
```
Jannify, line 12: With the construtor:
```suggestion
        private readonly IPacketSender packetSender;
        
```

Jannify (changes requested):

Measurity:
Is there a reason we don't just instantiate the PreflightChecks so it's never null? 
Example:
```cs
        [ProtoMember(3)]
        public ThreadSafeList<PreflightCheck> PreflightChecks { get; set; } = new();
```

Measurity (approved):
I don't know much about the rocket stages but the parts I did test worked.

dartasen (approved):
I'll need to rewrite those processors to use metadata system but for now, it looks good

PR #1659, opened 2021-11-30
Fixes the fast flashlight battery discharge

Shirkie01:
This changes fixes the fast flashlight battery discharge.

### DIFF
--- a/NitroxPatcher/Patches/Dynamic/EnergyMixin_ModifyCharge_Patch.cs
+++ b/NitroxPatcher/Patches/Dynamic/EnergyMixin_ModifyCharge_Patch.cs
@@ -26,7 +26,7 @@ namespace NitroxPatcher.Patches.Dynamic
                     NitroxId instanceId = NitroxEntity.GetId(__instance.gameObject);
                     ItemData batteryData = new(instanceId, NitroxEntity.GetId(battery), SerializationHelper.GetBytes(battery));
 
-                    NitroxServiceLocator.LocateService<StorageSlots>().EnergyMixinValueChanged(instanceId, UnityEngine.Mathf.Floor(__instance.charge), batteryData);
+                    NitroxServiceLocator.LocateService<StorageSlots>().EnergyMixinValueChanged(instanceId, __instance.charge, batteryData);
                 }
             }
         }


Coding-Hen:
Have you tested this? I cant verify this as a fix to the issue as it still seems present

Jannify:
I am fairly certain that the problem our player base is reporting is a different one and caused by simulating energy drain on each client individually (if owner of the item/vehicle or not) and the latency (which you need to enable in LiteNetLib setting to get a artificial one).

Shirkie01:
When I start two clients locally without this change, both clients discharge at the faster rate . When I have more time I'll try increasing the latency.

Jannify:
> When I start two clients locally without this change, both clients discharge at the faster rate . When I have more time I'll try increasing the latency.

I think you are right. The charge send with the packet is rounded down and therefore up to 0.9 of the charge is lost.

Shirkie01:
I think I figured out what's happening.

The intended design here is clearly " Only send the rounded number when the number ticks over to the next integer". However, on the next tick after being set, assume the client's charge is now 82.96, and the result value is -0.06. 

`Math.Abs(Math.Floor(82.96) - Math.Floor(82.96 - -0.06)) > 0`
`Abs(82 - 83) > 0`
`1 > 0`

This triggers the Math.Floor for the 82.96, resulting in the `0.96` of charge being lost, and then possibly repeating depending on the size of the `result` value. Accepting this change means that the charge isn't lost, though there may be a change that would be "more" correct by changing the formula, or maybe by doing `Math.Floor(instance.charge - result)` instead.

Jannify:
> I think I figured out what's happening.
> 
> The intended design here is clearly " Only update when the charge goes to a new integer." Since the update is only supposed to be sent when changing to a new integer, it makes sense to have the integer be rounded. However, on the next tick after being set, assume the client's charge is now 82.96, and the result value is -0.06. 
> 
> `Math.Abs(Math.Floor(82.96) - Math.Floor(82.96 - -0.06)) > 0`
> `Abs(82 - 83) > 0`
> `1 > 0`
> 
> This triggers the Math.Floor for the 82.96, resulting in the `0.96` of charge being lost, and then possibly repeating depending on the size of the `result` value. Accepting this change means that the charge isn't lost, though there may be a change that would be "more" correct by changing the formula, or maybe by doing `Math.Floor(instance.charge - result)` instead.

I think what you did is still the solution to fix it. We should send the real value to other clients and not the floored one. The rounding is only to trigger whole number steps to limit packet rate. The trigger can probably be optimised but should keep it's usage.

Coding-Hen (approved):

PR #1663, opened 2021-12-06
Refactored UPnP and simplified IP messages

Measurity:
Port forward rule will be created when server starts and removed when server exits. 
Reasoning: Some routers do not support a lease time and if a server is long running then it could exceed the lease time.

TODO:
 - [x] ~~[Write NuGet dependency to be used by Nitrox for port forwarding with either UPnP, NAT-PmP or PCP](https://github.com/Measurity/NAT)~~
 - [x] Remove UPnP class and use Mono.NAT internally to speed up this PR.

image


Measurity:
Clement tested this feature and it's not working for NAT-PmP setups. I need to change this to still use mono.nat.

dartasen (approved):
Working perfectly for me

EDIT: Is throwing an error while closing a server if a manual NAT rule is already active

dartasen (approved):

Issue #1653, opened 2021-11-22
[QoL] Overhaul Server Menu


killzoms:
- [ ] Add LAN Server Discovery

Bricklou:
Since the menu need to be enhanced, let's take *some* inspiration from Minecraft

![image](https://user-images.githubusercontent.com/15181236/142941172-10565d44-747f-4349-a93d-34e98b03c8ab.png)

here the legens of the picture above:
* WHITE: the local network server scan
* RED: button to add a new server
* CYAN: server ping, max players count and connected players count (with an option to disable it server side to prevent spamming)
* YELLOW: edit and delete server buttons
* GREEN: direct connection -> connect to a server without adding it to the server list

PR #1671, opened 2021-12-13
LAN server discovery

TwinBuilderOne:
When the server list menu is open, the client will automatically scan for servers hosted on devices on the same network and display them in the menu.
screenshot of menu showing a LAN server in the server list


killzoms:
@TwinBuilderOne while at it want to extend some of the features of the server browser? #1653

TwinBuilderOne:
Sure, I'd be open to that. Although it could be a little tricky designing a new GUI since I don't have the project set up in Unity.

Jannify:
If you do it it would be better in a new PR. Makes reviewing easier.

NitroxClient/Communication/LANDiscoveryClient.cs
```diff
@@ -0,0 +1,83 @@
...
+
+                client.SendDiscoveryRequest(writer, LANDiscoveryConstants.BROADCAST_PORT);
+            });
+            broadcastTimer.Change(0, 5000);
+
+            pollTimer = new Timer((state) =>
+            {
+                client.PollEvents();
+            });
+            pollTimer.Change(0, 15);
```
Jannify, line 60: Would increase this to something like 100ms bc it isn't that important to poll frequently.

NitroxClient/Communication/LANDiscoveryClient.cs
```diff
@@ -0,0 +1,83 @@
...
+
+        private static Action<IPEndPoint> foundServerCallback;
+
+        private static EventBasedNetListener listener;
+        private static NetManager client;
+
+        private static Timer broadcastTimer;
+        private static Timer pollTimer;
+
+        public static void SearchForServers(Action<IPEndPoint> callback)
```
Jannify, line 24: You never stop the discovery client, right? At least after joining LAN-Discovery should be disabled.

NitroxClient/MonoBehaviours/Gui/MainMenu/MainMenuMultiplayerPanel.cs
```diff
@@ -133,6 +135,17 @@ private void LoadSavedServers()
             }
         }
 
+        private void FindLANServers()
+        {
+            LANDiscoveryClient.SearchForServers(LANDiscoveryCallback);
```
Jannify, line 140:
```suggestion
            LANDiscoveryClient.SearchForServers(serverEndPoint =>
            {
                CreateServerButton($"{Language.main.Get("Nitrox_ConnectTo")} <b>LAN Server</b>\n{(NitroxPrefs.HideIp.Value ? "****" : serverEndPoint.Address)}:{(NitroxPrefs.HideIp.Value ? "****" : serverEndPoint.Port)}", serverEndPoint.Address.ToString(), serverEndPoint.Port.ToString());
            });
```

NitroxServer/Communication/LiteNetLib/LiteNetLibServer.cs
```diff
@@ -47,13 +47,15 @@ public override bool Start()
                 BeginPortForward(portNumber);
             }
 #endif
+            LANDiscoveryServer.Start();
```
Jannify, line 50: The `LANDiscoveryServer` should be an feature which can be disabled with `NitroxServer.Serialization.ServerConfig`. In big networks (like school) this feature is maybe not desired.

Jannify (changes requested):

NitroxClient/Communication/LANDiscoveryClient.cs
Measurity: I recommend turning this into an async task function with cancellation token parameter. Cancellation tokens can have a timeout.

NitroxClient/NitroxClient.csproj
```diff
@@ -374,6 +375,10 @@
       <Project>{0a377218-6b36-4522-89a3-a39cfc999209}</Project>
       <Name>NitroxModel-Subnautica</Name>
     </ProjectReference>
+    <ProjectReference Include="..\NitroxServer\NitroxServer.csproj">
+      <Project>{0FC864B0-694E-4FCA-B78C-8EF98BC6F262}</Project>
+      <Name>NitroxServer</Name>
+    </ProjectReference>
```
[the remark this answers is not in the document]
TwinBuilderOne: How did that happen lol

NitroxClient/Communication/LANDiscoveryClient.cs
TwinBuilderOne: Why would using an async function have an advantage here? Couldn't I just make a ServerFound event and fire it whenever a server is detected?

Measurity: It would be if you didn't use LiteNetLib to starts a new server but instead used UDP broadcast. LiteNetLib does not have async/await support (only event based).

NitroxServer/Communication/LANDiscoveryServer.cs
```diff
@@ -0,0 +1,65 @@
...
+            server.DiscoveryEnabled = true;
+            server.UnconnectedMessagesEnabled = true;
+
+            for (int i = 0; i < LANDiscoveryConstants.BROADCAST_PORTS.Length; i++)
+            {
+                if (server.Start(LANDiscoveryConstants.BROADCAST_PORTS[i]))
+                {
+                    break;
+                }
+            }
```
Jannify, line 31: Should we break at the first one started? I can imagine a "very professional" user network where client and server start their NetManager at different ports. Would it be possible to start the 5 servers and would it impact performance?

TwinBuilderOne: Good point. Luckily I don't think starting multiple NetManagers is necessary because the client can send a discovery request to all of the ports instead of just the one it is listening on.

Jannify (commented):
Code wise seems fine. Can't do "in-game" testing as I don't have a laptop right now.

Coding-Hen (approved):
Tested ingame between multiple computers on the network and works without issue

PR #1678, opened 2021-12-19
Added language versioning

dartasen:
Next step is to bring this language system to something like crowdin so we can easily manage translations.

- [x] Update existing translations
- [x] Fix some typos 
- [x] Reformat json files
- [x] Fix components that doesn't get live update (add buttons, options)
- [x] Add unit tests to check sanity of language files

image


Measurity:
Can you tell me why we need this version flag? If the language key does not exist then it should just default to English language for that key.

dartasen:
> Can you tell me why we need this version flag? If the language key does not exist then it should default to English language for that key.

We do default values to english ones, version is here to keep track of what's going on in the JSON and will be used in app like crowding.
Right now, it's pretty much useless since I bumped every languages to the latest version (as I could) but will not be the case anytime and those will need to be checked by native speakers though

Measurity:
> We do default values to english ones, version is here to keep track of what's going on in the JSON and will be used in app like crowding.

Never heard of crowding app. If it's useful for automation reasons then it's fine. But if we can have the same thing by writing a test for missing language keys then that could be a better option imo.

Issue #1493, opened 2021-05-30
[Cyclops] Sonar is executed from players position


Jannify:
The cyclops sonar is executed from the position of every other player even if outside the cyclops. 
![grafik](https://user-images.githubusercontent.com/23176718/120111517-79461280-c172-11eb-80dc-1b38d38de141.png)

Issue #1465, opened 2021-05-11
[Player] Spawning far away when relogging


dartasen:

Measurity:
Afaik we assume this is physics pushing the player away because it presumably spawns partly inside the escape pod. Might be enough to set player rigidbody to isKinematic/freeze while world is loading in. Correct me if I'm wrong.

dartasen:
> 
> 
> Afaik we assume this is physics pushing the player away because it presumably spawns partly inside the escape pod. Might be enough to set player rigidbody to isKinematic/freeze while world is loading in. Correct me if I'm wrong.

Can be the first we might need to look into indeed, but might be also related to weird vehicles position broadcasting (cyclops ?)

Jannify:
Fixed with #1679

PR #1679, opened 2021-12-19
Cyclops sync fix and improvements

tornac1234:
This PR will try to fix the existing bugs and maybe add some things that weren't synced yet

- [x] Fix cyclops sonar executing from everywhere [Issue#1493](https://github.com/SubnauticaNitrox/Nitrox/issues/1493)
- [x] Fix motor state not syncing
- [x] Fix vehicle loading errors about colors packets (silent errors in the logs but don't do anything wrong)
- [x] Fix motor state not restoring when spawning far away from the cyclops
- [x] Fix sonar button state not being synced
- [x] Fix spawning far away of the Cyclops when relogging (has to do with wrong calculations of the client position sent to the server)

### DIFF
--- a/Nitrox.Test/Patcher/Patches/CyclopsSonarButton_Update_PatchTest.cs
+++ b/Nitrox.Test/Patcher/Patches/CyclopsSonarButton_Update_PatchTest.cs
@@ -1,0 +1,52 @@
+﻿using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Linq;
+using System.Reflection.Emit;
+using HarmonyLib;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using NitroxModel.Helper;
+using NitroxPatcher.Patches.Dynamic;
+using NitroxTest.Patcher.Test;
+
+namespace Nitrox.Test.Patcher.Patches
+{
+    [TestClass]
+    public class CyclopsSonarButton_Update_PatchTest
+    {
+        [TestMethod]
+        public void Sanity()
+        {
+            List<CodeInstruction> instructions = PatchTestHelper.GenerateDummyInstructions(100);
+            instructions.Add(new CodeInstruction(CyclopsSonarButton_Update_Patch.INJECTION_OPCODE, CyclopsSonarButton_Update_Patch.INJECTION_OPERAND));
+            instructions.Add(new CodeInstruction(OpCodes.Callvirt, Reflect.Method((Player player) => player.GetMode())));
+            instructions.Add(new CodeInstruction(OpCodes.Brtrue));
+
+            IEnumerable<CodeInstruction> result = CyclopsSonarButton_Update_Patch.Transpiler(null, instructions);
+            Assert.AreEqual(instructions.Count + 2, result.Count());
+        }
+
+        [TestMethod]
+        public void InjectionSanity()
+        {
+            ReadOnlyCollection<CodeInstruction> beforeInstructions = PatchTestHelper.GetInstructionsFromMethod(CyclopsSonarButton_Update_Patch.TARGET_METHOD);
+            IEnumerable<CodeInstruction> result = CyclopsSonarButton_Update_Patch.Transpiler(CyclopsSonarButton_Update_Patch.TARGET_METHOD, beforeInstructions);
+
+            Assert.IsTrue(beforeInstructions.Count < result.Count());
+        }
+
+        [TestMethod]
+        public void CheckMethodValidity()
+        {
+            ReadOnlyCollection<CodeInstruction> instructions = PatchTestHelper.GetInstructionsFromMethod(CyclopsSonarButton_Update_Patch.TARGET_METHOD);
+            for (int i = 0; i < instructions.Count; i++)
+            {
+                CodeInstruction instruction = instructions[i];
+                if (instruction.opcode.Equals(CyclopsSonarButton_Update_Patch.INJECTION_OPCODE) && instruction.operand.Equals(CyclopsSonarButton_Update_Patch.INJECTION_OPERAND))
+                {
+                    CodeInstruction nextBr = instructions[i + 2];
+                    Assert.IsTrue(nextBr.opcode.Equals(OpCodes.Brtrue), "Looks like subnautica code has changed. Update jump offset!");
+                }
+            }
+        }
+    }
+}
--- a/Nitrox.Test/Patcher/Patches/CyclopsSonarButton_Update_PatchTest.cs
+++ b/Nitrox.Test/Patcher/Patches/CyclopsSonarButton_Update_PatchTest.cs
@@ -5,47 +5,45 @@ using System.Reflection.Emit;
 using HarmonyLib;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using NitroxModel.Helper;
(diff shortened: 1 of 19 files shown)


tornac1234:
Does the last commit fixes [Issue#1465](https://github.com/SubnauticaNitrox/Nitrox/issues/1465) ?

NitroxPatcher/Patches/Dynamic/CyclopsSonarButton_Update_Patch.cs
```diff
@@ -0,0 +1,82 @@
...
+using System.Reflection.Emit;
+using HarmonyLib;
+using NitroxClient.GameLogic;
+using NitroxClient.MonoBehaviours;
+using NitroxModel.DataStructures;
+using NitroxModel.Helper;
+
+namespace NitroxPatcher.Patches.Dynamic
+{
+    public class CyclopsSonarButton_Update_Patch : NitroxPatch, IDynamicPatch
```
Measurity, line 12: Do you know which logic is causing the sonar ping to not be visible when you exit the cyclops while another pilot is still driving?

What about adding a comment here? Something like this?
```suggestion
    /// <summary>
    /// Prevents sonar from turning off automatically when the player stops piloting the Cyclops.
    /// The sonar will stay on until the player leaves the vehicle and automatically turns on when they enter again (if sonar was on at that time). 
    /// </summary>
    public class CyclopsSonarButton_Update_Patch : NitroxPatch, IDynamicPatch
```

Measurity (approved):
Tested with multiple instances and couldn't find an issue.

NitroxPatcher/Patches/Dynamic/CyclopsSonarButton_SonarPing_Patch.cs
```diff
@@ -75,7 +77,7 @@ public static IEnumerable<CodeInstruction> Transpiler(MethodBase original, IEnum
                  */
                 if (instruction.opcode.Equals(OpCodes.Brtrue))
                 {
-                    if (instructionList[i - 1].opcode.Equals(OpCodes.Ldloc_1) && instructionList[i + 1].opcode.Equals(OpCodes.Ldarg_0))
+                    if (instructionList[i - 1].opcode.Equals(OpCodes.Call) && instructionList[i + 1].opcode.Equals(OpCodes.Ldarg_0))
```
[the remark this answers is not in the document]
tornac1234: As energy consumption sync seems weird for now, I don't think we should make changes about this. Also, when receiving the sonar ping packet, we should still simulate the energy loss as the cyclops did use the sonar ping for another client

NitroxPatcher/Patches/Dynamic/CyclopsSonarButton_Update_Patch.cs
```diff
@@ -0,0 +1,77 @@
...
+            CodeInstruction brInstruction = new(OpCodes.Brfalse);
+            for (int i = 0; i < codeInstructions.Count; i++)
+            {
+                CodeInstruction instruction = codeInstructions[i];
+
+                if (instruction.opcode.Equals(INJECTION_OPCODE) && instruction.operand.Equals(INJECTION_OPERAND))
+                {
+                    // The second line after the current instruction should be a Brtrue, we need its operand to have the same jump label for our brfalse
+                    CodeInstruction nextBr = codeInstructions[i + 2];
+                    Validate.IsTrue(nextBr.opcode.Equals(OpCodes.Brtrue), "Looks like subnautica code has changed. Update jump offset!");
```
Jannify, line 52: This (constant validation statements in general (like at the start of the transpiler)) should be checked in the test and not at runtime.

NitroxPatcher/Patches/Dynamic/CyclopsSonarButton_Update_Patch.cs
```diff
@@ -0,0 +1,77 @@
...
+            
+            /* Normally in the Update()
+             * if (Player.main.GetMode() == Player.Mode.Normal && this.sonarActive)
+		     * {
+		     *     this.TurnOffSonar();
+		     * }
+             * this part will be changed into:
+             * if (CyclopsSonarButton_Update_Patch.ShouldTurnOff() && Player.main.GetMode() == Player.Mode.Normal && this.sonarActive)
+             */
+            List<CodeInstruction> codeInstructions = new List<CodeInstruction>(instructions);
```
Jannify, line 41: Is there a reason you casting to List? Afaik you van iterate a IEnumerable with foreach.

Jannify (changes requested):

NitroxPatcher/Patches/Dynamic/CyclopsSonarButton_Update_Patch.cs
tornac1234: There's a moment I need to get the code instruction that's 2 lines further, that's why I don't use a foreach

NitroxPatcher/Patches/Dynamic/CyclopsSonarButton_Update_Patch.cs
```diff
@@ -16,8 +16,8 @@ public class CyclopsSonarButton_Update_Patch : NitroxPatch, IDynamicPatch
     {
         public static readonly MethodInfo TARGET_METHOD = Reflect.Method((CyclopsSonarButton t) => t.Update());
 
-        private static readonly OpCode INJECTION_OPCODE = OpCodes.Ldsfld;
-        private static readonly object INJECTION_OPERAND = Reflect.Field(() => Player.main);
+        public static readonly OpCode INJECTION_OPCODE = OpCodes.Ldsfld;
+        public static readonly object INJECTION_OPERAND = Reflect.Field(() => Player.main);
```
Measurity, line 20: `internal` is preferred. The Nitrox.Test project has exclusive access to all `internal` members for this reason.

Nitrox.Test/Patcher/Patches/CyclopsSonarButton_Update_PatchTest.cs
```diff
@@ -0,0 +1,52 @@
...
+using System.Linq;
+using System.Reflection.Emit;
+using HarmonyLib;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using NitroxModel.Helper;
+using NitroxPatcher.Patches.Dynamic;
+using NitroxTest.Patcher.Test;
+
+namespace Nitrox.Test.Patcher.Patches
+{
```
Jannify, line 12:
```suggestion
namespace NitroxPatcher.Patches.Dynamic;
```

NitroxPatcher/Patches/Dynamic/CyclopsSonarButton_Update_Patch.cs
```diff
@@ -0,0 +1,76 @@
...
+
+        public static void Prefix(CyclopsSonarButton __instance)
+        {
+            currentCyclopsId = NitroxEntity.GetId(__instance.subRoot.gameObject);
+        }
+
+        public static IEnumerable<CodeInstruction> Transpiler(MethodBase methodBase, IEnumerable<CodeInstruction> instructions)
+        {
+            Validate.NotNull(INJECTION_OPERAND);
+
```
Jannify, line 32:
```suggestion
```

Jannify (changes requested):

PR #1680, opened 2021-12-19
Fix server not removing join queue entries after the client disconnects

TwinBuilderOne:


Measurity (approved):

dartasen (approved):
Looks good to me, however is it possible to write some tests about a basic server queue simulation so we don't have those bad surprises afterwards ?

TwinBuilderOne:
It would probably be possible to use NSubstitute to get a substitute for NitroxConnection to use in the queue, but that could be tricky to get right.

dartasen (approved):

PR #1686, opened 2021-12-22
Automatically configure firewall

### LINKED ISSUE #1526: Ensure firewall is configured properly and try automate this process
- Add rules for `Subnautica.exe`, `NitroxServer.exe`.
- If Hamachi exists, set `Hamachi` network to be ignored by Windows Firewall:
1. Go to Firewall advanced security window (search for it through windows). 

2. Click properties on the right-side 

3. Do the following for all 3 profiles: Domain Profile, Private Profile and Public Profile (where applicable):  

- Click "Customize..." button of "protected network connections"  
- Untick "Hamachi"  
- Press OK, repeat for other profiles.

TwinBuilderOne:
When the server or client is started, the launcher will check if the correct firewall rules are in place and add them if necessary using [Windows Firewall Helper](https://www.nuget.org/packages/WindowsFirewallHelper/). Resolves #1526.


Measurity:
Thanks for working on this. Some feedback/questions:

 - I'd prefer if firewall rules were added when the launcher starts and letting it happen in the background (new Task.Run or using an already running background task). So that it isn't an extra delay before Nitrox can be started.
 - Did you test the code against Hamachi where it was blocked by Windows Firewall at first?
 - If the firewall rules do not exist yet you'll get an UnauthorizedException error. The error isn't user friendly and will give us trouble in help-channel on Discord. We should `try...catch` the error and rewrite the message (example):
> "Windows Firewall is blocking Nitrox. If you want to solve this issue automatically you can restart Nitrox as admin or add the firewall rules manually to allow Subnautica.exe and NitroxServer-Subnautica.exe"

TwinBuilderOne:
Do you know what's causing that error in the picture? It seems like the error results from attempting to create the rules, not from starting the server.

Measurity:
It happened after I removed all the nitrox related firewall rules and tried to start server through the launcher. If you don't get an error it's likely the task.run that hides the error.

NitroxLauncher/Models/Utils/WindowsHelper.cs
```diff
@@ -55,6 +64,100 @@ internal static void RestartAsAdmin()
...
+
+        private static void AddExclusiveFirewallRule(string name, string filePath)
+        {
+            if (!FirewallRuleExists(name))
+            {
+                AddFirewallRule(name, filePath);
+            }
+        }
+
+        private static bool FirewallRuleExists(string name) => FirewallManager.Instance.Rules.Any(rule => rule.FriendlyName == name);
```
Measurity, line 151: This check ignores the direction (inbound or outbound) so an existing outbound rule for Subnautica would mean that no inbound rule is created.

Measurity (changes requested):

Measurity (approved):

NitroxLauncher/Models/Utils/WindowsHelper.cs
```diff
@@ -55,6 +64,119 @@ internal static void RestartAsAdmin()
...
+                Log.Warn($"Tried to add firewall rule for program that does not exist: {ex.FileName}");
+            }
+            catch (UnauthorizedAccessException)
+            {
+                if (!Settings.Default.FirewallWarningShown)
+                {
+                    Settings.Default.FirewallWarningShown = true;
+                    Settings.Default.Save();
+                    MessageBox.Show("Try restarting the launcher as administrator or manually adding firewall rules for Nitrox programs. This warning won't be shown again.", "Error adding Windows Firewall rules", MessageBoxButton.OK, MessageBoxImage.Warning);
+                }
```
dartasen, line 86: Maybe we can remove this state setting by using LauncherNotifier.Warning(), will also allow us to redirect users to the upcoming troubleshoot page. And in this case without loosing any errors by hidding them

NitroxLauncher/Models/Utils/WindowsHelper.cs
```diff
@@ -55,6 +64,119 @@ internal static void RestartAsAdmin()
...
+        private static void CheckServerFirewallRules(FirewallDirection direction)
+        {
+            string serverPath = Path.Combine(Path.GetDirectoryName(Assembly.GetEntryAssembly().Location), ServerLogic.SERVER_EXECUTABLE);
+
+            AddExclusiveFirewallRule(Path.GetFileName(serverPath), serverPath, direction);
+        }
+
+        private static void CheckClientFirewallRules(FirewallDirection direction)
+        {
+            string clientPath = Path.Combine(LauncherLogic.Config.SubnauticaPath, "Subnautica.exe");
```
dartasen, line 106: We can replace the game exe by the constants defined in `NitroxModel.GameInfo.Subnautica.ExeName`

NitroxLauncher/Models/Utils/WindowsHelper.cs
```diff
@@ -55,6 +64,119 @@ internal static void RestartAsAdmin()
...
+                                return addressInfo.Address;
+                            }
+                        }
+                    }
+                }
+
+                return null;
+            }
+
+            const string NAME = "HamachiNitrox";
```
dartasen, line 132: It's better to put all constants and variables in the top of a file, and maybe give it a more explicit name, since NAME isn't really clear on what is it used for

NitroxLauncher/Models/Utils/WindowsHelper.cs
```diff
@@ -55,6 +64,119 @@ internal static void RestartAsAdmin()
...
+            };
+
+            FirewallManager.Instance.Rules.Add(rule);
+        }
+
+        private static void AddExclusiveFirewallRule(string name, string filePath, FirewallDirection direction)
+        {
+            if (!File.Exists(filePath))
+            {
+                throw new FileNotFoundException("Unable to add firewall rule to non-existent program", filePath);
```
dartasen, line 162: Do we want to show a popup of this error, since this error will end up being catch up by the error handler of WPF resulting in a MessageBox ?

Maybe we should just log this error and/or putting a little toaster to warn client ?

NitroxLauncher/ServerLogic.cs
```diff
@@ -3,6 +3,7 @@
 using System.IO;
 using System.Reflection;
 using NitroxLauncher.Models.Events;
+using NitroxLauncher.Models.Utils;
```
dartasen, line 6: minor: I think this is not used

dartasen (changes requested):

NitroxLauncher/Models/Utils/WindowsHelper.cs
dartasen: Yeah maybe we shouldn't be checking firewall rules at any startup since some firewalls don't need rules like this to work properly with Nitrox so we would end up adding useless rules. Adding an optional way of checking through the troubleshoot page seems fair

NitroxLauncher/Models/Utils/WindowsHelper.cs
Measurity: Showing a toast each time isn't ideal either since you need admin for this to not error out. Once we have the proper validation page in the launcher we can notify users there.

NitroxLauncher/Models/Utils/WindowsHelper.cs
Measurity: This error can happen when the subnautica path isn't defined. Since this firewall change is secondary issue, since you won't be able to use Nitrox, I don't see it necessary and could confuse the user unless we really explain it well.

NitroxLauncher/Models/Utils/WindowsHelper.cs
```diff
@@ -55,6 +67,117 @@ internal static void RestartAsAdmin()
...
+            CheckClientFirewallRules(direction);
+            CheckServerFirewallRules(direction);
+            CheckHamachiFirewallRules(direction);
+        }
+
+        private static void CheckServerFirewallRules(FirewallDirection direction)
+        {
+            string serverPath = Path.Combine(Path.GetDirectoryName(Assembly.GetEntryAssembly().Location), ServerLogic.SERVER_EXECUTABLE);
+
+            AddExclusiveFirewallRule(Path.GetFileName(serverPath), serverPath, direction);
```
TwinBuilderOne, line 104: Using the file name as the rule name will cause the code to break if the path to the server file changes because the launcher will think the rule already exists and the rule will not be updated.

PR #1688, opened 2021-12-26
New Icon designs because I don't think they have been changed since 2019

EnumInt:


EnumInt:
nevermind lmao the old icons work fine but I would use the server icons, closing

PR #1702, opened 2022-01-07
Update all the projects to use the Treat Warnings as Errors flag.

Shirkie01:
Updating the csproj files to set the Treat Warnings as Errors flag. Since there are currently no warnings in the codebase, no other changes were required.


Measurity:
We like to keep warnings as warnings since errors will block builds. If we were to do this it should be in Directory.Build.props file too.

PR #1706, opened 2022-01-14
Added server check for invalid username

dartasen:
This check is only client side for now, which can be problematic with evil persons

### DIFF
--- a/NitroxServer/GameLogic/PlayerManager.cs
+++ b/NitroxServer/GameLogic/PlayerManager.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text.RegularExpressions;
 using System.Threading;
 using NitroxModel.DataStructures;
 using NitroxModel.DataStructures.GameLogic;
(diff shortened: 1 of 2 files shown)


NitroxServer/GameLogic/PlayerManager.cs
```diff
@@ -55,12 +56,25 @@ public IEnumerable<Player> GetAllPlayers()
...
+            }
+
             if (!string.IsNullOrEmpty(serverConfig.ServerPassword) && (!authenticationContext.ServerPassword.HasValue || authenticationContext.ServerPassword.Value != serverConfig.ServerPassword))
             {
                 MultiplayerSessionReservationState rejectedState = MultiplayerSessionReservationState.REJECTED | MultiplayerSessionReservationState.AUTHENTICATION_FAILED;
                 return new MultiplayerSessionReservation(correlationId, rejectedState);
             }
 
+            //https://regex101.com/r/eTWiEs/2/
+            if (!Regex.IsMatch(authenticationContext.Username, @"^[a-zA-Z0-9._-]{3,25}$"))
```
Jannify, line 72: We need to also check if the name isn't "Player", like on the client side.

Jannify (approved):
Code wise seems fine. I didn't do any in-game testing, I believe you did for sure.

Issue #1522, opened 2021-07-11
Seamoth gets stuck on nothing.


lazackna:
### Version

1.4.0.0

### Bug

Sometimes the seamoth gets stuck while playing on a server.

### Expected

You just play the game and the seamoth gets stuck all of a sudden.

### Steps to reproduce

Just explore with the seamoth. After a random amount of time it will get stuck on nothing.

### Other details:

It can be fixed by leaving the server while in the seamoth and rejoining. This puts you next to the seamoth. But this fix is a gamble. When rejoining you will loose oxygen. Depending on how deep you are you might die.

TheLxkas:
you can also fix this by making another player drive it.
And yes this is already known.

Measurity:
Can you still slightly look around with seamoth when it is stuck? This could tell us something about the cause.

lazackna:
Won't budge even a bit. It will become completely stuck.

Droow:
For me, it can slightly move to the sides using the mouse but will immediately return to its stuck position. And yes, it can be fixed by another player driving it, but then the seamoth upgrades sometimes disappear for one of the players.

PR #1713, opened 2022-01-26
Fix the vehicle stuck on nothing bug

tornac1234:
Title says it all, I hope it doesn't sound too much like a workaround (#1522)

To test, you need to force the issue, you can do this:
add the following code at line 353 (end of **OnVehiclePrefabLoaded**) in **Vehicles.cs**
```
switch (techType)
{
    case TechType.Seamoth:
        gameObject.EnsureComponent<MultiplayerSeaMoth>();
        break;
    case TechType.Exosuit:
        gameObject.EnsureComponent<MultiplayerExosuit>();
        break;
}
```
Then follow these instructions:
- Spawn an exosuit or a seamoth, leave the server then rejoin
- If you disable the fix (comment the "Exit()" in Vehicle_OnPilotModeEnd_Patch), you will see that there's the vehicle stuck bug
- Enable back the fix, then when you come back, if you join for the first time you will have the stuck bug for the vehicle, but as soon as you leave it and enter it again, the bug will be gone

### DIFF
--- a/NitroxClient/GameLogic/Vehicles.cs
+++ b/NitroxClient/GameLogic/Vehicles.cs
@@ -351,6 +351,21 @@ namespace NitroxClient.GameLogic
 
             VehicleChildObjectIdentifierHelper.SetInteractiveChildrenIds(gameObject, interactiveChildIdentifiers);
 
+            // TODO: Remove this when merging
+            // Testing instructions:
+            // Spawn an exosuit or a seamoth, leave the server then rejoin
+            // If you disable the fix (comment the "Exit()" in Vehicle_OnPilotModeEnd_Patch), you will see that there's the vehicle stuck bug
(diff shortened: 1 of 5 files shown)


NitroxClient/GameLogic/Vehicles.cs
```diff
@@ -351,6 +351,21 @@ private void OnVehiclePrefabLoaded(TechType techType, GameObject prefab, NitroxI
 
             VehicleChildObjectIdentifierHelper.SetInteractiveChildrenIds(gameObject, interactiveChildIdentifiers);
 
+            // TODO: Remove this when merging
```
tornac1234, line 354: this is for testing purpose

dartasen: This should be in PR description, no need to include this inside the code with directives to remove afterwards

PR #1693, opened 2021-12-29
Fix lockers label disappearing

tornac1234:
- [x] Fix the bug [Issue#1568](https://github.com/SubnauticaNitrox/Nitrox/issues/1568)

PR #1715, opened 2022-01-28
Fix a Sign metadata case that wasn't treated

tornac1234:
Title says it all


dartasen (approved):
Fix for a regression introduced in #1693

PR #1719, opened 2022-01-30
Fix leaving bench when writing in the chat

### LINKED ISSUE #1707: Chair/bench exit key will be triggered while chatting
If you're sitting on a chair/bench, if you're typing some word like the E default exit key, the exit behaviour will be triggered.

tornac1234:
Would fix #1707


NitroxPatcher/Patches/Dynamic/FreecamController_FreecamToggle_Patch.cs
```diff
@@ -0,0 +1,22 @@
...
+
+namespace NitroxPatcher.Patches.Dynamic
+{
+    public class FreecamController_FreecamToggle_Patch : NitroxPatch, IDynamicPatch
+    {
+        private static readonly MethodInfo TARGET_METHOD = Reflect.Method((FreecamController t) => t.FreecamToggle());
+
+        public static void Postfix(FreecamController __instance)
+        {
+            Resolve<LocalPlayer>().FreecamEnabled = __instance.mode;
```
dartasen, line 14: What's the point of the state syncing of freecam ? Seems totally out of scope to me

dartasen: image

This method is called by freecam command, this should help remove useless state patch

tornac1234: i need to verify but doesn't sitting on the bench also sets you as locked ?

Jannify (approved):

Issue #1403, opened 2021-01-19
The user is not properly told the server is shutting down.


Exso12Ecesis:
### Version
1.4.0.0 ALPHA

### Bug
When restarting or shutting down the server, the clients connected do not see anything is happening, other than a momentary "lost connection" dialog message, which goes away after the server restarts. The client may continue to conduct business, despite the server not actually receiving any sync messages.

### Expected
The client should attempt to reconnect, or simply kick the user back to the menu after the server restarts.

### Steps to reproduce
Restart the server with clients in it.

PR #1720, opened 2022-02-02
Improved the modal system

tornac1234:
Will fix Issue #1403

image

Please tell me if you'd like another color / another place to put the text.
Personally I like it right there so that if the player still wants to play like this (why would they ?) they don't have some big text in the middle of their screen

**EDIT: This is no longer a thing**


NitroxClient/Communication/Packets/Processors/ServerStoppedProcessor.cs
```diff
@@ -0,0 +1,17 @@
...
+
+public class ServerStoppedProcessor : ClientPacketProcessor<ServerStopped>
+{
+    public override void Process(ServerStopped packet)
+    {
+        // This packet is just meant to tell the modal to modify its message
+        if (LostConnectionModal.Instance)
+        {
+            LostConnectionModal.ServerStopped = true;
+        }
```
Jannify, line 15: Instead of botching into the LostConnectionModal we should create a ServerStoppedModal. To prevent the LostConnectionModa from firing we can disconnect the client properly beforehand (but that needs investigation if something will break).

NitroxClient/MonoBehaviours/Gui/InGame/LostConnectionModal.cs
```diff
@@ -80,5 +102,31 @@ private IEnumerator Show_Impl()
...
+
+            uGUI_TextFade textFade = gameObject.GetComponent<uGUI_TextFade>();
+            textFade.SetColor(Color.red);
+            // We jump 2 lines so that it's written after the versison text
+            textFade.SetText($"\n\n{text}");
+            textFade.SetAlignment(TextAnchor.UpperRight);
+            textFade.FadeIn(1f, null);
+
+            return textFade;
+        }
```
Jannify, line 119: Afaik the LostConnectionModel is unavoidable (you have to click it). Therefor this text is imo not providing more info than there already is and can be deleted.

Jannify (changes requested):

NitroxClient/MonoBehaviours/Gui/InGame/Modal.cs
```diff
@@ -0,0 +1,170 @@
...
+        }
+
+        Button noButton = buttonNoObject.GetComponent<Button>();
+        noButton.onClick = new Button.ButtonClickedEvent();
+        noButton.onClick.AddListener(() => { ClickNo(); });
+        buttonNoObject.GetComponentInChildren<Text>().text = NoButtonText;
+    }
+
+    public abstract void ClickYes();
+    public abstract void ClickNo();
```
dartasen, line 145: Maybe those should be virtual instead so you don't need to override it everytime for nothing

NitroxClient/MonoBehaviours/Multiplayer.cs
```diff
@@ -137,8 +137,9 @@ public void InitMonoBehaviours()
             gameObject.AddComponent<ThrottledBuilder>();
 
             // UI.
+            gameObject.AddComponent<KickedModal>();
             gameObject.AddComponent<LostConnectionModal>();
-            gameObject.AddComponent<PlayerKickedModal>();
+            gameObject.AddComponent<ServerStoppedModal>();
```
dartasen, line 142: Maybe we can be more flexible than adding by hand those modals ?

NitroxClient/MonoBehaviours/Gui/InGame/Modal.cs
tornac1234: oh ok thanks, was looking for a way to make them optional

NitroxClient/MonoBehaviours/Multiplayer.cs
tornac1234: What about some class that detects all the Modals from this solution and registers them automatically ?

dartasen: You can already do it with dependency injection

tornac1234: Yea this is what I meant, something like this [NitroxPatchesModule](https://github.com/SubnauticaNitrox/Nitrox/blob/master/NitroxPatcher/Modules/NitroxPatchesModule.cs)

NitroxServer/Communication/LiteNetLib/LiteNetLibServer.cs
```diff
@@ -73,6 +73,7 @@ private async Task PortForwardAsync(ushort port)
 
         public override void Stop()
         {
+            playerManager.SendPacketToAllPlayers(new ServerStopped());
```
Jannify, line 76: Maybe we can wait here for like half a second so everyone has received the ServerStopped packet.

NitroxClient/MonoBehaviours/Gui/InGame/Modal.cs
```diff
@@ -0,0 +1,170 @@
...
+        StartCoroutine(Show_Impl());
+    }
+
+    /// <summary>
+    /// Removes the Modal from the screen
+    /// </summary>
+    public void Hide()
+    {
+        FreezeTime.End($"Nitrox{SubWindowName}");
+        if (!IsAvoidable)
```
Jannify, line 73: reverse condition

NitroxClient/MonoBehaviours/Gui/InGame/Modal.cs
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    public abstract void ClickYes();
+    public abstract void ClickNo();
+
+    private IEnumerator Show_Impl()
+    {
+        // Execute frame-by-frame to allow UI scripts to initialize.
+        InitSubWindow();
+        yield return null;
```
Jannify, line 151: `yield break;` or if this isn't working `yield return new WaitForEndOfFrame();`

Jannify (changes requested):

NitroxPatcher/Patches/Dynamic/FreezeTime_Begin_Patch.cs
```diff
@@ -1,17 +1,24 @@
...
     {
         private static readonly MethodInfo TARGET_METHOD = Reflect.Method(() => FreezeTime.Begin(default(string), default(bool)));
 
-        public static bool Prefix(string userId, bool dontPauseSound)
+        // We don't want to prevent from freezing the game if the opened modal wants to freeze the game
+        public static bool Prefix(string userId)
         {
-            return userId.Equals("FeedbackPanel");
+            // If we ask to freeze from a Nitrox modal, userId will be like this: NitroxServerStoppedModal, so we need to remove "Nitrox" to access the modal's name
+            return userId.Equals("FeedbackPanel") ||
```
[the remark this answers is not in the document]
tornac1234: that's some legacy code I don't know about, should we remove it ?

NitroxClient/MonoBehaviours/Gui/InGame/Modal.cs
tornac1234: `yield break;` makes following code unreachable apparently

NitroxPatcher/Patches/Dynamic/FreezeTime_Begin_Patch.cs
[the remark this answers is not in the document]
tornac1234: tho we disabled the feedback panel so there's no point in keeping this

NitroxPatcher/Patches/Dynamic/IngameMenu_OnDeselect_Patch.cs
```diff
@@ -0,0 +1,31 @@
...
+public class IngameMenu_OnDeselect_Patch : NitroxPatch, IDynamicPatch
+{
+    private static MethodInfo TARGET_METHOD = Reflect.Method((IngameMenu t) => t.OnDeselect());
+
+    // OnDeselect happens when you deselect the modal (by clicking outside or pressing escape)
+    // Therefore, if we prevent it from happening it certain cases, it will force the user to click one of the buttons of the modal
+    public static bool Prefix(IngameMenu __instance)
+    {
+        // Cancel if the modal is set to be non-avoidable (IsAvoidable) and if it's not the modal that wants to close itself (IsAvoidableBypass)
+        if (Modal.TryGetByName(__instance.currentScreen.name, out Modal modal) && !modal.IsAvoidable && !modal.IsAvoidableBypass)
```
Jannify, line 20: What gamobject is `__instance.currentScreen` exactly? Could we use `GetComponent<Modal>` to get the current Modal?

Jannify (changes requested):

NitroxPatcher/Patches/Dynamic/IngameMenu_OnDeselect_Patch.cs
tornac1234: they're just the visual modal, the component is located in the Multiplayer object

Jannify: Mhh can we have a static "currentModal" field in `Modal.cs`? I would only use the string to type conversation if really necessary.

tornac1234: ok I'll go for it

NitroxClient/MonoBehaviours/Gui/InGame/Modal.cs
```diff
@@ -0,0 +1,189 @@
...
+using UnityEngine;
+using UnityEngine.UI;
+using UWE;
+
+namespace NitroxClient.MonoBehaviours.Gui.InGame;
+
+/// <summary>
+/// Base class for Modal components, which are dialog boxes that appear in the middle of the screen
+/// </summary>
+public abstract class Modal : MonoBehaviour
```
Jannify, line 13: Except from `StartCoroutine()`  in L76, which can be replaced by `CoroutineHost.StartCoroutineAsync()`, I don't see a reason this is inheriting MonoBehaviour.

tornac1234: We used to add the modals as components in the Multiplayer GameObject. Do you think we don't need to add them as components anymore ? (since we now have a static dictionary with their references)

Jannify: Yes as we don't need Start(), Update, etc. we can (and should) remove the inheritance of MonoBehaviour.

Jannify (approved):
Code wise seems fine, needs in-game testing.

Coding-Hen (approved):

Issue #1723, opened 2022-02-04
Steam features (like Controller mapping) won't load with direct start


joshuatam:
### Version
master at 9e9382d

### Bug
Controller is not working.

### Expected
It was working in last stable build 1.5.0.1

### Steps to reproduce
Build the project from master branch, open NitroxLauncher.exe after plugging in a controller, after opening the game the cursor will move randomly and clicked buttons.

### Other details:
Windows 10
I have already opened Steam in the background.

Jannify:
Could you upload your game log (it's located inside the launcher folder)?a

joshuatam:
> Could you upload your game log (it's located inside the launcher folder)?

Sure.

[launcher-20220205.log](https://github.com/SubnauticaNitrox/Nitrox/files/8004543/launcher-20220205.log)

[game-20220205.log](https://github.com/SubnauticaNitrox/Nitrox/files/8004541/game-20220205.log)

joshuatam:
Btw, this is the log from 1.5.0.1

[launcher-20220205.log](https://github.com/SubnauticaNitrox/Nitrox/files/8004625/launcher-20220205.log)

[game-20220205.log](https://github.com/SubnauticaNitrox/Nitrox/files/8004629/game-20220205.log)

joshuatam:
I also notice another behavior, when clicking "Play Multiplayer" in v1.5.0.1, it will first open Steam in the background (Steam icon appeared at the Task Bar and disappeared shortly), and then start the game. While using the master branch, it will open the game directly without Steam icon on Task Bar.

joshuatam:
I am not sure whether it is related to https://github.com/SubnauticaNitrox/Nitrox/commit/fbd4229a057b751c26a93d4be51d299d01e2af0c

`NitroxLauncher/LauncherLogic.cs` logic is changed from `"steam://run/264710"` to the real path of the game exe.

joshuatam:
Finally to workaround by reviesing launch method of Steam platform.
Please see if it is a suitable modification. #1724

dartasen:
> I also notice another behavior, when clicking "Play Multiplayer" in v1.5.0.1, it will first open Steam in the background (Steam icon appeared at the Task Bar and disappeared shortly), and then start the game. While using the master branch, it will open the game directly without Steam icon on Task Bar.

Yeah, since 1.5 we reworked the whole logic behind platform management

Jannify:
[...]

PR #1724, opened 2022-02-05
Patch platform steam to use steam exe to launch game

joshuatam:
Revised to use steam exe to launch game, related to #1723

### DIFF
--- a/NitroxClient/MonoBehaviours/MultiplayerExosuit.cs
+++ b/NitroxClient/MonoBehaviours/MultiplayerExosuit.cs
@@ -28,7 +28,7 @@ namespace NitroxClient.MonoBehaviours
             base.Enter();
         }
 
-        internal override void Exit()
+        public override void Exit()
         {
             GetComponent<Rigidbody>().freezeRotation = true;
             exosuit.SetIKEnabled(false);
--- a/NitroxClient/MonoBehaviours/MultiplayerSeaMoth.cs
+++ b/NitroxClient/MonoBehaviours/MultiplayerSeaMoth.cs
@@ -38,7 +38,7 @@ namespace NitroxClient.MonoBehaviours
             }
         }
 
-        internal override void Exit()
+        public override void Exit()
         {
             seamoth.bubbles.Stop();
             base.Exit();
--- a/NitroxClient/MonoBehaviours/MultiplayerVehicleControl.cs
+++ b/NitroxClient/MonoBehaviours/MultiplayerVehicleControl.cs
@@ -74,7 +74,7 @@ namespace NitroxClient.MonoBehaviours
             enabled = true;
         }
 
-        internal virtual void Exit()
+        public virtual void Exit()
         {
             enabled = false;
         }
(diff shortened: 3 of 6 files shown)


Jannify:
After 1.5.0.1 We switched from the "steam launch command" to the game exe so we have full control of the subnautica process. Therefor we are currently discussing options and their pros and cons internally with the author of this part of code.

Jannify:
We discussed this internally and would like to stick with the current solution for now. But we will investigate a solution where you can choose the start method, so you can switch between direct start (current) and steam-link. For now there is a workaround which should work:

- Add `-nitrox <path to Nitrox launcher folder>` to the games launch options in steam
- Launch the game with the launcher and close it once the menu showed up
- Now start the game with steam, it should load Subnautica with Nitrox loaded.

kgleken:
> We discussed this internally and would like to stick with the current solution for now. But we will investigate a solution where you can choose the start method, so you can switch between direct start (current) and steam-link. For now there is a workaround which should work:
> 
> * Add `-nitrox <path to Nitrox launcher folder>` to the games launch options in steam
> * Launch the game with the launcher and close it once the menu showed up
> * Now start the game with steam, it should load Subnautica with Nitrox loaded.

I have 1.6.0.0 and I want to use this workaround so I can play with a controller. I can't tell if I'm doing something wrong.
> * When I launch the game using the nitrox launcher first, do I need to start the server and connect to the multiplayer session before exiting?
> * Is there a certain way that I need to exit in order to relaunch the game and have it work? I can see the server log says "Server has paused. Waiting for players to connect. World state saved."
> * I tried using the -nitrox launch argument. Can you show me how to format it so that I know if I am doing it right? For example I tried -nitrox "C:/Users/Glek/Downloads/Nitrox_1.6.0.0", with and without parenthesis, with both types of slashes and it still isn't working. The game still just launches in single player mode for me.

Jannify:
You don't need to connect to a server in your first launch. Important is that you enter the main menu when you start the game with the Nitrox launcher the first time. Then close the game with the main menu quit button. Then start it with steam, I use backslashes and no parenthesis. The Nitrox Launcher needs to be open the whole time.

kgleken:
Thank you so much for the fast response. I tried what you said, and it worked this time. For some reason I had it in my head that you needed to exit the game not through main menu.

Some other interesting stuff I found out that might be useful to you or others:
* Ultimately my end goal was to get it working on 1.6.0.0 so I could play with VR using a controller. The problem I was having was that the controller actually worked, it just doesn't work on the multiplayer button on the main menu once the game is launched, so you can't join a game. The d-pad kinda works but not on any of the stuff that is added by the Nitrox mod.
* Although the trick above worked to allow me to relaunch the game from steam with multiplayer mode on, it did not fix my problem. Perhaps the controller issue I had was different than [#1723](url) 
* What I was able to do however was use the d-pad/look around using my oculus to select menu options. To join a server I could click using my mouse, then on the next screen I was able to join by pressing the "Enter" key on my keyboard. Once I was in the game I could use my controller like normal
* I'm guessing that anybody can repeat these steps as a workaround as long as they are the one hosting server (the options for 127.0.0.1 are already pre-filled). Not sure if it still works to join another person's server, but my friends aren't using vr anyway, so I can just be the one to host

PR #1725, opened 2022-02-05
Fix for freecam nullpointer spam

dartasen:
Freecam does set the transform of the player to null and disable the camera gameobject causing various nullpointer inside our movement handler code.

```
[22:20:44.762] [ERR-UNITY] NullReferenceException: Object reference not set to an instance of an object
NitroxClient.MonoBehaviours.PlayerMovement.GetVehicleMovement () (at <76f1b38f81a64b09bc01bdb03a9086bb>:0)
NitroxClient.MonoBehaviours.PlayerMovement.Update () (at <76f1b38f81a64b09bc01bdb03a9086bb>:0)
```

### DIFF
--- a/NitroxClient/MonoBehaviours/PlayerMovement.cs
+++ b/NitroxClient/MonoBehaviours/PlayerMovement.cs
@@ -30,19 +30,24 @@ namespace NitroxClient.MonoBehaviours
             {
                 time = 0;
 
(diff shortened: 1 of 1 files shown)


Jannify (approved):
Code wise seems fine. Would include it in RC-test-session.

PR #1727, opened 2022-02-08
Adding a way to specify the installation path through environment variables

LionZXY:
**Background**
Right now we have no way to specify the install location explicitly. It would be useful to have a way to do this, as other ways can fail and the user can specify the path to the game himself. 

**Changes**
- Added `SUBNAUTICA_INSTALLATION_PATH`, which doesn't break the current functionality in any way.

**How to test**
- Change environment variable `SUBNAUTICA_INSTALLATION_PATH` (see how [here](https://www.architectryan.com/2018/08/31/how-to-change-environment-variables-on-windows-10/))
- Run server and check that path from environment variable catch correctly

If you have the game installed and one of the classes above has already found its way, you can test this change by moving that class up.

### DIFF
--- a/NitroxModel/Discovery/InstallationFinders/EnvironmentGameFinder.cs
+++ b/NitroxModel/Discovery/InstallationFinders/EnvironmentGameFinder.cs
@@ -1,0 +1,30 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+
+namespace NitroxModel.Discovery.InstallationFinders
+{
+    /// <summary>
+    ///     Trying to find the path in environment variables by the key SubnauticaInstallationPath that contains the installation directory of Subnautica.
+    /// </summary>
+    public class EnvironmentGameFinder : IFindGameInstallation
+    {
+        public string FindGame(IList<string> errors = null)
+        {
+            string path = Environment.GetEnvironmentVariable("SubnauticaInstallationPath");
+            if (string.IsNullOrEmpty(path))
+            {
+                errors?.Add(@"Configured game path was found empty. Please enter the path to the Subnautica installation.");
+                return null;
+            }
+
+            if (!Directory.Exists(Path.Combine(path, "Subnautica_Data", "Managed")))
+            {
+                errors?.Add($@"Game installation directory config '{path}' is invalid. Please enter the path to the Subnautica installation.");
+                return null;
+            }
+
+            return path;
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


NitroxModel/Discovery/InstallationFinders/EnvironmentGameFinder.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+
+namespace NitroxModel.Discovery.InstallationFinders
+{
```
Jannify, line 6: Small code consistency point here: We recently decided to use [file scoped namespace](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-10.0/file-scoped-namespaces) in new files.

LionZXY: Fix it, check again, please

NitroxModel/Discovery/InstallationFinders/EnvironmentGameFinder.cs
```diff
@@ -1,30 +1,29 @@
-﻿using System;
+﻿namespace NitroxModel.Discovery.InstallationFinders;
```
Measurity, line 1: Namespace declaration should be below usings (using statements should always come first)

Measurity:
In which cases does finding the Subnautica path fail? I'm interested in how we can fix the other lookups.

NitroxModel/Discovery/InstallationFinders/EnvironmentGameFinder.cs
```diff
@@ -0,0 +1,29 @@
...
+///     Trying to find the path in environment variables by the key SUBNAUTICA_INSTALLATION_PATH that contains the installation directory of Subnautica.
+/// </summary>
+public class EnvironmentGameFinder : IFindGameInstallation
+{
+    public string FindGame(IList<string> errors = null)
+    {
+        string path = Environment.GetEnvironmentVariable("SUBNAUTICA_INSTALLATION_PATH");
+        if (string.IsNullOrEmpty(path))
+        {
+            errors?.Add(@"Configured game path was found empty. Please enter the path to the Subnautica installation.");
```
Jannify, line 17:
```suggestion
            errors?.Add(@"Configured game path with environment variable SUBNAUTICA_INSTALLATION_PATH was found empty.");
```

Jannify (changes requested):

PR #1728, opened 2022-02-13
Add log level to console output

killzoms:


Measurity:
I'm 50/50 on this so what about making the console log template configurable in server.cfg? We could add a comment with a link to the template docs for Serilog: https://github.com/serilog/serilog/wiki/Configuration-Basics#output-templates

Jannify:
Closing this due to counter-proposal and inactivity.

PR #1729, opened 2022-02-13
Fixes null ref in PlayerManager

killzoms:
A player could leave while in the JoinQueue and it would cause a null reference when trying to reserve that player later


Coding-Hen (commented):
My concerns around the null check of the player has now been addressed. I've tested without errors and pending feedback from server owners who are testing this I think it's ready to be merged.

PR #1735, opened 2022-02-21
Prevent Ctrl+C from closing the server

### LINKED ISSUE #1489: Ctrl + C is closing the server
### Bug
When pressing `Ctrl+C` in the server console it stops the server

### Expected
It should clear the line with possible entered commands like in normal command-line-interfaces

### Steps to reproduce
Start the server, type something and press `Ctrl+C`

TwinBuilderOne:
Resolves #1489

### DIFF
--- a/NitroxServer-Subnautica/Program.cs
+++ b/NitroxServer-Subnautica/Program.cs
@@ -297,8 +297,8 @@ public class Program
 
     private static void OnCtrlCPressed(object sender, ConsoleCancelEventArgs e)
     {
-        // Prevents process from terminating
-        e.Cancel = true;
+        e.Cancel = true; // Prevents process from terminating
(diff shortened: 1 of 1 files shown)


NitroxServer-Subnautica/Program.cs
```diff
@@ -297,7 +297,8 @@ private static bool ConsoleEventCallback(int eventType)
 
     private static void OnCtrlCPressed(object sender, ConsoleCancelEventArgs e)
     {
-        StopServer();
+        // Prevents process from terminating
+        e.Cancel = true;
```
Jannify, line 301:
```suggestion
        e.Cancel = true; // Prevents process from terminating
        Console.Write("\r" + new string(' ', Console.WindowWidth - 1) + "\r"); // Clears current line
```

PR #1736, opened 2022-02-22
Fix an issue where duplicated item id in modules blocked world loading

tornac1234:


Jannify:
I think we speculated that the duplication is from one client which doesn't get the new container id. For that reason we should not assign a new item id because then we duplicate the item. Instead maybe we should remove one of the two item instances with the same id.

Coding-Hen (changes requested):
Needs to be moved to a new file as it will be a new release so the version will change. Presumably Upgrade_V1601

Jannify (approved):
Code wise seems fine

PR #1737, opened 2022-02-22
Added more logging to better determine "same key" world loading error…

Pysis868:
…s for potential later repair.

Without opening an issue, thought this code would be useful, and did not see existing functionality that would help with diagnosing the exact problem in the case I was investigating.

I saw a few other cases reported, but this one was produced specifically with the server upgrade from 1.5.0.1 to 1.6.0.0 using the JSON data file format.

The issue was contained in the WorldData file, concerning the "PowerCell(Charger2)" item under the below path that had duplicate sibling entries specifying different containerIds, the problem being their ItemId was identical when creating the modules of the 3 main world dictionaries.  Unsure how it arrived in this state, but the 1.5.0.1 server still seemed to at least load correctly.
Fixed using this targeted command: `jq -c --arg ItemId "$ItemId" '.InventoryData.Modules |= unique_by(.ItemId)' "$jsonFile"`

### DIFF
--- a/NitroxModel/DataStructures/GameLogic/EquippedItemData.cs
+++ b/NitroxModel/DataStructures/GameLogic/EquippedItemData.cs
@@ -26,7 +26,7 @@ namespace NitroxModel.DataStructures.GameLogic
 
         public override string ToString()
         {
-            return "[EquippedItemData ContainerGuid: " + ContainerId + "Id: " + ItemId + " Slot: " + Slot + " TechType: " + TechType + "]";
+            return "[EquippedItemData ContainerGuid: " + ContainerId + " Id: " + ItemId + " Slot: " + Slot + " TechType: " + TechType + "]";
         }
     }
 }
--- a/NitroxServer/GameLogic/Items/InventoryManager.cs
+++ b/NitroxServer/GameLogic/Items/InventoryManager.cs
@@ -13,9 +13,55 @@ namespace NitroxServer.GameLogic.Items
 
         public InventoryManager(List<ItemData> inventoryItems, List<ItemData> storageSlotItems, List<EquippedItemData> modules)
         {
-            inventoryItemsById = new ThreadSafeDictionary<NitroxId, ItemData>(inventoryItems.ToDictionary(item => item.ItemId), false);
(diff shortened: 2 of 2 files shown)


Jannify:
In #1736 we already implement a fix for this issue while upgrading save files. We will investigate the underlying issue, as all saves we received from users have problems with PowerCells we are not sure if this is a item system error or has something to do with bases and power cell chargers.
Will close this pr for now because it's just logging and will be (hopefully) fixed with the mentioned pr above.

Pysis868:
Would it hurt to include the logging?  What if a similar issue appears later?

tornac1234:
> Would it hurt to include the logging? What if a similar issue appears later?

we'd probably release another hotfix

Pysis868:
But why not add logging code that could only help, and is already done?  More information is always better, as it could help developers, supporters, and users arrive at any solution faster.

dartasen:
> But why not add logging code that could only help, and is already done? More information is always better, as it could help developers, supporters, and users arrive at any solution faster.

We do have better way to find the issue than logging, for example for a json file you only need to analyze the structure of the json with a online validator to find the issue right away.

Pysis868:
That's not what fixed it though.  I thought my jq tool would help, but it, like many, apparently silently discards duplicate object keys.  I tried jsonlint-mod since the original author of the base tool would not add the code people wanted, and that returned successfully.

The problem in this case was duplicate items as interpreted by the game/server engine, as indicated by my fixing command's method above.  If there was a better way of fixing this, then the supporters in the help channel would not be telling users to just play a new world whenever an error like this occurs, or keep various backups that, since is not automatic, many of the users asking for help did not have, and would also not help in the case of this upgrade bug.

Not having a solution that avoids data loss that is also readily available to users is what started this code request..

dartasen:
Jq is a good idea, however command line for lambda user isn't something we can imagine being possible (and we'll not ask our support staff to download files from users and patch them for us), I saw that they're hosting a online version of Jq ([](https://jqplay.org/)) which already sounds better.

> I also still don't understand the amount of effort in opposition to adding logging code..

We would need to rewrite the code somehow since he does not meet our code rules, and we also don't need logging for things we already know about and for things that has been merged in master yesterday.

Pysis868:
The potential is for cases that could happen in the future, and would shorten time to solution.
This was an upgrade bug, but there were other unsolved cases in the past in the support chat.

How does it not meet the coding guidelines then?

PR #1739, opened 2022-02-23
Reactivate GitHub workflows

Jannify:
- [x] Wait for Meas to setup the self-hosted runner
- [x] Wait for #1738 to be merged
grafik


Jannify:
We need to wait for the upcoming SN update because tests fail with Mono on Linux.

PR #1754, opened 2022-03-01
Fix cyclops not moving while controlling it through a camera

tornac1234:
The check for freecam detection may be not enough precise


Jannify (approved):
Also did in-game testing

Coding-Hen (approved):
In game testing shows working as intended

PR #1759, opened 2022-03-12
Improve player's experience during initial sync

tornac1234:
- [x] Prevent player from loosing health during game loading
- [x] Prevent player from loosing oxygen during game loading
- [x] Prevent player from clipping through the map when loading
- [x] Add progress for building initial sync
- [x] Prevent player from clipping out of the cyclops during InitialSync
- [x] Remember player's rotation

So, this is why these will occur in only certain cases:
When local player is first initial synced, he's moved to the spawn pod, but as soon as `PlayerPositionInitialSyncProcessor` happens, he's moved to where he was before, which means maybe in a dangerous place or underwater.
If following initial syncs take some time / make the game unloadable, player will die before having its game loaded.

Therefore, we prevent player from loosing oxygen/health before he has fully loaded and synced his game.

In order to test the fix:
- World setup:
    - Go to any world and go underwater somewhere you can loose oxygen or even health
    - Disconnect and shutdown server
- Code setup:
  - Test without the fix
    - Go to `OxygenManager_RemoveOxygen_Patch` and write `return true;` before what's currently inside the `Patch` method
    - Go to `PlayerModulesInitialSyncProcessor` inside the `Process` method and add `yield return new WaitForSeconds(15);` at the end of it
    - Go to `InitialPlayerSyncProcessor` at the end of `PlayerInitialSyncProcessor` at line L84 where there's the invincibility setting and make it false
    - Start game and you already lost some oxygen / health
  - Test with the fix
    - Just revert back the changes we made earlier but not the one inside `PlayerModulesInitialSyncProcessor`


TwinBuilderOne:
On the topic of things breaking while the game is loading, I would recommend making the Seamoth, P.R.A.W.N., and Cyclops not lose health/hull integrity while loading as well.

tornac1234:
strategic strike incoming...
with about 30 modified files

NitroxPatcher/Patches/Dynamic/OxygenManager_RemoveOxygen_Patch.cs
```diff
@@ -0,0 +1,26 @@
...
+using NitroxClient.MonoBehaviours;
+using NitroxModel.Helper;
+
+/// <summary>
+/// Prevents Local Player from loosing oxygen while he hasn't loaded yet
+/// Oxygen removal starts after PlayerPositionInitialSyncProcessor. If player is moved in the water because before, he'll still be in the spawn pod
+/// </summary>
+namespace NitroxPatcher.Patches.Dynamic
+{
+    public class OxygenManager_RemoveOxygen_Patch : NitroxPatch, IDynamicPatch
```
dartasen, line 12: Not a huge fan of those patches that will trigger at each frame.

I think we can use the same system as precursor teleporter which freeze your stats when you're teleporting and waiting for world settling.
If you look at Player::Update() : 
```csharp
	if (!this.CanBreathe() && !this.IsFrozenStats())
	{
		float breathPeriod = this.GetBreathPeriod();
		if (this.timeMonitor.DidChangeInterval(breathPeriod, 0f))
		{
			this.oxygenMgr.RemoveOxygen(this.GetOxygenPerBreath(breathPeriod, this.depthClass.value));
			this.tookBreathEvent.Trigger(this);
		}
	}
```
Might need to check but you can play with Player::FreezeStats() and Player::UnfreezeStats() to do the same thing (and also target all player stats, not only oxygen)

NitroxClient/MonoBehaviours/ThrottledBuilder.cs
```diff
@@ -69,6 +70,7 @@ private void ProcessBuildEventsUntilFrameBlocked()
                 try
                 {
                     ActionBuildEvent(nextEvent);
+                    QueueLengthChanged(this, new EventArgs());
```
dartasen, line 73: If this method called during playtime ?

dartasen (changes requested):

NitroxClient/GameLogic/InitialSync/PlayerPositionInitialSyncProcessor.cs
```diff
@@ -44,33 +48,53 @@ public override IEnumerator Process(InitialPlayerSync packet, WaitScreen.ManualW
...
             }
             Transform rootTransform = subRoot.transform;
             Quaternion vehicleAngle = rootTransform.rotation;
             position = vehicleAngle * position;
             position = position + rootTransform.position;
             Player.main.SetPosition(position);
+            yield return WaitForWorldLoad();
+        }
+
+        public static IEnumerator WaitForWorldLoad()
```
dartasen, line 85: Maybe we should extract this Enumerator to a proper class, so PlayerTeleportedProcessor and maybe future class are not coupled with that one

NitroxPatcher/Patches/Dynamic/OxygenManager_RemoveOxygen_Patch.cs
tornac1234: That's really my bad for not seing it since I litteraly took inspiration from that code

NitroxClient/Communication/Packets/Processors/PlayerTeleportedProcessor.cs
```diff
@@ -3,11 +3,19 @@
...
 using UnityEngine;
+using UWE;
+using Terrain = NitroxClient.GameLogic.Terrain;
 
 namespace NitroxClient.Communication.Packets.Processors
 {
     public class PlayerTeleportedProcessor : ClientPacketProcessor<PlayerTeleported>
     {
+        private readonly Terrain terrain;
+        public PlayerTeleportedProcessor(Terrain terrain)
```
dartasen, line 14:
```suggestion
        private readonly Terrain terrain;
        
        public PlayerTeleportedProcessor(Terrain terrain)
```

NitroxClient/GameLogic/Terrain.cs
```diff
@@ -93,5 +94,19 @@ private IEnumerator WaitAndSyncCells(float delay)
                 yield return new WaitForSeconds(0.05f);
             }
         }
+
+        public IEnumerator WaitForWorldLoad()
```
dartasen, line 98: WaitForWorldLoad can be static to remove the need of storing an instance and since it's only accessing static singleton

Jannify, line 98: Would mention in a xml comment that this also unfreezes the player as this can't be obtained by its name.

dartasen:
Looks good code wise, need a test session now to validate

Coding-Hen:
This may not be an issue (I'm reading this on my phone and code doesn't play too well). Does this take into account someone moving say a Cyclops while someone is loading in and frozen?

It might be one more for the test session. Other than that no issues I can see initially from reading the code

tornac1234:
> This may not be an issue (I'm reading this on my phone and code doesn't play too well). Does this take into account someone moving say a Cyclops while someone is loading in and frozen?
> 
> It might be one more for the test session. Other than that no issues I can see initially from reading the code

Waking up in a Cyclops is currently not supported

tornac1234:
As discussed in the Discord server, the PositionMaintainer seems out of scope and will probably be postponed to a future cyclops improvements PR.
In the meantime, I'll let this code here:
```c#
/// <summary>
/// Maintains the local player at certain relative location inside a subRoot, even if it moves.
/// </summary>
public class PositionMaintainer : MonoBehaviour
{
    private SubRoot subRoot;
    private Vector3 position;
    private Quaternion rotation;

    public void Initialize(SubRoot subRoot, Vector3 position, Quaternion rotation)
    {
        this.subRoot = subRoot;
        this.position = position;
        this.rotation = rotation;
    }

    public void Deinitialize()
    {
        subRoot = null;
        position = default;
        rotation = default;
    }

    public void LateUpdate()
    {
        if (!subRoot)
        {
            return;
        }
        Player.main.SetPosition(CalculatePosition(), CalculateRotation());
    }

    public void SetEnabled(bool state)
    {
        Player.main.cinematicModeActive = state;
        enabled = state;
    }

    public Vector3 CalculatePosition()
    {
        Transform rootTransform = subRoot.transform;
        Quaternion vehicleAngle = rootTransform.rotation;
        return vehicleAngle * position + rootTransform.position;
    }

    public Quaternion CalculateRotation()
    {
        return subRoot.transform.rotation * rotation;
    }
}
```

dartasen (approved):
Test sessions was successfull 

Oxygen stat is not synced on reload

PR #1771, opened 2022-04-02
Steam.cs: Skip StartPlatformAsync() if launcher is running on Wine

zimsneexh:
The launcher fails to start the game if it's running in Wine because StartPlatformAsync() fails to start steam.

Skipping it makes the Launcher work fine on Linux.

### DIFF
--- a/NitroxModel/Platforms/Store/Steam.cs
+++ b/NitroxModel/Platforms/Store/Steam.cs
@@ -2,6 +2,7 @@
 using System.Diagnostics;
 using System.IO;
 using System.Threading.Tasks;
+using Microsoft.Win32;
 using NitroxModel.Discovery;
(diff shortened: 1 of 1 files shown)


Measurity:
Thanks for the PR. Is it possible to detect steam on wine properly? I would prefer to make it work rather than skip.

zimsneexh:
I am not exactly sure if SOFTWARE\Valve\Steam\ActiveProcess\pid is added to the Wine Prefix's registry if it's running in Proton, since the windows steam client isn't actually running.

A warning message telling the user to start Steam manually might work?

Measurity:
I'd like to properly support linux and not skip functionality to make it work. So I'll close this and we'll revisit support for linux sometime later.

PR #1784, opened 2022-04-17
Use ISO639 language codes

Jannify:
This pr changes our language file names to the [ISO639-1](https://www.andiamo.co.uk/resources/iso-language-codes/) standard. As weblate only outputs these we are kinda required to implement it this way.


NitroxPatcher/Patches/Persistent/Language_LoadLanguageFile_Patch.cs
```diff
@@ -10,50 +12,79 @@ namespace NitroxPatcher.Patches.Persistent;
...
+            JsonData json = JsonMapper.ToObject(streamReader);
 
             foreach (string key in json.Keys)
             {
-                if (json[key].IsString)
+                JsonData jsonKey = json[key];
+                if (jsonKey.IsString)
                 {
-                    ___strings[key] = (string)json[key];
+                    strings[key] = (string)jsonKey;
```
dartasen, line 56:
```suggestion
                    strings[key] = (string)json[key];
```

NitroxPatcher/Patches/Persistent/Language_LoadLanguageFile_Patch.cs
```diff
@@ -10,50 +12,79 @@ namespace NitroxPatcher.Patches.Persistent;
...
+            {
+                languageToIsoCode.Add(culture.EnglishName, new Tuple<string, string>(culture.Name, culture.TwoLetterISOLanguageName));
+            }
+        }
+
+        // This language isn't registered in CultureInfo
+        if (existingLanguageNames.Contains("Spanish (Latin America)"))
+        {
+            languageToIsoCode.Add("Spanish (Latin America)", new Tuple<string, string>("es-419", "es"));
+        }
```
dartasen, line 86: Even if european spanish and latin america spanish does use different words for the same thing, I don't think it's worth doing special treatment for this one since both countries will understand the sentences (similar for US and UK english)

NitroxPatcher/Patches/Persistent/Language_LoadLanguageFile_Patch.cs
Jannify: If I change it to that I have two dictionary lookups.

NitroxLauncher/NitroxLauncher.csproj
```diff
@@ -6,6 +6,7 @@
         <Nullable>disable</Nullable>
         <UseWPF>true</UseWPF>
         <ApplicationIcon>icon.ico</ApplicationIcon>
+		<DisableFastUpToDateCheck>True</DisableFastUpToDateCheck>
```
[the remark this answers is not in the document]
Jannify: The result of this change is that it executes the AfterBuild targets of NitroxLauncher even if just a dependency of it (like Nitrox client) has changed code.

dartasen (approved):

NitroxPatcher/Patches/Persistent/Language_LoadLanguageFile_Patch.cs
Jannify: While I agree with the thought that it's normally not worth I think we should support any language SN also supports, even if it's just a en vs en-UK thing.

PR #1789, opened 2022-04-21
Added some code docs to NitroxModel project

Measurity:
Removed AtomicBool because it isn't used.


NitroxModel/DataStructures/GameLogic/Entity.cs
```diff
@@ -25,6 +25,14 @@ public class Entity
...
         public int Level { get; set; }
 
+        /// <summary>
+        ///     Gets the prefab class id assigned by Unity Engine. This is a unique <see cref="Guid"/>. 
+        /// </summary>
+        /// <remarks>
+        ///     A prefab is a Unity Engine object that acts as a blueprint to instantiate Game Objects.
+        ///     Game developers use prefabs to create similar copies of game objects in the scene/world as opposed to manually attaching
+        ///     the required components to each game object when similar behaviour is required.
+        /// </remarks>
```
Jannify, line 35: I don't know if we should explain GameObjects/Prefabs .A general understanding of Unity is definitely required to even start modding.

Measurity: Ok, I think some mention of what a prefab is should be given. What about a link to this doc instead of the text? https://docs.unity3d.com/Manual/Prefabs.html

Jannify: Yeah that's better

Jannify (approved):

PR #1790, opened 2022-04-23
Added persistence of KnownTech.analyzedTech

DV8FromTheWorld:
This fixed the issue where previously discovered tech would appear to be "newly discovered" upon rejoining and refinding the tech.
This was previously easily observed by:
 - joining a server
 - grab a peeper, which triggered the "New Creature Discovered" popup
 - leave and rejoin the server
 - grab a peeper and observe the popup again.

This change addresses this issue.

As a note, KnownTech has 2 types of tech: KnownTech.knownTech and KnownTech.analyzedTech

KnownTech.analyzedTech is tech that has been found and acquired. This includes things like Creatures, Resources, etc
KnownTech.knownTech on the other hand are TechTypes that have been "learned" (unlocked) due to finding/analyzing/scanning/unlocking a previous dependency.
For example, CreepvinePiece is a KnownTech.analyzedTech, and finding it unlocks the FiberMesh KnownTech.knownTech TechType.


NitroxModel/Packets/KnownTechEntry.cs
```diff
@@ -6,11 +6,19 @@ namespace NitroxModel.Packets
     [Serializable]
     public class KnownTechEntryAdd : Packet
```
Measurity, line 7: Apparent, filename and class name do not match.

NitroxModel/Packets/KnownTechEntry.cs
```diff
@@ -6,11 +6,19 @@ namespace NitroxModel.Packets
...
     {
+        public enum EntryCategory
+        {
+            KNOWN = 0,
+            ANALYZED = 1
```
Measurity, line 12:
```suggestion
            /// <summary>
            /// <b>KnownTech.knownTech</b><br/>
            /// Tech items that you learn about after fulfilling a requirement. Often times, discovering a new KnownTech.analyzedTech item will provide 1 or more KnownTech.knownTech items. (Peeper -> CookedPeeper)
            /// </summary>
            KNOWN = 0,
            /// <summary>
            /// <b>KnownTech.analyzedTech</b><br/>
            /// Tech items that you find in the world and acquire. They often show a notification saying you found/learned about them. (Fish, Resources, etc)
            /// <br/>
            /// </summary>
            ANALYZED = 1
```

Measurity (approved):
Code looks good and can confirm it works.

Coding-Hen:
I haven't tested this yet but looking at this change do we need to add something to the upgrade script or are we expecting this to be a breaking change or silently fix itself as things are scanned?

DV8FromTheWorld:
@Coding-Hen 
We do not need to add an upgrade script because there actually isn't anything to migrate from. We've never stored this data before and it is not possible to determine it completed from KnownTech.knownTech.

The expected flow for existing worlds will be that they will see the popup for each tech one more time, and after that the tech will have been persisted.

Measurity (approved):

dartasen (approved):

PR #1800, opened 2022-05-14
Added call to ensure that clients get the completed construction amou…

Shirkie01:
Added call to ensure that clients get the completed construction amount when finishing a base piece.


Jannify:
Was fixed with #1751

PR #1803, opened 2022-05-15
Fix remote players' lighting

### LINKED ISSUE #644: [Base] Player shading is incorrect
Players are always shaded as if they were outside causing them to appear pitch black during the night

tornac1234:
Some screenshots of the testing session in the thread ["light fix PR test session"](https://discord.com/channels/525437013403631617/975468417018658847)
(Tested with 3 players and seemed to work as expected)

How the fix works, we add a component named ``SkyApplier`` which goal is to adjust the lighting level of all the renderers inside a model depending on the lighting source.

Will fix #644

### DIFF
--- a/NitroxClient/GameLogic/RemotePlayer.cs
+++ b/NitroxClient/GameLogic/RemotePlayer.cs
@@ -84,6 +84,7 @@ namespace NitroxClient.GameLogic
             playerModelManager.BeginApplyPlayerColor(this);
             playerModelManager.RegisterEquipmentVisibilityHandler(PlayerModel);
             UpdateEquipmentVisibility();
+            SetupSkyAppliers();
 
             ErrorMessage.AddMessage($"{PlayerName} joined the game.");
         }
@@ -161,6 +162,7 @@ namespace NitroxClient.GameLogic
         {
             if (SubRoot != newSubRoot)
             {
+                SkyEnvironmentChanged.Broadcast(Body, newSubRoot);
                 if (newSubRoot)
                 {
                     Attach(newSubRoot.transform, true);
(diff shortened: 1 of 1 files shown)


NitroxClient/GameLogic/RemotePlayer.cs
```diff
@@ -268,5 +270,19 @@ private void UpdateEquipmentVisibility()
...
+        /// </summary>
+        private void SetupSkyAppliers()
+        {
+            // SkyAppliers are the components that apply the light effects on gameobjects
+            SkyApplier skyApplier = Body.AddComponent<SkyApplier>();
+            skyApplier.anchorSky = Skies.Auto;
+            skyApplier.emissiveFromPower = false;
+            skyApplier.dynamic = true;
+            skyApplier.renderers = Body.GetComponentsInChildren<SkinnedMeshRenderer>(true);
+            Log.Debug($"Successfully setup Sky Appliers on {PlayerName}");
```
Jannify, line 285: Minor: I wouldn't use the log here. You aren't (and probably can't) check if the SkyAppliers is successfully setup. Therefore this log will show up independent from the success of the method, only exception is a hard throw which would be logged anyway.

Jannify (changes requested):

NitroxClient/GameLogic/RemotePlayer.cs
tornac1234: Maybe I can insert a try catch with a first call to the ``SkyApplier.ApplySkybox()`` method and print if it was successful or an error

Jannify: The question is do we expect the method to fail? If not we shouldn't try catch it imo.

tornac1234: The ``SkyApplier.ApplySkybox()`` will fail if the skyApplier isn't correctly setup

Jannify: Looking at the code I don't see where it could fail as it checks for null for every variable used. Am I missing something?

tornac1234: I guess it's because of the renderers part, when I first created the function, I didn't include the ``.renderers = ...`` part and for some reasons, errors were thrown and you couldn't even load into the game.

Jannify: That's true if `renderers` is null it can throw but now that you use `Body.GetComponentsInChildren<SkinnedMeshRenderer>(true)` it should never be null, only not filled with anything in the worst case.

tornac1234: Ok, I'll just remove the logging then

dartasen (approved):

PR #1805, opened 2022-05-16
Added first .NET analyzer for validating packet deserialization

Measurity:
- [x] Ensure build fails when analyzer emits an error. [This is broken by design when using project referenced analyzer.](https://github.com/dotnet/roslyn/issues/6195#issuecomment-149951740)
- [x] Solution rebuild should not cause "file not found" for analyzer project (analyzer project should be rebuilt first).

Might be that to fix both these issues we need to move this analyzer to another repo and deploy on NuGet. But I wonder if this will make it harder to iterate while developing the analyzers.

Feature request for allowing project referenced analyzers to work: https://github.com/dotnet/roslyn/issues/18093

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -14,6 +14,7 @@
         <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
         <ImplicitUsings>disable</ImplicitUsings>
         <PathMap>$(MSBuildProjectDirectory)=$(MSBuildProjectName)</PathMap>
+        <EnableNETAnalyzers>true</EnableNETAnalyzers>
     </PropertyGroup>
     
     <PropertyGroup Condition="$([System.Text.RegularExpressions.Regex]::IsMatch($(MSBuildProjectName), '^Nitrox.*$'))">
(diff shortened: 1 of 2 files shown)


TwinBuilderOne:
Should be merged after #1798 as the current packet serializer does not require public parameterless constructors for packet types.

Jannify:
Did you try that https://github.com/dotnet/roslyn/issues/18093#issuecomment-1039415533 ?

Measurity:
[Yes, I added that](https://github.com/SubnauticaNitrox/Nitrox/pull/1805/files#diff-3153e604e6cb845e969e2a9fa8468be55e35ab9ab11aaaf3d258ce85ca898a83R33-R37) but the IDE integration isn't good enough that it updates analyzer warnings/errors immediately.

I'll go with NuGet or similar solution as this is what is supported by IDEs.

dartasen (approved):

Jannify (approved):
What's the plan for this pr? Merging now as an "enabling analyzer support" or wait to include the packet analyzer?

Measurity:
I'm not happy with the code analysis dependency requirement but let's merge it since it won't cause trouble either.

PR #1819, opened 2022-06-05
Fix many building bugs

tornac1234:
- [x] Fix ``BasePieceSpawnProcessor`` when they are rerun
- [x] Fix an issue where an object being built already has the NitroxId that is being added to it
- [x] Investigate about geometry respawning causing NRE for Multiplayer cinematic components
- [x] Change the way of identifying GameObjects
- [x] Fix hatches not correctly building on alien containments
- [x] Fix hatches not correctly building on MapRooms with something already built onto them
- [x] Could be cool to investigate more about building that replaces stuff which is not counted as deleted (e.g building a corridor on a window) **[FIXED]**
- ~~Repair an instance of ladders not being built on reconnection only~~

**This may need more testing.**
Concerning the `AreSameTechType()` part, I've tested manually for all the pieces.
Also I've made a messy reordering function in `BasePieceSpawnPrioritizer.cs` because I didn't know how to make it, can someone indicate me a better way of doing this please ?
The ladders issue will remain since it can happen and not happen in the same conditions, I won't look more into it

### DIFF
--- a/NitroxClient/GameLogic/Bases/GeometryRespawnManager.cs
+++ b/NitroxClient/GameLogic/Bases/GeometryRespawnManager.cs
@@ -110,14 +110,14 @@ public class GeometryRespawnManager
                    Equals(rotation, other.rotation);
         }
 
-        public bool Equals(Vector3 _position, Vector3 otherPosition)
+        public bool Equals(Vector3 position, Vector3 otherPosition)
         {
-            return Vector3.SqrMagnitude(_position - otherPosition) < 0.1f;
+            return Vector3.SqrMagnitude(position - otherPosition) < 0.1f;
         }
 
-        public bool Equals(Quaternion _rotation, Quaternion otherRotation)
+        public bool Equals(Quaternion rotation, Quaternion otherRotation)
         {
-            return Math.Abs(Quaternion.Angle(_rotation, otherRotation)) < 0.1f;
+            return Math.Abs(Quaternion.Angle(rotation, otherRotation)) < 0.1f;
         }
 
         public override bool Equals(object obj)
--- a/NitroxClient/GameLogic/Bases/GeometryRespawnManager.cs
+++ b/NitroxClient/GameLogic/Bases/GeometryRespawnManager.cs
@@ -29,23 +29,18 @@ public class GeometryRespawnManager
 
         foreach (Transform cellObject in cellObjects)
         {
-            if (cellObject != null)
+            if (!cellObject)
             {
-                for (int i = 0; i < cellObject.childCount; i++)
-                {
-                    Transform child = cellObject.GetChild(i);
+                continue;
+            }
 
-                    if (child != null && child.gameObject != null)
-                    {
-                        // Ensure there is already a nitrox id, we don't want to go creating one
-                        // which happens if you call GetId directly and it is missing.
-                        if (child.gameObject.GetComponent<NitroxEntity>() != null)
-                        {
-                            NitroxId id = NitroxEntity.GetId(child.gameObject);
-                            ObjectKey key = GetObjectKey(child.gameObject);
-                            nitroxIdByObjectKey[key] = id;
-                        }
-                    }
+            foreach (Transform child in cellObject)
+            {
+                // We only want to save NitroxIds
+                if (child.gameObject.TryGetComponent(out NitroxEntity nitroxEntity))
+                {
+                    ObjectKey key = GetObjectKey(child.gameObject);
+                    nitroxIdByObjectKey[key] = nitroxEntity.Id;
                 }
             }
         }
@@ -61,7 +56,10 @@ public class GeometryRespawnManager
     {
         // face based pieces are handled a bit differently on respawn as the BaseDeconstructable is not
         // yet attached to the game object.  So we'll pass in the known fields instead.
-        ObjectKey key = new(gameObject.name, gameObject.transform.position, faceCell, direction);
+        ObjectKey key = new(gameObject.name,
+                            gameObject.transform.parent.position,
+                            gameObject.transform.localPosition,
+                            gameObject.transform.localRotation);
         UpdateBasePieceIdIfPossible(gameObject, key);
     }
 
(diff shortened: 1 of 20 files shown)


Stelios210:
LOL

Stelios210:
@tornac1234 you are trash at fortnite!

NitroxClient/GameLogic/Bases/GeometryRespawnManager.cs
```diff
@@ -81,44 +79,45 @@ private void UpdateBasePieceIdIfPossible(GameObject gameObject, ObjectKey key)
...
+            return name == other.name &&
+                   Equals(parentCellPosition, other.parentCellPosition) &&
+                   Equals(position, other.position) &&
+                   Equals(rotation, other.rotation);
+        }
+
+        public bool Equals(Vector3 _position, Vector3 otherPosition)
+        {
+            return Vector3.SqrMagnitude(_position - otherPosition) < 0.1f;
+        }
```
dartasen, line 116:
```suggestion
        public bool Equals(Vector3 position, Vector3 otherPosition)
        {
            return Vector3.SqrMagnitude(position - otherPosition) < 0.1f;
        }
```

NitroxClient/GameLogic/Bases/GeometryRespawnManager.cs
```diff
@@ -81,44 +79,45 @@ private void UpdateBasePieceIdIfPossible(GameObject gameObject, ObjectKey key)
...
+
+        public bool Equals(Quaternion _rotation, Quaternion otherRotation)
+        {
+            return Math.Abs(Quaternion.Angle(_rotation, otherRotation)) < 0.1f;
```
dartasen, line 120:
```suggestion
        public bool Equals(Quaternion rotation, Quaternion otherRotation)
        {
            return Math.Abs(Quaternion.Angle(rotation, otherRotation)) < 0.1f;
```

NitroxClient/GameLogic/Bases/Spawning/BasePiece/BaseBioReactorSpawnProcessor.cs
```diff
@@ -19,15 +19,32 @@ public class BaseBioReactorSpawnProcessor : BasePieceSpawnProcessor
             TechType.BaseBioReactor
```
dartasen, line 19: Not related, but we should probably add a processor for BaseNuclearReactor techtype, it's almost a duplicate class of BioReactor in subnautica's code

NitroxClient/GameLogic/Bases/Spawning/BasePiece/BaseWaterParkProcessor.cs
```diff
@@ -45,5 +54,15 @@ protected override void SpawnPostProcess(Base latestBase, Int3 latestCell, GameO
...
+
+        private IEnumerator DelayModuleDetection(Base latestBase, Int3 latestCell, GameObject finishedPiece)
+        {
+            if (!finishedPiece)
+            {
+                yield break;
+            }
+            yield return new WaitForSeconds(0.1f);
+            SpawnPostProcess(latestBase, latestCell, finishedPiece);
+        }
```
dartasen, line 66: This should probably moved to base class since this piece code is duplicated in multiple processor

NitroxClient/GameLogic/Bases/Spawning/BasePiece/MapRoomSpawnProcessor.cs
```diff
@@ -18,6 +18,8 @@ public class MapRoomSpawnProcessor : BasePieceSpawnProcessor
             TechType.BaseMapRoom
         };
 
+        protected override bool ShouldRerunSpawnProcessor => false;
```
dartasen, line 21: Default boolean value in c# is false

NitroxClient/MonoBehaviours/ThrottledBuilder.cs
```diff
@@ -274,6 +278,13 @@ private static GameObject FindFinishedPiece(Transform cellTransform)
             return null;
         }
 
+        private static bool AreSameTechType(TechType actual, TechType expected)
+        {
+            return actual.Equals(expected) || // normal case
+                   (actual.Equals(TechType.BaseConnector) && expected.Equals(TechType.BaseCorridor)) || // a specific case
+                   expected.ToString().Contains(actual.ToString()); // for cases like: actual = BaseCorridor but expected = BaseCorridorT
```
dartasen, line 285: Maybe you should split into multiple line instead of fitting logic  inside a single return, it's quite hard to read right now

NitroxClient/GameLogic/Bases/Spawning/BasePiece/BasePieceSpawnPrioritizer.cs
```diff
@@ -77,5 +81,54 @@ private int ComputeBasePiecePriority(BasePieceData basePiece)
 
             return piecesToPriority.TryGetValue(techType, out int position) ? position : int.MaxValue;
         }
+
+        private List<BasePieceData> OrderByHeight(IEnumerable<BasePieceData> orderedList, TechType techType)
```
dartasen, line 85: I don't know for sure how to improve this piece of code, but this should probably tested

NitroxClient/GameLogic/Bases/Spawning/BasePiece/BaseBioReactorSpawnProcessor.cs
tornac1234: This should be made in a future "sync reactors" PR

NitroxClient/GameLogic/Bases/Spawning/BasePiece/BasePieceSpawnPrioritizer.cs
```diff
@@ -77,5 +81,54 @@ private int ComputeBasePiecePriority(BasePieceData basePiece)
...
+            }
+            
+            specificBasePieces = specificBasePieces.OrderBy(piece => -piece.ItemPosition.Y).ToList();
+
+            for (int index = minIndex; index <= maxIndex; index++)
+            {
+                pieces[index] = specificBasePieces[index - minIndex];
+            }
+            
+            i = 0;
```
Coding-Hen, line 130: Assigning i back to 0 isn't needed due to returning immediately after

NitroxClient/GameLogic/Bases/Spawning/BasePiece/BasePieceSpawnPrioritizer.cs
```diff
@@ -77,5 +81,54 @@ private int ComputeBasePiecePriority(BasePieceData basePiece)
 
             return piecesToPriority.TryGetValue(techType, out int position) ? position : int.MaxValue;
         }
+
+        private List<BasePieceData> OrderByHeight(IEnumerable<BasePieceData> orderedList, TechType techType)
+        {
+            int i = 0;
+            int minIndex = -1;
+            int maxIndex = -1;
+            foreach (BasePieceData basePieceData in orderedList)
```
Coding-Hen, line 90: If you need an idex here would a traditional for loop not work just as well? Saying that I'm not sure on how c# will optimise the loop here on access to the list

NitroxClient/GameLogic/Bases/Spawning/BasePiece/BaseWaterParkProcessor.cs
```diff
@@ -30,6 +33,12 @@ protected override void SpawnPostProcess(Base latestBase, Int3 latestCell, GameO
             }
 
             WaterPark waterPark = waterParkPiece.GetWaterParkModule();
+            // When reruning the spawn processor, the module will not be found at first so we need to delay its detection
+            if (!waterPark)
+            {
+                DelayModuleDetection(latestBase, latestCell, finishedPiece);
```
Coding-Hen, line 39: Is this a correct call? Should it not be the same as you have before calling a new coroutine? Here you are calling a aync method but not waiting on it. I would have thought a compiler warning would be thrown if nothing else as I believe this is an unadvised call?

Coding-Hen (commented):
I need to do a test of this ingame but visually looks like it might solve some issues

NitroxClient/GameLogic/Bases/Spawning/BasePiece/BasePieceSpawnPrioritizer.cs
tornac1234: We can probably let it like this for now and when we have a better solution we'll correct it. Also remember this is only happening one time at the beginning

tornac1234:
There's a known issue about building hatches on a scanner room that is already connected to corridor

tornac1234:
From the latest commit, we'll need to decide if we want to save some operations or not (may be more "dangerous")

dartasen (approved):

PR #1820, opened 2022-06-05
Remove ProtoBuf as save file serializer

Jannify:
This pr removes `ProtoBuf` as a save file serializer while keeping it's functionality for the client and the server resource parser.
It also moves some classes in `NitroxServer.GameLogic` and `NitroxServer.Serialization` to better reflect their behavior.

Tested in-game with a small base and a seamoth.


NinjaPedroX (commented):
```
[PropertyDescription("Possible values:", typeof(ServerSerializerMode))]
public ServerSerializerMode SerializerMode { get; set; } = ServerSerializerMode.JSON;
```

Shouldn't we remove this in `Nitrox/NitroxServer/Serialization/Serverconfig.cs`?

dartasen:
I'm not sure that's a good idea right now

PR #1808, opened 2022-05-24
Add a World manager to the launcher

NinjaPedroX:
- Changed the "Servers" tab of the launcher into a save file manager
- Changed the location of the save folders to `C:\Users\(Username)\AppData\Roaming\Nitrox\saves`
- Changed the location of the "server.cfg" file to be in each save folder
- Moved save location of the ingame server list file "servers" to `C:\Users\(Username)\AppData\Roaming\Nitrox`
- Created WorldManager class for most of the world manager logic (Credits to @Measurity)

TODO:
- [x] - Make UI more user-friendly
- [x] - Order world listing based on when they were last used
- [x] - List worlds that are out of date, but don't allow them to be clicked (show message saying that they cannot be used)
	- [x] - Change the world version text to red in the world listing if it's an out-of-date version
- [x] - Get the randomly generated world seed to set itself in "server.cfg" when a seed isn't specified during world creation
- [x] - Create functionality to allow users to import save files from other locations (i.e. from versions previous to a release with the Save Manager, or when someone sends their friend their save file)
- [x] - Fix server messages to handle new save files (as in not displaying that the save files are invalid when first starting up a world)
- [x] - Make code more efficient

Many thanks to Measurity, Clément, TwinBuilderOne, and Shirkie01 for helping me make this PR!

Current Design:
World Manager Design

Old Design:
image
image

PR #1824, opened 2022-06-18
Improve Backup System and add UI to Restore Backups

NinjaPedroX:
NOTE: This PR contains almost all of the original commits from the [World Manager PR](https://github.com/SubnauticaNitrox/Nitrox/pull/1808) due to being created while dependent of that PR before it was merged, of which I have not been able to remove as I thought I would be able to.

- Made the server back-up the save file every time the server starts up with that selected world
- Added UI and functionality to allow the user to select and restore a backup from within the world settings
- Changed the backups directory to the "Backups" folder within each save file
- Allowed a maximum of only 5 backups
- Changed the server error message for corrupted worlds to tell the user to restore a backup, and it now backs-up the corrupted save files so that devs can troubleshoot

TODO:

- [x] - Add a "backup" command to the server console so that the user can manually back up their save
- [ ] - Fix backup system to be able to deal with save name changes (Changing the name of the world would invalidate the names of the already existing backups)
- [ ] - Allow the save import system to import the save's backup folder if it contains one

image
image
image


dartasen:
As discussed with rux, we'll focus on world-manager itself so we can improve your design and user experience, especially for non-dev. I opened a PR to replace all the icons to fit the design. In the meantime rux is working on your figma prototype so we can focus on the overhaul (you'll be able to work on it if you want).

Since we're getting closer every day to the Subnautica 2.0 update, I wanna make sure the launcher is ready so we can ship this feature with 1.7. For now it's better to focus on this particular feature to make it 100% perfect and then we'll look at backup.

I'm closing this for now

PR #1839, opened 2022-08-05
Added fallback when external IP returned by router is a private network IP

Measurity:
Subnet calculator: https://jodies.de/ipcalc


Nitrox.Test/Model/Helper/NetHelperTest.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System.Net;
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using NitroxModel.Helper;
+
+namespace Nitrox.Test.Model.Helper
```
Jannify, line 6: Would give it the same namespace as NetHelper.

Nitrox.Test/Model/Helper/NetHelperTest.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    [TestClass]
+    public class NetHelperTest
+    {
+        [TestMethod]
+        public void ShouldMatchPrivateIps()
+        {
+            IPAddress.Parse("192.168.2.1").IsPrivate().Should().BeTrue();
```
Jannify, line 14: Shouldn't we test all the ips (or just the start and end of the ranges) defined in NetHelper? Why only a few manually selected?

Measurity: Just laziness. Will add the others when I can.

Jannify (approved):

NitroxModel/Helper/NetHelper.cs
```diff
@@ -13,6 +15,17 @@ namespace NitroxModel.Helper
 {
     public static class NetHelper
     {
+        private static readonly ReadOnlyCollection<string> privateNetworks =
```
dartasen, line 18: Do we really need ReadOnlyCollection wrapper here ?

NitroxModel/Helper/NetHelper.cs
```diff
@@ -124,5 +137,31 @@ public static IPAddress GetHamachiIp()
...
+            {
+                string[] parts = mask.Split('/');
+
+                int ipNum = BitConverter.ToInt32(ipAddress.GetAddressBytes(), 0);
+                int cidrAddress = BitConverter.ToInt32(IPAddress.Parse(parts[0]).GetAddressBytes(), 0);
+                int cidrMask = IPAddress.HostToNetworkOrder(-1 << (32 - int.Parse(parts[1])));
+
+                return (ipNum & cidrMask) == (cidrAddress & cidrMask);
+            }
+
```
dartasen, line 156: Maybe we can check if the adress.IsIPv4MappedToIpv6 to convert the adress to a standard IPV4 using MapToIpv4() method, will allow us to cover more case

tornac1234 (approved):

NitroxModel/Helper/NetHelper.cs
Measurity: This seems too niche to be worth adding at this moment.

dartasen (approved):

PR #1842, opened 2022-08-09
Updates local state of muted player if it's not synced with server in case of desync

tornac1234:
We could use true/false in the packet instantiations but anyway


Measurity:
Did the desync happen or is this a precaution?

tornac1234:
> Did the desync happen or is this a precaution?

Both, Clement (dartasen) found a way to mess up the system and come to that desync without even packet loss. But as we know, packet loss in itself is a reason to let the system unblock itself in case

dartasen (approved):

PR #1848, opened 2022-08-23
Improve linux support (new launcher)

### LINKED ISSUE #988: [Core] Linux Support
- [x] Remove/Replace usage of the windows registry
- [x] Wait for UWE to release the mod-pocalypse update
- [x] Implement the cross platform launcher

Measurity:
- [x] Allow build to complete with `dotnet build` on [dotnet SDK 8+](https://dotnet.microsoft.com/en-us/download)
 - [x] Alternative to `NitroxLauncher` to play Subnautica with Nitrox.
 - [x] Remove original launcher project (`NitroxLauncher`)
 - [x] Fix solution build errors/warnings (individual projects succeed)
 - [x] Add instant launch to new launcher (when this PR is merged #2154)
 - [x] Design review from @zanethefox 
 - [x] MacOS support @dartasen  
 - [x] Use dark theme for pop-ups
 - [x] Add loading indicators for IO tasks as they can take unpredictable amounts of time (e.g. anti virus interrupting file writes)
 - [x] Add embedded server option as it was in old launcher
 - [x] Don't show errors immediately for modals (wait for user to defocus the input)

This PR includes solution-wide changes to multi-target for .net 9 (server, launcher) and .net 4.7.2 (game client), allowing benefits of modern .net where possible.

Closes #988

## Resources to assist with implementation
- https://lutris.net/games/install/27324/view: Lutris config used by some for Nitrox 1.6


NinjaPedroX:
With that, we present to you the Cross-Platform launcher. Have fun reviewing this 🌚

IsaiahPetrichor:
I have just tested this build on Linux, it does not support server save files from earlier versions or launch on Linux. While the new launch looks and feels great it fails to address the original issue of Linux support. The current version of the launcher is more functional on Linux, however, it fails to detect a running Steam instance so it doesn't launch the game. If we just fixed the code for detecting Steam it should let us avoid rewriting the entire launcher.

NinjaPedroX:
> I have just tested this build on Linux, it does not support server save files from earlier versions or launch on Linux.
Nitrox v1.8.0.0 is not meant to support older save files since there were major changes made to the build system.

Schweeeeeeeeeeeeeeee:
@zanethefox bitte guck mal duch <3 und approve des

NitroxModel/Platforms/OS/Shared/ProcessEx.cs
```diff
@@ -102,21 +147,55 @@ public static ProcessEx GetFirstProcess(string procName, Func<ProcessEx, bool> p
...
+                    ProcessEx processEx = new ProcessEx(pid);
+                    if (comparer.Compare(procName, processEx.Name) == 0)
+                    {
+                        if (predicate == null || predicate(processEx))
+                        {
+                            return processEx;
+                        }
+                    }
+                }
+                catch (Exception)
```
dartasen: @jumpers775 Hello there, just wanted to know the reason behind that change, why do we need our implementation for getting process by name ?
Which is already implemented by the .NET SDK with Process.GetProcessesByName, tried on Windows, Linux, MacOS and it works like a charm

Noticed that because I tried building a version from a mac, and was getting an error that process_listpids doesn't exist (https://github.com/apple-opensource/xnu/blob/24525736ba5b8a67ce3a8a017ced469abe101ad5/libsyscall/wrappers/libproc/libproc.c#L44)

Thanks 😄

jumpers775: this PR was my introduction to c# so I just didn't know about that. my bad.

zanethefox (approved):

VazarinX:
:soonTM:

Nitrox.Launcher/Models/Extensions/ScreenExtensions.cs
```diff
@@ -0,0 +1,83 @@
...
+            navigationStack.RemoveAllFast(screen.ActiveViewModel, (item, param) => item.GetType() == param.GetType());
+            navigationStack.Add(routableViewModelBase);
+        }
+        Stopwatch sw = Stopwatch.StartNew();
+        Task contentLoadTask = routableViewModel.ViewContentLoadAsync();
+        // Only show loading screen if page isn't loading super quickly.
+        await Task.Delay(50);
+        if (!contentLoadTask.IsCompleted)
+        {
+            screen.ActiveViewModel = "Content is loading.. please wait";
```
dartasen, line 40: I'd rather have a WaitIndicator or a Spinner, the loading text is quite confusing.
Or maybe we could use some Shimmers view on heavy data loading

Nitrox.Launcher/ViewModels/ServersViewModel.cs
```diff
@@ -0,0 +1,230 @@
...
+        watcher.Deleted += OnDirectoryChanged;
+        watcher.Renamed += OnDirectoryChanged;
+
+        await Task.Run(async () =>
+        {
+            watcher.EnableRaisingEvents = true; // Slowish (~2ms) - Moved into Task.Run.
+
+            while (true)
+            {
+                cancellationToken.ThrowIfCancellationRequested();
```
dartasen, line 208: I would rather check for IsCancellationRequested to gracefully shut down the Task, we don't really need to take advantage of the exception system here

Nitrox.Launcher/ViewModels/ServersViewModel.cs
```diff
@@ -0,0 +1,230 @@
...
+                while (shouldRefreshServersList)
+                {
+                    try
+                    {
+                        await GetSavesOnDiskAsync();
```
dartasen, line 213: Maybe we should feed it with the cancellation token as well ?

Nitrox.Launcher/Models/Converters/DateToRelativeDateConverter.cs
```diff
@@ -0,0 +1,44 @@
...
+        DateTimeOffset date = value switch
+        {
+            DateTime dateTime => dateTime,
+            DateTimeOffset dateTimeOffset => dateTimeOffset,
+            DateOnly dateOnly => dateOnly.ToDateTime(TimeOnly.MinValue),
+            string text when DateTimeOffset.TryParse(text, out DateTimeOffset offset) => offset,
+            _ => throw new ArgumentException($"Value must be a {nameof(DateTime)} or {nameof(DateTimeOffset)}", nameof(value))
+        };
+
+        TimeSpan delta = DateTimeOffset.UtcNow - date;
```
dartasen, line 25: is date always a UTC ?

Nitrox.Launcher/Models/Design/MultiDataTemplate.cs
```diff
@@ -0,0 +1,46 @@
...
+using Avalonia.Markup.Xaml.Templates;
+using Avalonia.Metadata;
+using JB::JetBrains.Annotations;
+
+namespace Nitrox.Launcher.Models.Design;
+
+/// <summary>
+///     Selects a <see cref="DataTemplate" /> based on its <see cref="DataTemplate.DataType" />.
+/// </summary>
+public class MultiDataTemplate : AvaloniaList<DataTemplate>, IDataTemplate
```
dartasen, line 15: Didn't look at how it's used so I might be missing a point, but what's the difference between that and a DataTemplateSelector ?

Nitrox.Launcher/Models/Design/NitroxChangelog.cs
```diff
@@ -0,0 +1,30 @@
...
+
+namespace Nitrox.Launcher.Models.Design;
+
+[Serializable]
+public class NitroxChangelog
+{
+    public string Version { get; }
+
+    public DateTime Released { get; }
+
```
dartasen, line 11: Could be turned into a record I guess

Nitrox.Launcher/Models/Utils/GameInspect.cs
```diff
@@ -0,0 +1,67 @@
...
+    ///     Check to ensure the Subnautica is not in legacy.
+    /// </summary>
+    public static async Task<bool> IsOutdatedGameAndNotify(string gameInstallDir, IDialogService dialogService = null)
+    {
+        try
+        {
+            ArgumentException.ThrowIfNullOrWhiteSpace(gameInstallDir);
+            
+            string gameVersionFile = Path.Combine(gameInstallDir, GameInfo.Subnautica.DataFolder, "StreamingAssets", "SNUnmanagedData", "plastic_status.ignore");
+            if (int.TryParse(await File.ReadAllTextAsync(gameVersionFile), out int gameVersion) && gameVersion <= 68598)
```
dartasen, line 24: Maybe the version is something we might wanna have in GameInfo or whatever ?

Nitrox.Launcher/AppViewLocator.cs
```diff
@@ -0,0 +1,63 @@
...
+
+namespace Nitrox.Launcher;
+
+public sealed class AppViewLocator : ViewLocatorBase, ReactiveUI.IViewLocator
+{
+    private static IServiceProvider serviceProvider;
+
+    public AppViewLocator(IServiceProvider serviceProvider)
+    {
+        if (serviceProvider is IServiceScope)
```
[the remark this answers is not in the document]
Measurity: Indeed an Avalonia thing. If the `AppViewLocator` is used in AXAML, Avalonia will provide its own DI service provider. But we no longer use it in AXAML directly, so I'll remove if check.

Nitrox.Launcher/Models/Converters/PlatformToIconConverter.cs
```diff
@@ -0,0 +1,22 @@
...
+public class PlatformToIconConverter : Converter<PlatformToIconConverter>
+{
+    public override object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+    {
+        return BitmapAssetValueConverter.GetBitmapFromPath(GetIconPathForPlatform(value as Platform?));
+    }
+
+    private static string GetIconPathForPlatform(Platform? platform) => platform switch
+    {
+        Platform.EPIC => "/Assets/Images/store-icons/epic.png",
```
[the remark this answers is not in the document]
Measurity: True it's not ideal. I think we should change it to how `.notifications` style is done in MainWindow.axaml which uses conditional classes to select the icon we want.

Nitrox.Launcher/ViewModels/ServersViewModel.cs
Measurity: This is a .NET API thing and how `Task.Delay` works too. If `IsCancellationRequested` is used, the task will not have a canceled state as it exited normally, so it's better to throw. We could use `Task.FromCanceled` but then you'd still need to `try...catch` anyway in case a `Task.Delay` was canceled.

Nitrox.Launcher/Models/Design/MultiDataTemplate.cs
Measurity: WPF has that but Avalonia doesn't provide one by default, see bottom of page here <https://docs.avaloniaui.net/docs/get-started/wpf/datatemplates#data-template-selector>

NitroxModel/Helper/NitroxUser.cs
```diff
@@ -122,5 +162,53 @@ public static string GamePath
...
+                {
+                    return currentExecutablePath;
+                }
+
+                // File URI works different on Linux/OSX so just return path directly.
+                if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                {
+                    return Path.GetDirectoryName(Assembly.GetEntryAssembly()?.Location ?? ".") ?? Directory.GetCurrentDirectory();
+                }
+                return currentExecutablePath = new Uri(Path.GetDirectoryName(Assembly.GetEntryAssembly()?.CodeBase ?? Assembly.GetEntryAssembly()?.Location ?? ".") ?? Directory.GetCurrentDirectory()).LocalPath;
```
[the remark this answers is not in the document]
Measurity: Will be fixed in next commit I'm working on.

Nitrox.Launcher/Models/Utils/GameInspect.cs
Measurity: Think I'll leave this as is for. This code isn't proper for GameInfo support as is and want to see what HeN is doing.

OhmV-IR:
I think we were talking about putting the port checker in launcher, see https://github.com/SubnauticaNitrox/Nitrox/pull/2210 and https://discord.com/channels/525437013403631617/528915769107415040/1319975520946884609 Not a dealbreaker for launcher merge though can definitely wait to integrate until after new launcher is merged, this PR is already a pretty big monolith

NinjaPedroX:
> I think we were talking about putting the port checker in launcher, see https://github.com/SubnauticaNitrox/Nitrox/pull/2210 and https://discord.com/channels/525437013403631617/528915769107415040/1319975520946884609 Not a dealbreaker for launcher merge though can definitely wait to integrate until after new launcher is merged, this PR is already a pretty big monolith

Probably something for a later PR. The focus of this one is just to rewrite the old launcher with cross-platform compatibility + better world management stuff and other improvements.

dartasen (approved):

Dinhero21:
from what I understand, from this PR being merged, it means V1.8.0.0 will have native Linux support?

RivalsRebels:
> from what I understand, from this PR being merged, it means V1.8.0.0 will have native Linux support?

Yes, it will natively support it.

PR #1867, opened 2022-10-11
Fix player color for different players

dartasen:
Fix an issue where joining players would have the same color as you while their ping has the correct color.

This is achieved by early init of the body prototype before the first color swapping operation. The body prototype shouldn't be colored in the first place since color swapping operations is mapped on the original texture pixels (which are orange)

PS: Might be worth to investigate in a cache for SwapColorOperation since it's the exact same process everytime


tornac1234 (approved):
Tested it with 4 local players and it worked fine
image

Coding-Hen (approved):
My only question is if we want to have the logging  other than testing?

dartasen:
The logging is just here to avoid the "unused warning" thing

Issue #1841, opened 2022-08-09
Duplicate items on /back


prodigyrick:
### Are you on the latest version of Nitrox?

Yes, I'm on the latest commit of the repo (only for experts).

### Which OS are you playing on?

Windows

### On which platform did you bought Subnautica?

Steam

### Describe the issue.

Type /back after death.

### Describe the steps you took to encounter the bug.

Friend typed /back after dying. Ended up with duped items.

### Relevant log output

Game: http://improdigy.com/files/game-20220809.log
Server: http://improdigy.com/files/server-20220809.log

tornac1234:
I can confirm that bug

dartasen:
@ywgATustcbbs we did not adress this issue yet

PR #1878, opened 2022-10-24
Fix some occurrences of player inventory clear

tornac1234:
- [x] Prevent players to build on other players, or to destroy a building with players inside
> When the RemotePlayer object is destroyed, all of its inventory is destroyed too and will send parasite packets with fake informations (like item removal)
- [x] Prevent a player to modify another player's inventory

### DIFF
--- a/Nitrox.Assets.Subnautica/LanguageFiles/en.json
+++ b/Nitrox.Assets.Subnautica/LanguageFiles/en.json
@@ -65,5 +65,5 @@
     "Nitrox_WaitingPassword": "Waiting for server password input…",
     "Nitrox_WaitingUserInput": "Waiting for user input…",
     "Nitrox_WorldSettling": "Awaiting World Settling…",
-    "RemotePlayerObstacle": "Another player's either inside or too close to the deconstructable target."
+    "Nitrox_RemotePlayerObstacle": "Another player's either inside or too close to the deconstructable target."
 }
--- a/Nitrox.Assets.Subnautica/LanguageFiles/en.json
+++ b/Nitrox.Assets.Subnautica/LanguageFiles/en.json
@@ -64,5 +64,6 @@
     "Nitrox_Waiting": "Waiting in join queue…",
     "Nitrox_WaitingPassword": "Waiting for server password input…",
     "Nitrox_WaitingUserInput": "Waiting for user input…",
-    "Nitrox_WorldSettling": "Awaiting World Settling…"
+    "Nitrox_WorldSettling": "Awaiting World Settling…",
+    "RemotePlayerObstacle": "Another player's either inside or too close to the deconstructable target."
 }
--- a/NitroxClient/GameLogic/PlayerLogic/RemotePlayerIdentifier.cs
+++ b/NitroxClient/GameLogic/PlayerLogic/RemotePlayerIdentifier.cs
@@ -1,0 +1,17 @@
+﻿using UnityEngine;
+
+namespace NitroxClient.GameLogic.PlayerLogic;
+
+/// <summary>
+/// Attached to a RemotePlayer. Useful to determine that this script's GameObject is in the root of a RemotePlayer.
+/// </summary>
+public class RemotePlayerIdentifier : MonoBehaviour, IObstacle
+{
+    public RemotePlayer RemotePlayer;
+
+    public bool CanDeconstruct(out string reason)
+    {
+        reason = Language.main.Get("RemotePlayerObstacle");
+        return false;
+    }
+}
--- a/NitroxClient/GameLogic/PlayerLogic/RemotePlayerIdentifier.cs
+++ b/NitroxClient/GameLogic/PlayerLogic/RemotePlayerIdentifier.cs
@@ -3,15 +3,18 @@
 namespace NitroxClient.GameLogic.PlayerLogic;
 
 /// <summary>
-/// Attached to a RemotePlayer. Useful to determine that this script's GameObject is in the root of a RemotePlayer.
+/// Attached to a RemotePlayer. Useful to determine that this script's GameObject is in the EntityRoot of a RemotePlayer.
 /// </summary>
+/// <remarks>
+/// The EntityRoot of an object is defined as the top most GameObject as when an object hierarchy first was spawned in. Either from a prefab, or in Nitrox' case, a cloned root game object.
+/// </remarks>
 public class RemotePlayerIdentifier : MonoBehaviour, IObstacle
 {
     public RemotePlayer RemotePlayer;
 
     public bool CanDeconstruct(out string reason)
     {
-        reason = Language.main.Get("RemotePlayerObstacle");
+        reason = Language.main.Get("Nitrox_RemotePlayerObstacle");
         return false;
     }
 }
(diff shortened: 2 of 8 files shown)


tornac1234:
Cool thing to test is that RemotePlayerIdentifier is a MonoBehaviour so you can add anything inside to be related to the RemotePlayer instance. For example to test this PR you can add the following lines and use the debugger to verify if destroying the player will no longer send fake packets.
```c#
public void DestroyNow()
{
    Destroy(gameObject);
}
```

Nitrox.Assets.Subnautica/LanguageFiles/en.json
```diff
@@ -64,5 +64,6 @@
     "Nitrox_WaitingUserInput": "Waiting for user input…",
     "Nitrox_WorldSettling": "Awaiting World Settling…",
     "Nitrox_ThankForPlaying": "Thanks for using Nitrox!",
-    "Nitrox_Failure": "An error occured"
+    "Nitrox_Failure": "An error occured",
+    "RemotePlayerObstacle": "Another player's either inside or too close to the deconstructable target."
```
Measurity, line 68: All other keys start with `Nitrox_`
```suggestion
    "Nitrox_RemotePlayerObstacle": "Another player's either inside or too close to the deconstructable target."
```

NitroxClient/GameLogic/PlayerLogic/RemotePlayerIdentifier.cs
```diff
@@ -0,0 +1,17 @@
...
+/// <summary>
+/// Attached to a RemotePlayer. Useful to determine that this script's GameObject is in the root of a RemotePlayer.
+/// </summary>
+public class RemotePlayerIdentifier : MonoBehaviour, IObstacle
+{
+    public RemotePlayer RemotePlayer;
+
+    public bool CanDeconstruct(out string reason)
+    {
+        reason = Language.main.Get("RemotePlayerObstacle");
```
Measurity, line 14:
```suggestion
        reason = Language.main.Get("Nitrox_RemotePlayerObstacle");
```

NitroxClient/GameLogic/RemotePlayer.cs
```diff
@@ -283,6 +287,45 @@ private void UpdateEquipmentVisibility()
...
+                // This layer lets us have a collider as a trigger without preventing its detection as an obstacle
+                Body.layer = LayerID.Useable;
+                Collider = Body.AddComponent<CapsuleCollider>();
+                
+                Collider.center = Vector3.zero;
+                Collider.radius = refCollider.radius;
+                Collider.direction = refCollider.direction;
+                Collider.contactOffset = refCollider.contactOffset;
+                Collider.isTrigger = true;
+            }
```
Jannify, line 309: We probably should log if it fails the if

NitroxClient/GameLogic/RemotePlayer.cs
```diff
@@ -283,6 +287,45 @@ private void UpdateEquipmentVisibility()
...
+        }
+
+        /// <summary>
+        /// Change two parameters of the collider depending on the state of the player
+        /// </summary>
+        private void UpdateCollider(AnimChangeType type, AnimChangeState state)
+        {
+            if (type == AnimChangeType.UNDERWATER && state == AnimChangeState.ON)
```
Jannify, line 317: That way if the type is not UNDERWATER we always jump into the else, is this intended?

NitroxClient/GameLogic/PlayerLogic/RemotePlayerIdentifier.cs
```diff
@@ -0,0 +1,17 @@
+﻿using UnityEngine;
+
+namespace NitroxClient.GameLogic.PlayerLogic;
+
+/// <summary>
+/// Attached to a RemotePlayer. Useful to determine that this script's GameObject is in the root of a RemotePlayer.
+/// </summary>
```
Measurity, line 7: Might be good to explain what a root object is, for new devs especially.
```suggestion
/// <summary>
/// Attached to a RemotePlayer. Useful to determine that this script's GameObject is in the root of a RemotePlayer.
/// </summary>
/// <remarks>
/// The root of an object is defined as the top most GameObject as when an object hierarchy first was spawned in. Either from a prefab, or in Nitrox' case, a cloned root game object.
/// </remarks>
```

NitroxPatcher/Patches/Dynamic/Base_OnPreDestroy_Patch.cs
```diff
@@ -0,0 +1,29 @@
...
+public class Base_OnPreDestroy_Patch : NitroxPatch, IDynamicPatch
+{
+    private static readonly MethodInfo TARGET_METHOD = Reflect.Method((Base t) => t.OnPreDestroy());
+
+    public static void Prefix(Base __instance)
+    {
+        foreach (RemotePlayerIdentifier remotePlayerIdentifier in __instance.GetComponentsInChildren<RemotePlayerIdentifier>(true))
+        {
+            remotePlayerIdentifier.RemotePlayer.Detach();
+            SkyEnvironmentChanged.Broadcast(remotePlayerIdentifier.RemotePlayer.Body, (GameObject)null);
```
Jannify, line 21: Minor: What about moving this line into Detach()?

NitroxPatcher/Patches/Dynamic/Utils_GetEntityRoot_Patch.cs
```diff
@@ -0,0 +1,30 @@
...
+{
+    private static readonly MethodInfo TARGET_METHOD = Reflect.Method(() => UWE.Utils.GetEntityRoot(default));
+
+    public static bool Prefix(GameObject go, ref GameObject __result)
+    {
+        if (go.TryGetComponent(out RemotePlayerIdentifier remotePlayerIdentifier) || go.TryGetComponentInParent(out remotePlayerIdentifier))
+        {
+            __result = remotePlayerIdentifier.gameObject;
+        }
+        return __result == null;
```
Jannify, line 23: Would be better to return true/false to not compare gameobjects directly to null.

Jannify (changes requested):

NitroxClient/GameLogic/RemotePlayer.cs
tornac1234: yes, I'll modify the comment above to explain that there are two collisions behaviours depending on whether or not the player is underwater

NitroxPatcher/Patches/Dynamic/Base_OnPreDestroy_Patch.cs
tornac1234: In this case, I'll also move the same behaviour (but adapted to the furnished component) in Attach()

dartasen:
Can probably fix #1841

tornac1234:
> Can probably fix #1841

I could reproduce the duplicate items even alone in the server, so I'm pretty sure that this PR won't fix it

tornac1234:
Tests proved it worked as intended

Jannify (approved):

dartasen (approved):

PR #1881, opened 2022-10-25
Changed join process to be async

Measurity:


NitroxClient/MonoBehaviours/Gui/MainMenu/MainMenuMultiplayerPanel.cs
```diff
@@ -76,10 +78,21 @@ private void CreateServerButton(string text, string joinIp, string joinPort, boo
...
+                    return;
+                }
+                isJoining = true;
+                
                 txt.GetComponent<Text>().color = prevTextColor; // Visual fix for black text after click (hover state still active)
-                OpenJoinServerMenu(joinIp, joinPort);
+                await OpenJoinServerMenuAsync(joinIp, joinPort)
+                    .ContinueWith(t =>
+                    {
+                        isJoining = false;
```
Coding-Hen, line 94: With false only being set if this completes, if the connection errors out, will isJoining be stuck in true and therefore stop you connecting to another server?

Measurity: No, `ContinueWith` is always called. The `t` variable has the state with which it completed (successful or faulted) but in either case it's not joining anymore.

Coding-Hen: Okay sounds good, I just wanted to check that was how the functionality behaved

Coding-Hen (approved):
I haven't tested in game but code wise seems fine. I can't see this causing in-game issues, if anything I suspect it will unfreeze the ui

dartasen (approved):
Tested IG, it does fix some freeze you can experience while connecting to a server

PR #1884, opened 2022-10-28
Sync fire extinguishers FX

tornac1234:
Only issue is player rotation. On the local player, we are able to turn around as much as we want and spray the fire extinguisher everywhere. But remote players don't have this rotation thing. They can only rotate to the lower side (a little angle is possible only). So we'll always see the extinguisher facing downwards.

EDIT: The fix is to only rotate the FX

~~And this branch is built on #1878~~

https://youtu.be/LyenVi8g9fk

To be reviewed and looked at after #2012


Jannify:
Needs rebase since #1878 is merged

tornac1234:
I'll probably look a bit more into it concerning the fire extinguisher's rotation

Jannify (changes requested):

tornac1234:
Will wait for a movement fix PR to be merged to continue this one

dartasen:
Closing it for now

PR #1885, opened 2022-10-28
Fixed regression of cancel button functionality

Measurity:


tornac1234 (approved):
Tests have proven this works

PR #1894, opened 2022-11-05
Changed packet suppression to use stack trace instead of manual def

Measurity:
- [x] Fix stack trace iteration in packet send implementation not returning correct true/false for all situations
 - [x] I removed the suppression for movement on vehicle docking. This breaks animations. ~~Investigating alternative of blocking movement process during an animation.~~ **Skipped for now (added TODO), vehicle spawning is broken and the way docking animation was fixed was unconventional (should be implemented using injected code at proper end of dock animation).**
- [x] ~~Fix regression sometimes getting stuck when mounting a vehicle platform.~~ same issue on master..

Reviewers: 
1. See https://github.com/SubnauticaNitrox/Nitrox/pull/1894/files#diff-dea4cf62b89b41b7bff3bc6660f6ed2b6c8291498d0aad707148a9a695141f55R126-R162 for main change.

This check takes about 20 to 30 microseconds (0.0030ms) in DEBUG build, though it improves maintainability and readability of packet sending (no longer manually checking if suppress is necessary or redundant).


NitroxClient/Communication/MultiplayerSession/MultiplayerSessionManager.cs
```diff
@@ -116,24 +118,48 @@ public void Disconnect()
             }
         }
 
-        public bool Send(Packet packet)
+        public void Send<T>(T packet) where T : Packet
+        {
+            Client.Send(packet);
+        }
+        
+        public bool SendIfGameCode<T>(T packet) where T : Packet
```
dartasen, line 126: Why does it need to be generic ?

Measurity: It doesn't need to be. But it does allow you to specify the type while sending the packet like this:
```cs
packetSender.Send<ServerStopped>(new());
```

NitroxPatcher/Patches/Dynamic/PDAScanner_NotifyProgress_Patch.cs
```diff
@@ -13,7 +16,7 @@ public static void Prefix(PDAScanner.Entry entry)
         {
             if (entry != null)
             {
-                Resolve<PDAManagerEntry>().Progress(entry, null);
+                Resolve<IPacketSender>().SendIfGameCode<PDAEntryProgress>(new(entry.techType.ToDto(), entry.progress, entry.unlocked, null));
```
Jannify, line 19:
```suggestion
                SendPacket<PDAEntryProgress>(new(entry.techType.ToDto(), entry.progress, entry.unlocked, null));
```

Jannify (changes requested):

PR #1902, opened 2022-11-11
Added simple rigidbody editor, added hierarchy searchbar...

Shirkie01:
… for searching through the hierarchy when attempting to find a gameobject with a specific name or component. The search will start at the selected gameobject, or if there is none selected, search all. The search runs on a separate thread so it won't block the UI, and clicking the `X` will cancel the search.

Rigidbody Editor:
image

Hierarchy Searchbar:
image


NitroxClient/Debuggers/SceneDebugger.cs
```diff
@@ -213,7 +222,73 @@ private void RenderTabHierarchy()
...
+                            isSearching = false;
+                            currentSearchItem = string.Empty;
+                        }, searchTaskCancelTokenSource.Token);
+                    }
+                    // Make sure to re-enable the GUI for everyone else
+                    GUI.enabled = true;
+
+                    // Display some sort of indication to the user that we haven't just frozen up
+                    GUILayout.Label("Current Search Item: " + currentSearchItem);                    
+                }
```
Measurity, line 285: The hierarchy tab is getting complicated. Could you refactor the code to use a new inner class (to the SceneDebugger) that handles the filtering? Similarly to the `IDrawer` system but just for the search bar.

NitroxClient/MonoBehaviours/NitroxDebugManager.cs
```diff
@@ -111,10 +111,10 @@ private void DoWindow(int windowId)
                 }
             }
             if (showDebuggerList)
-            {
+            {                
                 foreach (BaseDebugger debugger in debuggers)
-                {
-                    debugger.Enabled = GUILayout.Toggle(debugger.Enabled, $"{debugger.DebuggerName} debugger{debugger.HotkeyString}");
+                {
```
Jannify, line 116:
```suggestion
                {
```

NitroxClient/Debuggers/Drawer/Unity/RigidbodyDrawer.cs
```diff
@@ -0,0 +1,50 @@
...
+            {
+                GUILayout.Label("Angular Velocity", NitroxGUILayout.DrawerLabel, GUILayout.Width(LABEL_WIDTH));
+                NitroxGUILayout.Separator();
+                VectorDrawer.DrawVector3(rb.velocity, VECTOR_MAX_WIDTH);
+            }
+
+            using (new GUILayout.HorizontalScope())
+            {
+                GUILayout.Label("Mass", NitroxGUILayout.DrawerLabel, GUILayout.Width(LABEL_WIDTH));
+                if (float.TryParse(GUILayout.TextField(rb.mass.ToString()), out float newMass))
```
Jannify, line 37: Can you set the width of the float field so it is as long as the Vector3 above?

Jannify (changes requested):

NitroxClient/Debuggers/Drawer/Unity/RigidbodyDrawer.cs
Shirkie01: Done

NitroxClient/MonoBehaviours/NitroxDebugManager.cs
Shirkie01: Removed

NitroxClient/Debuggers/Drawer/Unity/RigidbodyDrawer.cs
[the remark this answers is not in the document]
Shirkie01: Changed to use the existing Float field

NitroxClient/MonoBehaviours/NitroxDebugManager.cs
```diff
@@ -16,7 +16,7 @@ public class NitroxDebugManager : MonoBehaviour
     private List<BaseDebugger> debuggers;
     private readonly HashSet<BaseDebugger> prevActiveDebuggers = new();
 
-    private bool showDebuggerList;
```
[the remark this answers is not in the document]
Shirkie01: Removed

Shirkie01:
Since the functionality I was looking for already existed in the Scene Extras debugger, I've instead added pagination to that search.

image

NitroxClient/Debuggers/Drawer/Unity/RigidbodyDrawer.cs
```diff
@@ -0,0 +1,48 @@
...
+                NitroxGUILayout.Separator();
+                VectorDrawer.DrawVector3(rb.velocity, VALUE_MAX_WIDTH);
+            }
+
+            using (new GUILayout.HorizontalScope())
+            {
+                GUILayout.Label("Angular Velocity", NitroxGUILayout.DrawerLabel, GUILayout.Width(LABEL_WIDTH));
+                NitroxGUILayout.Separator();
+                VectorDrawer.DrawVector3(rb.velocity, VALUE_MAX_WIDTH);
+            }
```
Jannify, line 32: Normally we only expose all fields (if possible) that are also shown in the Unity Editor (see [here](https://docs.unity3d.com/2019.4/Documentation/Manual/class-Rigidbody.html) for rigidbody). So if you have the time it would be great if you could implement the rest of the fields. And as velocity and angular velocity are not part of this I would move them to the end.

NitroxClient/Debuggers/Drawer/Unity/RigidbodyDrawer.cs
```diff
@@ -0,0 +1,48 @@
...
+                GUILayout.Label("Velocity", NitroxGUILayout.DrawerLabel, GUILayout.Width(LABEL_WIDTH));
+                NitroxGUILayout.Separator();
+                VectorDrawer.DrawVector3(rb.velocity, VALUE_MAX_WIDTH);
+            }
+
+            using (new GUILayout.HorizontalScope())
+            {
+                GUILayout.Label("Angular Velocity", NitroxGUILayout.DrawerLabel, GUILayout.Width(LABEL_WIDTH));
+                NitroxGUILayout.Separator();
+                VectorDrawer.DrawVector3(rb.velocity, VALUE_MAX_WIDTH);
```
Jannify, line 31: I think `rb.velocity` should be `rb.angularVelocity`

NitroxClient/Debuggers/SceneExtraDebugger.cs
```diff
@@ -76,27 +81,73 @@ private void RenderTabTools()
...
                 {
-                    if (index > 30)
+                    hierarchyScrollPos = scroll.scrollPosition;
+
+                    int startIndex = resultsPerPage * searchPage;
+                    for (int index = startIndex; index < gameObjectResults.Count; index++)
                     {
-                        GUILayout.Label($"There are {gameObjectResults.Count - index} more results which aren't displayed", "fillMessage");
-                        break;
+                        if (index > startIndex + resultsPerPage)
```
Jannify, line 93: I would swap the condition of the for loop and the if. The for is counting from startIndex to resultsPerPage while the if should break the loop before the array access if we are out of bounds.

NitroxClient/Debuggers/SceneExtraDebugger.cs
```diff
@@ -76,27 +81,73 @@ private void RenderTabTools()
...
+                                searchPage = 0;
                             }
                         }
+                        GUI.enabled = true;
+
+                        // Get the maximum page number based on the size of the results
+                        int maxPage = gameObjectResults.Count / resultsPerPage;
+
+                        GUILayout.FlexibleSpace();
+                        GUILayout.Label("Page " + (searchPage + 1) + " of " + (maxPage + 1), GUILayout.ExpandHeight(true));
```
Jannify, line 134: Would use string interpolation

NitroxClient/Debuggers/SceneExtraDebugger.cs
```diff
@@ -3,6 +3,7 @@
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Text.RegularExpressions;
+using NitroxClient.Debuggers.Drawer;
```
Jannify, line 6: Is this still needed?

Jannify (changes requested):

NitroxClient/Debuggers/Drawer/Unity/RigidbodyDrawer.cs
Shirkie01: Added all fields except freeze rotation by xyz and freeze rotation xyz. For some reason those are not part of the Rigidbody API, but given that they can be set through the transform I don't view this as a problem.

NitroxClient/Debuggers/Drawer/Unity/RigidbodyDrawer.cs
Shirkie01: Corrected. Also made sure to actually assign it

NitroxClient/Debuggers/SceneExtraDebugger.cs
Shirkie01: Nope. Gone.

Shirkie01:
Adding my own comment here: Apparently the EnumPopup was never implemented, so using it for the Rigidbody doesn't work. Will look into this more.

Jannify (changes requested):
Code wise seems fine (besides the one change I request). Need to see how it looks and interacts in game now.

NitroxClient/Debuggers/Drawer/Unity/RigidbodyDrawer.cs
```diff
@@ -0,0 +1,97 @@
...
+                NitroxGUILayout.Separator();
+                rb.velocity = VectorDrawer.DrawVector3(rb.velocity, VALUE_MAX_WIDTH);
+            }
+
+            using (new GUILayout.HorizontalScope())
+            {
+                GUILayout.Label("Angular Velocity", NitroxGUILayout.DrawerLabel, GUILayout.Width(LABEL_WIDTH));
+                NitroxGUILayout.Separator();
+                rb.angularVelocity = VectorDrawer.DrawVector3(rb.angularVelocity, VALUE_MAX_WIDTH);
+            }
```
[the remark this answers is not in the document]
Shirkie01: Made these be read-only.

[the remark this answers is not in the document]
Jannify: Making `DrawVector3Label()` wasn't necessary :D Since vector is a struct you just need to change

`rb.velocity = VectorDrawer.DrawVector3(rb.velocity, VALUE_MAX_WIDTH);`
to
`VectorDrawer.DrawVector3(rb.velocity, VALUE_MAX_WIDTH);`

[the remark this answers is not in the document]
Shirkie01: Are you sure we want to have editable textboxes that can't be edited?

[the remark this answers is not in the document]
Jannify: Yes. It's at least better then the current label and used in NitroxTransformDrawer.

Jannify:
I have problems with the current solution with the EnumDrawer:
- Displaying all Enum values is very very dangerous. `TechType` for example has hundreds of values.
- When selecting a `Pickupable` MB of an item the drawer crashes, probably bc of the issue above.

NitroxClient/Debuggers/Drawer/Unity/RigidbodyDrawer.cs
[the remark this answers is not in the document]
Shirkie01: Done

Shirkie01:
The reason TechType was crashing was not due to the large amount of values, but instead due to SceneDebugger using the `Enum` class. That has been resolved. I added a "Show / Hide" button for any raw Enum types that defaults to false. This ensures that a user doesn't have to scroll through the entire list if they are looking for some other field.

Since the visibility has to be per component and per field, this was done by adding another visibility dictionary to SceneDebugger.

Jannify (approved):

dartasen (approved):

Issue #1903, opened 2022-11-24
Removed base pieces do not stay removed on the server.


CoreyShay:
### Are you on the latest version of Nitrox?

Yes, I'm on the latest commit of the repo (only for experts).

### Which OS are you playing on?

Windows

### On which platform did you buy Subnautica?

Steam

### Describe the issue.

Any base piece that is removed on clients will not stay removed if the client is restarted and re-joins the server.

### Describe the steps you took to encounter the bug.

1. Construct a base piece (any).
2. Remove the base piece.
3. Quit and restart the client and rejoin the server.

Results:
Base piece is still present after re-joining the server.

Note in game log, client was run twice. First run already contains a single base piece. It was removed during that run. 2nd run reports the same message:

1st run:
[20:33:48.467] [INF] Received initial sync packet with 1 base pieces

2nd run:
[20:34:57.659] [INF] Received initial sync packet with 1 base pieces

[game-20221123.log](https://github.com/SubnauticaNitrox/Nitrox/files/10080839/game-20221123.log)
[server-20221123.log](https://github.com/SubnauticaNitrox/Nitrox/files/10080840/server-20221123.log)

PR #1904, opened 2022-11-24
Fix for base pieces failing to be removed on server (#1903)

CoreyShay:
https://github.com/SubnauticaNitrox/Nitrox/issues/1903

### DIFF
--- a/Nitrox.Test/Model/DataStructures/NitroxIdTest.cs
+++ b/Nitrox.Test/Model/DataStructures/NitroxIdTest.cs
@@ -1,0 +1,45 @@
+﻿using System;
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace NitroxModel.DataStructures;
+
+[TestClass]
+public class NitroxIdTest
+{
+    private NitroxId id1;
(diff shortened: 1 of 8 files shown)


CoreyShay:
Hello, devs. I'm a big fan of this mod, and I really enjoy playing it with my kids. Since I'm an engineer myself, I normally just pull from the master on GitHub. I encountered this issue, and poked around the code until I figured out what the problem was. Thank you very much for all you do, and let me know where else I can look in the code to fix more issues, since I've barely scratched the surface so far.

TwinBuilderOne:
While we appreciate the enthusiasm, we need to verify the issue before we can review and merge this PR. Can you include your server log in the original issue?

If you are interested in contributing to the project, you should head to our Discord server and request the Jr-Alterra role there.

CoreyShay:
Thank you. I'll do that. Attached logs to the issue. The server log does not seem very telling. The game log on 2 consecutive runs on the same server shows on each run that there is 1 base piece synched to the client, but that piece was removed on the 1st run so it should not exist on the 2nd.

tornac1234 (approved):

dartasen (approved):

PR #1906, opened 2022-11-29
Cleaned up publicizer code, reducing allocations

Measurity:
New:
image


dartasen (approved):
image

Faster than old version since it's publishing both assembly in parallel
Can compile and launch the game as well

PR #1908, opened 2022-12-03
Refactoring NitroxEntity.GetId() to better reflect behavior

Jannify:
`NitroxEntity.GetId` just generates new nitrox ids (which are then out of sync) if it can not find a NitroxEntity on the GameObject. This pr removes `NitroxEntity.GetId` and adds methods for different use cases and "throw levels".

- <GameObject/Component>.TryGetNitroxEntity()
- <GameObject/Component>.TryGetNitroxId()
- <GameObject/Component>.TryGetIdOrWarn()
- <GameObject/Component>.GetId()


Sunrunner37:
One thing that would be nice is to introduce a new contract: NitroxEntity.Assign(gameObject, entity).  Then we would be able to see the entity information in the scene debugger.

Sunrunner37:
I'm very mixed on this PR and would like to hear other's opinions.

My main issue is that we are trading off reliability for a small increase in conditional debugability.  I.e. the benefit is that it can detect an unknown entity exception closer to the caller instead of the server and give early detection.   

Nitrox used to be extremely aggressive on "Require/Validate" in the earlier days (esp in packet processors); However, we found that throwing exceptions is user logs causes bad things to propagate, There are a lot of code paths that throwing any exception or deviating causes the game to lock, be unresponsive, or in a state we can't really recover.  For example, if I use the crafter and throw an exception (maybe because I can't resolve the crafter's id), will cause the crafting window to stay open on my screen.  This might have prevented some minor 'bad' behavior (such as another user not able to find my crafter) but at the risk of destabilizing the session itself.  

Overall, there are benefits to not doing this:

1) In the current approach the user is at least given the chance to partially recover or operate in a degraded state. 

2) Missing ids will not block the threads of main hooks such as patches.  Risking breaking the game.

3) While the increase in debuggability is nice there are only a few rare cases where it makes a significant difference over just receiving an unknown entity packet (haven't really seen one in awhile anyways).

4) This refactor touches so many code paths, we are guaranteed to miss at least a couple GetIds that are actually the first time it should be assigned. 

5) It is generally easier for developers to not think about the status of the id.  There are a few edge cases when we don't want tagging (though this could be solved with better naming like EnsureId())

tornac1234:
When I first pointed out this issue, my point only that we were having local occurrences of NitroxEntity.GetId which would create a new unsynced id for an object it certain places and therefore would disrupt the actual NitroxEntity component adding. 
Therefore, in the places where we meant to add the nitroxentity and get it, we should simply rename it to EnsureEntity and in other places where we didnt mean to create the nitroxentity, we should be using TryGetEntity instead to not accidentaly give an unsynced id to the object.

Jannify:
To 1. While there is the potential to recover from a badly generated id the outcome is completely unpredictable which makes debugging and error analysis for code which is using `GetId()` very hard. In general it's likely that it is very buggy during the session and can resolve itself after a server or client restart, artifacts from duplicate id for the same object are possible.

To 2. `RequireId()` is the current safe one-to-one replacement of `GetId()`. `TryGetId()` is the better solution and should be generally used to not break the game/patches/etc

To 3. Reliable id getting should be the top priority imo as it's the very core of our tooling. Problems with setting or getting Ids should be identified instantly as nearly ALL of Nitrox code depends on it's functionality. We shouldn't have a holey foundation.

To 4. Yeah that is kinda likely even through I tried to find every occurrence. The problem is that `GetId()` had two completely different use cases through our code base and untangling the spaghetti now is not that straight forward. 

To 5. In my experience it's the complete opposite (Tornacs comment is confirming that). As I was working on the animation/sound sync I spend hours debugging why the clients had different ids for some GameObjects (I think it was the lifepod or something in it). While using an unsynced id can be not problematic with not critical stuff like 🐟, other objects (like base part, vehicles, etc.) break the sync completely and usually give very bad ux.

---

In general I want to move away from `RequireId()` and widely adapt `TryGetId()` as mentioned in (2). I think that working on a broken/usynced id is bad and can more or less only result in unwanted randomness. Therefor we should return/stop the code/patch/etc. early if we encounter an entity with no id and log it accordingly (like Meas said). If such "soft try catch" isn't possible we can rely on "RequireId" or generate a new (unsyced) id if practicable.

Sunrunner37:
I agree that unknown ids have the capacity to create havoc; however, when faced with two options: "stop everything" vs "take a chance at degraded behavior", I favor the latter unless we are explicitly changing the code path to perform a `Try` operation.  In my opinion, a `Validate` in the new `RequireId` codepath is a non-starter because it is used everywhere and can result in game breaking behavior (esp in critical game loops).  

I like Measurity's suggestion of keeping `GetId` as-is and logging an internal warning (with debug pragma).  This makes sense because the contract does not return an optional.  In all cases the caller expects it to return a value and it will, just logging an error if it is a new assignment. 

As for initially setting the id, I agree with Tornac that we can transform all of the `GetId`s that are intended to generate a new value to a new api called `EnsureId`. This is a common pattern that you already have in the PR, just behind an api, example:

```
                    if (!NitroxEntity.TryGetIdFrom(mono.gameObject, out NitroxId id))
                    toPopulate.Add(new PathBasedChildEntity(relativePathName, id, null, null, vehicleId, new()));
                    {
                        id = new NitroxId();
                        NitroxEntity.SetNewId(mono.gameObject, id);
                    }
```

NitroxClient/MonoBehaviours/NitroxEntity.cs
```diff
@@ -76,16 +77,77 @@ public static bool TryGetObjectFrom(NitroxId id, out GameObject gameObject)
...
+            if (gameObject && gameObject.TryGetComponent(out NitroxEntity nitroxEntity))
+            {
+                nitroxId = nitroxEntity.Id;
+                return true;
+            }
+
+            Log.Warn($"[{typeof(T).Name}] Couldn't find an id on {gameObject.GetFullHierarchyPath()}");
+            nitroxId = null;
+            return false;
+        }
```
Measurity, line 118: Having to pass in a `T` for the calling class makes it quite verbose. I recommend using `CallerFilePath` as a last parameter instead.

Jannify: That `CallerFilePath` is really nice.
If I have to create NitroxEntity from the ground up I would move the id getting to GameObjectExtension. But I wouldn't do so now because all the id logic is located in NitrroxEntity. Using a static method also highlights the method because NitroxEntity gets colored a different color in the IDE.

Measurity (approved):
LGTM CW

Sunrunner37:
I'm still confused as I am seeing RequireId everywhere in codepaths that are invoked via patches.

Measurity (approved):
I changed some things and looked through the entire PR:
 - Some of the GetId APIs are now extension methods on `GameObject` and `Component`.
 - I removed the warn when a `GameObject`/`Component` is dead since this is normal Unity flow and nothing happening is fine in this case.

NitroxClient/GameLogic/Items.cs
```diff
@@ -111,7 +116,7 @@ public static IEnumerable<Entity> GetPrefabChildren(GameObject gameObject, Nitro
 
                     if (metadata.HasValue)
                     {
-                        NitroxId id = NitroxEntity.GetId(prefab.gameObject);
+                        NitroxId id = NitroxEntity.GetIdOrGenerateNew(prefab.gameObject);
```
Sunrunner37, line 119: Should we do this before calling meta data extractors above?  Someone may introduce a metadata extractor that needs the id for something.

NitroxClient/GameLogic/Items.cs
```diff
@@ -41,7 +42,11 @@ public void PickedUp(GameObject gameObject, TechType techType)
             // We want to remove any remote tracking immediately on pickup as it can cause weird behavior like holding a ghost item still in the world.
             RemoveAnyRemoteControl(gameObject);
 
-            NitroxId id = NitroxEntity.GetId(gameObject);
+            if (!gameObject.TryGetNitroxId(out NitroxId id))
```
Sunrunner37, line 45: Let's verify that the id exists already for picking up items from a crafter and that we weren't using GetId for that.

NitroxClient/Helpers/NitroxEntityExtensions.cs
```diff
@@ -0,0 +1,116 @@
...
+            Log.Warn($"[{filePath[(filePath.LastIndexOf("\\", StringComparison.Ordinal) + 1)..^2] + methodName}():L{lineNumber}] Couldn't find an id on {component.GetFullHierarchyPath()}");
+            nitroxId = null;
+            return false;
+        }
+
+        nitroxId = nitroxEntity.Id;
+        return true;
+    }
+
+    public static Optional<NitroxId> GetOptionalId(this GameObject gameObject)
```
Sunrunner37, line 97: `GetId` is probably sufficient.

NitroxPatcher/Patches/Dynamic/Creature_ChooseBestAction_Patch.cs
```diff
@@ -16,7 +15,10 @@ public class Creature_ChooseBestAction_Patch : NitroxPatch, IDynamicPatch
 
         public static bool Prefix(Creature __instance, ref CreatureAction __result)
         {
-            NitroxId id = NitroxEntity.GetId(__instance.gameObject);
+            if (!__instance.TryGetIdOrWarn(out NitroxId id))
```
Sunrunner37, line 18: This will probably be very noisy if it happens.  Do we have a way to combat this?

Sunrunner37 (changes requested):

NitroxPatcher/Patches/Dynamic/Creature_ChooseBestAction_Patch.cs
```diff
@@ -44,7 +44,7 @@ public static void Postfix(Creature __instance, ref CreatureAction __result)
 
         public override void Patch(Harmony harmony)
         {
-            PatchMultiple(harmony, TARGET_METHOD, prefix:true, postfix:true);
+            //PatchMultiple(harmony, TARGET_METHOD, prefix:true, postfix:true);
```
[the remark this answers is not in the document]
Jannify: Yes the patch doesn't do anything at the moment. We could also delete it if that is better.

NitroxModel/DataStructures/Optional.cs
```diff
@@ -150,14 +150,21 @@ public void GetObjectData(SerializationInfo info, StreamingContext context)
         }
     }
 
+
     public struct OptionalEmpty
     {
+        [Obsolete("Use Optional.Empty instead")]
```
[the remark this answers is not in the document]
Jannify: @Measurity

NitroxClient/GameLogic/MobileVehicleBay.cs
```diff
@@ -7,47 +7,46 @@
...
-        public void BeginCrafting(ConstructorInput constructor, GameObject constructedObject, TechType techType, float duration)
-        {
-            MostRecentlyCrafted = constructedObject;
-            Vehicles.RemoveNitroxEntityTagging(constructedObject);
-
-            if (!TransmitLocalSpawns)
-            {
-                return;
-            }
+        constructor.constructor.TryGetIdOrWarn(out NitroxId constructorId);
```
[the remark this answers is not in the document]
Jannify: In `VehicleWorldEntitySpawner.cs` the `spawnerObj` is handled as an optional GameObject so i would say yes

NitroxClient/GameLogic/Building.cs
```diff
@@ -34,28 +34,28 @@ public void PlaceBasePiece(BaseGhost baseGhost, ConstructableBase constructableB
                 return;
             }
 
-            NitroxId id = NitroxEntity.GetId(constructableBase.gameObject);
+            NitroxId id = NitroxEntity.GetIdOrGenerateNew(constructableBase.gameObject);
```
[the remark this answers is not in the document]
Jannify: `EnsureId` is not completely conveying what the method does imo. What about `GetIdOrNew`?

NitroxClient/GameLogic/Spawning/WorldEntities/ReefbackWorldEntitySpawner.cs
```diff
@@ -25,15 +25,15 @@ public IEnumerator SpawnAsync(WorldEntity entity, Optional<GameObject> parent, E
...
+
+            if (!reefback.Value.TryGetComponent(out ReefbackLife life))
             {
                 result.Set(Optional.Empty);
                 yield break;
             }
 
             life.initialized = true;
-            life.SpawnPlants();
+            yield return life.SpawnPlants();
```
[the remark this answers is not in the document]
Jannify: Why should this causing duplicate entities? Because of the `CoroutineUtils.waitForNextFrame`?

NitroxPatcher/Patches/Dynamic/Creature_ChooseBestAction_Patch.cs
Jannify: Like mentioned above this patch is not used. In theory we could use `Log.WarnOnce` which would only warn once per id.

NitroxClient/GameLogic/Items.cs
Jannify: Sry I don't get what you are saying here. Could maybe explain?

NitroxPatcher/Patches/Dynamic/Creature_ChooseBestAction_Patch.cs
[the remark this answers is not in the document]
Measurity: If we delete this commented out code, we should probably just delete the entire patch and make a detailed issue for it.

NitroxModel/DataStructures/Optional.cs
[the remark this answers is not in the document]
Measurity: A struct constructor can't be non-public, you can always do `default(MyStruct)`. It's because a struct is just a sequence of bits and you can always initialize them to all 0, unlike the concept of a `class`.

NitroxPatcher/Patches/Dynamic/Creature_ChooseBestAction_Patch.cs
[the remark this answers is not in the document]
Sunrunner37: This patch does do something.  In the prefix, it only allows `ChooseBestAction` to run when there is a simulation lock.  If we comment this out, I don't know that fish will perform their actions anymore.

NitroxPatcher/Patches/Dynamic/Creature_ChooseBestAction_Patch.cs
Sunrunner37: This sounds good, let's make it so this method can only ever log once per id.

NitroxClient/GameLogic/Items.cs
Sunrunner37: I was saying that this may have been assigning an id when a user was initially picking up an item from a crafter.  Can you test picking up an item from a crafter and ensure it works?

NitroxClient/GameLogic/Spawning/WorldEntities/ReefbackWorldEntitySpawner.cs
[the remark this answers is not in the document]
Sunrunner37: Because the code right below it loops through child entities and spawns them.  This might include plants, and, if so, you would have two plants in the world (one with NitroxEntity and the other without).

Sunrunner37 (approved):

PR #1915, opened 2022-12-11
C# 11

Measurity:


NitroxLauncher/MainWindow.xaml.cs
```diff
@@ -85,12 +85,8 @@ public MainWindow()
...
                     FrameContent = webBrowser;
 
-                    string embed = "<html><head>" +
-                                   "<meta http-equiv=\"X-UA-Compatible\" content=\"IE=Edge\"/>" +
-                                   "</head><body>" +
-                                   $"<iframe width=\"{webBrowser.Width - 24}\" height=\"{webBrowser.Height - 24}\" src=\"{{0}}\"" +
-                                   "frameborder = \"0\" allow = \"autoplay; encrypted-media\" allowfullscreen></iframe>" +
-                                   "</body></html>";
+                    string embed =
+                        $"<html><head><meta http-equiv=\"X-UA-Compatible\" content=\"IE=Edge\"/></head><body><iframe width=\"{webBrowser.Width - 24}\" height=\"{webBrowser.Height - 24}\" src=\"{{0}}\"frameborder = \"0\" allow = \"autoplay; encrypted-media\" allowfullscreen></iframe></body></html>";
```
Coding-Hen, line 89: Personally I preferred the former embed for the readability of the html but I have no problem with the change. Maybe we should consider breaking out this html again, just in case we need to modify it

Coding-Hen (approved):
I didn't realise that we still had all the strings like this. It's good to get rid of them

dartasen (approved):

NitroxLauncher/MainWindow.xaml.cs
Measurity: Good call. It's now this:
```cs
webBrowser.NavigateToString($"""
                    <html>
                        <head>
                            <meta http-equiv="X-UA-Compatible" content="IE=Edge" />
                        </head>
                        <body>
                            <iframe width="{webBrowser.Width - 24}" height="{webBrowser.Height - 24}" src="https://www.youtube.com/embed/i8ju_10NkGY?autoplay=1&loop=1&showinfo=0&controls=0" frameborder="0" allow="autoplay; encrypted-media" allowfullscreen></iframe>
                        </body>
                    </html>
                    """);
```

PR #1925, opened 2022-12-15
Fix chat enter

Measurity:


NitroxPatcher/Patches/Dynamic/DevConsole_Update_Patch.cs
```diff
@@ -1,19 +1,56 @@
...
+
+        /// <summary>
+        ///     Prevent console from opening if server wants it disabled.
+        /// </summary>
         public static void Prefix()
         {
             PlatformUtils.SetDevToolsEnabled(!NitroxConsole.DisableConsole);
         }
 
+        public static IEnumerable<CodeInstruction> Transpiler(MethodBase original, IEnumerable<CodeInstruction> instructions) => instructions
```
Jannify, line 39: Since we are adding a transpiler we should also add an test for it

NitroxPatcher/PatternMatching/InstructionsPattern.cs
```diff
@@ -0,0 +1,140 @@
...
+        if (matchCount != expectedMatches)
+        {
+            throw new Exception($"Expected pattern to match {expectedMatches} times but was {matchCount}. {Environment.NewLine}Pattern:{Environment.NewLine}{this}{Environment.NewLine}IL:{Environment.NewLine}{il.ToPrettyString()}");
+        }
+
+        // Apply operations on index of IL or return the original instruction.
+        for (int i = 0; i < il.Length; i++)
+        {
+            yield return il[i];
+            if (insertOperations.ContainsKey(i))
```
Jannify, line 100: Maybe use `TryGetValue` here?

Jannify (changes requested):

NitroxPatcher/Patches/Dynamic/DevConsole_Update_Patch.cs
```diff
@@ -1,19 +1,56 @@
...
+                    case "TrueArgument":
+                        instruction.opcode = Ldc_I4_0;
+                        break;
+                }
+            });
+
         public override void Patch(Harmony harmony)
         {
-            PatchPrefix(harmony, Reflect.Method((DevConsole t) => t.Update()));
+            PatchPrefix(harmony, TargetMethod);
```
dartasen, line 52: Do we need to keep a prefix method since we're transpiling the method ?

Measurity: I kept it because prefix is easier to add new code. You prefer I include the prefix change into the transpiler for performance reasons?

NitroxPatcher/Patches/Dynamic/DevConsole_Update_Patch.cs
Measurity: The way transpiler tests are done is dumb. They don't prevent us from breaking the code and neither do the seem to help us fixing it for new game code. The way this pattern system works will error out if the code changes significantly (but at runtime, not CI/build time). 

I'll add it just for posterity but we should evaluate if these tests provide value.

NitroxPatcher/Patches/Dynamic/DevConsole_Update_Patch.cs
dartasen: If we can avoid dynamic patches on Update method, it definitely sounds better.

NitroxConsole.Disable is only set once during session policies exchange. So we don't really need to be reactive on it's value. Unless we want to allow dynamic session policies, but right now it requires a server restart.

Jannify (approved):

PR #1937, opened 2022-12-24
Several stabilization changes to get us closer to a playable build, including:

Sunrunner37:
image

Entities are mostly spawning - still some slowness in starting chunks though.  It also looks like simulation ownership system is working fine.  This PR does have some benign unauthenticated packets due to moving the dynamic patching location - we can work through these with suppression when we get to them. 

* Changed dynamic patching hook to be earlier on.  This allows patches that require early initialization to function (such as uGUI_PDA)
* Fixed several bugs in entity spawning async transformation
* Fixes to allow users to spawn alongside their remote vitals
* Prevented PDA from erroring out initial syncs with multiple players
* Added a new patch to prevent escape pod supplies from spawning
* Small tweak to logger to only log a warning instead of throwing when attempting to re-initialize.
* Fixed a bug preventing us from finding the remote player's inventory.

### DIFF
--- a/NitroxClient/GameLogic/Spawning/DefaultEntitySpawner.cs
+++ b/NitroxClient/GameLogic/Spawning/DefaultEntitySpawner.cs
@@ -57,11 +57,9 @@ namespace NitroxClient.GameLogic.Spawning
 
         IEnumerator CreateGameObject(TechType techType, string classId, TaskResult<GameObject> result)
         {
-            GameObject prefab = null;
-
             IPrefabRequest prefabCoroutine = PrefabDatabase.GetPrefabAsync(classId);
             yield return prefabCoroutine;
-            prefabCoroutine.TryGetPrefab(out prefab);
+            prefabCoroutine.TryGetPrefab(out GameObject prefab);
 
             if (prefab == null)
             {
--- a/NitroxClient/GameLogic/Spawning/DefaultEntitySpawner.cs
+++ b/NitroxClient/GameLogic/Spawning/DefaultEntitySpawner.cs
@@ -58,10 +58,10 @@ namespace NitroxClient.GameLogic.Spawning
         IEnumerator CreateGameObject(TechType techType, string classId, TaskResult<GameObject> result)
         {
             GameObject prefab = null;
-            if (PrefabDatabase.TryGetPrefabFilename(classId, out string filename))
-            {
-                prefab = Resources.Load<GameObject>(filename);
-            }
+
+            IPrefabRequest prefabCoroutine = PrefabDatabase.GetPrefabAsync(classId);
+            yield return prefabCoroutine;
+            prefabCoroutine.TryGetPrefab(out prefab);
 
             if (prefab == null)
             {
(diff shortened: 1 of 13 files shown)


dartasen (approved):

Coding-Hen (approved):

Jannify:
I encountered a few errors:

- Changing the patch trigger point errors out `CyclopsHelmHUDManager_Start_Patch` on the cyclops prototype scene which is loaded additive
- I got a NRE at `RemotePlayerVitals.CreateForPlayer`. I think this was also present before the SN update
- `uGUI_PlayerListTab` is erroring out hard when a second player joins. if you open the PDA it displays nothing and you are stuck in it forever!
- Strangely the Flare threw an NRE error in Awake on pickup

Would definitely want to have the PDA fixed, rest is optional/can be done later.

NitroxPatcher/Patches/Dynamic/SpawnEscapePodSupplies_OnNewBorn_Patch.cs
```diff
@@ -0,0 +1,25 @@
+using System.Reflection;
+using HarmonyLib;
+using NitroxModel.Helper;
+using UWE;
+
+/// <summary>
+/// Don't allow the game to spawn initial supplies in the escape pod.
+/// </summary>
+namespace NitroxPatcher.Patches.Dynamic
+{
```
Jannify, line 10: Would prefer file scoped namespaces

NitroxClient/GameLogic/Spawning/DefaultEntitySpawner.cs
```diff
@@ -58,10 +58,10 @@ public IEnumerator SpawnAsync(Entity entity, Optional<GameObject> parent, Entity
...
         {
             GameObject prefab = null;
-            if (PrefabDatabase.TryGetPrefabFilename(classId, out string filename))
-            {
-                prefab = Resources.Load<GameObject>(filename);
-            }
+
+            IPrefabRequest prefabCoroutine = PrefabDatabase.GetPrefabAsync(classId);
+            yield return prefabCoroutine;
+            prefabCoroutine.TryGetPrefab(out prefab);
```
Jannify, line 64: Would move variable declaration inside the TryGetPrefab()

Jannify (changes requested):

Sunrunner37:
Fixed the PDA issue.  Will address cyclops issues as a part of a larger vehicle refactoring in next PR.

NitroxClient/GameLogic/Spawning/DefaultEntitySpawner.cs
Sunrunner37: Done

NitroxPatcher/Patches/Dynamic/SpawnEscapePodSupplies_OnNewBorn_Patch.cs
Sunrunner37: Done

Jannify (approved):

PR #1938, opened 2022-12-24
Removed OS specific directives inside csproj

dartasen:


Measurity:
If this isn't fixing an issue atm, I suggest we leave it for now since this will be replaced when we have linux launcher anyway. I believe this OS condition was added by Marijn way back to fix linux setup.

PR #1941, opened 2022-12-25
Version/1.7.1.0

dartasen:
- [x] Fix firewall configurator for older windows version that was crashing the launcher everytime, for now it's just dodging the process since the feature needs a complete rework to get it working on Windows 7
- [x] Added warning message upon game start (only if the game is updated)
image
image
image
- [x] Added warning message upon server start (if the game is updated)

### DIFF
--- a/NitroxLauncher/LauncherLogic.cs
+++ b/NitroxLauncher/LauncherLogic.cs
@@ -88,12 +88,54 @@ namespace NitroxLauncher
         [Conditional("RELEASE")]
         public async void ConfigureFirewall()
         {
-            Task task = Task.Run(() => WindowsHelper.CheckFirewallRules());
-            await task;
+            Log.Info($"Using {Environment.OSVersion}");
 
-            if (task.Exception != null)
+            if (Environment.OSVersion.Platform != PlatformID.Win32NT)
             {
-                MessageBox.Show($"An error occurred configuring the firewall: {task.Exception}");
+                return;
+            }
+
+            /**
+                This feature won't work in older windows version and will crash the launcher instantly
+                
+                Windows Vista : 6.0
+                Windows 7 : 6.1
+                Windows 8 : 6.2, Windows 8.1 : 6.3
+                Windows 10/11 : 10.0
+            **/
+            if (Environment.OSVersion.Version.Major <= 6)
+            {
+                return;
+            }
+
+            CancellationTokenSource cancellationTokenSource = new();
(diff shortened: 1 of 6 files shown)


Measurity (approved):

NitroxLauncher/LauncherLogic.cs
```diff
@@ -88,12 +88,52 @@ public async void CheckNitroxVersion()
...
-                MessageBox.Show($"An error occurred configuring the firewall: {task.Exception}");
+                return;
+            }
+
+            /**
+                This feature won't work in older windows version and will crash the launcher instantly
+                Windows Vista : 6.0
+                Windows 7 : 6.1
+                Windows 8 : 6.2, Windows 8.1 : 6.3
+            **/
```
Coding-Hen, line 103: To me this is slightly unclear. Is that it only works on these operating systems or it doesn't work on these? I think maybe a small tweak to the text to explain further, maybe to say it wont work on the below, which is what I believe the if statement below is checking. 

Does this mean we only support windows 10 and upwards now?

NitroxModel/Helper/NitroxUser.cs
```diff
@@ -100,5 +100,31 @@ public static string GamePath
...
+
+                string streamingAssetsFolder = Path.Combine(GamePath, "Subnautica_Data", "StreamingAssets");
+                string aaFolder = Path.Combine(streamingAssetsFolder, "aa");
+
+                if (!Directory.Exists(streamingAssetsFolder)) {
+                    // Probably authorization exception
+                    return null;
+                }
+
+                if (File.Exists(Path.Combine(aaFolder, "catalog.json")) && File.Exists(Path.Combine(aaFolder, "settings.json")))
```
Coding-Hen, line 121: If we are just checking that the settings and catalog exist as files is there any point checking the folder exists and not just checking the final path? I would assume if the folder doesn't exist it will throw an error saying no file in the file check as it won't be able to find the paths

NitroxLauncher/LauncherLogic.cs
dartasen: Seems pretty clear to me, we're checking the windows major version here. If it's 6.x then it's a return

dartasen: Right now you can't open the launcher on W7 and older due to this feature

NitroxModel/Helper/NitroxUser.cs
dartasen: No error is thrown by File.Exist in any way

PR #1942, opened 2022-12-27
Entity Metadata Extraction

Sunrunner37:
Entity metadata is the short list of attributes that Nitrox cares about for an object in the game.  Consider an in-game sign, which has an entire hierarchy of unity objects and data, but Nitrox only wants to sync the text, color, and font.  When entities update, Nitrox needs to resolve the entity metadata (using the hierarchy) and sync that to the server.  Currently, entity metadata is constructed in patches, resulting in duplicate metadata instantiation for every type of update action (ex: starship door open/close/lock status).  While this is minor at the moment, it will become difficult when consolidating other systems like vehicles and bases.

Nitrox also requires the ability to fetch metadata when a user drops an item.  For example, a dropped flashlight should have metadata about the `EnergyMixin` and the on/off status.  This was previously achieved by serializing the entire gameObject hierarchy when an item dropped.  However, this had serious limitations because these items could be interacted with after being dropped.  An example is the constructor, which deploys after dropping.  The only way to achieve this in previous Nitrox was a complex `ItemDropActions` system that post-processed each dropping, looking for specific types. The entire `DroppedItem` system was almost an identical clone to the existing entity system but just specific to that one data type.

This PR entirely removes the Dropped Item system (packets, processors, drop actions) and replaces it with the existing `EntitySpawnedByClient` codepath.  The `DroppedItem` data structure is now a simple `WorldEntity` and any required post-processing can be done by `EntityMetadataProcessors`.  This also removes the need to serialize the entire gameObject hierarchy; instead, when an item is dropped, we will extract Metadata off and send it alongside the `WorldEntity`.  This extract logic can also be used for other places that send entity updates, they simply call a function with the updated object and the patches no longer needs to know how to resolve the metadata itself.  

As a first implementation, the constructor is now working properly (previously, it would bomb out the entire game).  I also moved over Sealed doors as a small example.  Planning to follow this PR up with another that solves much more sophisticated use-cases such as the seaglide, flashlights, and pipes.


NitroxClient/GameLogic/Spawning/Metadata/ConstructorMetadataProcessor.cs
```diff
@@ -0,0 +1,21 @@
+using NitroxModel.DataStructures.GameLogic.Entities.Metadata;
+using UnityEngine;
+
+namespace NitroxClient.GameLogic.Spawning.Metadata;
+
+class ConstructorMetadataProcessor : GenericEntityMetadataProcessor<ConstructorMetadata>
```
Jannify, line 6:
```suggestion
public class ConstructorMetadataProcessor : GenericEntityMetadataProcessor<ConstructorMetadata>
```

Jannify, line 6:
```suggestion
public class ConstructorMetadataProcessor : GenericEntityMetadataProcessor<ConstructorMetadata>
```

NitroxClient/GameLogic/Spawning/Metadata/Extractor/GenericEntityMetadataExtractor.cs
```diff
@@ -0,0 +1,63 @@
...
+
+    static EntityMetadataExtractor()
+    {
+        processors = Assembly.GetExecutingAssembly()
+                             .GetTypes()
+                             .Where(t => typeof(EntityMetadataExtractor).IsAssignableFrom(t) &&
+                                    t.IsClass && !t.IsAbstract)
+                             .Select(Activator.CreateInstance)
+                             .Cast<EntityMetadataExtractor>()
+                             .ToDictionary(p => p.GetType().BaseType.GetGenericArguments()[0]);
```
Jannify, line 37: Is there a reason you are using plain reflection and not Autofac?

NitroxClient/GameLogic/Spawning/Metadata/Extractor/GenericEntityMetadataExtractor.cs
```diff
@@ -0,0 +1,63 @@
...
+    {
+        if (processors.TryGetValue(o.GetType(), out EntityMetadataExtractor extractor))
+        {
+            return extractor.From(o);
+        }
+
+        return Optional.Empty;
+    }
+
+    public static Optional<EntityMetadata> Extract(GameObject o)
```
Jannify, line 50: That could be problematic with multiple viable Components on the GameObject. I would favor using a Component as the parameter.

NitroxPatcher/Patches/Dynamic/Constructor_Deploy_Patch.cs
```diff
@@ -0,0 +1,32 @@
...
+    public static readonly MethodInfo TARGET_METHOD = Reflect.Method((Constructor t) => t.Deploy(default(bool)));
+
+    public static void Prefix(Constructor __instance, bool value)
+    {
+        // only trigger updates when there is a valid state change.
+        if(value != __instance.deployed)
+        {
+            NitroxId id = NitroxEntity.GetId(__instance.gameObject);
+            Entities entities = NitroxServiceLocator.LocateService<Entities>();
+            entities.EntityMetadataChanged(__instance, id);
```
Jannify, line 22:
```suggestion
            Resolve<Entities>().EntityMetadataChanged(__instance, id);
```

NitroxPatcher/Patches/Dynamic/Constructor_Deploy_Patch.cs
```diff
@@ -0,0 +1,32 @@
...
+        }
+    }
+
+    public override void Patch(Harmony harmony)
+    {
+        PatchPrefix(harmony, TARGET_METHOD);
+    }
+}
+
```
Jannify, line 31:
```suggestion
```

NitroxPatcher/Patches/Dynamic/Sealed_Weld_Patch.cs
```diff
@@ -16,10 +15,7 @@ public class Sealed_Weld_Patch : NitroxPatch, IDynamicPatch
         public static void Postfix(Sealed __instance)
         {
             NitroxId id = NitroxEntity.GetId(__instance.gameObject);
-            SealedDoorMetadata doorMetadata = new(__instance._sealed, __instance.openedAmount);
-            Entities entities = NitroxServiceLocator.LocateService<Entities>();
-
-            entities.BroadcastMetadataUpdate(id, doorMetadata);
+            NitroxServiceLocator.LocateService<Entities>().EntityMetadataChanged(__instance, id);
```
Jannify, line 18:
```suggestion
            Resolve<Entities>().EntityMetadataChanged(__instance, id);
```

Jannify (changes requested):

NitroxPatcher/Patches/Dynamic/Constructor_Deploy_Patch.cs
dartasen: Note: We're now using Resolve which also uses NitroxServiceLocator, but it's providing a cache

NitroxClient/GameLogic/Spawning/Metadata/ConstructorMetadataProcessor.cs
Sunrunner37: done

NitroxClient/GameLogic/Spawning/Metadata/Extractor/GenericEntityMetadataExtractor.cs
Sunrunner37: swapped to autofac

NitroxClient/GameLogic/Spawning/Metadata/Extractor/GenericEntityMetadataExtractor.cs
Sunrunner37: Yes, this is one issue that I currently see.  However, the only codepath that invokes this is when we drop an item.  In that case, we don't actually know which component will source metadata (if any).  I will explore ways to improve this in a follow-up when implementing the seaglide.  Potentially we can use the techtype to prioritize specific components or we can define a global priority.  I'm not sure the best approach until implementing a few more and seeing what works best.

NitroxPatcher/Patches/Dynamic/Constructor_Deploy_Patch.cs
Sunrunner37: Thanks this is better.

Jannify (approved):
Code wise looks fine

Coding-Hen (approved):
I agree code wise looks good, removes a lot of repeated code and simplifies entity metadata updates which is nice. I think testing would be good but might be worth doing on all of the new changes at once and we can then iron out any issues then

PR #1943, opened 2022-12-27
Vehicle Construction

Sunrunner37:
This PR fixes a variety of issues preventing the constructor from spawning the Seamoth and Cyclops.  In the current state, spawning anything from the constructor causes everything to freeze with a flood of exceptions. I was also able to use this opportunity to simplify the spawning hook to no longer use transient object manager.

I've verified that the EnergyMixins are loaded in and appropriately tagged with a NitroxEntity.  However, there may be more follow-up work to verify their full functionality. 

Note: The constructor deployment was synced in #1942.  Thus, it will not deploy for remote players until that is merged.  However, I have verified that the construction started packet is emitted.  It also results in a message on the remote player as well.


NitroxClient/GameLogic/Vehicles.cs
```diff
@@ -47,10 +47,11 @@ public VehicleModel BuildVehicleModelFrom(GameObject gameObject, TechType techTy
         {
             if (VehicleHelper.IsVehicle(techType))
             {
+                UnityEngine.Component.DestroyImmediate(gameObject.GetComponent<NitroxEntity>()); // template has pre-existing NitroxEntity, remove.
+                NitroxId constructedObjectId = NitroxEntity.GetId(gameObject);
```
dartasen, line 51: This is something that was started with #1908 but we shouldn't use GetId as an ID setter anymore (it will attach a random ID when it's not existing, which not something we want in most of the cases)

NitroxClient/GameLogic/MobileVehicleBay.cs
```diff
@@ -22,36 +20,22 @@ public MobileVehicleBay(IPacketSender packetSender, Vehicles vehicles)
...
-                    monoBehaviour.StartCoroutine(vehicles.UpdateVehiclePositionAfterSpawn(vehicleModel, constructedObject, duration + 10.0f));
-                }
-            }
-            else
+            MonoBehaviour monoBehaviour = constructor.GetComponent<MonoBehaviour>();
+            //We want to store the fallen position of the object to avoid flying object on reload 
+            if (monoBehaviour)
             {
-                Log.Error("Could not send packet because there wasn't a corresponding constructed object!");
+                monoBehaviour.StartCoroutine(vehicles.UpdateVehiclePositionAfterSpawn(vehicleModel, constructedObject, duration + 10.0f));
```
dartasen, line 38: We should definetely find something more sexier than that. I added that as a temporary workaround for people

NitroxPatcher/Patches/Dynamic/ConstructorInput_OnCraftingBegin_Patch.cs
```diff
@@ -28,15 +28,24 @@ public static IEnumerable<CodeInstruction> Transpiler(MethodBase original, IEnum
...
+                     * Callback(constructor, gameObject, techType, duration);
                      */
-                    yield return new CodeInstruction(OpCodes.Ldc_I4_0);
+                    yield return original.Ldloc<ConstructorInput>(0);
                     yield return original.Ldloc<GameObject>(0);
-                    yield return new CodeInstruction(OpCodes.Call, Reflect.Method(() => Add(default(TransientObjectType), default(object))));
+                    yield return new CodeInstruction(OpCodes.Ldarg_0);
+                    yield return new CodeInstruction(OpCodes.Ldfld, TARGET_METHOD.DeclaringType.GetField("techType", BindingFlags.Instance | BindingFlags.Public));
+                    yield return new CodeInstruction(OpCodes.Ldarg_0);
+                    yield return new CodeInstruction(OpCodes.Ldfld, TARGET_METHOD.DeclaringType.GetField("duration", BindingFlags.Instance | BindingFlags.Public));
```
dartasen, line 38: You can use Reflect.Property((ConstructorInput c) => c.TechType), we're no longer using raw reflection

Will also work on private/protected attributes since Nitrox is now using Publicized assemblies

NitroxPatcher/Patches/Dynamic/ConstructorInput_OnCraftingBegin_Patch.cs
```diff
@@ -28,15 +28,24 @@ public static IEnumerable<CodeInstruction> Transpiler(MethodBase original, IEnum
...
+                    yield return new CodeInstruction(OpCodes.Call, Reflect.Method(() => Callback(default, default, default, default)));
                 }
             }
         }
 
+        public static void Callback(ConstructorInput constructor, GameObject constructedObject, TechType techType, float duration)
+        {
+            NitroxServiceLocator.LocateService<MobileVehicleBay>().BeginCrafting(constructor, constructedObject, techType, duration);
```
dartasen, line 46:
```suggestion
            Resolve<MobileVehicleBay>().BeginCrafting(constructor, constructedObject, techType, duration);
```

NitroxClient/GameLogic/Vehicles.cs
Jannify: Is saw this too but NitroxEntity has no correct method for it right now. I will just update it with #1908.

Sunrunner37: We should talk a bit about that.  Throwing when it doesn't exist is very rigid.  It is what we used to do in a lot of places (like RequireComponent or Validate.NotNull) but in the end it causes functionality to bomb out entirely instead of operate in a semi-degraded state.

NitroxPatcher/Patches/Dynamic/ConstructorInput_OnCraftingBegin_Patch.cs
Sunrunner37: For clarity, the TARGET_METHOD.DeclaringType is not ConstructorInput, it is the class automatically generated for the IEnumerable.  I don't think there is any way to access it via the anonymous function route.

NitroxPatcher/Patches/Dynamic/ConstructorInput_OnCraftingBegin_Patch.cs
Sunrunner37: Sure

NitroxClient/GameLogic/MobileVehicleBay.cs
Sunrunner37: There is a good chance this my get solved when I move vehicles over to the entity system.  We'll have to see.

dartasen (approved):

Issue #1945, opened 2022-12-28
Backlog Item: Support Multiple Escapepods


Sunrunner37:
### Describe the issue

The multi-lifepod functionality has resided in nitrox for a long time but has never worked correctly.  It is currently disabled by setting the required players to a large amount.  We will need to figure out the right experience to support this feature.  Large servers (such as SMG) don't want to have hundreds of lifepods floating around. Thus, we may need some kind of way to set them inactive based on if players are online.

Jannify:
#2043 brings this issue out of the backlog. While the intro cinematic is playing the escape pod needs to be locked and the joining players animation look strange for already online player

PR #1946, opened 2022-12-28
Escape Pod Refactoring

Sunrunner37:
Currently, Nitrox maintains custom infrastructure to manage the lifecycle of escape pods.  This includes an initial sync processor, multiple repair events, and a special persistence section.  This PR treats the escape pod as a global root `WorldEntity` and each of it's children (fabricator, medical, etc) as `PrefabPlaceholderEnities`.  Players actions, such as repairing the radio, will be treated as a simple entity metadata update. This allows us to leverage all the existing `Entity` code and deprecate custom escape pod implementations.

The PR accomplishes a few things:

1) Performs all the refactoring to represent the escape pod as an entity.
2) Fixes functionality that was not working due to the subnautica update: tagging children with the right ids.  You can now repair the radio and have that synced.  

There were a couple other misc changes that had to be included because they facilitated the switch:

1) Changed the invocation point of `Multiplayer.SubnauticaLoadingComplete()`.  The existing patch on `PAXTerrainController` was VERY complex and it ultimately was only invoking after terrain loading.  This new patch point should be the very last thing executed before the loading screen goes away.

2) Skipped the cutscene, I think this somehow  broke during the update.  

One thing to note: multiple escape pods remains 'off' as there are some challenges with loading multiple escape pod prefabs.  I added a backlog item to correctly implement this feature (it never really worked), see #1945.


Sunrunner37:
Quick update.  Found the issue causing the common repair point not working.  Now all items in the escape pod have the right guids and have synced repairs. 

image

NitroxPatcher/Patches/Dynamic/Player_SetCurrentEscapePod_Patch.cs
```diff
@@ -17,10 +17,11 @@ public static void Prefix(ref Player __instance, EscapePod value)
         {
             NitroxId podId = null;
 
-            if (value != null)
+            if (value && value.GetComponent<NitroxEntity>() != null)
```
tornac1234, line 20: I think we should not be using null equality checks with components, which is why we have `AliveOrNull`
```suggestion
            if (value && value.GetComponent<NitroxEntity>().AliveOrNull())
```

Measurity: This is fine because Unity overrides the == and != operators. We have an analyzer that will warn you on incorrect usage.

NitroxServer/GameLogic/EscapePodManager.cs
```diff
@@ -38,70 +42,53 @@ public NitroxId AssignPlayerToEscapePod(ushort playerId, out Optional<EscapePodM
...
-                                         new NitroxId(),
-                                         new NitroxId(),
-                                         new NitroxId(),
-                                         new NitroxId(),
-                                         true,
-                                         true);
-
-            EscapePods.Add(escapePod);
+            entityRegistry.AddEntity(escapePod);
+            entityRegistry.AddEntities(escapePod.ChildEntities);
```
Coding-Hen, line 67: Will this add the child entities twice being part of the escapePod already? If not might we want to add this into the AddEntity code so we don't need to add them both individualy or do you not think it worth it?

NitroxClient/GameLogic/Spawning/Metadata/RepairedComponentMetadataProcessor.cs
```diff
@@ -0,0 +1,30 @@
+using NitroxModel.DataStructures.GameLogic.Entities.Metadata;
+using UnityEngine;
+
+namespace NitroxClient.GameLogic.Spawning.Metadata;
+
+public class RepairedComponentMetadataProcessor : GenericEntityMetadataProcessor<RepairedComponentMetadata>
+{
+    public override void ProcessMetadata(GameObject gameObject, RepairedComponentMetadata metadata)
+    {
+        Log.Info("RepairedComponentMetadataProcessor");
```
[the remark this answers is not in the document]
Sunrunner37: nope - removed.

NitroxClient/GameLogic/Spawning/WorldEntities/EscapePodWorldEntitySpawner.cs
```diff
@@ -0,0 +1,129 @@
...
+            GameObject escapePod = EscapePod.main.gameObject;
+            UnityEngine.Component.DestroyImmediate(escapePod.GetComponent<NitroxEntity>()); // if template has a pre-existing NitroxEntity, remove.
+            NitroxEntity.SetNewId(escapePod, escapePodEntity.Id);
+
+            EntityMetadataProcessor.ApplyMetadata(escapePod, escapePodEntity.Metadata);
+
+            Rigidbody rigidbody = escapePod.GetComponent<Rigidbody>();
+            if (rigidbody != null)
+            {
+                Log.Debug("Freezing escape pod rigidbody");
```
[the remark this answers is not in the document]
Sunrunner37: nope - removed.

NitroxClient/GameLogic/InitialSync/PlayerInitialSyncProcessor.cs
```diff
@@ -27,7 +29,11 @@ public override IEnumerator Process(InitialPlayerSync packet, WaitScreen.ManualW
             yield return null;
 
             SetPlayerGameObjectId(packet.PlayerGameObjectId);
-            waitScreenItem.SetProgress(0.33f);
+            waitScreenItem.SetProgress(0.20f);
```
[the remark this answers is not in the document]
Sunrunner37: Sure

NitroxServer/GameLogic/EscapePodManager.cs
Sunrunner37: This case will not add redundant entities because the escape pod is only broadcasted to the client.  I think there might be implications to adding the children as well in AddEntities.  Off the top, other code paths using it might start throwing on duplicates and I have to check to see if there are any reasons that they shouldn't be in the registry. Will consider in a follow-up PR though.

Coding-Hen (approved):

PR #1947, opened 2022-12-28
Fix remote player stuttering/teleporting when standing still or moving.

Sunrunner37:
Nitrox uses a velocity-based system to update remote player locations.  This provides a fluid movement and doesn't incur large costs associated with setting the transform position over and over.  However, with the recent Subnautica update, it appears that we update the velocity too quickly.  Movement updates don't appear to apply velocity for multiple packets then cause the player to over correct.  It is likely that Nitrox is applying velocity updates too frequently for the physics engine to react (0.05s). 

In this PR, I've updated the frequency of movement packets to 0.12s.  This caused normal movement behavior; however, not very smooth.  To counteract this, I also enabled some interpolation to bring back the smoothness.  We can play with these a bit more; however, this configuration looks about the same as the last update to me.  

This change also has the added benefit of significantly reducing the amount of packets that Nitrox sends.

### DIFF
--- a/NitroxClient/MonoBehaviours/PlayerMovementBroadcaster.cs
+++ b/NitroxClient/MonoBehaviours/PlayerMovementBroadcaster.cs
@@ -18,7 +18,7 @@ public class PlayerMovementBroadcaster : MonoBehaviour
     public const int LOCATION_BROADCAST_TICK_SKIPS = 1;
 
     private LocalPlayer localPlayer;
-    private int locationBroadcastSkipCounter;
+    private int locationBroadcastSkipThreshold = LOCATION_BROADCAST_TICK_SKIPS;
 
     public void Awake()
     {
@@ -28,12 +28,12 @@ public class PlayerMovementBroadcaster : MonoBehaviour
     public void FixedUpdate()
     {
         // Throttle location broadcasts to not run on every physics tick.
-        if (locationBroadcastSkipCounter++ < LOCATION_BROADCAST_TICK_SKIPS)
+        if (locationBroadcastSkipThreshold-- > 0)
         {
             return;
         }
-        // Reset counter.
-        locationBroadcastSkipCounter = 0;
+        // Reset skip threshold.
+        locationBroadcastSkipThreshold = LOCATION_BROADCAST_TICK_SKIPS;
 
         // Freecam does disable main camera control
         // But it's also disabled when driving the cyclops through a cyclops camera (content.activeSelf is only true when controlling through a cyclops camera)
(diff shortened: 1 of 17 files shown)


NitroxClient/MonoBehaviours/PlayerMovementBroadcaster.cs
```diff
@@ -11,7 +11,7 @@ namespace NitroxClient.MonoBehaviours;
 
 public class PlayerMovementBroadcaster : MonoBehaviour
 {
-    public const float BROADCAST_INTERVAL = 0.05f;
+    public const float BROADCAST_INTERVAL = 0.12f;
```
dartasen, line 14: Maybe we should put a lil comment there to explain that we shouldn't go down this value because of weird behaviours of the physic engine

Measurity: We might need to change how movement is handled by putting it in a [FixedUpdate](https://docs.unity3d.com/ScriptReference/MonoBehaviour.FixedUpdate.html). So when a movement update is received, wait until physics frame and set velocity then. Might solve this weird behavior. Or at least something to look into in the future by adding a TODO comment here.

Measurity (approved):

dartasen (approved):

Coding-Hen (approved):

NitroxClient/MonoBehaviours/PlayerMovementBroadcaster.cs
```diff
@@ -11,62 +11,66 @@ namespace NitroxClient.MonoBehaviours;
...
 
-    public void Update()
+    public void FixedUpdate()
     {
-        time += Time.deltaTime;
-
-        // Only do on a specific cadence to avoid hammering server
-        if (time >= BROADCAST_INTERVAL)
+        // Throttle location broadcasts to not run on every physics tick.
+        if (locationBroadcastSkipCounter++ < LOCATION_BROADCAST_TICK_SKIPS)
```
Coding-Hen, line 31: Would it potentially be safer to invert this? Just incase it goes over the location broadcast tick skips and then it will never limit it

Coding-Hen (changes requested):

NitroxClient/MonoBehaviours/PlayerMovementBroadcaster.cs
Measurity: Like this?
```cs
// Throttle location broadcasts to not run on every physics tick.
if (--locationBroadcastSkip >= 0)
{
    return;
}
// Reset counter.
locationBroadcastSkip = LOCATION_BROADCAST_TICK_SKIPS;
```

Coding-Hen: I do like the design you have proposed but I think doing it that way would mean the first update would be skipped with the value of locationBroadcastSkip being -1. If we increment it after it will allow it to run on the first update and then skip others until it reaches the LOCATION_BROADCAST_TICK_SKIPS. 
```
// Throttle location broadcasts to not run on every physics tick.
if (locationBroadcastSkip-- > 0)
{
    return;
}
// Reset counter.
locationBroadcastSkip = LOCATION_BROADCAST_TICK_SKIPS;
```

Coding-Hen (approved):

PR #1949, opened 2022-12-29
Ability to derive child entities from dropped items. 

Sunrunner37:
In past PRs, we've talked about how Nitrox has relied on serialized objects to facilitate item mechanics.  It was applied inconsistently and resulted in entities losing 'state' frequently.  For example, placing a battery in a storage container would only record the tech type (the charge ended up being lost).  Dropped batteries were slightly better, they at least maintained state unless nested inside other game objects like tools.  Moving items between the world and inventory always lost state eventually. 

This is the first workable implementation of how this will work in the future.  I've implemented entity mechanics for the flashlight and battery.  The new structure will look like this:

```
WorldEntity (Metadata: Flashlight [on: true])
    - PrefabChildEntity (Metadata: Battery [charge: 50%])
```

When inventories are on the new entity system this hierarchy can be seamlessly transitioned into an `InventoryItemEntity` while keeping the entire structure and metadata, example:

``` 
StorageEntity
    - InventoryItemEntity (Metadata: Flashlight [on: true])
         - PrefabChildEntity (Metadata: Battery [charge: 50%])
```

This PR syncs every item with a battery without us having to list out which items have batteries.  When an item is dropped, Nitrox will convert it into a `WorldEntity`.  Then, it will look for prefab children that can produce `EntityMetadata`, in this case `BatteryMetadata`.  Qualifying children will be assigned as a `PrefabChildEntity` under the `WorldEntity`. This allows us to sync metadata without having to be familiar with the root type.  An example is a welder, which doesn't have notable state that Nitrox will store.  This PR will store the welder in the following format:

```
WorldEntity [TechType = TechType.Welder)
    - PrefabChildEntity (Metadata: Battery [charge: 50%])
```

In this summary, this PR includes the following:

1) Implemented syncing for all dropped items that have batteries.
2) Implemented syncing for a dropped flashlight (on/off state) as an example.
3) Fixed a major bug preventing children spawning (fixes some missing fragments).
4) Reworked how PrefabChildEntity spawning works. 

This is probably the first 'new' functionality since the large update.  Players can now drop flashlights, take out the battery and drop the battery.  All while maintaining the charge state in a simple entity metadata entry.


Sunrunner37:
Updated the description.  If it seems like I am picking random people, I am trying to spread out the PRs.

NitroxClient/GameLogic/Entities.cs
```diff
@@ -166,6 +171,24 @@ private void AddPendingParentEntity(Entity entity)
...
+        // IEnumerator (async) and levarage async Prefab/CraftData instantiation functions.  However, even though these
+        // functions are successful, it doesn't mean the entity is fully setup.  Subnautica is known to spawn coroutines 
+        // in the start() method of objects to spawn prefabs or other objects. An example is anything with a battery, 
+        // which gets configured after the fact.  In most cases, Nitrox needs to wait for objets to be fully spawned in 
+        // order to setup ids.  Historically we would persist metadata and use a patch to later tag the item, which gets
+        // messy.  This function will allow us wait on any type of instantiation necessary; this can be optimized later
+        // to move on to other spawning and come back when this item is ready.  
+        private IEnumerator AwaitAnyRequiredEntitySetup(GameObject gameObject)
+        {
+            EnergyMixin energyMixin = gameObject.GetComponent<EnergyMixin>();
```
Jannify, line 184: Component lookup is fine for now I think. But in the future if this method is getting out of hand with many component lookups and checks we maybe want to sort with flags saved in the WorldEntity or linked to the Techtype. Especially since this is executed for each WorldEntity.

NitroxClient/GameLogic/Entities.cs
```diff
@@ -166,6 +171,24 @@ private void AddPendingParentEntity(Entity entity)
...
+
+            if (energyMixin)
+            {
+                yield return new WaitUntil(() => energyMixin.battery != null);
```
Jannify, line 188: I guess energyMixin.battery is a MonoBehaviour?
```suggestion
                yield return new WaitUntil(() => energyMixin.battery);
```

NitroxClient/GameLogic/InitialSync/GlobalRootInitialSyncProcessor.cs
```diff
@@ -16,7 +18,7 @@ public GlobalRootInitialSyncProcessor(Entities entities)
         public override IEnumerator Process(InitialPlayerSync packet, WaitScreen.ManualWaitItem waitScreenItem)
         {
             Log.Info($"Received initial sync packet with {packet.GlobalRootEntities.Count} global root entities");
-            yield return entities.SpawnAsync(packet.GlobalRootEntities);
+            yield return entities.SpawnAsync(packet.GlobalRootEntities.Cast<Entity>().ToList());
```
Jannify, line 21: Is this cast necessary?

NitroxClient/GameLogic/Spawning/PrefabChildEntitySpawner.cs
```diff
@@ -8,24 +9,26 @@ namespace NitroxClient.GameLogic.Spawning
...
     {
-        // When we first encounter a PrefabChildEntity, we simply need to assign it the right id matching the server
+        // When we encounter a PrefabChildEntity, we need to assign the id to a prefab with the same class id and index.
         public override IEnumerator SpawnAsync(PrefabChildEntity entity, TaskResult<Optional<GameObject>> result)
         {
             GameObject parent = NitroxEntity.RequireObjectFrom(entity.ParentId);
+            PrefabIdentifier prefab = parent.GetAllComponentsInChildren<PrefabIdentifier>()
+                                            .Where(prefab => prefab.classId == entity.ClassId)
+                                            .Skip(entity.ComponentIndex)
+                                            .First();
```
Jannify, line 19: Is there a reason you didn't choose `ElementAt()`? Seems more intuitive to me.

Jannify (changes requested):

NitroxClient/GameLogic/Entities.cs
Sunrunner37: Yeah this is my thought as well.  It isn't a *great* solution, but the applicable cases are quite limited.  Will monitor it as more are implemented.

NitroxClient/GameLogic/Entities.cs
Sunrunner37: sounds good.

Sunrunner37: actually this is not a mono :)

NitroxClient/GameLogic/InitialSync/GlobalRootInitialSyncProcessor.cs
Sunrunner37: Yes, entities in global root are `WorldEntity`.  C# won't autocast the collection to the base collection.

NitroxClient/GameLogic/Spawning/PrefabChildEntitySpawner.cs
Sunrunner37: makes sense.

NitroxClient/GameLogic/InitialSync/GlobalRootInitialSyncProcessor.cs
Sunrunner37: I take it back.  After the merge, this is now an `Entity` (due to the serialization issues with `List`).  Updated.

Jannify (approved):
Code wise looks fine 👍

PR #1950, opened 2022-12-29
Rework WorldPersistenceTest

Jannify:

PR #1953, opened 2022-12-30
First pass at onboarding inventories to the new entity system.

Sunrunner37:
The crusade to remove serialized game objects and migrate them to entities continues.  Nitrox has a significant amount of tech debt and redundancy with inventories.  The application has four separate systems for inventory, equipment, modules, and storage slots - all with the same packets (add/remove), packet processors, indexes, and special persistence logic.  All of this is because there is slightly different logic when spawning on the client.  

This PR starts with migrating just the inventory over to entities. In this, I slightly tweaked the escape pod structure to leverage the PrefabChild refactoring from the dropped item PR.  The resulting structure is the following (attached the real dump so it can be seen in the actual json entity format).

```
EscapePodEntity
  - PrefabChildEntity [TechType.Radio]
  - PrefabChildEntity [TechType.Fabricator]
  - PrefabChildEntity [TechType.Medkit]
  - InventoryEntity
       - InventoryItemEntity [TechType.Water]
```

Previously when Nitrox moved an item from one inventory to another it would re-serialize the entire gameobject hierarchy and persist it on the server.  Now, moving an item from the player's inventory to a storage chest will result in a single `EntityReparented` packet. In the next update to this PR, I will ensure the player structure looks as follows:

```
PlayerEntity
  - InventoryEntity
       - InventoryItemEntity [TechType.Water]
  - EquipmentInventoryEntity
       - InventoryItemEntity [TechType.Flashlight] (Metadata: Charge 50%)
```

One major change is only hooking into `container.add()` calls and no longer `container.remove()`.  This is because extracting an item from an inventory always results in a new add event.  The only exceptions are dropped items (already handled) and deleting items from reactors (we should not use the container code to track this).  

There are some outstanding items that I need to solve before merging.  First, is ensuring that the player has an entity to record inventory data (later they can be swapped to the system). Next, there are some edge cases such as plants that need to be accounted for.  However, without building working it may be difficult to test those.  When performing my tests I was able to successfully sync the inventory of the escape pod between multiple remote players.

[structure.txt](https://github.com/SubnauticaNitrox/Nitrox/files/10323572/structure.txt)


Sunrunner37:
Update: the first PR blocker has been resolved.  Items are now successfully registered as `InventoryItemEntity` under each `PlayerEntity`.  I was able to test this by putting my entire inventory into the escape pod storage, then placing it back in my inventory, and vise-versa a couple times (while restarting the client in between). 

image

Sunrunner37:
Updated to take into account the latest `DataContract` settings on master.  I've tested with the available code paths (buildings are completely broken on master). Waterproof locker storage will require some new functionality to the dropped item actions to appropriately tag it - I see this as a follow-up action item for the next PR.

NitroxPatcher/Patches/Dynamic/Planter_IsAllowedToAdd_Patch.cs
```diff
@@ -0,0 +1,30 @@
...
+using NitroxClient.GameLogic;
+using NitroxClient.MonoBehaviours;
+using NitroxModel.DataStructures;
+using NitroxModel.Helper;
+
+namespace NitroxPatcher.Patches.Dynamic;
+
+public class Planter_IsAllowedToAdd_Patch : NitroxPatch, IDynamicPatch
+{
+    public static readonly MethodInfo TARGET_METHOD = Reflect.Method((Planter p) => p.IsAllowedToAdd(default(Pickupable), default(bool)));
```
tornac1234, line 12:
```suggestion
    public static readonly MethodInfo TARGET_METHOD = Reflect.Method((Planter p) => p.IsAllowedToAdd(default, default));
```

tornac1234 (changes requested):
Overall LGCW (didn't test yet)
I just think that you forgot to edit `WorldPersistenceTest.cs` which was updated with #1950

NitroxPatcher/Patches/Dynamic/Planter_IsAllowedToAdd_Patch.cs
dartasen: We should keep those explicit types, if you use infering type, compilation will still work even there's a minor signature refactor

Sunrunner37:
Rebased off of master and modified the WorldPersistenceTest - they are passing.  Also, force pushed to get this into nice separate commits for a merge, please don't squash.  Thanks,

NitroxModel/DataStructures/GameLogic/Entities/PlayerWorldEntity.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Collections.Generic;
+using BinaryPack.Attributes;
+using NitroxModel.DataStructures.Unity;
+using System.Runtime.Serialization;
+
+namespace NitroxModel.DataStructures.GameLogic.Entities;
+
+[Serializable]
+[DataContract]
+public class PlayerWorldEntity : WorldEntity
```
Jannify, line 12: No ProtoInclude for this in WorldEntity

NitroxModel/Packets/EntityReparented.cs
```diff
@@ -0,0 +1,24 @@
+using System;
+using NitroxModel.DataStructures;
+
+namespace NitroxModel.Packets
```
Jannify, line 4: Filescoped namespace

NitroxModel/Packets/EntityReparented.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    [Serializable]
+    public class EntityReparented : Packet
+    {
+        //make sure to update the parent
+        public NitroxId Id { get; }
```
Jannify, line 10: The comment sound to me like this id is the one of the old parent but that isn't the case, right? If it is the old parent id I would rename this field to `OldParentId`

NitroxPatcher/Patches/Dynamic/CrafterLogic_TryPickupSingleAsync_Patch.cs
```diff
@@ -0,0 +1,64 @@
...
+    public static readonly MethodInfo TARGET_METHOD = AccessTools.EnumeratorMoveNext(TARGET_METHOD_ORIGINAL);
+
+    public static readonly OpCode INJECTION_OPCODE = OpCodes.Call;
+    public static readonly object INJECTION_OPERAND = Reflect.Method(() => CrafterLogic.NotifyCraftEnd(default(GameObject), default(TechType)));
+
+    private static readonly MethodInfo COMPONENT_GAMEOBJECT_GETTER = Reflect.Property((Component t) => t.gameObject).GetMethod;
+
+    public static IEnumerable<CodeInstruction> Transpiler(MethodBase original, IEnumerable<CodeInstruction> instructions)
+    {
+        Validate.NotNull(INJECTION_OPERAND);
```
Jannify, line 30: Minor: `INJECTION_OPERAND` can't be null as we get it with the lambda expression

NitroxPatcher/Patches/Dynamic/CrafterLogic_TryPickupSingleAsync_Patch.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        foreach (CodeInstruction instruction in instructions)
+        {
+            yield return instruction;
+
+            if (instruction.opcode.Equals(INJECTION_OPCODE) && instruction.operand.Equals(INJECTION_OPERAND))
+            {
+                /*
+                 * Callback(this.gameObject, gameObject);
+                 */
```
Jannify, line 40: Would write here what line exactly we replace with which, e.g.:
```suggestion
                /*
                 * CrafterLogic.NotifyCraftEnd(this.gameObject, gameObject)
                 * is replaced with
                 * Callback(this.gameObject, gameObject);
                 */
```

NitroxServer/Communication/Packets/Processors/EntityReparentedProcessor.cs
```diff
@@ -0,0 +1,57 @@
+using NitroxModel.DataStructures.GameLogic;
+using NitroxModel.DataStructures.Util;
+using NitroxModel.Packets;
+using NitroxServer.Communication.Packets.Processors.Abstract;
+using NitroxServer.GameLogic;
+using NitroxServer.GameLogic.Entities;
+
+namespace NitroxServer.Communication.Packets.Processors
```
Jannify, line 8: Filescoped namespace

NitroxServer/Communication/Packets/Processors/EntityReparentedProcessor.cs
```diff
@@ -0,0 +1,57 @@
...
+            Optional<Entity> newParent = entityRegistry.GetEntityById(packet.NewParentId);
+
+            // old parent may not exist anymore, so don't expect it.
+            if (newParent.HasValue)
+            {
+                newParent.Value.ChildEntities.Add(entity);
+            }
+            else
+            {
+                Log.Info($"Could not find new parent to reparent an entity for: {packet}");
```
Jannify, line 51: If this happens is the Item in a limbo? Where is it stored?

Nitrox.Test/Server/Serialization/WorldPersistenceTest.cs
```diff
@@ -102,7 +102,6 @@ public void VehicleDataTest(PersistedWorldData worldDataAfter, string serializer
     [DataTestMethod, DynamicWorldDataAfter]
     public void InventoryDataTest(PersistedWorldData worldDataAfter, string serializerName)
     {
-        AssertHelper.IsListEqual(worldData.WorldData.InventoryData.InventoryItems.OrderBy(x => x.ItemId), worldDataAfter.WorldData.InventoryData.InventoryItems.OrderBy(x => x.ItemId), ItemDataTest);
```
Jannify, line 105: Tests are missing for:
- InventoryEntity
- InventoryItemEntity
- CrafterMetadata
- PlantableMetadata

NitroxClient/GameLogic/Spawning/Metadata/CrafterMetadataProcessor.cs
```diff
@@ -0,0 +1,44 @@
...
+using NitroxModel.DataStructures.GameLogic.Entities.Metadata;
+using NitroxModel_Subnautica.DataStructures;
+using UnityEngine;
+
+namespace NitroxClient.GameLogic.Spawning.Metadata;
+
+public class CrafterMetadataProcessor : GenericEntityMetadataProcessor<CrafterMetadata>
+{
+    // small increase to prevent this player from swiping item from remote player
+    public static float ANTI_GRIEF_DURATION_BUFFER = 0.2f;
```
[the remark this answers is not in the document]
Measurity: The `static` keyword is incompatible if it's `const` and should be removed.

NitroxClient/GameLogic/Spawning/InventoryEntitySpawner.cs
```diff
@@ -0,0 +1,36 @@
...
+            NitroxEntity.SetNewId(container.gameObject, entity.Id);
+            result.Set(Optional.OfNullable(container.gameObject));
+        }
+        else
+        {
+            Log.Error($"Unable to find prefab for: {entity}");
+            result.Set(Optional.Empty);
+        }
+
+        yield break;
```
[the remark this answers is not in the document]
tornac1234: This one is necessary to keep the function as an IEnumerator

NitroxClient/GameLogic/Spawning/Metadata/CrafterMetadataProcessor.cs
```diff
@@ -0,0 +1,44 @@
...
+    private void SpawnItemInCrafter(GameObject gameObject, CrafterMetadata metadata)
+    {
+        GhostCrafter ghostCrafter = gameObject.RequireComponentInChildren<GhostCrafter>(true);
+
+        float ellaspedFromStart = (DayNightCycle.main.timePassedAsFloat - metadata.StartTime);
+
+        // If the crafter started way in the past, set duration to 0.01 to instantly have the item crafted (the  
+        // craft function will not work with 0).  Keeping track of both the duration and start time allows us to
+        // solve use-cases such as reloading when an item is being crafted or not picked up yet. 
+        float duration = Mathf.Max(metadata.Duration - ellaspedFromStart + ANTI_GRIEF_DURATION_BUFFER, 0.01f);
```
tornac1234, line 40:
```suggestion
        float elapsedFromStart = (DayNightCycle.main.timePassedAsFloat - metadata.StartTime);

        // If a craft started in the past, set duration to 0.01 (0 doesn't work) to instantly have the item crafted.
        // Keeping track of both the duration and start time allows us to solve use-cases such as
        // reloading when an item is being crafted or not picked up yet.
        float duration = Mathf.Max(metadata.Duration - elapsedFromStart + ANTI_GRIEF_DURATION_BUFFER, 0.01f);
```

tornac1234 (dismissed):
After testing, it seems that the escape pod didn't register the item put inside its storage (when reconnecting, it was empty when I filled it before leaving).

Sunrunner37:
Resolved all open feedback. Let me know if there are any additional blockers.

Nitrox.Test/Server/Serialization/WorldPersistenceTest.cs
```diff
@@ -436,14 +450,18 @@ private static PersistedWorldData GeneratePersistedWorldData()
                 {
                     Entities = new List<Entity>()
                     {
-                        new PrefabChildEntity(new NitroxId(), "pretty class id", new NitroxTechType("Radio"), 1, null, new NitroxId()),
+                        new PrefabChildEntity(new NitroxId(), "pretty class id", new NitroxTechType("Fabrifactor"), 1, new CrafterMetadata(new NitroxTechType("FilteredWater"), 100, 10), new NitroxId()),
```
Measurity, line 453: What is a Fabrifactor? 👀

Sunrunner37: It is new in the BZ update

Sunrunner37:
Updated the PR based on latest testing.  To summarize there were a few things to tweak:

1) The server did not know about items that are spawned on pickup (creepvine fruit, fire extinguisher, and spawn item command).  Implemented some generic logic that unfamiliar pickups go to the server first.  This solves for lots and lots of different codepaths that spawn prefabs with a click event.

2) The code was not properly translating from WorldEntity to InventoryItemEntity on pickup.  This has been corrected.

I was able to test a range of items from around the world.  Everything that the server knows about works, example:

image

I was also able to test with multiple players at the same time.  This is working besides an ordering issue (RemoteInitial sync requires GlobalRoot and vise versa); this will be corrected in the next PR when I migrate the rest of the remote player logic to the new system.  However, this is out of scope for this PR and things are working as I intend.

Measurity (approved):
Works on my side as well.
Items didn't enter in real time in container except for seed bundle but on rejoin they showed up.

dartasen (approved):
Works on my side too

NitroxServer/Communication/Packets/Processors/EntityReparentedProcessor.cs
Jannify: Probably this could help: https://www.meziantou.net/updating-your-project-to-use-file-scoped-namespaces.htm

Jannify (approved):
LGTM CW :)

PR #1959, opened 2023-01-03
Fix the base ghost material name.

Shirkie01:
Updates the material name used by the base construct ghost models.


Measurity:
`ghostmodel` was changed to addressable asset (moved out of `resources.assets` file to `main-discrete_assets_materials_a35c903f46c99d7d13561070823d1dc2`) see:
image


`constructingGhost` also existed in legacy Subnautica and matches with current one, so it's likely not correct to change to `constructingGhost`.

PR #1960, opened 2023-01-04
Fixed dispose of assemblies loaded via asset parsing on server

Measurity:
Assemblies loaded by reference of the targeted asset assemblies can't be unloaded without changing AssetTools.NET code here. (We only have a reference to some of the assemblies, and can only dispose those):
https://github.com/nesrak1/AssetsTools.NET/blob/upd21-with-inst/AssetTools.NET/Extra/MonoDeserializer/MonoCecilTempGenerator.cs#L63-L69

**Preview of handles after full load of server**
image

### DIFF
--- a/NitroxServer-Subnautica/Resources/Parsers/PrefabPlaceholderGroupsParser.cs
+++ b/NitroxServer-Subnautica/Resources/Parsers/PrefabPlaceholderGroupsParser.cs
@@ -8,17 +8,17 @@ using AddressablesTools;
 using AddressablesTools.Catalog;
 using AssetsTools.NET;
 using AssetsTools.NET.Extra;
-using Mono.Cecil;
 using NitroxServer_Subnautica.Resources.Parsers.Helper;
 using NitroxServer.Resources;
 
 namespace NitroxServer_Subnautica.Resources.Parsers;
 
-public class PrefabPlaceholderGroupsParser
+public class PrefabPlaceholderGroupsParser : IDisposable
 {
     private readonly string prefabDatabasePath;
     private readonly string aaRootPath;
     private readonly AssetsBundleManager am;
+    private readonly ThreadSafeMonoCecilTempGenerator monoGen;
 
     public PrefabPlaceholderGroupsParser()
     {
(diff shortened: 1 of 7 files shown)


NitroxServer-Subnautica/Resources/Parsers/PrefabPlaceholderGroupsParser.cs
```diff
@@ -48,17 +48,9 @@ public PrefabPlaceholderGroupsParser()
 
         // Select only prefabs with a PrefabPlaceholdersGroups component in the root ans link them with their dependencyPaths
         ConcurrentDictionary<string, string[]> prefabPlaceholdersGroupPaths = GetAllPrefabPlaceholdersGroupsFast(loadAddressableCatalog);
-        am.UnloadAll();
```
Jannify, line 51: This line should stay bc the internal cache list is slowing down the process more than loading a few assets again. This is maybe better in new AssetToolsNetVersion but we need a byte to texture library bc ATNs sub-package is only for netStadnard.

Jannify (changes requested):

NitroxServer-Subnautica/Resources/Parsers/ThreadSafeMonoCecilTempGenerator.cs
```diff
@@ -0,0 +1,41 @@
+using System;
+using System.Collections.Generic;
+using AssetsTools.NET;
+using AssetsTools.NET.Extra;
+using Mono.Cecil;
+
+namespace NitroxServer_Subnautica.Resources.Parsers;
```
Jannify, line 7: Would move it to the helpers dir.

NitroxServer-Subnautica/Resources/Parsers/PrefabPlaceholderGroupsParser.cs
Measurity: This would have been great as comment in code. Will add it back later. On my PC server starts in 3.6s so didn't notice.

Jannify (approved):
LGTM CW

PR #1961, opened 2023-01-05
Fix EntitySlots spawning

Jannify:
This fixes EntitySlots (=> all random spawned loot and fragments) not spawning

~~I am not happy how `PlaceholderGroupWorldEntity` turned out. Feedback is appreciated, espacially since the current solution is not working with ProtoBuf (bc null).~~

### DIFF
--- a/NitroxClient/GameLogic/Spawning/WorldEntities/PlaceholderGroupWorldEntitySpawner.cs
+++ b/NitroxClient/GameLogic/Spawning/WorldEntities/PlaceholderGroupWorldEntitySpawner.cs
@@ -11,10 +11,12 @@ namespace NitroxClient.GameLogic.Spawning.WorldEntities;
 
 public class PlaceholderGroupWorldEntitySpawner : IWorldEntitySpawner
 {
+    private readonly WorldEntitySpawnerResolver spawnerResolver;
     private readonly DefaultWorldEntitySpawner defaultSpawner;
 
-    public PlaceholderGroupWorldEntitySpawner(DefaultWorldEntitySpawner defaultSpawner)
+    public PlaceholderGroupWorldEntitySpawner(WorldEntitySpawnerResolver spawnerResolver, DefaultWorldEntitySpawner defaultSpawner)
     {
+        this.spawnerResolver = spawnerResolver;
         this.defaultSpawner = defaultSpawner;
     }
 
@@ -34,7 +36,10 @@ public class PlaceholderGroupWorldEntitySpawner : IWorldEntitySpawner
             result.Set(Optional.Empty);
             yield break;
         }
+        
+        result.Set(prefabPlaceholderGroupGameObject);
 
+        // Spawning PrefabPlaceholders
         PrefabPlaceholdersGroup prefabPlaceholderGroup = prefabPlaceholderGroupGameObject.Value.GetComponent<PrefabPlaceholdersGroup>();
 
         for (int index = 0; index < placeholderGroupEntity.PrefabPlaceholders.Length; index++)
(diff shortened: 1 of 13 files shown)


NitroxClient/GameLogic/Spawning/WorldEntities/PlaceholderGroupWorldEntitySpawner.cs
```diff
@@ -7,75 +7,93 @@
...
 
-        public bool SpawnsOwnChildren()
-        {
-            return true;
-        }
-}
-
+    public bool SpawnsOwnChildren()
+    {
+        return true;
```
Sunrunner37, line 97: One idea I had is to introduce a new spawner for the placeholder children, called PrefabPlaceholderEntitySpawner.  That way we do not need to spawn our own children and if it is a world entity then it will automatically be routed to the correct WorldEntity spawner.  Then the logic here is very simple: spawn the parent, tag it with nitrox id, and allow children to be spawned upstream.

Jannify: Problem is that we need the group to get the correct parent and position. We can not completely handle even the world entities independent from the group.

Sunrunner37: Is this because we need to do the slot transformation after the item is spawned (lines 83-85 above)?  Is it possible that we can do this as a pre-processing step when creating the entity and not at spawn time?

Jannify: It's not 83-85, that is just a fix for a strange bug.
Ideally prefabs are spawned parented to the corresponding PrefabPlaceholder. Problem is that the server doesn't know about these placeholders nor the hierarchy.

tornac1234:
It works on my machine, fragments are spawning right, and I didn't encounter the duplicated entities issues that we had last time we had this issue so it's fine.

NitroxClient/GameLogic/Spawning/WorldEntities/PlaceholderGroupWorldEntitySpawner.cs
```diff
@@ -63,19 +69,31 @@ public IEnumerator SpawnAsync(WorldEntity entity, Optional<GameObject> parent, E
...
+                    IWorldEntitySpawner spawner = spawnerResolver.ResolveEntitySpawner(slotEntity);
+                    Optional<GameObject> slotEntityParent = Optional.Of(prefabPlaceholder.gameObject);
+
+                    TaskResult<Optional<GameObject>> slotEntityTaskResult = new();
+                    yield return spawner.SpawnAsync(slotEntity, slotEntityParent, cellRoot, slotEntityTaskResult);
+
+                    if (slotEntityTaskResult.value.HasValue)
+                    {
+                        // For some reasons it's not zero after spawning so we reset it here
+                        slotEntityTaskResult.value.Value.transform.localPosition = Vector3.zero;
```
dartasen, line 84: Is it possible that the local rotation isn't equal to the Identity?

NitroxModel/DataStructures/GameLogic/Entities/WorldEntity.cs
```diff
@@ -14,10 +14,10 @@ namespace NitroxModel.DataStructures.GameLogic.Entities
...
     [Serializable]
     [DataContract]
-    [ProtoInclude(10, typeof(PlaceholderGroupWorldEntity))]
-    [ProtoInclude(20, typeof(EscapePodWorldEntity))]
-    [ProtoInclude(30, typeof(PlayerWorldEntity))]
-    [ProtoInclude(40, typeof(VehicleWorldEntity))]
+    [ProtoInclude(51, typeof(PlaceholderGroupWorldEntity))]
+    [ProtoInclude(52, typeof(EscapePodWorldEntity))]
+    [ProtoInclude(53, typeof(PlayerWorldEntity))]
+    [ProtoInclude(54, typeof(VehicleWorldEntity))]
```
dartasen, line 20: Not sure to understand

Measurity: This edit makes sense:

https://stackoverflow.com/a/984325/1277156
> [ProtoInclude] works a lot like [XmlInclude] for XmlSerializer - or [KnownType] for DataContractSerializer - it allows it to recognise subclasses of a type during (de)serialization. The only additional thing is that it needs a tag (number) to identify each sub-type (that must be unique, and not clash with any of the fields from the parent type).

In other words, `ProtoInclude` is like a `[DataMember(Order = x)]` in that you don't need gaps in between includes. **But** they should be unique in the same type. Using a starting offset of 50 for protoincludes will help not conflict with the `[DataMember(Order = x)]` on the fields *and* still leave room for more fields without becoming binary incompatible.

NitroxClient/GameLogic/Spawning/WorldEntities/PlaceholderGroupWorldEntitySpawner.cs
Jannify: Need to check this with a side by side comparison to base SN.

NitroxModel/DataStructures/GameLogic/Entities/WorldEntity.cs
Jannify: From the docs I found that the key needs to be unique (scope here being `ProtoInclude` as well as `DataMember`) for the type it's declared in. No docs explained what happens for types which are parents while being a subtype at the same time. My guess is that they will collide. Anyway I changed it to 51+ to better reflect the range we have here. While Entity.cs increments it's `ProtoInclude` key by 10, `WorldEntity` increments now by 1 starting from the Key it has in `Entity` (50).

Measurity: <https://stackoverflow.com/a/56994082/1277156>

PR #1968, opened 2023-01-10
Vehicles Initial Entity Refactoring.  [1/3]

Sunrunner37:
Note: This is a single PR in a series.  I've split them up and held them back to reduce cognitive load.  However, one disadvantage is that some functionality may be assumed missed.  I will try to clarify through discussion comments and the description.

This is the first pass at bringing vehicles onto the new entity system.  I've consolidated all spawning logic that was spread across Vehicles.cs and constructor patches into a `VehicleWorldEntitySpawner`.  The previous concept of `InteractiveChildObjectIdentifier` has been moved to simple `ChildEntities` (note: the logic needs to be optimized but it identical to the existing logic, just using a new `PathBasedChildEntity` instead).  

After the living large patch, vehicle spawning does not work at all.  This PR will enable vehicle spawning to work again for the constructor and logging back into the world.  This does not yet enable the moving pieces or components yet; however, it does ensure all vehicle children are correctly tagged with ids.  In the next PR, I consolidate all of the cyclops actions (lights, silent running, sonar, etc) into entity metadata and remove all of those packets, processors, and bloat. 

image


NitroxClient/GameLogic/MobileVehicleBay.cs
```diff
@@ -20,20 +26,27 @@ public MobileVehicleBay(IPacketSender packetSender, Vehicles vehicles)
...
-            Log.Debug($"Building item from constructor with id: {constructorId}");
+            if (!TransmitLocalSpawns)
+            {
+                return;
+            }
 
-            VehicleModel vehicleModel = vehicles.BuildVehicleModelFrom(constructedObject, techType);
-            vehicles.AddVehicle(vehicleModel);
+            // Sometimes build templates, such as the cyclops, are already tagged with IDs.  Remove any that exist to retag.
+            UnityEngine.Component.DestroyImmediate(constructedObject.GetComponent<NitroxEntity>());
```
Measurity, line 35: Think we should prefix this with "TODO" here. (Figure out why NitroxEntity is in the template). Recreating the NitroxEntity can be problematic if NitroxEntity gets more importance/state in the future.

NitroxClient/GameLogic/MobileVehicleBay.cs
```diff
@@ -20,20 +26,27 @@ public MobileVehicleBay(IPacketSender packetSender, Vehicles vehicles)
...
+                
+            packetSender.Send(new EntitySpawnedByClient(vehicleEntity));
 
             MonoBehaviour monoBehaviour = constructor.GetComponent<MonoBehaviour>();
             //We want to store the fallen position of the object to avoid flying object on reload 
             if (monoBehaviour)
             {
-                monoBehaviour.StartCoroutine(vehicles.UpdateVehiclePositionAfterSpawn(vehicleModel, constructedObject, duration + 10.0f));
+                monoBehaviour.StartCoroutine(vehicles.UpdateVehiclePositionAfterSpawn(constructedObjectId, techType, constructedObject, duration + 10.0f));
             }
```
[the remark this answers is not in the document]
Sunrunner37: It looks like this code was originally written with only game object passed in.  I've updated it to remove the mono fetching.

Sunrunner37:
Any comments from anyone else?  Looking to submit 2/3 in the series.

NitroxClient/GameLogic/Helper/VehicleChildEntityHelper.cs
```diff
@@ -0,0 +1,61 @@
...
+        NitroxId parentId = NitroxEntity.GetId(vehicle);
+        List<Entity> children = new List<Entity>();
+
+        string constructedObjectsName = vehicle.GetFullHierarchyPath();
+
+        foreach (Type type in interactiveChildTypes)
+        {
+            Component[] components = vehicle.GetComponentsInChildren(type, true);
+
+            for(int i = 0; i < components.Length; i++)
```
Coding-Hen, line 41: Is there a reason to use a indexed for loop here and not just loop over components? It would remove the assignment line below and any potential array index issues

Sunrunner37: This file is just a rename of an existing file `VehicleInteractiveIdentifierHelper".  I didn't touch any of the logic in this PR.  However, I have fully optimized it in a subsequent commit (waiting on this PR).

tornac1234 (approved):
I'm fine with the code, also tested and cyclops reappeared successfully (if it was the only point of this PR, then it's good)

NitroxClient/GameLogic/MobileVehicleBay.cs
```diff
@@ -20,20 +26,27 @@ public MobileVehicleBay(IPacketSender packetSender, Vehicles vehicles)
...
+            vehicleEntity.ChildEntities = VehicleChildEntityHelper.ExtractChildren(constructedObject);
+                
+            packetSender.Send(new EntitySpawnedByClient(vehicleEntity));
 
             MonoBehaviour monoBehaviour = constructor.GetComponent<MonoBehaviour>();
             //We want to store the fallen position of the object to avoid flying object on reload 
             if (monoBehaviour)
             {
-                monoBehaviour.StartCoroutine(vehicles.UpdateVehiclePositionAfterSpawn(vehicleModel, constructedObject, duration + 10.0f));
+                monoBehaviour.StartCoroutine(vehicles.UpdateVehiclePositionAfterSpawn(constructedObjectId, techType, constructedObject, duration + 10.0f));
```
tornac1234: Idk if you have plans for this next, but maybe this 10.0f should be replaced by ``VehicleWorldEntitySpawner.CONSTRUCTION_DURATION_IN_SECONDS`` if possible.

Coding-Hen (approved):
I haven't tested in game but code wise looks good to me.

PR #1975, opened 2023-01-20
Added source generator for removing boilerplate with serializable types

Measurity:
- [ ] Rework serializable generator to generate model implementation function for protobuf and JSON serializers.


Measurity:
Source generator doesn't seem well suited for handling project-wide and hierarchy classes at once while still having good performance so I'll close this for now.

PR #1977, opened 2023-01-24
Fixed Reflect API for out parameters

Measurity:
Expressions can't create and capture variables. But they can refer to a preexisting variable.

### DIFF
--- a/NitroxModel/Helper/Reflect.cs
+++ b/NitroxModel/Helper/Reflect.cs
@@ -2,127 +2,134 @@
 using System.Linq.Expressions;
 using System.Reflection;
 
-namespace NitroxModel.Helper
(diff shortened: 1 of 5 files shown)


NitroxModel/Helper/Reflect.cs
```diff
@@ -1,128 +1,137 @@
-﻿using System;
+﻿extern alias JB;
```
Jannify, line 1: What is the reason you include `JetBrains.Annotations` here?

Measurity: No reason anymore. Forgot to remove it.

Jannify (approved):
LGTM CW

tornac1234 (approved):
CWLGTM and WOMM

PR #1982, opened 2023-02-11
Fix respawning and simulation issues when reloading batches.

Sunrunner37:
Since the last update there were issues respawning entities when swimming to a previously accessed batch.  Additionally, there were also issues locking in simulated entities on the client.

This PR has several fixes for these issues:

1. Allow simulated gameobejcts to be resolved before transmitting transforms.  Previously, we required game objects to be present when receiving a simulation order from the server.  However, this doesn't work too well with new async entity spawning when the object may not be there until later.  An additional benefit of resolving before transform transmission is the ability to resync or respawn entities without worrying about retagging entity broadcasting. 
2. Re-introduce ProtoContract to NitroxEntity, which must have fallen off during the large refactoring for the update.
3. Cleared out the player's chunks when they reconnect allowing the server to resend previously seen entities.  This is a long standing bug but never occurred before the update because we would receive an entity cell change on disconnect. 
4. Moved the patch point that removes entity cells to a more reliable patch point.


NitroxPatcher/Patches/Dynamic/EntityCell_SleepAsync_Patch.cs
```diff
@@ -0,0 +1,53 @@
...
+using System.Reflection;
+using System.Reflection.Emit;
+using HarmonyLib;
+using NitroxModel.Helper;
+
+namespace NitroxPatcher.Patches.Dynamic;
+
+/**
+ * Entity cells will go sleep when the player gets out of range.  This needs to be reported to the server so they can lose simulation locks.
+ */
```
tornac1234, line 11:
```suggestion
/// <summary>
/// Entity cells will go sleep when the player gets out of range. This needs to be reported to the server so they can lose simulation locks.
/// </summary>
```

NitroxPatcher/Patches/Dynamic/EntityCell_SleepAsync_Patch.cs
```diff
@@ -0,0 +1,53 @@
...
+                 */
+                yield return TranspilerHelper.Ldloc<EntityCell>(original);
+                yield return new CodeInstruction(OpCodes.Call, Reflect.Method(() => Callback(default(EntityCell))));                
+            }
+        }
+    }
+
+    public static void Callback(EntityCell entityCell)
+    {
+        Resolve<NitroxClient.GameLogic.Terrain>().CellUnloaded(entityCell.BatchId, entityCell.CellId, entityCell.Level);
```
tornac1234, line 45:
```suggestion
        Resolve<Terrain>().CellUnloaded(entityCell.BatchId, entityCell.CellId, entityCell.Level);
```

NitroxPatcher/Patches/Dynamic/EntityCell_SleepAsync_Patch.cs
```diff
@@ -0,0 +1,53 @@
+using System.Collections.Generic;
+using System.Reflection;
+using System.Reflection.Emit;
+using HarmonyLib;
```
tornac1234, line 4:
```suggestion
using HarmonyLib;
using NitroxClient.GameLogic;
```

NitroxPatcher/Patches/Dynamic/EntityCell_SleepAsync_Patch.cs
tornac1234: I'm pretty sure that the compiler originates it from the right package even when inserted in Subnautica as a transpiler

tornac1234 (approved):
Nothing much to say (just visual stuff in the patch). Looks good to me code-wise and didn't receive errors when swimming around (alone) apart from one issue I encountered but I'm pretty sure it isn't related to the PR.

PR #1984, opened 2023-02-12
Option 1: Added IPacketSender.SendIfGameCode and used it instead of .Send()

Measurity:
See option 2 here: #1994 

The prior PR (#1894) altered too many elements, therefore this PR only includes the addition of `IPacketSender.SendIfGameCode(Packet)`.

`IPacketSender.Send` should only be employed if it is not located in the Patch or GameLogic namespaces. Currently, it is utilized in only two instances.

Going forward, we should eliminate the usage of `.Suppress` wraps, which have been marked as deprecated.

## Why this PR?
We expect packet sends to only happen to "notify" server, but we also call `.Send(packet)` indirectly when we use UWE APIs due to our injected code into UWE APIs, and also when we use packet sends in our "reusable" API. This makes it impossible to limit a solution to just patch classes.

The mixed use of UWE API in 2 domains (info poking & UWE API reuse) is the broader problem.

Of these solutions I went with option 2:
 1. Duplicate every patched method: one with packet sending used by game domain and one without for UWE API convenience in Nitrox code. This also requires a change to `GameLogic` namespaced classes that are used in `_Patch` classes to never send packets or opt out of them via parameter.
 2. Have a method that can swap/branch its implementation based on who called it.


NitroxClient/GameLogic/ChatUI/PlayerChatManager.cs
```diff
@@ -83,25 +83,25 @@ public void SendMessage()
...
-                multiplayerSession.Send(new ServerCommand(trimmedInput.Substring(1)));
+                multiplayerSession.SendIfGameCode(new ServerCommand(trimmedInput.Substring(1)));
                 playerChat.InputText = "";
                 playerChat.Select();
                 return;
             }
 
             // We shouldn't add the message to the local chat instantly but instead let the server tell us if this message is added or not
-            multiplayerSession.Send(new ChatMessage(multiplayerSession.Reservation.PlayerId, trimmedInput));
+            multiplayerSession.SendIfGameCode(new ChatMessage(multiplayerSession.Reservation.PlayerId, trimmedInput));
```
Sunrunner37, line 93: Can these paths echo?

NitroxClient/Helpers/ThrottledPacketSender.cs
```diff
@@ -24,7 +24,7 @@ public void Update()
                     continue;
                 }
 
-                if (packetSender.Send(throttledPacket.Packet))
+                if (packetSender.SendIfGameCode(throttledPacket.Packet))
```
Sunrunner37, line 27: Will this ever work since it always originates from our Update()?

NitroxClient/Helpers/ThrottledPacketSender.cs
```diff
@@ -48,8 +48,12 @@ public bool SendThrottled(Packet packet, Func<Packet, object> dedupeMethod, floa
                 return true;
             }
 
+            if (!packetSender.SendIfGameCode(packet))
+            {
+                return false;
```
Sunrunner37, line 53: Contract of this method isn't very clear.  In this case, the caller will assume that the packet was not sent due to throttling and may branch/do other logic based on this assumption. (worst case, tuck the event away to send later, which would not be wanted in this case).

NitroxModel/IPatch.cs
```diff
@@ -0,0 +1,8 @@
+namespace NitroxModel;
+
+/// <summary>
+///     Interface to mark a type as a patcher, which inject code.
+/// </summary>
+public interface IPatch
```
Sunrunner37, line 6: Why not just `INitroxPatch`?

NitroxClient/Helpers/ThrottledPacketSender.cs
Measurity: In order to not change logic I basically renamed `Send` to `SendIfGameCode`, keeping the original suppression check until it can be removed (see TODO in SendIfGameCode).

NitroxModel/IPatch.cs
Measurity: `INitroxPatch` uses Harmony reference. I need it in NitroxModel which, rightfully, does not have `Harmony`.

NitroxClient/Helpers/ThrottledPacketSender.cs
Measurity: I have removed the bool return and changed it to an if..else because it had an either/or intend before.

Measurity:
I didn't go with a flag-based solution since it would have to be specified manually by each call to patch/inject-related code, undoing the convenience of this PR. I updated the top comment to explain.

PR #1985, opened 2023-02-13
Fix prefab placeholders not showing up when restarting the server.

Sunrunner37:
The prefab placeholder code had a strong dependency on the ordering of children to execute successfully.  However, the transmission and persistence has no ordering guarantees.  Instead, I swapped out the custom `PrefabPlaceholderEntity` type to use the existing `PrefabChildEntity` type.  This type has a built in component index that can used to maintain the ordering.  With this change, prefab placeholders correctly respawn when restarting the server.

In addition to this change, I also included two small tweaks to make entity spawning a little better:

1) Do not attempt to spawn children entities unless the parent spawner returned a game object. 

2) In the default entity spawner, ensure that the cellRoot is not being unloaded before assigning the entity as a child.


Sunrunner37:
Switching to draft as there are still a subset of placeholders that aren't working.  It seems that `PlaceholderGroupWorldEntity` can have nested `PlaceholderGroupWorldEntity` and doesn't seem to be handled by the current spawner. 

example entity `82dc6b9e-22ed-28f4-b32a-706af8e1a572`(supply crate) inside `wreck1` (`e1eb89eb-a220-f457-971b-a5a83104d13f`)

Jannify:
Nested `PlaceholderGroupWorldEntity` especially the supply grates, have worked on my branch so it's probably something else interfering not working.

Sunrunner37:
> Nested PlaceholderGroupWorldEntity especially the supply grates, have worked on my branch so it's probably something else interfering/not working.

From my testing, the nested supply grates will spawn the first time but not on subsequent restarts.  Still looking into the issue; however, I suspect that it might only work the first time due to the legacy 'pendingParent' code that is in the entity spawner (
https://github.com/SubnauticaNitrox/Nitrox/blob/master/NitroxClient/GameLogic/Entities.cs#L153).  The `PlaceholderGroupWorldEntitySpawner` only iterates over direct children and doesn't recursively traverse the hierarchy.  This works when the `PlaceholderGroupWorldEntity` is a direct children but if there is any other type in the middle then the grandchildren may not be traversed.

> Also why did you swap the existing entity type instead of updating it? Now we need to serialize two unused extra fields for every PlaceholderGroup Entity.

Which extra fields? I believe it already had classId and componentIndex is the new thing that fixs the ordering issue).

Sunrunner37:
Moving this out of draft as the experience is now significantly better than master.  There were a couple issues at play here:

1) EntitySlot placeholders were being assigned to the wrong parent, typically the cell root.  This caused orphans when pruning out cell root entities because children could not be updated.

2) `PlaceholderGroupWorldEntitySpawner` was only looping through direct children to spawn.  However, many of the children `WorldEntity` had nested children inside them.  These seem to spawn on the first load because they would be yielded before the parent and be considered pending children.  This doesn't seem to happen on subsequent runs.  The parent behavior in the first point may have also contributed to some of the odd behavior.  I think as a rule of thumb, if a spawner emits SpawnsOwnChildren=true then it needs to traverse the entire hierarchy unless it knows for certain that it is only one level deep. 

Some pictures below.  These represent a before/after going to wreck1, restarting, and back to wreck1.  Master vs PR:

image

image

Sunrunner37:
Moving out of draft for review.  The functionality isn't perfect but significantly better than master.

With the latest push, seems that even nested groups are fine when deserialized from a batch (i.e. player went out of range and came back).  Here is an example supply crate that I opened, took the items, warped out, and came back:

image

Sunrunner37:
Moving out of draft for review. The functionality isn't perfect but significantly better than master.

With the latest push, seems that even nested groups are fine when deserialized from a batch (i.e. player went out of range and came back). Here is an example supply crate that I opened, took the items, warped out, and came back:

image

Sunrunner37:
One thing I noticed is that crashfish homes are also PrefabPlaceholderGroups.  They were throwing an exception because the new spawner was not compatible with the existing crashfish bootstrapper (the code that actually adds the crash on the server side).  

Fixed the issue and now fish and powder are spawning correctly.  And yes, they pop out and attack correctly:

image

Also added in a protection to only allow placeholder groups to respawn when coming from a batch refresh.

dartasen (approved):
Works for me in game

NitroxClient/GameLogic/Spawning/WorldEntities/PlaceholderGroupWorldEntitySpawner.cs
```diff
@@ -1,16 +1,26 @@
...
+/// Group placeholders are unique in that the game cleans up their children each time a player moves out of its batch. In single player, the placeholders
+/// would be respawned by iterating over the placeholders - this is patched out in nitrox because we need to tag them. Due to this behavior, we need to 
+/// respawn the placeholders when the batch is reloaded.  The top level group entity will still exist in the batch; thus, we will try to load the object
+/// by ID first and if we can't find it then spawn the entity in.  
+/// </summary>
 public class PlaceholderGroupWorldEntitySpawner : IWorldEntitySpawner
 {
+    // A simple way to detect if the respawn is triggered from a batch deserialization is to tag the group with a mono that is not registered in protobuf.
+    // If the group is untagged on a subsequent call then we know to respawn children again. 
+    private class BatchRespawnDetection : MonoBehaviour { };
```
Jannify, line 22: Would call it `NitroxBatchRespawnDetection` so its origin is clear in the SceneDebugger.

NitroxClient/GameLogic/Spawning/WorldEntities/PlaceholderGroupWorldEntitySpawner.cs
```diff
@@ -22,9 +32,33 @@ public PlaceholderGroupWorldEntitySpawner(WorldEntitySpawnerResolver spawnerReso
...
     public IEnumerator SpawnAsync(WorldEntity entity, Optional<GameObject> parent, EntityCell cellRoot, TaskResult<Optional<GameObject>> result)
     {
-        TaskResult<Optional<GameObject>> prefabPlaceholderGroupTaskResult = new();
-        yield return defaultSpawner.SpawnAsync(entity, parent, cellRoot, prefabPlaceholderGroupTaskResult);
-        Optional<GameObject> prefabPlaceholderGroupGameObject = prefabPlaceholderGroupTaskResult.Get();
+        Optional<GameObject> prefabPlaceholderGroupGameObject = NitroxEntity.GetObjectFrom(entity.Id);
+
+        if (prefabPlaceholderGroupGameObject.HasValue)
+        {
+            /**
```
Jannify, line 39:
```suggestion
            /*
```

NitroxServer-Subnautica/GameLogic/Entities/Spawning/CrashFishBootstrapper.cs
```diff
@@ -4,23 +4,27 @@
...
-        private WorldEntity SpawnChild(WorldEntity parentEntity, DeterministicGenerator deterministicBatchGenerator, TechType techType, string classId)
-        {
-            NitroxId id = deterministicBatchGenerator.NextId();
+    private WorldEntity SpawnChild(Entity parentEntity, DeterministicGenerator deterministicBatchGenerator, TechType techType, string classId, int level)
+    {
+        NitroxId id = deterministicBatchGenerator.NextId();
 
-            return new WorldEntity(new NitroxVector3(0, 0, 0), new NitroxQuaternion(0, 0, 0, 1), new NitroxVector3(1, 1, 1), techType.ToDto(), parentEntity.Level, classId, true, id, parentEntity, false, null);
-        }
+        return new WorldEntity(new NitroxVector3(0, 0, 0), new NitroxQuaternion(0, 0, 0, 1), new NitroxVector3(1, 1, 1), techType.ToDto(), level, classId, true, id, parentEntity, false, null);
```
Jannify, line 28:
```suggestion
        return new WorldEntity(NitroxVector3.Zero, NitroxQuaternion.Identity, NitroxVector3.One, techType.ToDto(), level, classId, true, id, parentEntity, false, null);
```

NitroxClient/GameLogic/Spawning/WorldEntities/PlaceholderGroupWorldEntitySpawner.cs
```diff
@@ -39,55 +73,94 @@ public IEnumerator SpawnAsync(WorldEntity entity, Optional<GameObject> parent, E
...
+        IWorldEntitySpawner spawner = spawnerResolver.ResolveEntitySpawner(worldEntity);
+
+        yield return spawner.SpawnAsync(worldEntity, parent, cellRoot, worldEntityResult);
+
+        if (worldEntityResult.value.HasValue)
+        {
+            EntityMetadataProcessor.ApplyMetadata(worldEntityResult.value.Value, worldEntity.Metadata);
+
+            // For some reasons it's not zero after spawning so we reset it here
+            worldEntityResult.value.Value.transform.localPosition = Vector3.zero;
```
Jannify, line 144: As clement mentioned in my previous pr he was asking himself if the rotation is also wrong/not zero. I couldn't check this so maybe you could give it a look.

Jannify (changes requested):

NitroxClient/GameLogic/Spawning/WorldEntities/PlaceholderGroupWorldEntitySpawner.cs
Sunrunner37: I don't personally see anything off with the rotations, should be good.

Sunrunner37:
Updated PR w/ requested changes.

Jannify (approved):
LGTM CW

PR #1990, opened 2023-02-18
Persisting and syncing quick bind slots

Sunrunner37:
This should be another new feature.  Nitrox existing code for recording quick slot bindings; however, these never got replayed during the initial sync.  Also, this existing code recorded UWE unique identifiers so needed some slight refactoring to use nitrox ids.

Example adding several duplicate items, setting them into quick slots 1/2 and 3/5, then reloading the server several times:

image

### DIFF
--- a/NitroxClient/GameLogic/InitialSync/QuickSlotInitialSyncProcessor.cs
+++ b/NitroxClient/GameLogic/InitialSync/QuickSlotInitialSyncProcessor.cs
@@ -1,21 +1,16 @@
 using System.Collections;
-using NitroxClient.Communication.Abstract;
+using System.Collections.Generic;
 using NitroxClient.GameLogic.InitialSync.Base;
 using NitroxClient.MonoBehaviours;
 using NitroxModel.DataStructures;
-using NitroxModel.DataStructures.Util;
 using NitroxModel.Packets;
 
 namespace NitroxClient.GameLogic.InitialSync;
 
 public class QuickSlotInitialSyncProcessor : InitialSyncProcessor
 {
-    private readonly IPacketSender packetSender;
-
-    public QuickSlotInitialSyncProcessor(IPacketSender packetSender)
+    public QuickSlotInitialSyncProcessor()
     {
-        this.packetSender = packetSender;
-
         DependentProcessors.Add(typeof(PlayerInitialSyncProcessor));  // the player needs to be configured before we can set quick slots.
         DependentProcessors.Add(typeof(EquippedItemInitialSyncProcessor)); // we need to have the items spawned into our inventory before we can quick slot them.
     }
@@ -24,16 +19,17 @@ public class QuickSlotInitialSyncProcessor : InitialSyncProcessor
     {
         int nonEmptySlots = 0;
 
+        Dictionary<NitroxId, InventoryItem> inventoryItemsById = GetItemsById();
+ 
         for (int i = 0; i < packet.QuickSlotsBindingIds.Length; i++)
         {
             waitScreenItem.SetProgress(i, packet.QuickSlotsBindingIds.Length);
 
             NitroxId id = packet.QuickSlotsBindingIds[i];
-            Optional<InventoryItem> item = getItem(id);
 
-            if (item.HasValue)
+            if (id != null && inventoryItemsById.TryGetValue(id, out InventoryItem inventoryItem) )
             {
-                Inventory.main.quickSlots.binding[i] = item.Value;
+                Inventory.main.quickSlots.binding[i] = inventoryItem;
                 Inventory.main.quickSlots.NotifyBind(i, state: true);
             }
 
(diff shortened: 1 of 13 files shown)


NitroxClient/GameLogic/InitialSync/QuickSlotInitialSyncProcessor.cs
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    private Optional<InventoryItem> getItem(NitroxId id)
+    {
+        if (id == null)
+        {
+            return Optional.Empty;
+        }
+
+        foreach (InventoryItem inventoryItem in Inventory.main.container)
```
Jannify, line 53: What about indexing them once into a dict instead of looping over them for each quick slot? I doubt it's impacting performance much but probably better nonetheless.

NitroxClient/GameLogic/LocalPlayer.cs
```diff
@@ -14,13 +13,15 @@
 using UnityEngine;
 using UnityEngine.Rendering;
 using Object = UnityEngine.Object;
+using NitroxClient.Helpers;
```
Jannify, line 16: Should be moved to line 5

Jannify (changes requested):

NitroxClient/GameLogic/InitialSync/QuickSlotInitialSyncProcessor.cs
Sunrunner37: Done

Jannify (approved):
LGTM CW

NitroxClient/GameLogic/InitialSync/QuickSlotInitialSyncProcessor.cs
```diff
@@ -0,0 +1,57 @@
...
+{
+    public QuickSlotInitialSyncProcessor()
+    {
+        DependentProcessors.Add(typeof(PlayerInitialSyncProcessor));  // the player needs to be configured before we can set quick slots.
+        DependentProcessors.Add(typeof(EquippedItemInitialSyncProcessor)); // we need to have the items spawned into our inventory before we can quick slot them.
+    }
+
+    public override IEnumerator Process(InitialPlayerSync packet, WaitScreen.ManualWaitItem waitScreenItem)
+    {
+        int nonEmptySlots = 0;
```
tornac1234, line 20: This isn't used

NitroxClient/GameLogic/InitialSync/QuickSlotInitialSyncProcessor.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        Dictionary<NitroxId, InventoryItem> inventoryItemsById = GetItemsById();
+ 
+        for (int i = 0; i < packet.QuickSlotsBindingIds.Length; i++)
+        {
+            waitScreenItem.SetProgress(i, packet.QuickSlotsBindingIds.Length);
+
+            NitroxId id = packet.QuickSlotsBindingIds[i];
+
+            if (id != null && inventoryItemsById.TryGetValue(id, out InventoryItem inventoryItem) )
```
tornac1234, line 30: If the id is null, we should probably still unbind the slot because by default, there'll be a flare in there

tornac1234 (changes requested):
Tried the following:
- Join game, and use "tools loot" to give some items to play with
- Move the tools in the EMPTY SLOTS (there'll still be the flare in the first slot)
- Empty the first slot containing the flare (but keep the flare)
- Leave another slot empty (where there was nothing at first)
- Reconnect
- The first slot contains a flare and all the other slots are filled

It seems that the slot unbinding doesn't work properly, and as mentionned in another comment, during initial sync, null entity id should be understood as an unbound slot, so the slot should be unbound manually.

NitroxClient/GameLogic/InitialSync/QuickSlotInitialSyncProcessor.cs
Sunrunner37: All slots are unbound by default when starting a new game.  The issue above is that we only patch QuickSlot.Bind and not QuickSlot.Unbind (thus the server thinks we still have everything full).  I can add this new patch.

NitroxClient/GameLogic/InitialSync/QuickSlotInitialSyncProcessor.cs
Sunrunner37: It is used below at the final log line.

NitroxClient/GameLogic/InitialSync/QuickSlotInitialSyncProcessor.cs
Sunrunner37: I take it back, unbinds were properly registered with the server.  it was just the initial equipment sync items.  added the unbind line here on null and tools seems to work.

NitroxClient/GameLogic/InitialSync/QuickSlotInitialSyncProcessor.cs
tornac1234: Well yeah but it is not modificated after being initialized

NitroxClient/GameLogic/InitialSync/QuickSlotInitialSyncProcessor.cs
```diff
@@ -0,0 +1,62 @@
...
+            else
+            {
+                // Unbind any default stuff from equipment addition.
+                Inventory.main.quickSlots.Unbind(i);
+            }
+
+            yield return null;
+        }
+
+        Log.Info($"Recieved initial sync with {nonEmptySlots} quick slots populated with items");
```
tornac1234, line 44:
```suggestion
        Log.Info($"Received initial sync with {nonEmptySlots} quick slots populated with items");
```

NitroxClient/GameLogic/InitialSync/QuickSlotInitialSyncProcessor.cs
```diff
@@ -0,0 +1,62 @@
...
+        for (int i = 0; i < packet.QuickSlotsBindingIds.Length; i++)
+        {
+            waitScreenItem.SetProgress(i, packet.QuickSlotsBindingIds.Length);
+
+            NitroxId id = packet.QuickSlotsBindingIds[i];
+
+            if (id != null && inventoryItemsById.TryGetValue(id, out InventoryItem inventoryItem) )
+            {
+                Inventory.main.quickSlots.binding[i] = inventoryItem;
+                Inventory.main.quickSlots.NotifyBind(i, state: true);
```
tornac1234, line 33:
```suggestion
                Inventory.main.quickSlots.NotifyBind(i, state: true);
                nonEmptySlots++;
```

tornac1234 (changes requested):
Apart from that, it'll just need a rebase and it'll be fine.

Sunrunner37:
Performed the two minor comments and did the rebase.  Additional rounds of testing look good.

PR #1994, opened 2023-02-24
Packet suppression to use static member on generic type (part 1/2)

Measurity:
This change lets the compiler deal with type lookups instead of a dictionary lookup. The later is slower and harder for compiler to optimize.

This PR is the first part. The second part (in a future PR) will add an analyzer to detect packet sends in injected code and warn if UWE code is called that will send packets without a proper `.Suppress()`.

This is an alternative to: https://github.com/SubnauticaNitrox/Nitrox/pull/1984 which uses the stacktrace and is slower but removes the need for littering the code with `packetSender.Suppress(packet)` calls.

### Benchmark
|       Method |      Mean |     Error |    StdDev | Ratio | RatioSD |
|------------- |----------:|----------:|----------:|------:|--------:|
|       Struct |  4.773 ns | 0.0913 ns | 0.0854 ns |  1.00 |    0.00 |
| StaticStruct |  1.754 ns | 0.0037 ns | 0.0035 ns |  0.37 |    0.01 |
|        Queue | 20.422 ns | 0.0121 ns | 0.0095 ns |  4.28 |    0.07 |
|         Func | 43.988 ns | 0.0997 ns | 0.0883 ns |  9.21 |    0.17 |
|      Hashmap | 46.375 ns | 0.0365 ns | 0.0323 ns |  9.71 |    0.17 |

[Benchmark code](https://gist.github.com/Measurity/5fdf79eb2153782326af2a3b194badec)


NitroxClient/Communication/PacketSuppression.cs
```diff
@@ -1,21 +1,22 @@
...
-        }
+/// <summary>
+///     Suppresses the given packet type from being sent. Disables the suppression when disposed.
+/// </summary>
+/// <typeparam name="T">The packet type to suppress.</typeparam>
+public readonly struct PacketSuppressor<T> : IDisposable
+    where T : Packet
+{
+    private static bool isSuppressed;
+    public static bool IsSuppressed => isSuppressed;
```
Sunrunner37, line 14: Do you think we would get better performance if we stuck with the previous method but used a static hashmap (that does NOT vary with T)?  We would have to hack it to allow multiple disposes (or the api may already support but not sure).  This would reduce object creation in favor of hashmap lookup.

Sunrunner37 (approved):

Sunrunner37 (approved):

PR #1997, opened 2023-03-09
Fixed placeholder nesting causing batch issues and broken precursor objects.

Sunrunner37:
Subnautica spawns placeholder objects as siblings of placeholder groups.  Example taken from debugger in single player:

image

The Nitrox behavior was spawning these as direct children.  This behavior was resulting in two major issues:

1) The game wipes children of PrefabPlaceholderGroups at serialization time causing us to maintain custom logic to respawn the children after batch deserilization.  

2) Some child components rely on the unity messaging framework to communicate.  For example the `PrecursorKeyTerminal` needs to send a `ToggleDoor` message through the parent to `PrecursorDoor`.  This is not possible if the parenting is incorrect.  

This PR fixes instances of precursor objects not working, for example:

image

It also removes the custom batch deserialization code as it is no longer required.

### DIFF
--- a/NitroxClient/GameLogic/Spawning/WorldEntities/PlaceholderGroupWorldEntitySpawner.cs
+++ b/NitroxClient/GameLogic/Spawning/WorldEntities/PlaceholderGroupWorldEntitySpawner.cs
@@ -5,22 +5,13 @@ using NitroxClient.MonoBehaviours;
 using NitroxModel.DataStructures.GameLogic;
 using NitroxModel.DataStructures.GameLogic.Entities;
 using NitroxModel.DataStructures.Util;
+using NitroxModel_Subnautica.DataStructures;
 using UnityEngine;
 
 namespace NitroxClient.GameLogic.Spawning.WorldEntities;
 
-/// <summary>
-/// Group placeholders are unique in that the game cleans up their children each time a player moves out of its batch. In single player, the placeholders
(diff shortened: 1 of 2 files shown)


Coding-Hen (approved):

NitroxClient/GameLogic/Spawning/WorldEntities/PlaceholderGroupWorldEntitySpawner.cs
```diff
@@ -140,8 +106,8 @@ private IEnumerator SpawnWorldEntityChild(WorldEntity worldEntity, EntityCell ce
         {
             EntityMetadataProcessor.ApplyMetadata(worldEntityResult.value.Value, worldEntity.Metadata);
 
-            // For some reasons it's not zero after spawning so we reset it here
-            worldEntityResult.value.Value.transform.localPosition = Vector3.zero;
+            worldEntityResult.value.Value.transform.localPosition = worldEntity.Transform.LocalPosition.ToUnity();
+            worldEntityResult.value.Value.transform.localRotation = worldEntity.Transform.Rotation.ToUnity();
```
Jannify, line 110:
```suggestion
            worldEntityResult.value.Value.transform.localRotation = worldEntity.Transform.LocalRotation.ToUnity();
```
Should be both local, right?

Sunrunner37:
Addressed PR feedback

Jannify (approved):
LGTM CW

PR #1999, opened 2023-03-11
InstalledBatteryEntity Implementation

Sunrunner37:
Background: Nitrox previously used a custom system called "storage slots" for managing the state of batteries in vehicles.  This was extremely confusing for maintenance development as it wasn't clear the difference between containers, storage slots, inventories, modules slots, and all the other various locations items could be tucked away.  Whenever a battery state changed, the entire battery game object was serialized and transmitted for persistence on the server.  A similar operation happens for items that contain batteries - although in this case the entire item gameobject was serialized including the battery.

This PR introduces an InstalledBatteryEntity that lives within the Vehicle and Item entity hierarchies.  This completes one of the major milestones in the entity refactor to allow players to maintain battery state across all of their tools, which was previously only half implemented due to technical limitations. 

UWE loads batteries into tools and vehicles in an async nature.  This PR uses a similar approach to the previous method and suppresses UWE default battery loading.  However, unlike the previous implementation, the PR will force battery load to be sync, allowing us to send a single entity payload to the server.  In previous versions of nitrox, this would send a packet for every battery to the server when the async loading was done (up to 6 for the cyclops). 

Finally, the PR includes all of the deletion of storage slots.  These were the last remaining piece of the InventoryManager server side; thus, that is also removed. 

Example entities in debug mode, showing guids for slots and batteries. 

image

image


NitroxClient/ClientAutoFacRegistrar.cs
```diff
@@ -14,6 +14,7 @@
 using NitroxClient.GameLogic.Bases.Spawning.BasePiece;
 using NitroxClient.GameLogic.ChatUI;
 using NitroxClient.GameLogic.FMOD;
+using NitroxClient.GameLogic.Helper;
```
Coding-Hen, line 17: If we are removing a line of code do we need this dependency adding?

NitroxClient/GameLogic/Spawning/InstalledBatteryEntitySpawner.cs
```diff
@@ -0,0 +1,55 @@
...
+public class InstalledBatteryEntitySpawner : EntitySpawner<InstalledBatteryEntity>
+{
+    public override IEnumerator SpawnAsync(InstalledBatteryEntity entity, TaskResult<Optional<GameObject>> result)
+    {
+        Optional<GameObject> parent = NitroxEntity.GetObjectFrom(entity.ParentId);
+
+        if (!parent.HasValue)
+        {
+            Log.Info($"Unable to find parent to install battery {entity}");
+            result.Set(Optional.Empty);
```
Coding-Hen, line 21: If we are setting the results here do we not want to return? If we continue we will send up always having the result set further down the method

NitroxClient/GameLogic/Spawning/InstalledBatteryEntitySpawner.cs
```diff
@@ -0,0 +1,55 @@
...
+        }
+
+        EnergyMixin energyMixin = parent.Value.GetComponent<EnergyMixin>();
+
+        if (!energyMixin)
+        {
+            Log.Info($"Unable to find EnergyMixin on parent to install battery {entity}");
+            result.Set(Optional.Empty);
```
Coding-Hen, line 29: And here. Though if this check doesn't return the lines below will error accessing energymixin

Sunrunner37: done

NitroxClient/GameLogic/Spawning/InstalledBatteryEntitySpawner.cs
Sunrunner37: good catch, forgot to return in case the object is not in a good state.

NitroxClient/ClientAutoFacRegistrar.cs
Sunrunner37: removed

NitroxClient/GameLogic/Helper/BatteryChildEntityHelper.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Collections.Generic;
+using UWE;
+using UnityEngine;
+using static WorldForces;
+
+namespace NitroxClient.GameLogic.Helper;
+
+/// <summary>
+/// Vehicles and items are created without a battery loaded into them.Subnautica usually spawns these in async; however, this
+/// is disabled in nitrox so we can properly tag the id.  Here we create the installed battery (with a new NitroxId) and have the
```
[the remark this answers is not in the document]
Sunrunner37: updated

Sunrunner37 (commented):
Addressed comments

NitroxPatcher/Patches/Dynamic/EnergMixin_SpawnDefaultAsync_Patch.cs
```diff
@@ -0,0 +1,45 @@
...
+    public static readonly MethodInfo TARGET_METHOD = AccessTools.EnumeratorMoveNext(TARGET_METHOD_ORIGINAL);
+
+    public static readonly OpCode INJECTION_OPCODE = OpCodes.Call;
+    public static readonly object INJECTION_OPERAND = Reflect.Method(() => CrafterLogic.NotifyCraftEnd(default(GameObject), default(TechType)));
+
+    public static IEnumerable<CodeInstruction> Transpiler(MethodBase original, IEnumerable<CodeInstruction> instructions)
+    {
+        Validate.NotNull(INJECTION_OPERAND);
+
+        // Blank out the method and replace with:
```
Jannify, line 27:
```suggestion
        // Blanks out the generated MoveNext() and replaces it with:
```
Minor: To better reflect that we are blanking/replacing the generated state machine.

Jannify (changes requested):

Sunrunner37:
Addressed PR comments

Jannify (approved):
LGTM CW, did not do any in-game testing

PR #2000, opened 2023-03-15
Adding fixes for multiple small bugs.

Sunrunner37:
1) Players not being able to enter a newly constructed cyclops due to missing cinematic controllers.
2) Player pings were not showing up due to inappropriate parenting of the remote player game object in addition to unnecessary offset on ping position.
3) The player entity was never having its position updated on the server side.  This makes some calculations incorrect (such as isSwimming).  Currently these positions are in two places, there is a todo item to consolidate.  
4) Fixed a long-standing record keeping bug with simulation lock type not stored on the client properly.  To my knowledge this has no effect as singular lock requests go through a different code path.  However, it was a land mine that should be fixed.  
5) Fixed multiple bugs causing player ping manager to not register changes in ping status or show the action popups.
6) Fixed freeze time so it correctly functions with the big update's enum value.

### DIFF
--- a/NitroxPatcher/Patches/Dynamic/FreezeTime_Begin_Patch.cs
+++ b/NitroxPatcher/Patches/Dynamic/FreezeTime_Begin_Patch.cs
@@ -1,4 +1,5 @@
-﻿using System.Reflection;
+using System.Collections.Generic;
+using System.Reflection;
 using HarmonyLib;
 using NitroxModel.Helper;
 using UWE;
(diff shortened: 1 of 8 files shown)


NitroxPatcher/Patches/Dynamic/FreezeTime_Begin_Patch.cs
```diff
@@ -1,28 +1,28 @@
...
-        // We don't want to prevent from freezing the game if the opened modal wants to freeze the game
-        public static bool Prefix(string userId)
-        {
-            // If we ask to freeze from a Nitrox modal, userId will be like this: NitroxServerStoppedModalFreeze
-            return userId.Contains("Nitrox") && userId.EndsWith("Freeze");
-        }
+    // We don't want to prevent from freezing the game if the opened modal wants to freeze the game
+    public static bool Prefix(FreezeTime.Id id)
+    {
+        return allowedFreezeIds.Contains(id);
```
dartasen, line 21: It's likely we will never have update on this enum, maybe we can just do some ==

Sunrunner37: I think this is mostly stylistic as there isn't any performance implications

Measurity: Ran a benchmark. First two rows return true, last return false. Performance difference is relatively huge but let's say anything 0.001ms (1000ns) or lower should be fine.

The hash set adds extra context with its variable name. It's very minor. But if I had to decide I agree with Clement here because the comments are already explaining the code well.

|      Method |         id |      Mean |     Error |    StdDev | Ratio | Allocated | Alloc Ratio |
|------------ |----------- |----------:|----------:|----------:|------:|----------:|------------:|
|     Hashset | WaitScreen | 7.4625 ns | 0.0170 ns | 0.0159 ns | 1.000 |         - |          NA |
| Conditional | WaitScreen | 0.0146 ns | 0.0036 ns | 0.0034 ns | 0.002 |         - |          NA |
|             |            |           |           |           |       |           |             |
|     Hashset |  Something | 5.3859 ns | 0.0153 ns | 0.0136 ns | 1.000 |         - |          NA |
| Conditional |  Something | 0.0372 ns | 0.0031 ns | 0.0026 ns | 0.007 |         - |          NA |

Sunrunner37: Sorry I left the comment ambiguous. I think it is obvious that a hashset has poorer performance characteristics over a hard coded conditional.  The context is that when someone is playing then FreezeTime.Begin is invoked when dialogs like save/quit are opened.  The performance implication is negligible.  You probably get more overhead from the harmony method proxy.

Jannify (approved):
LGTM CW

dartasen (approved):

Coding-Hen (approved):

PR #2001, opened 2023-03-15
Reducing jitter and stuttering as a cyclops passenger.

Sunrunner37:
This an experimental PR to see if we can reduce the poor cyclops experience.
 
I've included two potential fixes:

1) When riding as a passenger, the cyclops interpolation switches rapidly between none/Interpolate each frame.  The root cause is `Player.RequiresHighPrecisionPhysics()` going in and out of criteria every few frames.  The change here is to give a constant value of 'true' when another player is piloting the sub.

2) Even with the above fix the cyclops would stick on 'Interpolate'.  This results in poor visual performance over 'None' when surfacing or diving in the cyclops (and can also launch the player more often!). This was due to the `FreezeRigidbodyWhenFar` causing permanent interpolate when in range although `RequiresHighPrecisionPhysics` should have locked interpolate to 'None'.  This seems like a potential oversight from UWE.  The fix is to patch out `FreezeRigidbodyWhenFar` for the cyclops after construction time.

### DIFF
--- a/NitroxClient/MonoBehaviours/MultiplayerCyclops.cs
+++ b/NitroxClient/MonoBehaviours/MultiplayerCyclops.cs
@@ -8,7 +8,7 @@ namespace NitroxClient.MonoBehaviours
         private ISubThrottleHandler[] subThrottleHandlers;
         private float previousAbsYaw;
 
-        internal RemotePlayer CurrentPlayer { get; set; }
+        public RemotePlayer CurrentPlayer { get; set; }
 
         protected override void Awake()
         {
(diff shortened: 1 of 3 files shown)


tornac1234 (approved):
Tests have shown that there's no longer jittering (congrats that's insane !) but still it doesn't prevent the player from being launched at the other side of the planet (but that's yet another issue).

killzoms:
Speaking of the player being launched I have a potential fix in mind, I'll do some testing to see if I can pull it off...

PR #2002, opened 2023-03-16
Added entity deletion accounting at several locations.

Sunrunner37:
Nitrox currently has several codepaths that don't report a destroyed entity.  I went through and plugged some of the major gaps. 

This PR also includes a minor bug fix to remove broadcasting on picked up items and a circular dependency between server classes.


Coding-Hen (approved):
Lgcw

PR #2004, opened 2023-03-18
Quick fix for EntityCell Visibility

killzoms:


Sunrunner37 (approved):

Sunrunner37:
Tested this out locally, it works really well.  I think it was definitively a good idea to reduce the delay as it was very noticeable.  We may want to experiment with different delays but for now it is good.

PR #2006, opened 2023-03-18
Join queue despaghettification

TwinBuilderOne:
- The join queue logic is now centralized in a single async method and no longer relies on fallible callbacks
- It is now impossible for the queue to get stuck even if a player crashes their game while connecting
- Players enter the queue once the base game has finished loading to reduce unnecessary delays

Work still needs to be done around the loading screen because the UI freezes at several points during the connection process and some messages are no longer visible because `WaitScreen` does nothing in the latest version.


tornac1234 (changes requested):
Probably some things I've not seen yet and didn't test that many figure cases.
One issue I've found is PlayerDisconnected() doesn't trigger a queue continuation. If you join with 2 players, and when both are in the waiting queue and you suddenly Atl+F4 one of the clients, the server will be put in pause state without resuming for the other one (it waited for 5 minutes tho but that's a "worst case" situation while we would have just been able to resume the queue once we know that the loading player was disconnected).

Also there's no way to tell that you're waiting once you're in the loading. You should get informations with like the number of player in front of you in the queue and maybe a maximum wait time calculated by ``number of players in queue * InitialSyncTimeout``, we should definitely add some more visual elements for the player instead of locking him in there without any information. I think we discussed this once, I proposed that we would have some sort of UI on top of the screen showing the informations we want (those proposed above and the timing passed waiting also) and a way to leave the queue and head back to the main menu.

I would not want this merged without at least some information given to the player, and then we'd set the UI I proposed for a later PR.

TwinBuilderOne:
I agree, this PR needs a dedicated UI screen to be production-ready. I think we should go ahead and add it now because it's bad practice to merge a PR that says "Finish me later with another PR."

tornac1234:
Would like to hear about where this is going. Should we try to restrict the scope of this PR (e.g. postpone early loading which needs more work as said above) to at least have the fixes merged ?

TwinBuilderOne:
I think so, adding new menus sounds more like a beta feature. We do need to add at least some kind of additional UI to the loading screen, however.

tornac1234:
How's this going ?

TwinBuilderOne:
I took a very long break from this project because I was busy and my interests had changed, so I haven't been keeping up with things very closely. Actually, I was thinking of getting back into Nitrox, and this is a good time to do it. This PR is functional (last time I checked) but to be "polished" it still needs some sort of UI on the loading screen to show progress, or just to modify the vanilla progress bar. That would be better suited for a separate PR. Other than that, it should work (I just need to fix the conflicts).

TwinBuilderOne:
It still works! (The PR is bugged right now and isn't showing the latest commit. If you check the branch on my fork, everything is there.)

Jannify:
Would this make sense to be merge before or after #2023 ?

TwinBuilderOne:
I don't think it matters. They shouldn't break each other, right?

NitroxClient/Communication/Packets/Processors/PlayerSyncTimeoutProcessor.cs
```diff
@@ -0,0 +1,22 @@
...
+public class PlayerSyncTimeoutProcessor : ClientPacketProcessor<PlayerSyncTimeout>
+{
+    public override void Process(PlayerSyncTimeout packet)
+    {
+        Multiplayer.Main.StartCoroutine(TimeoutRoutine());
+    }
+
+    private IEnumerator TimeoutRoutine()
+    {
+        Log.InGame("Error: Initial sync timeout, closing game");
```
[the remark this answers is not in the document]
TwinBuilderOne: I tried this but the modal wasn't appearing most of the time for some reason. When it did appear there was only a box that said "Please wait...". I think the game has problems showing modals while the loading screen is still showing.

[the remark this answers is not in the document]
TwinBuilderOne: I had the logger give a small countdown before closing the game so the user can at least see what happened. It's the best we can do without a custom loading screen.

TwinBuilderOne:
The PR is ready for review now. Just one thing - currently, if the client is in the queue, the loading screen disappears and they can play as normal. However, when they start syncing it reappears and anything they did is undone. This is unintentional, but I left it in for now since it's nice to be able to jump/swim around while waiting instead of just looking at a static image on a loading screen. Should we keep this in? Maybe add something to the queue enter message to let the player know what is happening?

NitroxServer/Communication/Packets/PacketHandler.cs
```diff
@@ -12,26 +13,28 @@ namespace NitroxServer.Communication.Packets
...
         {
             Player player = playerManager.GetPlayer(connection);
-            if (player == null)
+            if (player != null)
             {
-                ProcessUnauthenticated(packet, connection);
+                ProcessAuthenticated(packet, player);
             }
-            else
+            else if (!joiningManager.GetQueuedPlayers().Contains(connection))
```
tornac1234, line 35: Isn't it too harsh on the connection ? o(n) on packet receiving (a lot per second) might be a load too bad for the server

tornac1234 (dismissed):
Was my previous review comment treated ?

SpaceMonkeyy86:
The join queue takes place entirely on the loading screen now, so we don't have a way to show UIs to the player besides the in-game log. We can add a UI that also shows more information on loading progress like how pre-2.0 Subnautica did it, but that would be best in another PR.

NitroxModel/Packets/JoinQueueInfo.cs
```diff
@@ -0,0 +1,18 @@
+using System;
+
+namespace NitroxModel.Packets;
+
+[Serializable]
+public class JoinQueueInfo : Packet
+{
+    public int Position { get; }
+    public int Timeout { get; }
+    public bool ShowMaximumWait { get; }
```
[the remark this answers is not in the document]
SpaceMonkeyy86: We show the maximum wait time per person when the player enters the queue. This information never changes so there is no need to show it again.

NitroxServer/Communication/Packets/PacketHandler.cs
SpaceMonkeyy86: I highly doubt this could cause performance problems since it only runs for unauthenticated packets (sent very infrequently compared to normal packets) and there will realistically never be more than a few people in the queue at once. It's best to use a profiler to confirm that a piece of code is actually causing a bottleneck before trying to optimize it.

NitroxClient/MonoBehaviours/Multiplayer.cs
```diff
@@ -113,16 +113,39 @@ public static IEnumerator LoadAsync()
...
+        public void TimeOut()
+        {
+            multiplayerSession.Disconnect();
+            StartCoroutine(TimeOutRoutine());
+
+            IEnumerator TimeOutRoutine()
+            {
+                int timer = 5;
+
+                while (timer > 0)
```
Jannify, line 138: Is there a reason you are not using a for loop here?

NitroxServer/GameLogic/JoiningManager.cs
```diff
@@ -0,0 +1,245 @@
...
+            serverConfig.KeepInventoryOnDeath
+        );
+
+        player.SendPacket(initialPlayerSync);
+    }
+
+    public void JoiningPlayerDisconnected(INitroxConnection connection)
+    {
+        // They may have been queued, so just erase their entry
+        JoinQueue = new ThreadSafeQueue<(INitroxConnection, string)>(JoinQueue.Where(tuple => !Equals(tuple.Item1, connection)));
```
Jannify, line 237: This is not really tread safe. Question is if it can be problematic...

NitroxClient/Communication/Packets/Processors/JoinQueueInfoProcessor.cs
```diff
@@ -0,0 +1,24 @@
...
+        if (packet.ShowMaximumWait)
+        {
+            Log.InGame($"The maximum wait time per person is {MillisToMinutes(packet.Timeout)} minutes.");
+        }
+    }
+
+    private static string MillisToMinutes(int milliseconds)
+    {
+        double minutes = milliseconds / 60000.0;
+        return Math.Round(minutes, 1).ToString();
```
Coding-Hen, line 22: I find that this could be misleading, why would we not just have it in seconds or a combination of both. If the milliseconds came out to 20 seconds we would round it to 0.3 which would require the user to calculate what this actually means

NitroxServer/GameLogic/PlayerManager.cs
```diff
@@ -1,43 +1,52 @@
...
-        public PlayerManager(List<Player> players, ServerConfig serverConfig)
+        public PlayerManager(List<Player> players, World world, ServerConfig serverConfig)
         {
             allPlayersByName = new ThreadSafeDictionary<string, Player>(players.ToDictionary(x => x.Name), false);
             currentPlayerId = players.Count == 0 ? (ushort)0 : players.Max(x => x.Id);
 
+            this.world = world;
             this.serverConfig = serverConfig;
+
+            _ = JoinQueueLoop();
```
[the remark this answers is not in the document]
Measurity: Should use `.ContinueWithHandleError()` on it in case of exceptions.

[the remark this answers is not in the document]
SpaceMonkeyy86: Using an underscore is an explicit discard of the return value and avoids a compiler warning. The latest version of the code uses Task.Run so this doesn't matter anymore.

Jannify (changes requested):

NitroxClient/Communication/Packets/Processors/JoinQueueInfoProcessor.cs
```diff
@@ -0,0 +1,24 @@
+using NitroxClient.Communication.Packets.Processors.Abstract;
+using NitroxModel.Packets;
+using System;
```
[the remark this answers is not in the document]
SpaceMonkeyy86: Yeah I finally found that but this code was probably written about two years ago so I didn't realize at the time. It's fixed now.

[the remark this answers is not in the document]
SpaceMonkeyy86: Also how is this comment from February 14

NitroxServer/GameLogic/PlayerManager.cs
```diff
@@ -259,31 +402,10 @@ public void PlayerDisconnected(INitroxConnection connection)
             }
         }
 
-        public void FinishProcessingReservation(Player player = null)
+        public void NonPlayerDisconnected(INitroxConnection connection)
```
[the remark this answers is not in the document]
SpaceMonkeyy86: Renamed it to JoiningPlayerDisconnected.

SpaceMonkeyy86:
Ready for review again. This PR is a blocker for a few things such as loading screen info (and it's also just a lot faster/smoother than the current implementation), so I would like to see this in the upcoming release if possible. We should also do in-game testing, where we build a lot of bases and vehicles to make the sync take longer and try various combinations of joining at the same time and crashing the game to make sure the server holds up.

NitroxServer/GameLogic/JoiningManager.cs
```diff
@@ -0,0 +1,234 @@
...
+            catch (Exception e)
+            {
+                Log.Error($"Unexpected error during player connection: {e}");
+            }
+        }
+    }
+
+    public void AddToJoinQueue(INitroxConnection connection, string reservationKey)
+    {
+        // Necessary to avoid race conditions between this method and the queue count check
```
Jannify, line 121: Would copy this docu also to the Lock init line L27

NitroxServer/GameLogic/JoiningManager.cs
```diff
@@ -0,0 +1,234 @@
...
+                await Task.Run(() =>
+                {
+                    while (true)
+                    {
+                        if (syncFinished ||
+                            connection.State == NitroxConnectionState.Disconnected ||
+                            source.IsCancellationRequested)
+                        {
+                            return;
+                        }
```
Jannify, line 85: Would revert the if statement to use it directly in the while loop statement

NitroxServer/GameLogic/JoiningManager.cs
```diff
@@ -0,0 +1,234 @@
...
+                }
+                else
+                {
+                    Log.Info($"Player {name} joined successfully. Remaining requests: {joinQueue.Count}");
+                    BroadcastPlayerJoined(playerManager.GetPlayer(connection));
+                }
+            }
+            catch (Exception e)
+            {
+                Log.Error($"Unexpected error during player connection: {e}");
```
Jannify, line 114:
```suggestion
                Log.Error($"Unexpected error during player connection inside the join queue: {e}");
```

NitroxServer/GameLogic/JoiningManager.cs
```diff
@@ -0,0 +1,234 @@
...
+        return joinQueue.Select(tuple => tuple.Item1);
+    }
+
+
+    private void SendInitialSync(INitroxConnection connection, string reservationKey)
+    {
+        IEnumerable<PlayerContext> GetOtherPlayers(Player player)
+        {
+            return playerManager.GetConnectedPlayers().Where(p => p != player)
+                                        .Select(p => p.PlayerContext);
```
Jannify, line 149: Minor:
```suggestion
            return playerManager.GetConnectedPlayers().Where(p => p != player).Select(p => p.PlayerContext);
```

NitroxServer/GameLogic/JoiningManager.cs
```diff
@@ -0,0 +1,234 @@
...
+        }
+
+        PlayerWorldEntity SetupPlayerEntity(Player player)
```
Jannify, line 152:
```suggestion
        PlayerWorldEntity SetupNewPlayerEntity(Player player)
```

Jannify (changes requested):

SpaceMonkeyy86:
Ok, changes made 👍

NitroxPatcher/Patches/Dynamic/ErrorMessage_OnLateUpdate_Patch.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Reflection.Emit;
+using HarmonyLib;
+using NitroxModel.Helper;
+using NitroxPatcher.Patches;
+using UnityEngine;
+
+/// <summary>
+/// Ensures the in-game log is animated smoothly regardless of the time scale.
+/// </summary>
+public sealed partial class ErrorMessage_OnLateUpdate_Patch : NitroxPatch, IDynamicPatch
```
Jannify, line 12: Looking at `GameAnalytics_Patch` this should probably not be marked partial as our code generation doesn't need to generate the Patch method.

Jannify (approved):
LGTM CW

NitroxClient/MonoBehaviours/PlayerMovementBroadcaster.cs
```diff
@@ -17,6 +17,13 @@ public void Awake()
 
     public void Update()
     {
+        // HIGHLY TEMPORARY. There should be a way to block all packets from being sent when in the join queue
```
Jannify, line 20: Would add a `TODO:` so it is not lost in the code

OhmV-IR:
IG test went well

SpaceMonkeyy86:
In-game testing went perfectly, PR is ready to merge 👍

NitroxServer/GameLogic/JoiningManager.cs
```diff
@@ -0,0 +1,236 @@
...
+
+                await Task.Run(() =>
+                {
+                    while (!syncFinished &&
+                           connection.State != NitroxConnectionState.Disconnected &&
+                           !source.IsCancellationRequested)
+                    {
+                        Task.Delay(10).Wait();
+                    }
+                });
```
Measurity, line 86: Since this code isn't CPU-bound (most time spend on `Task.Delay`), it shouldn't use `Task.Run`.

```suggestion
                while (!syncFinished &&
                       connection.State != NitroxConnectionState.Disconnected &&
                       !source.IsCancellationRequested)
                {
                    await Task.Delay(10);
                }
```

NitroxServer/GameLogic/JoiningManager.cs
```diff
@@ -0,0 +1,236 @@
...
+                    return;
+                }
+            }
+
+            try
+            {
+                (INitroxConnection connection, string reservationKey) = joinQueue.Dequeue();
+                string name = playerManager.GetPlayerContext(reservationKey).PlayerName;
+
+                // Do this after dequeueing because everyone's position shifts forward
```
Measurity, line 62:
```suggestion
                // Do this after dequeuing because everyone's position shifts forward
```

NitroxServer/GameLogic/JoiningManager.cs
```diff
@@ -0,0 +1,236 @@
...
+                for (int i = 0; i < array.Length; i++)
+                {
+                    (INitroxConnection c, _) = array[i];
+                    c.SendPacket(new JoinQueueInfo(i + 1, serverConfig.InitialSyncTimeout));
+                }
+
+                Log.Info($"Starting sync for player {name}");
+                SendInitialSync(connection, reservationKey);
+
+                CancellationTokenSource source = new(serverConfig.InitialSyncTimeout);
```
Measurity, line 73: Should dispose of CTS.

```suggestion
                using CancellationTokenSource source = new(serverConfig.InitialSyncTimeout);
```

NitroxModel/DataStructures/ThreadSafeQueue.cs
```diff
@@ -121,6 +121,22 @@ public IEnumerable<T> Clone()
             }
         }
 
+        public void Filter(Func<T, bool> predicate)
```
Measurity, line 124: Would rename to `RemoveWhere`. `Filter` is usually used when a method makes no side effect.

Measurity (approved):

Measurity:
A minor issue: using the instant launch option, the player leaving the join queue is seen in swimming animation on other end.

SpaceMonkeyy86:
> Was my previous review comment treated ?

Yes, Alt+F4 immediately moves to the next player now, and the server sends in-game log messages to everyone in the queue with their current position and maximum wait time, with updates anytime that information changes.

Nitrox.Server.Subnautica/Models/GameLogic/PlayerManager.cs
```diff
@@ -44,19 +41,30 @@ public PlayerManager(List<Player> players, SubnauticaServerConfig serverConfig)
         this.serverConfig = serverConfig;
     }
 
-    public List<Player> GetConnectedPlayers()
+    public IEnumerable<Player> GetAllPlayers() => allPlayersByName.Values;
```
Jannify, line 45: I would like a xml comment here to clarify this is not connected players but all since forever. Better would be another name for this function to reflect that.

Nitrox.Server.Subnautica/Models/GameLogic/JoiningManager.cs
```diff
@@ -0,0 +1,231 @@
...
+{
+    private readonly PlayerManager playerManager;
+    private readonly SubnauticaServerConfig serverConfig;
+    private readonly World world;
+    private readonly SessionSettings sessionSettings;
+
+    private readonly ThreadSafeQueue<(INitroxConnection, string)> joinQueue = new();
+    private readonly Lock queueLocker = new(); // Necessary to avoid race conditions between JoinQueueLoop and AddToJoinQueue
+    private bool queueActive;
+    public Action SyncFinishedCallback { get; private set; }
```
Jannify:
```suggestion
    public Action? SyncFinishedCallback { get; private set; }
```

Nitrox.Server.Subnautica/Models/GameLogic/JoiningManager.cs
```diff
@@ -0,0 +1,231 @@
...
+                    Log.Info($"Initial sync timed out for player {name}");
+                    SyncFinishedCallback = null;
+
+                    if (connection.State == NitroxConnectionState.Connected)
+                    {
+                        connection.SendPacket(new PlayerKicked("Initial sync took too long and timed out"));
+                    }
+                    playerManager.PlayerDisconnected(connection);
+                }
+                else
```
Jannify, line 94: Instead of else should probably test nullability on `playerManager.GetPlayer(connection)`

Nitrox.Server.Subnautica/Models/GameLogic/JoiningManager.cs
```diff
@@ -0,0 +1,231 @@
...
+            }
+        }
+    }
+
+    public void AddToJoinQueue(INitroxConnection connection, string reservationKey)
+    {
+        // Necessary to avoid race conditions between JoinQueueLoop and AddToJoinQueue
+        lock (queueLocker)
+        {
+            Log.Info($"Added player {playerManager.GetPlayerContext(reservationKey).PlayerName} to queue");
```
Jannify:
```suggestion
            Log.Info($"Added player {playerManager.GetPlayerContext(reservationKey)?.PlayerName} to queue");
```

Nitrox.Server.Subnautica/Models/GameLogic/JoiningManager.cs
SpaceMonkeyy86: The player becomes non-null upon sending initial sync, that wouldn't work since we need to check whether they sent back a PlayerSyncFinished packet.

Jannify (approved):

PR #2009, opened 2023-03-19
add extra checks to RemotePlayerCinematicController

killzoms:


Coding-Hen (approved):

tornac1234 (approved):
Tests showed we can now see swimming animations for both players

PR #2013, opened 2023-03-31
Simplifying pickup system and removing some object duplication

Sunrunner37:
Picking up an item in the world (or through a fabricator) was causing duplicate `EntitySpawnedByClient` packets.  Ultimately this was causing object duplicate in some containers and the remote player.  The logic for pickup was also quite complex because it would try to convert the world entity to an inventory entity server side.  I moved this conversion client side which also allows us to avoid sending the additional `EntitySpawnedByClient` packet if it is unknown.

Demonstrating remote player picking up a bunch of items without duplication:
image

Additionally, it looks like we were not deleting resources server-side when crafting because reloading was causing resources to come back.  I'm not sure if this is happening in 1.7.0.1 or if it is due to the entity migration.  Either way, the issue is now resolved.  Example after cooking two peepers and logging back in:

image


Measurity (approved):
Tested and works on my end.

PR #2014, opened 2023-04-03
Optimized position broadcasting and more network debugger info.

Sunrunner37:
Currently, Nitrox broadcasts every simulated entity's position every 0.25 seconds.  This gives the appearance of fish synchronization but causes significant load - especially for non-LAN players.

This PR changes the mechanics behind movement simulation for most entities.  Things that have a SwimBehavior or WalkBehavior (that internally use a SplineFollow) will broadcast changes only when their destination changes.  This provides more fluid behavior and significantly reduces packet volume.

One thing to note is that some entities will still need the previous behavior.  For example, dropping an item that floats down to the bottom.  In this case, we can look towards future optimizations to dedupe packets when little movement actually occurred or perform a similar custom spline behavior. 

Benchmark results over 12 min:

Old System: 6669 packets (48.6MB)
New System: 6593 packets (8.8MB)

Screenshot (with the built in SimulateLatency enabled in LiteNetLib):

image

Also included in this PR is an update to add byte aggregation to the network debugger.  See below:

image

### DIFF
--- a/NitroxClient/Communication/PacketReceiver.cs
+++ b/NitroxClient/Communication/PacketReceiver.cs
@@ -16,11 +16,11 @@ namespace NitroxClient.Communication
             this.networkDebugger = networkDebugger;
         }
 
-        public void PacketReceived(Packet packet)
+        public void PacketReceived(Packet packet, int size)
         {
             lock (receivedPackets)
             {
-                networkDebugger?.PacketReceived(packet);
+                networkDebugger?.PacketReceived(packet, size);
                 receivedPackets.Enqueue(packet);
             }
         }
--- a/NitroxClient/Communication/PacketReceiver.cs
+++ b/NitroxClient/Communication/PacketReceiver.cs
@@ -16,11 +16,11 @@ namespace NitroxClient.Communication
             this.networkDebugger = networkDebugger;
         }
 
-        public void PacketReceived(Packet packet, int size)
+        public void PacketReceived(Packet packet, int byteSize)
         {
             lock (receivedPackets)
             {
-                networkDebugger?.PacketReceived(packet, size);
+                networkDebugger?.PacketReceived(packet, byteSize);
                 receivedPackets.Enqueue(packet);
             }
         }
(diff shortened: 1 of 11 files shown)


NitroxClient/Debuggers/NetworkDebugger.cs
```diff
@@ -208,6 +219,17 @@ private void AddPacket(Packet packet, bool isSent)
...
+        {
+            string[] suf = { "B", "KB", "MB", "GB", "TB" };
+            if (byteCount == 0)
+                return "0" + suf[0];
+            long bytes = Math.Abs(byteCount);
+            int place = Convert.ToInt32(Math.Floor(Math.Log(bytes, 1024)));
+            double num = Math.Round(bytes / Math.Pow(1024, place), 1);
+            return (Math.Sign(byteCount) * num).ToString() + suf[place];
+        }
+
```
Measurity, line 232: Recommend putting this in `NitroxModel.Extensions.cs` as an extension method for `uint`.
```suggestion
        public static string AsByteUnitText(this uint byteSize)
        {
            // Uint can't go past 4GB, so we don't need to worry about overflow.
            string[] suf = { "B", "KB", "MB", "GB" };
            if (byteSize == 0)
            {
                return $"0{suf[0]}";
            }
            int place = Convert.ToInt32(Math.Floor(Math.Log(byteSize, 1024)));
            double num = Math.Round(byteSize / Math.Pow(1024, place), 1);
            return num + suf[place];
        }
```

NitroxClient/Communication/PacketReceiver.cs
```diff
@@ -16,11 +16,11 @@ public PacketReceiver(INetworkDebugger networkDebugger = null)
             this.networkDebugger = networkDebugger;
         }
 
-        public void PacketReceived(Packet packet)
+        public void PacketReceived(Packet packet, int size)
         {
             lock (receivedPackets)
             {
-                networkDebugger?.PacketReceived(packet);
+                networkDebugger?.PacketReceived(packet, size);
```
Measurity, line 23: Recommend renaming argument to `byteSize` because `size` is too ambiguous in meaning.

Sunrunner37:
Applied the requested changes.

Measurity (approved):
Sometimes fish jump around even on LAN. But for the most part it's stable.

PR #2015, opened 2023-04-03
Added InternalsVisibleTo for all Nitrox projects in Directory.Build.props

Measurity:
Removed manual definitions of InternalsVisibleTo in each project.

We've been using the same namespace in Nitrox.Test as the code it tests but with [this attribute](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.internalsvisibletoattribute?view=net-8.0) that's not needed.


tornac1234 (approved):
Nothing strikes me in here, also I could build the solution successfully also so ig it's fine

dartasen (approved):

PR #2016, opened 2023-04-03
SpawnOnKill and BreakableResource replication added (e.g. Coral Disks or Limestones)

VELD-Dev:
SpawnOnKill behaviour is now replicated. It makes objects like Coral Disks spawn the object they are assigned to when they are killed.

Ultra super special thanks to @Measurity, @tornac1234 for helping me with programming this first contributions to Nitrox, and @Sunrunner37 for helping to tweak the end.
Thanks @dartasen, @Measurity, @Sunrunner37, @bud3699 and @tornac1234 for testing my edits when I was not able to

I hope I will be able to continue working with the Nitrox team !


NitroxPatcher/Patches/Dynamic/SpawnOnKill_OnKill_Patch.cs
```diff
@@ -0,0 +1,66 @@
...
+using NitroxModel.Packets;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Text;
+using System.Threading.Tasks;
+using UnityEngine;
+
+namespace NitroxPatcher.Patches.Dynamic
```
tornac1234, line 17: Inline the namespace declaration

NitroxPatcher/Patches/Dynamic/SpawnOnKill_OnKill_Patch.cs
```diff
@@ -0,0 +1,66 @@
...
+{
+    /// <summary>
+    /// Synchronizes entities that Spawn something when they are killed, e.g. Coral Disks.
+    /// </summary>
+    internal class SpawnOnKill_OnKill_Patch : NitroxPatch, IDynamicPatch
```
tornac1234, line 22: Make the class public

NitroxPatcher/Patches/Dynamic/SpawnOnKill_OnKill_Patch.cs
```diff
@@ -0,0 +1,66 @@
...
+    {
+        private static readonly MethodInfo TARGET_METHOD = Reflect.Method((SpawnOnKill t) => t.OnKill());
+
+        public static bool Prefix(SpawnOnKill __instance)
+        {
+            // I did this for safety reason, but I don't think it's necessary
+            if (__instance != null)
```
tornac1234, line 29: if the code is running, the __instance is not likely to be null, you probably won't need that

NitroxPatcher/Patches/Dynamic/SpawnOnKill_OnKill_Patch.cs
```diff
@@ -0,0 +1,66 @@
...
+                {
+                    Rigidbody component = gameObject.GetComponent<Rigidbody>();
+                    if (component)
+                    {
+                        component.AddForce(UnityEngine.Random.onUnitSphere * 1.4f, ForceMode.Impulse);
+                    }
+                }
+                
+                // Get the ID of the new game object
+                NitroxId newEntityId = NitroxEntity.GetId(gameObject);
```
tornac1234, line 43: Here you want to get the id from the __instance's gameobject only so you don't need this step (it will be counterproductive and do something you don't expect)

NitroxPatcher/Patches/Dynamic/SpawnOnKill_OnKill_Patch.cs
```diff
@@ -0,0 +1,66 @@
...
+
+                // get the ID of the destroyed object
+                NitroxId destroyedEntityId = NitroxEntity.GetId(__instance.gameObject);
+                // Send packet telling that it was destroyed to the server
+                Resolve<IPacketSender>().Send(new EntityDestroyed(destroyedEntityId));
```
tornac1234, line 48: You don't need to destroy the entity because the Items.Dropped action will update the entity that currently has the provided id with the newly furnished data.

NitroxPatcher/Patches/Dynamic/SpawnOnKill_OnKill_Patch.cs
```diff
@@ -0,0 +1,66 @@
...
+
+                // Give a new ID to the new game object
+                NitroxEntity.SetNewId(gameObject, newEntityId);
```
tornac1234, line 51:
```suggestion
                NitroxEntity.SetNewId(gameObject, destroyedEntityId);
```
Therefore here, you need to use the id found on the object that is getting destroyed

tornac1234 (changes requested):
Make the changes on a single commit then send it back once you're done with all the requested changes :)

NitroxPatcher/Patches/Dynamic/SpawnOnKill_OnKill_Patch.cs
VELD-Dev: Oooooh that's really interesting

Measurity (approved):
Tested and works for me. We should squash this PR as one commit.

Sunrunner37:
> 

Should we also make this configurable in the server cfg?  This behavior will pretty much kill the smg server if it is merged without a config.

VELD-Dev:
> > Tested and works for me. But as @tornac1234 notified me, would be better to add BreakableResource to this PR as well before squashing this PR.
> 
> Should we also make this configurable in the server cfg? This behavior will pretty much kill the smg server if it is merged without a config.

I don't know if this is a good idea, but maybe set the savability of this in the configs, for example it will reset the deletes resources after a restart, but it will keep syncing the items. However I don't know how to do that so I will not try that out, it looks a little tricky to put in place tho...
By the way, I'm renaming the PR to "SpawnOnKill and BreakableResource replication added (e.g. Coral Disks or Limestones)"
I also need someone to try the latest update, I cannot right now I'm not at home but I made the changes.

NitroxPatcher/Patches/Dynamic/SpawnOnKill_OnKill_Patch.cs
```diff
@@ -0,0 +1,59 @@
+using HarmonyLib;
+using NitroxClient.Communication.Abstract;
+using NitroxClient.GameLogic;
+using NitroxClient.MonoBehaviours;
+using NitroxModel.DataStructures;
+using NitroxModel.Helper;
+using NitroxModel.Packets;
+using NitroxPatcher.PatternMatching;
+using Oculus.Platform;
```
tornac1234, line 9:
```suggestion
```
unused

NitroxPatcher/Patches/Dynamic/SpawnOnKill_OnKill_Patch.cs
```diff
@@ -0,0 +1,59 @@
...
+    private static readonly InstructionsPattern spawnInstanceOnKillPattern = new()
+    {
+        Reflect.Method(() => Object.Instantiate(default(GameObject), default(Vector3), default(Quaternion))),
+        { Stloc_0, "DropOnKillInstance" },
+        Ldarg_0,
+    };
+
+    public static IEnumerable<CodeInstruction> Transpiler(MethodBase original, IEnumerable<CodeInstruction> instructions)
+    {
+        static IEnumerable<CodeInstruction> InsertCallbackCall(string label, CodeInstruction _)
```
tornac1234, line 33: I'm not sure why you created a function inside while you could just apply .Transform just like in [DevConsole_Update_Patch](https://github.com/SubnauticaNitrox/Nitrox/blob/master/NitroxPatcher/Patches/Dynamic/DevConsole_Update_Patch.cs)

NitroxClient/Communication/Packets/Processors/EntityDestroyedProcessor.cs
```diff
@@ -24,27 +24,25 @@ public EntityDestroyedProcessor(Entities entities, IPacketSender packetSender)
     public override void Process(EntityDestroyed packet)
```
tornac1234, line 24: Just a remark for this file as it was modified for this PR, you can remove any reference to ``packetSender`` (don't forget the "using" line also)

NitroxClient/Communication/Packets/Processors/EntityDestroyedProcessor.cs
```diff
@@ -24,27 +24,25 @@ public EntityDestroyedProcessor(Entities entities, IPacketSender packetSender)
...
     {
         entities.RemoveEntity(packet.Id);
+        if (!NitroxEntity.TryGetObjectFrom(packet.Id, out GameObject gameObject))
+        {
```
tornac1234, line 28: Can probably add some logging in case it happens

NitroxPatcher/Patches/Dynamic/BreakableResource_BreakIntoResources_Patch.cs
```diff
@@ -0,0 +1,33 @@
...
+using NitroxModel.DataStructures;
+using NitroxModel.Helper;
+using NitroxModel.Packets;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net.NetworkInformation;
+using System.Reflection;
+using System.Text;
+using System.Threading.Tasks;
```
tornac1234, line 13:
```suggestion
using System.Reflection;
```
all unused

NitroxPatcher/Patches/Dynamic/BreakableResource_BreakIntoResources_Patch.cs
```diff
@@ -0,0 +1,33 @@
...
+namespace NitroxPatcher.Patches.Dynamic;
+
+public class BreakableResource_BreakIntoResources_Patch : NitroxPatch, IDynamicPatch
+{
+    private static MethodInfo TARGET_METHOD = Reflect.Method((BreakableResource t) => t.BreakIntoResources());
+
+    public static void Prefix(BreakableResource __instance)
+    {
+        // Gets the entity ID
+        NitroxId id = NitroxEntity.GetId(__instance.gameObject);
```
tornac1234, line 24: Should be changed to an if clause with NitroxEntity.TryGetEntityFrom, and only send the packet if we found an entity on the object because there would be no point in creating a new id then asking the server to delete it as if it always existed.

NitroxPatcher/Patches/Dynamic/BreakableResource_SpawnResourceFromPrefab_Patch.cs
```diff
@@ -0,0 +1,59 @@
...
+using NitroxClient.MonoBehaviours;
+using NitroxModel.DataStructures;
+using NitroxModel.Helper;
+using NitroxPatcher.PatternMatching;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Text;
+using System.Threading.Tasks;
```
tornac1234, line 13:
```suggestion
using NitroxClient.GameLogic;
using NitroxClient.MonoBehaviours;
using NitroxModel.DataStructures;
using NitroxModel.Helper;
using NitroxPatcher.PatternMatching;
using System.Collections.Generic;
using System.Reflection;
```
all unused

NitroxPatcher/Patches/Dynamic/BreakableResource_SpawnResourceFromPrefab_Patch.cs
```diff
@@ -0,0 +1,59 @@
...
+using UnityEngine;
+using static System.Reflection.Emit.OpCodes;
+
+namespace NitroxPatcher.Patches.Dynamic;
+
+/// <summary>
+/// Synchronizes entities that can be broken and that will drop material, such as limestones...
+/// </summary>
+public class BreakableResource_SpawnResourceFromPrefab_Patch : NitroxPatch, IDynamicPatch
```
tornac1234, line 22: Just as for SpawnOnKill_OnKill_Patch, you need to create a test class. It will also help you debug the transpiler without launching your game.

NitroxPatcher/Patches/Dynamic/BreakableResource_SpawnResourceFromPrefab_Patch.cs
```diff
@@ -0,0 +1,59 @@
...
+{
+    public static readonly MethodInfo TARGET_METHOD_ORIGINAL = Reflect.Method(() => BreakableResource.SpawnResourceFromPrefab(default, default, default));
+    public static readonly MethodInfo TARGET_METHOD = AccessTools.EnumeratorMoveNext(TARGET_METHOD_ORIGINAL);
+
+    private static readonly InstructionsPattern SpawnResFromPrefPattern = new()
```
tornac1234, line 27: Pattern not working, make sure to not confuse between the "Call" and "Callvirt" opcodes, which are different

NitroxPatcher/Patches/Dynamic/BreakableResource_SpawnResourceFromPrefab_Patch.cs
```diff
@@ -0,0 +1,59 @@
...
+                    break;
+            }
+        }
+        return instructions.Transform(SpawnResFromPrefPattern, InsertCallback);
+    }
+
+    private static void Callback(GameObject __instance)
+    {
+        NitroxId newId = new();
+        NitroxEntity.SetNewId(__instance, newId);
```
tornac1234, line 51:
```suggestion
        NitroxEntity.SetNewId(__instance, new());
```

Nitrox.Test/Patcher/Patches/Dynamic/SpawnOnKill_OnKill_PatchTest.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Linq;
+using FluentAssertions;
+using HarmonyLib;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using NitroxTest.Patcher;
+using static NitroxPatcher.Patches.Dynamic.SpawnOnKill_OnKill_Patch;
+
+namespace Nitrox.Test.Patcher.Patches.Dynamic;
```
tornac1234, line 10:
```suggestion
namespace NitroxPatcher.Patches.Dynamic;
```
As explained by Measurity in #2017, we now have a defined namespace rule, which you'll also apply when making the test for BreakableResource's transpiler

tornac1234 (changes requested):
Will review again once you'll have made sure the transpiler is working

NitroxClient/Communication/Packets/Processors/EntityDestroyedProcessor.cs
VELD-Dev: It's @Measurity's changes

NitroxClient/Communication/Packets/Processors/EntityDestroyedProcessor.cs
VELD-Dev: @Measurity

NitroxClient/Communication/Packets/Processors/EntityDestroyedProcessor.cs
Measurity: 👍

NitroxPatcher/Patches/Dynamic/BreakableResource_SpawnResourceFromPrefab_Patch.cs
Measurity: When using `Reflect.*` APIs it should now match any call-like opcode.

NitroxPatcher/Patches/Dynamic/BreakableResource_BreakIntoResources_Patch.cs
Measurity: Since the breakable entity already should have a Nitrox ID it shouldn't assign a new one (`GetId` does assing one if there is no ID set). 

Here's an example:

```cs
        if (!NitroxEntity.TryGetEntityFrom(__instance.gameObject, out NitroxEntity destroyedEntity))
        {
            Log.Warn($"[{nameof(BreakableResource_BreakIntoResources_Patch)}] Could not find {nameof(NitroxEntity)} for breakable entity {__instance.gameObject.GetFullHierarchyPath()}.");
            return;
        }

        Resolve<IPacketSender>().Send(new EntityDestroyed(destroyedEntity.Id));
```

VELD-Dev: I still don't get the point, like why do you do that ?

Measurity: Code shouldn't do something unexpected like pretending a broken state (missing id) is fine and just assigning an unsynced id (only your client knows it) to the breakable entity.

Rather, we want to know that the state is unexpected. Because it means the server did a bad job synchronizing the id of the breakable entity to the client. Hence the logging and doing nothing (since there's no good way forward when the id missing).

Measurity: Hypothetically: in game it would leave things unsynchronized because we can't fix it client side. Meaning the breakable entity isn't broken but stays. The issue would be on the server and would require further investigation by us.

VELD-Dev: like this ?
```cs
    public static void Prefix(BreakableResource __instance)
    {
        // Don't ask me what it does...
        if (!NitroxEntity.TryGetEntityFrom(__instance.gameObject, out NitroxEntity destroyedEntity))
        {
            Log.Warn($"[{nameof(BreakableResource_BreakIntoResources_Patch)}] Could not find {nameof(NitroxEntity)} for breakable entity {__instance.gameObject.GetFullHierarchyPath()}.");
            return;
        }
        // Send packet to destroy the entity
        Resolve<IPacketSender>().Send(new EntityDestroyed(destroyedEntity.Id));
    }
```

Measurity: Yes, that's all

NitroxPatcher/Patches/Dynamic/BreakableResource_BreakIntoResources_Patch.cs
```diff
@@ -20,8 +20,12 @@ public class BreakableResource_BreakIntoResources_Patch : NitroxPatch, IDynamicP
 
     public static void Prefix(BreakableResource __instance)
     {
-        // Gets the entity ID
-        NitroxId id = NitroxEntity.GetId(__instance.gameObject);
+        // Don't ask me what it does...
```
Measurity, line 23: This can be removed now

VELD-Dev: indeed

Measurity (approved):

NitroxClient/GameLogic/Items.cs
```diff
@@ -59,25 +59,36 @@ public void PickedUp(GameObject gameObject, TechType techType)
...
+        /// </summary>
+        public void Dropped(GameObject gameObject, TechType techType = default)
         {
+            if (techType == default)
+            {
+                techType = CraftData.GetTechType(gameObject);
+                if (techType == TechType.None)
+                {
+                    Log.Error($"Could not find {nameof(TechType)} for {gameObject.name}");
+                    return;
```
Sunrunner37, line 73: Spawning primarily uses class id (below) for loading the prefab. Tech type is normally just a 'nice to have'.  I don't think we should get ourselves into a desync state if we resolve to TechType.None.

VELD-Dev: but it cannot happen ? @Measurity what do you think ?
In any case, it's just moving the CraftData.GetTechType(GameObject) from the exterior of the function to inside the function

Measurity: Removed the return and logging. Will not just pass in whatever TechType is fetched by `CraftData.GetTechType(gameObject)` unless a non-null `TechType` was already given.

Sunrunner37: Thanks Measurity.  @VELD-Dev just for clarity, via `CraftData.GetTechType`, it seems to be a normal possible case that the prefab identifier exists but there is no mapping of class id to tech type (defaults to none).  

image

VELD-Dev: > Thanks Measurity. @VELD-Dev just for clarity, via `CraftData.GetTechType`, it seems to be a normal possible case that the prefab identifier exists (thus a class id) but there is no mapping of class id to tech type (defaults to none).
> 
> image

Ohhh okay got it, thanks !

NitroxPatcher/Patches/Dynamic/SpawnOnKill_OnKill_Patch.cs
```diff
@@ -0,0 +1,58 @@
...
+            }
+        }
+
+        return instructions.Transform(spawnInstanceOnKillPattern, InsertCallbackCall);
+    }
+
+    private static void Callback(SpawnOnKill spawnOnKill, GameObject spawningItem)
+    {
+        NitroxId destroyedEntityId = NitroxEntity.GetId(spawnOnKill.gameObject);
+        Resolve<IPacketSender>().Send(new EntityDestroyed(destroyedEntityId));
```
VELD-Dev, line 50: where went `NitroxEntity.SetNewId(spawningItem, destroyedEntityId)` ???

Sunrunner37: We shouldn't re-use the old id because it represents a completely different type of object. Instead, we should set a new id to spawningItem.  Right now this is happening implicitly inside `Dropped` when it calls `GetId`

VELD-Dev: Ok, fine for me 👍

NitroxPatcher/Patches/Dynamic/SpawnOnKill_OnKill_Patch.cs
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        return instructions.Transform(spawnInstanceOnKillPattern, InsertCallbackCall);
+    }
+
+    private static void Callback(SpawnOnKill spawnOnKill, GameObject spawningItem)
+    {
+        if (!NitroxEntity.TryGetEntityFrom(spawnOnKill.gameObject, out NitroxEntity destroyedEntity))
+        {
+            Log.Warn($"[{nameof(BreakableResource_BreakIntoResources_Patch)}] Could not find {nameof(NitroxEntity)} for breakable entity {spawnOnKill.gameObject.GetFullHierarchyPath()}.");
```
Sunrunner37, line 51: minor: The comment here is inaccurate it says `nameof(BreakableResource_BreakIntoResources_Patch)`

NitroxPatcher/Patches/Dynamic/SpawnOnKill_OnKill_Patch.cs
```diff
@@ -0,0 +1,64 @@
...
+        } else
```
Sunrunner37, line 52: minor: formatting

NitroxPatcher/Patches/Dynamic/SpawnOnKill_OnKill_Patch.cs
VELD-Dev: oh yeah that's true, I just copied without thinking about changing that my bad, i'm correcting it right now

NitroxPatcher/Patches/Dynamic/SpawnOnKill_OnKill_Patch.cs
VELD-Dev: what needs to be formatted ?

Sunrunner37: We follow the standard c# convention: 

```
if (var)
{
}
else
{
}
```

PR #2018, opened 2023-04-08
Creating custom AutoBogus for WorldPersistenceTest and PacketsSerializableTest

Jannify:
The "official" AutoBogus nuget package provided many problems when using on PersistedWorldData and was also very slow. Therefor I implemented our own version with full customization.

While implementing I [fixed](https://github.com/SubnauticaNitrox/Nitrox/commit/91de4da217191169d2b96ceae7190dbe033be587) a few serialization errors which have slipped the pr reviews.


Nitrox.Test/Helper/Faker/NitroxAbstractFaker.cs
```diff
@@ -0,0 +1,66 @@
...
+        AssignableTypesCount = subTypes.Length;
+        FakerByType.Add(type, this);
+        foreach (Type subType in subTypes)
+        {
+            assignableFakers.Enqueue(GetOrCreateFaker(subType));
+        }
+    }
+
+    public INitroxFaker[] GetSubFakers() => assignableFakers.ToArray();
+
```
Measurity, line 59: A comment would be nice here to explain the logic.

```suggestion

    /// <summary>
    ///     Selects an implementing type in a round-robin fashion of the abstract type of this faker. Then creates an instance of it.
    /// </summary>
```

NitroxServer/Serialization/World/WorldPersistence.cs
```diff
@@ -93,26 +93,38 @@ internal Optional<World> LoadFromFile(string saveDir)
                 return Optional.Empty;
             }
 
-            try
+            UpgradeSave(saveDir);
```
Measurity, line 96: The UpgradeSave can throw null ref when `saveFileVersion` is null. This should probably be handled to run a `return` early.

Measurity (approved):

dartasen (approved):

PR #2021, opened 2023-04-11
Allow the server to download a prebaked entity file hosted by nitrox.

Sunrunner37:
Currently, Nitrox loads entities as the player moves around the world.  This causes notable lag when a player enters a batch for the first time because the server needs to invoke the batch spawning logic.  The alternative is for the user to enable `CreateFullEntityCache` in the config file, which loads all of the entities ahead of time but causes a significant start-up latency.

This change allows us to host a prebaked save file that contains all of the parsed entity data (identified as `BAKED`).  The previous functionality is still available under the modes `STREAMING` and `UPFRONT`. The release build will always use baked because it is the ideal user experience; however, developers will leverage streaming by default because it is easier on the hard disk.  If developers want to test features that do not require spawning they can set the config `prebakedSave` to `Dev-NoSpawningRequired.zip` and select baked behavior to skip all entity spawning. 

When baked is enabled, the server will download the baked save file to the app data folder (next to saves).  This download only needs to happen one time and will be used for any new server going forward.  Updating the config will cause a new file to download (in case we identify a bug in the template).

Baked mode will also prevent us from doing expensive resource parsing (such as prefab placeholder groups).  This optimization makes me comfortable disabling the concurrency in this class that was causing deadlock and potentially inconsistent spawning in game. 

Note: saves can be found here: https://github.com/SubnauticaNitrox/Saves


NitroxServer/Serialization/World/WorldManager.cs
```diff
@@ -130,12 +133,14 @@ public static string CreateEmptySave(string name)
 
         ServerConfig serverConfig = ServerConfig.Load(saveDir);
 
-        string fileEnding = "json";
         if (serverConfig.SerializerMode == ServerSerializerMode.PROTOBUF)
         {
-            fileEnding = "nitrox";
+            new ServerProtoBufSerializer().Serialize(Path.Combine(saveDir, $"Version.nitrox"), new SaveFileVersion());
```
Jannify, line 138: Shouldn't we use the created Serializers by AutoFac here?

NitroxServer/Serialization/World/WorldPersistence.cs
```diff
@@ -165,6 +174,32 @@ private World CreateFreshWorld()
...
+                string url = $"https://github.com/SubnauticaNitrox/Saves/raw/master/{Uri.EscapeDataString(config.prebakedSave)}";
+                Log.Info($"Downloading prebaked save from {url} to {prebakedPath}");
+                using WebClient client = new();
+                client.DownloadFile(url, prebakedPath);
+            }
+
+            using ZipArchive zip = new(File.OpenRead(prebakedPath), ZipArchiveMode.Read);
+            zip.ExtractToDirectory(savePath);
+
+            Log.Info($"Extracting prebaked save into {savePath}");
```
Jannify, line 194:
```suggestion
            Log.Info($"Extracted prebaked save into {savePath}");
```

Jannify (changes requested):

dartasen:
Closing it for now, it's likely we won't go in that direction for now

PR #2023, opened 2023-04-22
Upgrade LiteNetLib to v1.0.1.1

Jannify:


NitroxServer/Communication/ConnectionState.cs
```diff
@@ -1,11 +1,32 @@
...
+        }
+
+        if (connectionState.HasFlag(ConnectionState.Disconnected))
+        {
+            state = NitroxConnectionState.Disconnected;
+        }
+
+        return state;
     }
 }
```
Measurity, line 32: Would use bitwise operations here because HasFlag isn't generic and boxes the Enum argument (fixed in .NET Core+, but we're on mono)
Extension classes usually end with plural `s`

```suggestion
public static class NitroxConnectionStateExtensions
{
    public static NitroxConnectionState ToNitrox(this ConnectionState connectionState)
    {
        if ((connectionState & ConnectionState.Connected) == ConnectionState.Connected)
        {
            return NitroxConnectionState.Connected;
        }
        if ((connectionState & ConnectionState.Disconnected) == ConnectionState.Disconnected)
        {
            return NitroxConnectionState.Disconnected;
        }

        return NitroxConnectionState.Unknown;
    }
}
```

Measurity (changes requested):
Can you add the following after the try-catch in `UpgradeSave` function? When I create a new save it still NREs because the serializer returns null without throwing exceptions..

```cs
            if (saveFileVersion == null)
            {
                // No version file, do nothing.
                return;
            }
```

Measurity (approved):
Works for me. If it can serialize the first few packets I don't expect this to suddenly break 👍

NitroxServer/Communication/LiteNetLib/LiteNetLibServer.cs
```diff
@@ -122,13 +120,12 @@ private void PeerDisconnected(NetPeer peer, DisconnectInfo disconnectInfo)
 
     private void NetworkDataReceived(NetPeer peer, NetDataReader reader, byte channel, DeliveryMethod deliveryMethod)
     {
-        netPacketProcessor.ReadAllPackets(reader, peer);
-    }
+        int packetDataLength = reader.GetInt();
+        byte[] packetData = new byte[packetDataLength];
```
dartasen, line 124: We might wanna check if packetDataLength hasn't a weird value before allocating memory based on this int

NitroxClient/Communication/NetworkingLayer/LiteNetLib/LiteNetLibClient.cs
```diff
@@ -83,14 +81,13 @@ public void Stop()
 
     private void ReceivedNetworkData(NetPeer peer, NetDataReader reader, byte channel, DeliveryMethod deliveryMethod)
     {
-        netPacketProcessor.ReadAllPackets(reader, peer);
-    }
+        int packetDataLength = reader.GetInt();
+        byte[] packetData = new byte[packetDataLength];
```
dartasen, line 85: Same here : check int beforehand

dartasen (commented):
Otherwise looks good code wise

dartasen (approved):

PR #2032, opened 2023-04-29
Disable "Add Server" button when add server window overlay enabled

### LINKED ISSUE #1995: "add Server" window will not close on "discard" button.

Yes, i'm on commit: 48fe2be8a77bebebefa0726732e0981fc6fff6f3

Windows

Steam

### Describe the issue.

While adding a new Server from the multiplayer menu, clicking "abbrechen" (or "discard" in english prob), while hovering over the "add Server" button will not close the "add server" window.

I'm guessing it's probably got todo with, the "add Server" button beeing triggerred at the same time and therefore reopening the window while at the same time closing it.
Therefore it appears to not close.
Should be quite easy to fix.

image

Start Subnautica.
Open Multiplayer.
Click "add Server" button.
Try to close the pop-up window, while hovering over the button and the "add Server" button at the same time.

### Relevant log output

_No response_

MistaOmega:
Should close #1995 by disabling button when showingAddServer enabled

Main method for disabling button is handled in OnGUI but utilises a flag to prevent multiple expensive ops per frame. should mean it should run the expensive ops on 1 frame, and never again until the state changes

This fix required both modifying a variable name and moving this gameobject var to an outer scope. I have ran basic tests throughout the menu and have not observed any issues with this whatsoever. but if there is something deep-rooted that this would affect, I can try and approach it from a different angle.

### DIFF
--- a/NitroxClient/MonoBehaviours/Gui/MainMenu/MainMenuMultiplayerPanel.cs
+++ b/NitroxClient/MonoBehaviours/Gui/MainMenu/MainMenuMultiplayerPanel.cs
@@ -21,6 +21,7 @@ namespace NitroxClient.MonoBehaviours.Gui.MainMenu
         private GameObject savedGamesRef;
         private GameObject deleteButtonRef;
         private GameObject multiplayerButton;
+        private GameObject addServerButtonInst;
         private Transform savedGameAreaContent;
         public JoinServer JoinServer { get; private set; }
 
@@ -31,6 +32,7 @@ namespace NitroxClient.MonoBehaviours.Gui.MainMenu
         private bool shouldFocus;
         private bool showingAddServer;
         private bool isJoining;
+        private bool isAddServerEnabled;
 
         public void Setup(GameObject loadedMultiplayer, GameObject savedGames)
         {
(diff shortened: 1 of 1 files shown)


Measurity (changes requested):
Thanks for working on this issue.

The problem with this GUI is that clicks aren't blocked by the "add server UI" popup and pass through to the Canvas UI that Subnautica uses. Immediately opening the "add server UI" again because the cancel click happens to overlap with the "add server" button.

I propose to use `uGUI_MainMenu.main.canvasGroup.interactable = false/true` to disable the game UI input until after the popup is closed. There might be a way to block input with IMGUI UIs, if so that would be even better.

Here's an example, don't forget to also change `ShowAddServerWindow`:

```cs
private void HideAddServerWindow()
{
    IEnumerator Inner()
    {
        showingAddServer = false;
        shouldFocus = true;
        // Wait one frame so that the click event of main menu buttons aren't triggered (on mouse up event)
        yield return null;
        uGUI_MainMenu.main.canvasGroup.interactable = true;
    }

    CoroutineHost.StartCoroutine(Inner());
}
```

MistaOmega:
By doing the changes I put in place it disables the action triggering an event on pass through, which does prevent what the issue outlined 

However I see why this is less than optimal, especially when compared with your outlined method. I'll attempt that ASAP and modify the PR 

Thank you

NitroxClient/MonoBehaviours/Gui/MainMenu/MainMenuMultiplayerPanel.cs
```diff
@@ -132,37 +130,37 @@ public static async Task OpenJoinServerMenuAsync(string serverIp, string serverP
...
+            {
+                serverNameInput = "local";
+                serverHostInput = "127.0.0.1";
+                serverPortInput = ServerList.DEFAULT_PORT.ToString();
+                showingAddServer = true;
+                shouldFocus = true;
+                yield return null;
+                uGUI_MainMenu.main.canvasGroup.interactable = false;
+            }
+            CoroutineHost.StartCoroutine(SetWindowComponents());
```
Measurity, line 143: This works but it's not necessary to wait until next frame to disable input. Conceptually I prefer it like this, where the input is disabled as soon as possible.

```cs
            uGUI_MainMenu.main.canvasGroup.interactable = false;
            serverNameInput = "local";
            serverHostInput = "127.0.0.1";
            serverPortInput = ServerList.DEFAULT_PORT.ToString();
            showingAddServer = true;
            shouldFocus = true;
```

Measurity (commented):
Works and looks pretty good, just one minor request and then I'll approve.

NitroxClient/MonoBehaviours/Gui/MainMenu/MainMenuMultiplayerPanel.cs
MistaOmega: Understood, making the edits now, thank you

MistaOmega:
I am assume the coroutine logic for hiding needs to stay, otherwise the problem would persist, so I've left that part in, just removed the 1 frame delay on disabling input

Measurity (approved):
I'll let someone else from the team review and then we'll merge it.

dartasen (approved):

Issue #2020, opened 2023-04-10
Add an "Access to Folder" or move it to the world manager page of launcher


VELD-Dev:
As explained in the title, add a button to access the folder of the save directly from the launcher, for example into the save settings. I think it would greatly improve navigation to save files, and above this, it would be much better if it's a button on each server/save in the manager, it would automatically open *this* save. Maybe an idea for the next launcher ? I think this would be an interesting improvement for users and for interface simplicity

Measurity:
Implemented in new launcher. See #1848

PR #2033, opened 2023-05-01
Added ability to open explorer at Server World save file location on Launcher.

MistaOmega:
Addresses #2020 by adding a button to the server worlds menu to open the save file location of that world.
Used existing settings and icons wherever I could to make this addition as seamless as possible. 

I also propose replacing the ellipsis icon for the world settings with the cog icon, allowing me to then use the ellipsis icon for opening the save file location which think would make more sense. 

Increased the size of each list element to accommodate the new button whilst keeping the old button sizes the same. 83 -> 125 px in height.

### DIFF
--- a/NitroxLauncher/Pages/ServerPage.xaml.cs
+++ b/NitroxLauncher/Pages/ServerPage.xaml.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Reflection;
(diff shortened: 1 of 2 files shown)


NitroxLauncher/Pages/ServerPage.xaml.cs
```diff
@@ -405,7 +427,7 @@ private void TBWorldSeed_Input(object sender, KeyboardFocusChangedEventArgs e)
             Config.Seed = TBWorldSeed.Text;
         }
 
-        private void TBMaxPlayerCap_Input(object sender, System.Windows.Input.KeyboardFocusChangedEventArgs e)
+        private void TBMaxPlayerCap_Input(object sender, KeyboardFocusChangedEventArgs e)
```
MistaOmega, line 430: Unsure why, but I think the cleanup profile may have done these changes removing the specific import path. I can add these back if this layout is preferred

dartasen: World manager was developed by one of our support team (non-developer), so this probably due to adding event with XAML editor

dartasen (commented):
Thank's for working on this 😀
However we're currently rewriting an MVVM launcher with AvaloniaUI that will allow us to start porting the mod to Linux. So I don't know if we should merge this

Feel freeto join [our discord](https://discord.com/invite/E8B4X9s), so we can have a talk about it in dev channels !

dartasen:
I'm closing this one for now, as discussed on discord

PR #2043, opened 2023-05-21
Intro cinematic sync

Jannify:
https://cdn.discordapp.com/attachments/1038497385208365177/1120414323932680282/2023-06-19_20-03-44.mp4

### DIFF
--- a/NitroxClient/GameLogic/InitialSync/PlayerInitialSyncProcessor.cs
+++ b/NitroxClient/GameLogic/InitialSync/PlayerInitialSyncProcessor.cs
@@ -42,8 +42,11 @@ public class PlayerInitialSyncProcessor : InitialSyncProcessor
 
     private void SetPlayerIntroCinematicMode(IntroCinematicMode introCinematicMode)
     {
-        localPlayer.IntroCinematicMode = introCinematicMode;
-        Log.Info($"Received initial sync player IntroCinematicMode: {introCinematicMode}");
+        if (localPlayer.IntroCinematicMode < introCinematicMode)
+        {
+            localPlayer.IntroCinematicMode = introCinematicMode;
+            Log.Info($"Received initial sync player IntroCinematicMode: {introCinematicMode}");
+        }
     }
 
     private void SetPlayerGameObjectId(NitroxId id)
--- a/NitroxClient/GameLogic/InitialSync/PlayerInitialSyncProcessor.cs
+++ b/NitroxClient/GameLogic/InitialSync/PlayerInitialSyncProcessor.cs
@@ -28,6 +28,7 @@ public class PlayerInitialSyncProcessor : InitialSyncProcessor
         this.localPlayer = localPlayer;
 
         AddStep(sync => SetPlayerPermissions(sync.Permissions));
+        AddStep(sync => SetPlayerIntroCinematicMode(sync.IntroCinematicMode));
         AddStep(sync => SetPlayerGameObjectId(sync.PlayerGameObjectId));
         AddStep(sync => AddStartingItemsToPlayer(sync.FirstTimeConnecting));
         AddStep(sync => SetPlayerStats(sync.PlayerStatsData));
@@ -37,6 +38,12 @@ public class PlayerInitialSyncProcessor : InitialSyncProcessor
     private void SetPlayerPermissions(Perms permissions)
     {
         localPlayer.Permissions = permissions;
+    }
+
+    private void SetPlayerIntroCinematicMode(IntroCinematicMode introCinematicMode)
+    {
+        localPlayer.IntroCinematicMode = introCinematicMode;
+        Log.Info($"Received initial sync player IntroCinematicMode: {introCinematicMode}");
     }
 
     private void SetPlayerGameObjectId(NitroxId id)
--- a/NitroxClient/MonoBehaviours/IntroCinematicUpdater.cs
+++ b/NitroxClient/MonoBehaviours/IntroCinematicUpdater.cs
@@ -1,0 +1,164 @@
+using System.Collections;
+using NitroxClient.GameLogic;
+using NitroxModel.Core;
+using UnityEngine;
+
+namespace NitroxClient.MonoBehaviours;
+
+public class IntroCinematicUpdater : MonoBehaviour
+{
+    public static RemotePlayer Partner;
+    private static Transform modelRoot;
+
+    private static readonly Transform playerTransform = Player.main.transform;
+    private static readonly Quaternion rotateOtherSide = Quaternion.Euler(0, 180, 0);
+
+    private Transform introEndTarget;
+    private Vector3 introRemoteEndPosition;
+    private Transform[] seatPartsLeft;
+    private Transform[] seatPartsRight;
+    private Transform seatArmRestLeft;
+    private Transform seatArmRestRight;
+    private SkinnedMeshRenderer[] seatBarRendererRight;
+
+    private SkinnedMeshRenderer remotePlayerHeadRenderer;
+    private GameObject remotePlayerCustomHead;
+    private SkinnedMeshRenderer[] remoteRenders;
+
+    public void Awake()
+    {
+        if (Partner == null)
+        {
+            Log.Error($"[{nameof(IntroCinematicUpdater)}.Awake()]: Partner was null. Disabling MonoBehaviour.");
+            enabled = false;
+            return;
+        }
+        modelRoot = EscapePod.main.transform.Find("models/Life_Pod_damaged_03/root");
+        Transform seatLeft = modelRoot.Find("life_pod_seat_01_left_damaged_jnt1");
+        Transform seatRight = modelRoot.Find("life_pod_seat_01_right_damaged_jnt1");
+
+        seatPartsLeft = new[]
+        {
+            seatLeft.Find("life_pod_seat_01_left_damaged_jnt2"),
+            seatLeft.Find("life_pod_seat_01_left_damaged_jnt3")
+        };
+
+        seatPartsRight = new[]
+        {
+            seatRight.Find("life_pod_seat_01_right_damaged_jnt2"),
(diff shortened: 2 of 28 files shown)


Jannify:
Syncing the cinematic and basic team-up/skip UI is done.

Notes on the state of this pr:
- The intro is disabled in `DEBUG` by default, edit `MainGameController_ShouldPlayIntro_Patch` for testing
- The intro works better with multiple lifepods but that is out of scope for this pr, mainly bc:
- Interacting with the lifepod by a third player (using hatch ladder, etc.) will hard cut the intro cinematic with unknown side effects. Therefor the server should spawn one lifepod per player pair/intro cinematic and it should be locked for other players
- We need a config options so big server can disable it intro and spawning multiple lifepods

NinjaPedroX:
Instead of locking the lifepods, have them fall out of the sky during the intro cutscene :)

NitroxServer/Communication/Packets/Processors/SetIntroCinematicModeProcessor.cs
```diff
@@ -0,0 +1,32 @@
...
+{
+    private readonly PlayerManager playerManager;
+
+    public SetIntroCinematicModeProcessor(PlayerManager playerManager)
+    {
+        this.playerManager = playerManager;
+    }
+    public override void Process(SetIntroCinematicMode packet, Player player)
+    {
+        player.PlayerContext.IntroCinematicMode = packet.Mode;
```
tornac1234, line 19: Probably want to modify the packet to make sure that the provided ``packet.PlayerId`` is equals to ``player.Id`` just in case (security-wise).

Measurity (changes requested):
Code wise seems fine and working but did notice a side effect:
- On every join you get a "press to continue" screen which is different from master branch. I don't know if this was intended and if we want it. I personally prefer not to have the screen.

tornac1234 (changes requested):
Encountered an issue for the second player who joined.
Also my two local players didn't start the cinematic at the exact same time. Maybe it could be made so that the packet sent from the server tells them to start at a precise timing like Time.now + 3 seconds (but it would change the packet structure).
Another issue to me is the fact that a player can't just skip waiting for a partner and start the cinematic without other people (let's say a group of 3 people want to play, 2 players will see the cinematic but the other one will be stuck).
image

NitroxClient/GameLogic/InitialSync/PlayerInitialSyncProcessor.cs
```diff
@@ -23,30 +23,42 @@ public PlayerInitialSyncProcessor(Items item, ItemContainers itemContainers)
 
         public override IEnumerator Process(InitialPlayerSync packet, WaitScreen.ManualWaitItem waitScreenItem)
         {
-            SetPlayerPermissions(packet.Permissions);
-            waitScreenItem.SetProgress(0.16f);
+            LocalPlayer localPlayer = NitroxServiceLocator.LocateService<LocalPlayer>();
+
+            SetPlayerPermissions(localPlayer, packet.Permissions);
+            waitScreenItem.SetProgress(0.14f);
```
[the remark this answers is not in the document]
Jannify: Good idea. Best for separate pr. Will add a `TODO:`

NitroxPatcher/Patches/Dynamic/EscapePod_StopIntroCinematic_Patch.cs
```diff
@@ -0,0 +1,26 @@
...
+
+public class EscapePod_StopIntroCinematic_Patch : NitroxPatch, IDynamicPatch
+{
+    private static readonly MethodInfo targetMethod = Reflect.Method((EscapePod e) => e.StopIntroCinematic(default(bool)));
+
+    public static void Postfix(bool isInterrupted)
+    {
+        if (isInterrupted)
+        {
+            //TODO: Currently this is sent as an unauthenticated packet when skipping the intro up front (Configuration=Debug) and is therefor not processed by the server.
```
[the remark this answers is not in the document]
Jannify: Don't know what would be the best way to do that.

Measurity (approved):

dartasen (approved):
Works fine IG

NitroxPatcher/Patches/Dynamic/EscapePod_StopIntroCinematic_Patch.cs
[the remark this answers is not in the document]
tornac1234: Could probably have a ``packetsToSendOnceConnected" queue (would also contain ``PlayerSyncFinished`` sent by ``InitialPlayerSyncProcessor.ProcessInitialSyncPacket``) which would be all sent by ``Multiplayer.SetLoadingComplete``

NitroxClient/MonoBehaviours/IntroCinematicUpdater.cs
```diff
@@ -0,0 +1,164 @@
...
+        yield return new WaitForSecondsRealtime(50.5f - 40f); //Slowly move remote player backwards
+        shouldDisplaceRemotePlayer = true;
+
+        yield return new WaitForSecondsRealtime(50.72f - 50.5f); //Disable and move remote player
+        shouldDisplaceRemotePlayer = false;
+        remotePlayerDisplacement = Vector3.zero;
+        remoteRenders.ForEach(smr => smr.enabled = false);
+
+        yield return new WaitForSecondsRealtime(53f - 50.72f); //Slowly move local player to modified cin-end
+        shouldDisplaceLocalPlayer = true;
```
Drakonkinst, line 114: Could these values be moved into constants? Seems like some repeated values that would also benefit from being named constants.

NitroxClient/MonoBehaviours/IntroCinematicUpdater.cs
```diff
@@ -0,0 +1,164 @@
...
+        {
+            for (int i = 0; i < seatPartsLeft.Length; i++)
+            {
+                seatPartsRight[i].localRotation = seatPartsLeft[i].localRotation;
+            }
+
+            seatArmRestRight.localPosition = -seatArmRestLeft.localPosition;
+        }
+
+        if (shouldDisplaceRemotePlayer) remotePlayerDisplacement += new Vector3(0, 0, 0.05f);
```
Drakonkinst, line 146: There should be brackets here. Per contribution guidelines: "Encapsulate everything in braces, even single-line statements."

Also trying to figure out what this is doing for my own learning, why are we displacing by this amount? This seems too tiny to be the distance between the seats so I'm confused what its purpose is.

NitroxPatcher/Patches/Dynamic/uGUI_SceneIntro_IntroSequence_Patch.cs
```diff
@@ -0,0 +1,174 @@
...
+    {
+        // Return Time.time when starting solo and disable skip button when staring duo
+        return Resolve<LocalPlayer>().IntroCinematicMode == IntroCinematicMode.SINGLEPLAYER ? Time.time : -1f;
+    }
+
+    // ReSharper disable once UnusedMethodReturnValue.Local
+    private static bool IsRemoteCinematicReady(uGUI_SceneIntro uGuiSceneIntro)
+    {
+        if (callbackRun) return true;
+        if (GameModeUtils.currentGameMode.HasFlag(GameModeOption.Creative)) uGuiSceneIntro.Stop(true); // Stopping intro if Creative like in normal SN
```
Drakonkinst, line 96: See previous comment about braces

NitroxPatcher/Patches/Dynamic/uGUI_SceneIntro_IntroSequence_Patch.cs
```diff
@@ -0,0 +1,174 @@
...
+
+        if (Resolve<LocalPlayer>().IntroCinematicMode == IntroCinematicMode.COMPLETED)
+        {
+            uGuiSceneIntro.Stop(true);
+            EndRemoteCinematic();
+        }
+
+        if (!uGuiSceneIntro.moveNext) return false;
```
Drakonkinst, line 104: See previous comment about braces

NitroxClient/MonoBehaviours/IntroCinematicUpdater.cs
Jannify: The idea behind this is that the "time since cinematic started" is always at the left side of the subtraction. That way if we need to change the timing or insert something, we can just bring up a running stopwatch on the screen and know exactly when something is going to happen. I think making them to dedicated consts would just end up in "firstSegmentTime, secondSegmentTime".

NitroxClient/MonoBehaviours/IntroCinematicUpdater.cs
Jannify: We updated our guidelines and code style internally but didn't published this change to the contribution guidelines. This should now be updated 👍

NitroxPatcher/Patches/Dynamic/uGUI_SceneIntro_IntroSequence_Patch.cs
```diff
@@ -0,0 +1,187 @@
...
+            partner.AnimationController["escapepod_intro"] = true;
+
+            IntroCinematicUpdater.Partner = partner;
+            partner.Body.AddComponent<IntroCinematicUpdater>();
+        }
+    }
+
+    private static void EndRemoteCinematic()
+    {
+        if (partner != null)
```
tornac1234, line 169: "partner" is not null even when the partner disconnects to fix that we can either:
- replace: ``partner != null`` by ``IsPartnerValid()`` in the two methods
```cs
private static bool IsPartnerValid()
{
    return partner != null && Resolve<PlayerManager>().Find(partner.PlayerId).HasValue;
}
```
- have the patch listen to ``PlayerManager.OnRemove``, which might be a bit too much pain to find the right time to register and then to unregister the callback

tornac1234 (changes requested):
Tests revealed it was good apart from the critical issue (in one comment)

PR #2052, opened 2023-06-13
Implemented initial stages on BZ integration

Coding-Hen:
Missing lots and still a lot of work but should give a start for us allowing BZ compatibility

This really needs to be moved out into its seperate components and not with the compiler flags but that can be done in due time. 

Happy to make any changes before this goes in if people would prefer


20PercentRendered:
I tried porting Nitrox to Below Zero last year (before the 2.0 update was announced, started it around August), and a lot of the issues I faced were related to the major changes in Subnautica 2.0 (BZ has had 2.0 for a long time now)

This a writeup of the issues I encountered and at the end a way I think we could add BZ support cleanly. 

Most of the problem spots were:
 - Sync functions being removed and being forced to use coroutines for things that were synchronous before
 - Different asset paths
 - Switch to prefab system
 - TextMeshPro being used instead of the old unity system

BZ related bugs:
- Bug: Main menu's multiplayer button didn't show up
    - Issue: Wrong asset paths and color picker didn't exist in the same place in BZ
    - Patchwork Solution: Load color picker prototype with `Addressables.LoadAssetAsync<GameObject>` (I can provide full code if needed)
    - Long-term solution: Create a UI Framework/uGUI wrapper. Would also help mod devs in the future who want to mod Nitrox when mod support comes.
  
- Bug: Entities not spawning at all
    - Issue: NitroxServer loads Subnautica assets instead of BZ assets
    - Solution: In NitroxServer get the current game's path through AutoFac. In game-specific servers, provide the correct game path through AutoFac. 
    
- Bug: Entities don't spawn near the player when they join in
    - Issue: ???
    - Patchwork Solution: Walk away from your spawn area and walk back
   
- Bug: Building doesn't work
    - Issue: Nitrox doesn't yet support the 2.0 update's building systems. 
    - Solution: Wait for building support
   
- Bug: Tiny black cubes floating about
    - Issue: The models fail to load and are replaced with a placeholder lootcube
    - Solution: ????
  
- Bug: PDA doesn't save anything and player tab broken
    - Issue: ?????
    - Solution: ?????

Lots of these might already work with little to no effort due to the immense amount of work poured into adding 2.0 support to Nitrox. 

My idea for adding BZ support without spaghetti (take with multiple grains of salt, I'm not a very experienced Nitrox dev):

1. Split (Nitrox Project)-Subnautica:s into (Nitrox Project)-All (for both games, can be compiled with either of the games) and (Nitrox Project)-Subnautica for SN only. 
2. Create BZ versions of (Nitrox Project)s
3. Edit MSBuild scripts so that:
     - Support for a game is built if the user has that game
     - Projects can check for Subnautica/BZ through MSBuild properties
     - Launcher will only support the games the dev has, shouldn't be an issue since CI probably does have both games
4. Edit all references to GamePath replacing it with some other solution appropriate for that usecase:
     - Multiple variables
     - Function to get game path for SN/BZ
     - Employing AutoFac on Client/Server/Launcher code
5. Add UI support
     - Implement the Library tab inside the launcher (it's atleast partially implemented currently)
     - -or-
     - Add a second "Game" tab and rename them Subnautica and Subnautica: Below Zero if both games exist at build time

dartasen:
Closing for now

PR #2074, opened 2023-08-23
Improve playtest speed

tornac1234:
Really looks hacky right now, couldn't find a way to wait for the server to be ready without hiding the output from the console (it's really a mess)


dartasen:
I'm closing it for now, like said it's really an ad hoc solution and we're maybe willing to implement a more global solution if it really a need while developing

PR #2085, opened 2023-10-02
Added capture of server console input for hotkeys

Measurity:
### Hotkeys
 - `Escape` will clear the input line.
 - `Ctrl+C` will clear the input line. If no input, will terminate the server and console
 - `Ctrl+D` will terminate the server and console


NitroxServer-Subnautica/Program.cs
```diff
@@ -110,20 +120,138 @@ private static async Task StartServer(string[] args)
...
+        await handleConsoleInput;
+
+        Console.Write($"{Environment.NewLine}Server is closing..");
     }
 
-    private static async Task ListenForCommandsAsync(Server server)
+    /// <summary>
+    ///     Handles per-key input of the console and passes input submit to <see cref="ConsoleCommandProcessor"/>.
+    /// </summary>
+    static async Task HandleConsoleInput(Action<string> submitHandler, CancellationTokenSource cancellation = default)
```
dartasen, line 131:
```suggestion
    private static async Task HandleConsoleInputAsync(Action<string> submitHandler, CancellationTokenSource cancellation = default)
```

NitroxServer-Subnautica/Program.cs
```diff
@@ -110,20 +120,138 @@ private static async Task StartServer(string[] args)
             AppMutex.Release();
         }
 
-        await listenForCommands;
+        await handleConsoleInput;
+
+        Console.Write($"{Environment.NewLine}Server is closing..");
```
dartasen, line 125: In external mode, we cannot see this text appearing

dartasen (commented):
Some of the behaviors I find lacking when using a console : 
* Unable to use arrow to navigate between inputs
* Unable to use enter to create an empty line

Also noticed some errors while testing the feature;

Step to reproduce :  Press `tab` then press `delete`
```txt
Index was out of range. Must be non-negative and less than the size of the collection.
Parameter name: index
System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection.
Parameter name: index
   at System.Text.StringBuilder.Remove(Int32 startIndex, Int32 length)
   at NitroxServer_Subnautica.Program.<HandleConsoleInput>d__4.MoveNext() in NitroxServer-Subnautica\Program.cs:line 204
```

Step to reproduce : Open server in `launcher wrap` mode
```txt
System.IO.IOException: Invalid Descriptor
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.Console.set_TreatControlCAsInput(Boolean value)
   at NitroxServer_Subnautica.Program.<StartServer>d__3.MoveNext() in NitroxServer-Subnautica\Program.cs:line 53
```

NitroxServer-Subnautica/Program.cs
Measurity: It's there now after the errors were fixed. But in embedded, it closes faster than external so you won't be able to read it.

dartasen (approved):

Nitrox.Test/Model/DataStructures/CircularBufferTest.cs
```diff
@@ -0,0 +1,111 @@
...
+    public void ShouldBeEmptyWhenCleared()
+    {
+        CircularBuffer<string> buffer = new(10);
+        buffer.Count.Should().Be(0);
+        buffer.Add("1");
+        buffer.Add("1");
+        buffer.Add("1");
+        buffer.Count.Should().Be(3);
+        buffer.Clear();
+        buffer.Count.Should().Be(0);
```
Jannify, line 73: Would also check LastChangedIndex here to be it works with Clear()

Measurity: Yep good call, should have been `-1` after `.Clear()`.

Also moved the `SequenceEquals` check in tests to `ShouldOverwriteOldestItemInBufferWhenCapped` as it fits better there.

Jannify (approved):
LGTM CW

PR #2092, opened 2023-11-22
Add Docfx Code Doc generation support

killzoms:
This adds docFx code doc generation support, adds a new "Linux Release" configuration which does not build the NitroxLauncher project, and clears out some excessive error spam caused by NitroxAnalyzer not running.


Directory.Build.props
```diff
@@ -51,7 +51,7 @@
     <Choose>
         <When Condition="'$(NitroxLibrary)' == 'true'">
             <ItemGroup>
-                <ProjectReference Include="..\Nitrox.Analyzers\Nitrox.Analyzers.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" />
+                <ProjectReference Include="$(MSBuildThisFileDirectory)Nitrox.Analyzers\Nitrox.Analyzers.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" />
             </ItemGroup>
```
Measurity, line 55: Please check if you can use the new nuget package for our analyzers to solve the "fake" VS errors on build:
https://github.com/SubnauticaNitrox/Nitrox.Analyzers

Recommend removing this item group and changing the one below as the following:
```csproj
    <!-- Shared dependencies for all Nitrox.* projects -->
    <Choose>
        <When Condition="'$(NitroxProject)' == 'true'">
            <ItemGroup>
                <PackageReference Include="JetBrains.Annotations" Version="2022.3.1">
                    <Aliases>JB</Aliases>
                </PackageReference>
                <PackageReference Include="IsExternalInit" Version="1.0.3">
                    <PrivateAssets>all</PrivateAssets>
                    <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
                </PackageReference>
                <!-- Include our analyzer from NuGet -->
                <PackageReference Include="Nitrox.Analyzers" Version="1.0.5">
                    <PrivateAssets>all</PrivateAssets>
                    <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
                </PackageReference>
            </ItemGroup>
        </When>
    </Choose>
```

Measurity (changes requested):

Directory.Build.props
killzoms: Want me to revert my Patch method changes as well? or make a PR on the Nitrox.Analyzers Repo for the changes?

Measurity: The former, as it has less OOP complexity.

killzoms: changes made

Measurity: `override` and `TARGET_METHOD` changes are till here which aren't necessary anymore with the nuget packaged analyzers.

killzoms: Should have all other changes reverted now

NitroxServer/NitroxServer.csproj
```diff
@@ -3,6 +3,7 @@
     <PropertyGroup>
         <TargetFramework>net472</TargetFramework>
         <Nullable>disable</Nullable>
+        <Configurations>Debug;Release;LinuxRelease</Configurations>
```
Measurity, line 6: This changes can be safely removed without changing behavior.

Measurity (changes requested):
I think this PR should be draft until we know how to integrate it with the documentation repository. Right now, there's almost no team-wide knowledge, afaik, on how to maintain/use/host docfx or if this is even a useful step towards the overall "documentation" goal. 

Questions we need to answer are: 
 - Do we need another tool for docs? Why docfx? What does it do that docusaurus can't?
 - Can we export docfx for our docusaurus docs.
 - Instructions on how to export/how to keep the documentations in sync with our code.
 - Are we gonna use github actions to do it automatically to improve maintainability?

Directory.Build.props
```diff
@@ -7,7 +7,7 @@
         <TestLibrary>false</TestLibrary>
         <NitroxLibrary>false</NitroxLibrary>
         <UnityModLibrary>false</UnityModLibrary>
-        <BuildToolDir>$(SolutionDir)Nitrox.BuildTool\bin\</BuildToolDir>
+        <BuildToolDir>$(MSBuildThisFileDirectory)Nitrox.BuildTool\bin\</BuildToolDir>
         <BuildGenDir>$(BuildToolDir)generated_files\</BuildGenDir>
         <BuildGenDllDir>$(BuildGenDir)publicized_assemblies\</BuildGenDllDir>
         <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
```
Measurity, line 13: This file still contains references to the "Nitrox.Analyzers" project that can be removed:
```csproj
<PropertyGroup Condition="'$(NitroxProject)' == 'true' and '$(MSBuildProjectName)' != 'Nitrox.Analyzers'">
```

NitroxServer/NitroxServer.csproj
killzoms: Actually LinuxRelease is configured to not build the NitroxLauncher and removing it will actually drop the ability to be able to compile Nitrox on Linux in its current state, granted Nitrox cannot run on Linux as it stands right now...
image

killzoms:
> Questions we need to answer are:
> 
> * Do we need another tool for docs? Why docfx? What does it do that docusaurus can't?

DocFx can generate code docs from C# XML comments as opposed to Docusaurus which as far as I can tell from the limited browsing of its site, only appears to support manually written documentation

> * Can we export docfx for our docusaurus docs.

Yes we can, Here is a tool that does exactly [that](https://github.com/Jan0660/DocFxMarkdownGen)

> * Instructions on how to export/how to keep the documentations in sync with our code.

DocFx does not currently run automatically when the project is built, the documentation for docfx itself lies [here](https://dotnet.github.io/docfx/index.html)

> * Are we gonna use github actions to do it automatically to improve maintainability?

We don't strictly have to use github Actions as unfortunately it does require compiling Nitrox itself which in turn requires Subnautica.

NitroxServer/NitroxServer.csproj
Measurity: It compiles fine on windows when I removed the line. Did you confirm this behavior after you removed the line?

I ask because `Configuration` is set by an argument to the MSBuild CLI (called by the IDE), depending on which you selected in the IDE drop down. It should not be set by .props/.targets but can be used in conditionals to do something different (like optimize code in RELEASE config).

dartasen:
Closing for now, there's no need for that at the moment.

Worth looking when projects has been unified/cleaned up

PR #2096, opened 2023-12-01
Add basic save backup system

NinjaPedroX:
This PR focuses on adding a basic backup system to the Nitrox server, and that it would ideally be able to function without the world manager so that it can be included in v1.8.0.0 for use right away.

The first two commits were made on my copy of the v1.7.0.0 branch of the main repository, then cherry-picked them into this up-to-date branch. The third commit fixes the code for use with the world manager. I have done this in an attempt to make it easier for you guys to cherry pick the first two commits of this PR for use in v1.8.0.0 of Nitrox. Not sure if I did it effectively though.

The backup system functions as follows:
- Backups are kept within the save directory, in a subfolder called "Backups"
- Each backup is a .zip file with the title "Backup - yyyyMMddhhmmss". When the system detects a number of backup files higher than the max value defined in the config, it deletes the oldest one(s), unless there is no limit defined.
- There are two new server commands: `BackupCommand` and `AutoBackupCommand`.  The former performs a backup when it's run and the latter automatically backs up the save when the server autosaves.
- I have added two new config settings: `DisableAutoBackup` and `MaxBackups`. The former is there for a similar reason as the `DisableAutoSave` setting, while the latter defines the max number of backups the system keeps. The default is set to 10.
- Having save files that are corrupted no longer makes the server back them up and start anew; instead, it simply tells the user to restore one of their backups.

I think having this in the next release is essential since we get a lot of users that talk about their worlds being corrupted in the Discord server's help channel. This system will help them recover most of their progress and improve their overall experience as Nitrox continues to be more stable.

image


Drakonkinst:
After this feature is merged, would we want to wait until world manager is implemented to do further backup functionality (allowing players to load backups without having to just drag & drop files, maybe naming backups, etc.?) Is there a PR for the world manager or is this just something being discussed?

NinjaPedroX:
> After this feature is merged, would we want to wait until world manager is implemented to do further backup functionality (allowing players to load backups without having to just drag & drop files, maybe naming backups, etc.?) Is there a PR for the world manager or is this just something being discussed?

I did work on that a while back in [this PR](https://github.com/SubnauticaNitrox/Nitrox/pull/1824), but it was closed because we wanted to focus on improving the world manager design and implementing a cross-platform launcher (see [this PR](https://github.com/SubnauticaNitrox/Nitrox/pull/1848)).

The purpose of this PR is simply to implement the backend backup functionality to solve the issue I mentioned in the PR summary, but implementing the UI for it will come after the new launcher is finished.

NitroxServer/Server.cs
```diff
@@ -52,7 +52,14 @@ public Server(WorldPersistence worldPersistence, World world, ServerConfig serve
             saveTimer.AutoReset = true;
             saveTimer.Elapsed += delegate
             {
-                Save();
+                if (!serverConfig.DisableAutoBackup)
+                {
+                    BackUp();
```
Drakonkinst, line 57: I see that this method calls `Save()` after a conditional check. Why don't we just call `Save()` here either way? Is there a significant behavior change when checking `IsRunning`?

NinjaPedroX: The `BackUp();` function calls `Save();`, so doing that would mean saving the world twice (leading to log spam)

NitroxServer/Serialization/World/WorldPersistence.cs
```diff
@@ -86,6 +88,69 @@ internal bool Save(PersistedWorldData persistedData, string saveDir)
...
+                FileSystem.Instance.ZipFilesInDirectory(outZip, $"{outZip}.zip");
+                Directory.Delete(outZip, true);
+
+                // Check for total number of backups and prune as necessary
+                List<string> backups = new();
+                foreach (string file in Directory.EnumerateFiles(backupDir))
+                {
+                    FileInfo fileInfo = new(file);
+
+                    if (fileInfo.Extension != ".zip" || !fileInfo.Name.Contains($"Backup - "))
```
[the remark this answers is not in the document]
NinjaPedroX: We can definitely discuss that with the other devs in the developer channels of the Discord server (once you get the Jr-Alterra role).

I don't think allowing users to rename their backup files is necessary. Ideally, all they would have to do when their world corrupts is to backup up the most recent save file, and the extra backups are there in case some of the recently-made ones are also corrupted.

NitroxServer/Serialization/World/WorldPersistence.cs
```diff
@@ -86,6 +88,69 @@ internal bool Save(PersistedWorldData persistedData, string saveDir)
...
+                    int j = 0;
+                    for (int i = 0; i < backups.Count; i++)
+                    {
+                        if (i >= config.MaxBackups)
+                        {
+                            File.Delete(backups.ElementAt(j));
+                            j++;
+                        }
+                    }
+                }
```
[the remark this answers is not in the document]
NinjaPedroX: I will try this out. I was originally facing issues with it deleting the most recent file instead of the oldest.

NitroxServer/ConsoleCommands/AutoBackupCommand.cs
```diff
@@ -0,0 +1,42 @@
+using NitroxModel.DataStructures.GameLogic;
+using NitroxServer.ConsoleCommands.Abstract;
+using NitroxServer.ConsoleCommands.Abstract.Type;
+using NitroxServer.Serialization;
+using NitroxServer.Serialization.World;
+using System.IO;
+
+namespace NitroxServer.ConsoleCommands
+{
+    internal class AutoBackupCommand : Command
```
dartasen, line 10: I don't see any reason to have a dynamic autobackup enable/disable command, if the server owner wants to disable the backup behavior, he can change the config and restart the server

NitroxServer/ConsoleCommands/BackupCommand.cs
```diff
@@ -0,0 +1,18 @@
+﻿using NitroxModel.DataStructures.GameLogic;
+using NitroxServer.ConsoleCommands.Abstract;
+
+namespace NitroxServer.ConsoleCommands
+{
+    internal class BackupCommand : Command
+    {
+        public BackupCommand() : base("backup", Perms.ADMIN, "Creates a backup of the save")
```
dartasen, line 8: Same for that command, I don't find it useful

NitroxServer/Serialization/ServerConfig.cs
```diff
@@ -39,6 +41,21 @@ public int SaveInterval
...
         }
 
+        [PropertyDescription("Leave blank for no backup limit")]
+        public int MaxBackups
+        {
+            get => maxBackupsSetting;
+
+            set
+            {
+                if (!string.IsNullOrEmpty(value.ToString()))
```
dartasen, line 51: Is there a reason we're performing this check ?

NitroxServer/Serialization/World/WorldPersistence.cs
```diff
@@ -86,6 +86,65 @@ internal bool Save(PersistedWorldData persistedData, string saveDir)
             }
         }
 
+        public bool BackUp(string saveDir)
+        {
+            string backupDir = Path.Combine(saveDir, "Backups");
+            if (!Directory.Exists(backupDir))
+            {
+                Directory.CreateDirectory(backupDir);
```
dartasen, line 94: Directory.CreateDirectory already check for path before creating it

https://learn.microsoft.com/fr-fr/dotnet/api/system.io.directory.createdirectory?view=netframework-4.7.2

dartasen: The CreateDirectory should be performed in the try catch also, since it can throw an exception.

NinjaPedroX:
Will be closing this PR since I have included these changes within #1848 (https://github.com/SubnauticaNitrox/Nitrox/pull/1848/commits/f6e26c1cdffae2b0660a5c931219fa33026b3925). Please make any remaining suggestions over there.

PR #2099, opened 2023-12-18
Sync base hull

tornac1234:


NitroxClient/MonoBehaviours/BaseLeakManager.cs
```diff
@@ -0,0 +1,112 @@
...
+
+        idByRelativeCell[relativeCell] = cellId;
+
+        if (cellObject.TryGetComponent(out LiveMixin cellLiveMixin))
+        {
+            float deltaHealth = health - cellLiveMixin.health;
+            if (Mathf.Abs(deltaHealth) > 1)
+            {
+                // Useful part of BaseHullStrength.CrushDamageUpdate
+                this.Resolve<LiveMixinManager>().SyncRemoteHealth(cellLiveMixin, health, cellObject.position, DamageType.Pressure);
```
tornac1234, line 51: Will supress the packet sending from this

NitroxPatcher/Patches/Dynamic/BaseHullStrength_CrushDamageUpdate_Patch.cs
```diff
@@ -0,0 +1,58 @@
...
+        }
+        else
+        {
+            return false;
+        }
+    }
+
+    public static IEnumerable<CodeInstruction> Transpiler(IEnumerable<CodeInstruction> instructions)
+    {
+        return new CodeMatcher(instructions).MatchStartForward(new CodeMatch(OpCodes.Call, Reflect.Method(() => ErrorMessage.AddMessage(default))))
```
Jannify, line 33: Won't `CodeMatcher(instructions).End()` or `CodeMatcher(instructions).End().Advance(-1)` also work?

Jannify (changes requested):

NitroxClient/MonoBehaviours/BaseLeakManager.cs
```diff
@@ -0,0 +1,112 @@
...
+        {
+            return;
+        }
+
+        idByRelativeCell[relativeCell] = cellId;
+
+        if (cellObject.TryGetComponent(out LiveMixin cellLiveMixin))
+        {
+            float deltaHealth = health - cellLiveMixin.health;
+            if (Mathf.Abs(deltaHealth) > 1)
```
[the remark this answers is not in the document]
tornac1234: Health goes from 0 to 100

NitroxClient/MonoBehaviours/BaseLeakManager.cs
```diff
@@ -0,0 +1,112 @@
...
+                        {
+                            num = 2;
+                        }
+                        else if (baseHullStrength.totalStrength <= -2f)
+                        {
+                            num = 1;
+                        }
+                        if (baseHullStrength.crushSounds[num] != null)
+                        {
+                            Utils.PlayFMODAsset(baseHullStrength.crushSounds[num], cellObject, 20f);
```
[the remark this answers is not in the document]
tornac1234: Will add a comment

NitroxClient/MonoBehaviours/BaseLeakManager.cs
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    public Int3 Relative(Int3 absoluteCell)
+    {
+        return absoluteCell - @base.anchor;
+    }
+
+    public LeakRepaired RemoveLeakByAbsoluteCell(Int3 absoluteCell)
+    {
+        if (idByRelativeCell.TryGetValue(Relative(absoluteCell), out NitroxId cellId))
```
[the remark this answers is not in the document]
tornac1234: Relative cell "frequently" changes (when the base is modified) so it'd end up in clearing the said cache when required (more work for not much performance)

NitroxModel/DataStructures/GameLogic/Entities/Bases/BaseLeakEntity.cs
```diff
@@ -0,0 +1,48 @@
...
+    public BaseLeakEntity(float health, NitroxInt3 relativeCell, NitroxId leakId, NitroxId parentId)
+    {
+        Health = health;
+        RelativeCell = relativeCell;
+        Id = leakId;
+        ParentId = parentId;
+    }
+
+    /// <remarks>Used for deserialization</remarks>
+    public BaseLeakEntity(float health, NitroxInt3 relativeCell, NitroxId id, NitroxTechType techType, EntityMetadata metadata, NitroxId parentId, List<Entity> childEntities)
```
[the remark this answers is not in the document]
tornac1234: Entity has no base constructor

NitroxClient/MonoBehaviours/BaseLeakManager.cs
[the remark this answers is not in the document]
Jannify: I mean just a local field xD

NitroxClient/MonoBehaviours/BaseLeakManager.cs
```diff
@@ -0,0 +1,112 @@
...
+                        else if (baseHullStrength.totalStrength <= -2f)
+                        {
+                            num = 1;
+                        }
+                        if (baseHullStrength.crushSounds[num] != null)
+                        {
+                            Utils.PlayFMODAsset(baseHullStrength.crushSounds[num], cellObject, 20f);
+                        }
+                    }
+                    ErrorMessage.AddMessage(Language.main.GetFormat("BaseHullStrDamageDetected", baseHullStrength.totalStrength));
```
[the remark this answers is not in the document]
tornac1234: This would make a wall of text during initial sync. I'll move this line in the condition above so it only happens after initial sync and copy it to the base entity loading so that we still have at least one of those messages.

Jannify (approved):

tornac1234:
Rebased and retested, still works

NitroxPatcher/Patches/Dynamic/LiveMixin_AddHealth_Patch.cs
```diff
@@ -45,19 +51,61 @@ public static void Postfix(float __state, LiveMixin __instance, bool __runOrigin
...
+                    if (!processingPacket)
+                    {
+                        HandleBaseLeakRepair(baseCell, victim);
+                    }
+                    return false;
+            }
+        }
+
+        return !processingPacket;
+    }
```
Measurity, line 76: This logic seems to do nothing when `processingPacket` is false.

I recommend changing this to return early, or better yet, not call this function so it doesn't need the extra argument.

In addition, `Should*` prefixed methods should not have side effects like sending packets. Rename the method to something like `BroadcastHeal` or `HandleBroadcastHeal`.

NitroxClient/GameLogic/LiveMixinManager.cs
```diff
@@ -8,7 +7,7 @@ public class LiveMixinManager
 {
     private readonly SimulationOwnership simulationOwnership;
 
-    private bool processingRemoteHealthChange = false;
+    public bool ProcessingRemoteHealthChange { get; private set; } = false;
```
Measurity, line 10: Default for booleans is `false`, no need to init with `false`.
Recommend renaming to something more to-the-point:
```suggestion
    public bool IsRemoteHealthChanging { get; private set; }
```

NitroxServer/Communication/Packets/Processors/BuildingProcessor.cs
```diff
@@ -1,19 +1,24 @@
...
 namespace NitroxServer.Communication.Packets.Processors;
 
 public abstract class BuildingProcessor<T> : AuthenticatedPacketProcessor<T> where T : Packet
 {
     internal readonly BuildingManager buildingManager;
     internal readonly PlayerManager playerManager;
+    internal readonly EntitySimulation entitySimulation;
 
-    public BuildingProcessor(BuildingManager buildingManager, PlayerManager playerManager)
+    public BuildingProcessor(BuildingManager buildingManager, PlayerManager playerManager, EntitySimulation entitySimulation = null)
```
Measurity, line 17: Why `= null`? If it's required, add `Validate.NotNull(entitySimulation);` in the `ClaimBuildPiece` method.

tornac1234: It's null because many processors don't need this part but it's just bad design at this point so I'll just add it properly to all classes requiring this instance

tornac1234: The full constructor (with entitySimulation) should only be used when the processor needs to use ClaimBuildPiece but it'd need to make a 2nd constructor for the ones which don't need to use ClaimBuildPiece. So I made the 2 constructors into one.

NitroxPatcher/Patches/Dynamic/LiveMixin_AddHealth_Patch.cs
Measurity: Would this make sense?
```cs
private static bool TryBroadcastHeal(LiveMixin victim)
{
    foreach (MonoBehaviour monoBehaviour in victim.GetComponents<MonoBehaviour>())
    {
        switch (monoBehaviour)
        {
            case RadiationLeak radiationLeak:
                HandleRadiationLeakRepair(radiationLeak);
                return true;
            case BaseCell baseCell:
                HandleBaseLeakRepair(baseCell, victim);
                return true;
        }
    }
    return false;
}
```
and
```cs
public static void Postfix(float __state, LiveMixin __instance, bool __runOriginal)
{
    bool healthChanged = __state != __instance.health;
    if (!__runOriginal || !healthChanged)
    {
        return;
    }
    if (!Resolve<LiveMixinManager>().ProcessingRemoteHealthChange || TryBroadcastHeal(__instance))
    {
        return;
    }
    
    // Other code
```

tornac1234: No we can't. We need to trigger either of the Handle functions because they contain some important code to be executed (it's not all broadcast code). This is why I put the "processingremotehealthchange" check later

Measurity: Yea I messed up the boolean logic, should have been:
`if (Resolve<LiveMixinManager>().ProcessingRemoteHealthChange && TryBroadcastHeal(__instance))`

If processing, try broadcast heal and if it was handled do nothing more. If not processing, continue on.

tornac1234: But then you're missing the default case in which the object being healed is not a specific instance (so TryBroadcastHeal would return false in your situation) but still we're processing a packet (first condition returns true) then we're not returning while we should be.

Measurity: Not sure I get which default case you mean `switch` in `ShouldBroadcastHeal` has none.

If with default case you mean this code at the end of `Postfix`:
```cs
// Let others know if we have a lock on this entity
if (__instance.TryGetIdOrWarn(out NitroxId id) && Resolve<SimulationOwnership>().HasAnyLockType(id))
{
    Optional<EntityMetadata> metadata = Resolve<EntityMetadataManager>().Extract(__instance.gameObject);

    if (metadata.HasValue)
    {
        Resolve<Entities>().BroadcastMetadataUpdate(id, metadata.Value);
    }
}
```

Then it should still be called when either:
 - `Resolve<LiveMixinManager>().ProcessingRemoteHealthChange` is `false` (same as before, `ShouldBroadcastHeal` did nothing)
 - `ShouldBroadcastHeal(__instance, processingPacket)` failed to handle it before. (same as before but I inverted the return values).

tornac1234: Then I'll just mix everything in Postfix (the current last part of it is what I meant by "default case" and it'll now be handled by the "default" case itself in the switch)

Measurity: Sounds good, might be more verbose but it's hard to keep track otherwise as an outsider.

dartasen (approved):

Coding-Hen (approved):
LGTM

PR #2105, opened 2024-01-07
Add DenyOwnershipHand translation and refactor few MBs namespaces

Jannify:
Adds translation for DenyOwnershipHand 
Moves a few GUI files to the right place and adjusts namespaces


Measurity (approved):
LGTM 

~~but I'm not familiar with how to add new keys in Weblate~~ 
From Discord by Jannify: already done via Weblate as that's the source of truth to avoid merge conflicts.

dartasen (approved):

PR #2106, opened 2024-01-10
Sync leviathans gameplay

tornac1234:
**Creature sync:**
- [x] Sync Reaper Leviathans
- [x] Sync Sea Dragons
- [x] Sync Sea Treaders

**Creature behaviours:**
- [x] Sync leviathans attacking cyclops
- [x] Sync leviathans attacking a player or vehicle (seamoth or exosuit)

**Weapon sync:**
- [x] Sync statis rifle
- [x] Sync torpedos (better than current implementation)

**Other related issues:**
- [x] Fix player vitals not showing a creative player's name
- [x] Fix seamoth not spawning module inventory
- [x] Adds an InfectedMixin and a LiveMixin on remote players to ensure they're potential targets

TODO when everything is done:
~~- [ ] Ensure crash fishes can target other players~~ POSTPONED to another PR (will be short but required both this one and #2101 )


tornac1234:
Would require some tests :)

tornac1234:
There's still the following to be looked into:
- sometimes, move animation are chunky and you can see a small jump
- when another player gets eaten by a leviathan, they broadcast the related cinematic which is played on other players so a black screen appears as if they got eaten themselves by a leviathan (but they didn't)

NitroxPatcher/Patches/Dynamic/AggressiveWhenSeeTarget_ScanForAggressionTarget_Patch.cs
```diff
@@ -0,0 +1,64 @@
...
+    internal static readonly MethodInfo TARGET_METHOD = Reflect.Method((AggressiveWhenSeeTarget t) => t.ScanForAggressionTarget());
+
+    public static bool Prefix(AggressiveWhenSeeTarget __instance)
+    {
+        if (!__instance.TryGetNitroxId(out NitroxId creatureId) ||
+            Resolve<SimulationOwnership>().HasAnyLockType(creatureId))
+        {
+            return true;
+        }
+
```
Jannify, line 25:
```suggestion
```

NitroxPatcher/Patches/Dynamic/AttackCyclops_OnCollisionEnter_Patch.cs
```diff
@@ -0,0 +1,50 @@
...
+    public static bool ShouldCollisionAnnoyCreature(Collision collision)
+    {
+        return AttackCyclops_UpdateAggression_Patch.IsTargetAValidInhabitedCyclops(collision.gameObject);
+    }
+
+    public static bool Prefix(AttackCyclops __instance)
+    {
+        return !__instance.TryGetNitroxId(out NitroxId creatureId) ||
+               Resolve<SimulationOwnership>().HasAnyLockType(creatureId);
+    }
```
Jannify, line 49: Would put prefix to top

NitroxPatcher/Patches/Dynamic/SeamothTorpedoWhirlpool_Register_Patch.cs
```diff
@@ -0,0 +1,19 @@
...
+namespace NitroxPatcher.Patches.Dynamic;
+
+/// <summary>
+/// Prevents remote players from being drawn by torpedo's whirlpools
+/// </summary>
+public sealed partial class SeamothTorpedoWhirlpool_Register_Patch : NitroxPatch, IDynamicPatch
+{
+    private static readonly MethodInfo TARGET_METHOD = Reflect.Method((SeamothTorpedoWhirlpool t) => t.Register(default, ref Reflect.Ref<Rigidbody>.Field));
+
+    public static bool Prefix(Collider other, ref bool __result)
```
Jannify, line 15: `__result` is not used

NitroxPatcher/Patches/Dynamic/StasisSphere_OnHit_Patch.cs
```diff
@@ -0,0 +1,35 @@
...
+using NitroxClient.GameLogic;
+using NitroxModel.DataStructures.Unity;
+using NitroxModel.Helper;
+using NitroxModel.Packets;
+using NitroxModel_Subnautica.DataStructures;
+
+namespace NitroxPatcher.Patches.Dynamic;
+
+/// <summary>
+/// Prevents remote stasis sphere from triggering hit effets (they're triggered with packets)
```
Jannify, line 12: I find the wording here a bit misleading as we don't skip the base-method din the prefix

NitroxPatcher/Patches/Dynamic/StasisSphere_Shoot_Patch.cs
```diff
@@ -0,0 +1,26 @@
...
+using NitroxModel_Subnautica.DataStructures;
+using UnityEngine;
+
+namespace NitroxPatcher.Patches.Dynamic;
+
+public sealed partial class StasisSphere_Shoot_Patch : NitroxPatch, IDynamicPatch
+{
+    private static readonly MethodInfo TARGET_METHOD = Reflect.Method((StasisSphere t) => t.Shoot(default, default, default, default, default));
+
+    private static ushort LocalPlayerId => Resolve<LocalPlayer>().PlayerId;
```
Jannify, line 15: Don't know if this is needed. As it's technically a method nothing is cached

NitroxPatcher/Patches/Dynamic/Vehicle_TorpedoShot_Patch.cs
```diff
@@ -0,0 +1,41 @@
...
+using NitroxModel_Subnautica.DataStructures;
+using UnityEngine;
+
+namespace NitroxPatcher.Patches.Dynamic;
+
+public sealed partial class Vehicle_TorpedoShot_Patch : NitroxPatch, IDynamicPatch
+{
+    private static readonly MethodInfo TARGET_METHOD = Reflect.Method(() => Vehicle.TorpedoShot(default, default, default));
+
+    public static bool Prefix(Vehicle __instance, ref bool __result, ItemsContainer container, TorpedoType torpedoType, Transform muzzle)
```
Jannify, line 16: Locking at the Sn code it does look like a transpiler would be very easy. Or is this file not up to date?

Jannify (changes requested):

NitroxPatcher/Patches/Dynamic/StasisSphere_OnHit_Patch.cs
tornac1234: Fixed the patch to actually do what the summary explained.

NitroxPatcher/Patches/Dynamic/StasisSphere_Shoot_Patch.cs
tornac1234: It was just there for reading purposes

NitroxClient/Helpers/NitroxEntityExtensions.cs
```diff
@@ -85,7 +86,8 @@ public static bool TryGetNitroxId(this Component component, out NitroxId nitroxI
         }
         if (!component.TryGetComponent(out NitroxEntity nitroxEntity))
         {
-            Log.Warn($"[{filePath[(filePath.LastIndexOf("\\", StringComparison.Ordinal) + 1)..^2] + methodName}():L{lineNumber}] Couldn't find an id on {component.GetFullHierarchyPath()}");
+            Action<string> warnMethod = warnOnce ? Log.WarnOnce : Log.Warn;
+            warnMethod.Invoke($"[{filePath[(filePath.LastIndexOf("\\", StringComparison.Ordinal) + 1)..^2] + methodName}():L{lineNumber}] Couldn't find an id on {component.GetFullHierarchyPath()}");
```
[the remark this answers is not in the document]
tornac1234: I don't really like this method for what it does only give general context ("a certain gameobject happened not be marked with an id when it was required"). To keep it, we could just create a static cache in NitroxEntityExtensions with already printed GameObject paths to avoid repeating those warns (it would be absolutely useless to have this repeatedly logged).

NitroxPatcher/Patches/Dynamic/AttackCyclops_UpdateAggression_Patch.cs
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        // Has the local player inside
+        if (Player.main && Player.main.currentSub && Player.main.currentSub.gameObject == targetObject)
+        {
+            return true;
+        }
+
+        // Has a remote player inside
+        return targetObject.GetComponentInChildren<RemotePlayerIdentifier>(true);
```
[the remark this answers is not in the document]
tornac1234: Two cases here:
- Remote player is in Cyclops' root
- Remote player is FAR in the hierarchy, parented to the driver seat

Current issue: remote player is never unparented from the driver seat even after leaving it (???) so this would need a fix in another PR I guess

Possible solutions (that would come in the same PR mentionned above:
- have a MB (either MultiplayerCyclops or a new one) store in a list all players (local and remote) which are currently in the cyclops, and use this list instead of the components search
- One loop in the first children layer, and one in the driver seat layer

tornac1234:
The latest Sea Dragons sync code wasn't tested because it's really hard without #2113

NitroxServer/Communication/Packets/Processors/SeaDragonAttackTargetProcessor.cs
```diff
@@ -0,0 +1,13 @@
+using NitroxModel.Packets;
+using NitroxServer.Communication.Packets.Processors.Abstract;
+using NitroxServer.GameLogic;
+using NitroxServer.GameLogic.Entities;
+
+namespace NitroxServer.Communication.Packets.Processors;
+
+public class SeaDragonAttackTargetProcessor : TransmitIfCanSeePacketProcessor<SeaDragonAttackTarget>
```
dartasen, line 8: Can be simplified using primary constructors
```csharp
public class SeaDragonAttackTargetProcessor(
    PlayerManager playerManager,
    EntityRegistry entityRegistry
) : TransmitIfCanSeePacketProcessor<SeaDragonSwatAttack>(playerManager, entityRegistry)
```

I don't put comments on the others processors but same idea (SeaDragonSwatAttackProcessor, SeaDragonGrabExosuitProcessor, RangedAttackLastTargetUpdateProcessor, ...)

NitroxServer/Communication/Packets/Processors/AggressiveWhenSeeTargetChangedProcessor.cs
```diff
@@ -0,0 +1,13 @@
+using NitroxModel.Packets;
+using NitroxServer.Communication.Packets.Processors.Abstract;
+using NitroxServer.GameLogic;
+using NitroxServer.GameLogic.Entities;
+
+namespace NitroxServer.Communication.Packets.Processors;
```
dartasen, line 6: Maybe we could regroup the different processors related to creature sync under a same folder ?
Thoughts on it ?

NitroxPatcher/Patches/Dynamic/SeaDrargon_ReleaseExosuit_Patch.cs
```diff
@@ -0,0 +1,22 @@
+using System.Reflection;
+using NitroxClient.MonoBehaviours;
+using NitroxModel.Helper;
+
+namespace NitroxPatcher.Patches.Dynamic;
+
+/// <summary>
+/// Enables exosuits's position sync when they're released from sea dragons
+/// </summary>
+public sealed partial class SeaDrargon_ReleaseExosuit_Patch : NitroxPatch, IDynamicPatch
```
dartasen, line 10: small typo Dragon

NitroxClient/GameLogic/RemotePlayer.cs
```diff
@@ -17,6 +17,10 @@ namespace NitroxClient.GameLogic
 {
     public class RemotePlayer : INitroxPlayer
     {
+        /// <summary>
+        /// Marks <see cref="Player.mainObject"/> and every <see cref="Body"/> so they can be precisely queried (e.g. by sea dragons).
+        /// </summary>
+        public const EcoTargetType PLAYER_ECO_TARGET_TYPE = (EcoTargetType)5050;
```
dartasen, line 23: Looking at the EcoTargetType enum, not sure to fully understand where the 5050 is coming from, maybe we can add some details in the comments

NitroxPatcher/Patches/Dynamic/RangedAttackLastTarget_StartCasting_Patch.cs
```diff
@@ -0,0 +1,34 @@
...
+/// Prevents non simulating players from running locally <see cref="RangedAttackLastTarget.StartCasting"/>.
+/// Broadcasts this event on the simulating player.
+/// </summary>
+public sealed partial class RangedAttackLastTarget_StartCasting_Patch : NitroxPatch, IDynamicPatch
+{
+    internal static readonly MethodInfo TARGET_METHOD = Reflect.Method((RangedAttackLastTarget t) => t.StartCasting(default));
+
+    public static void Prefix(RangedAttackLastTarget __instance)
+    {
+        if (!Resolve<AI>().IsCreatureWhitelisted(__instance.creature) ||
```
dartasen, line 20: This `if` is really hard to read.It should be splitted imo. The game won't go slower for few more IL lines generated

NitroxPatcher/Patches/Dynamic/RangedAttackLastTarget_StartCharging_Patch.cs
```diff
@@ -0,0 +1,34 @@
...
+/// Prevents non simulating players from running locally <see cref="RangedAttackLastTarget.StartCasting"/>.
+/// Broadcasts this event on the simulating player.
+/// </summary>
+public sealed partial class RangedAttackLastTarget_StartCharging_Patch : NitroxPatch, IDynamicPatch
+{
+    internal static readonly MethodInfo TARGET_METHOD = Reflect.Method((RangedAttackLastTarget t) => t.StartCharging(default));
+
+    public static void Prefix(RangedAttackLastTarget __instance)
+    {
+        if (!Resolve<AI>().IsCreatureWhitelisted(__instance.creature) ||
```
dartasen, line 20: same here

NitroxPatcher/Patches/Dynamic/AttackCyclops_UpdateAggression_Patch.cs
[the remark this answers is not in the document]
Jannify: If it's only the two cases could we either check only all direct children of the cyclops and from the driver seat down?

NitroxPatcher/Patches/Dynamic/SeaDragonAggressiveTowardsSharks_GetAggressionTarget_Patch.cs
```diff
@@ -0,0 +1,40 @@
...
+        }
+
+        // To redirect the call to base.GetAggressionTarget(), we ensure the if is skipped in the original method
+        float timeLastPlayerAttack = __instance.timeLastPlayerAttack;
+        __instance.timeLastPlayerAttack = Time.time;
+        
+        __result = __instance.GetAggressionTarget();
+
+        __instance.timeLastPlayerAttack = timeLastPlayerAttack;
+        return false;
```
Jannify, line 38: Will this now also target the local player? I don't know if the local player is included in `RemotePlayer.PLAYER_ECO_TARGET_TYPE` so maybe this comment is pointless.

NitroxPatcher/Patches/Dynamic/SeaDragonMeleeAttack_OnTouchFront_Patch.cs
```diff
@@ -0,0 +1,117 @@
...
+     * 2nd injection:
+     *     global::Utils.PlayEnvSound(this.attackSound, collider.transform.position, 20f);
+     * }
+     * component3.gameObject.GetComponent<LiveMixin>().TakeDamage(this.biteDamage, default(Vector3), DamageType.Normal, base.gameObject);
+     * BroadcastSeaDragonAttackTarget(this, target);       <---- INSERTED LINE
+     * 
+     * 3rd injection:
+     * Exosuit component4 = target.GetComponent<Exosuit>();
+     * BroadcastSeaDragonAttackRemotePlayer(this, target);            <---- INSERTED LINE
+     * if (component4 != null)
```
Jannify, line 52: This seems a little to early, no? Broadcasting before all checks are done if the exosuit can even be grabed

NitroxPatcher/Patches/Dynamic/SeaDragonMeleeAttack_SwatAttack_Patch.cs
```diff
@@ -0,0 +1,49 @@
...
+        // 1st injection
+        return new CodeMatcher(instructions).End()
+                                            .InsertAndAdvance([
+                                                new CodeInstruction(OpCodes.Ldarg_0),
+                                                new CodeInstruction(OpCodes.Ldarg_1),
+                                                new CodeInstruction(OpCodes.Ldarg_2),
+                                                new CodeInstruction(OpCodes.Call, Reflect.Method(() => BroadcastSwatAttack(default, default, default)))
+                                            ])
+                                            .InstructionEnumeration();
+    }
```
Jannify, line 39: Are you sure this works? won't this add the instructions after the ret?

Jannify (changes requested):

NitroxClient/GameLogic/BulletManager.cs
```diff
@@ -0,0 +1,167 @@
...
+
+    public void TorpedoHit(NitroxId bulletId, Vector3 position, Quaternion rotation)
+    {
+        if (NitroxEntity.TryGetComponentFrom(bulletId, out SeamothTorpedo torpedo))
+        {
+            torpedo.tr.position = position;
+            torpedo.tr.rotation = rotation;
+            torpedo.OnHit(default);
+            torpedo.Deactivate();
+        }
```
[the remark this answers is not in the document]
tornac1234: On the local player, the torpedo might have already exploded while the packet is received with latency. i'll add a comment explaining it

NitroxServer/Communication/Packets/Processors/Abstract/TransmitIfCanSeePacketProcessor.cs
```diff
@@ -0,0 +1,59 @@
...
+        List<Entity> entities = [];
+        foreach (NitroxId entityId in entityIds)
+        {
+            if (entityRegistry.TryGetEntityById(entityId, out Entity entity))
+            {
+                entities.Add(entity);
+            }
+            else
+            {
+                return;
```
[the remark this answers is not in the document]
tornac1234: This would probably be a useless spam giving no other info that some creatures are present on a client but not on the server. As we're aware of this issue, it's not required that we log anything concerning it

NitroxPatcher/Patches/Dynamic/SeaDragonAggressiveTowardsSharks_GetAggressionTarget_Patch.cs
tornac1234: This target type is applied to the local player too in ``PlayerInitialSyncProcessor.SetPlayerGameObjectId``, so yes the local player will still be targetted

NitroxPatcher/Patches/Dynamic/SeaDragonMeleeAttack_OnTouchFront_Patch.cs
tornac1234: At this point, all that counts is whether or not ``component4`` is null, which I check in ``BroadcastSeaDragonAttackRemotePlayer`` so the broadcast is fine at this place

NitroxPatcher/Patches/Dynamic/SeaDragonMeleeAttack_SwatAttack_Patch.cs
tornac1234: Yes I am. It does add instructions before the ret.

Measurity (commented):
Looks good CW but I need to do in-game testing.

dartasen:
Did a playtest with Ninja & Tornac : 
* Seems to work well
* Waiting on a commit to remove debug messages for actions
* (Optional) Waiting on a commit for cyclops beacon death sync

NitroxClient/GameLogic/AI.cs
```diff
@@ -1,36 +1,170 @@
...
+    public void BroadcastNewAction(NitroxId creatureId, Creature creature, CreatureAction newAction)
+    {
+        if (!syncedCreatureWhitelist.Contains(creature.GetType()))
         {
-            this.packetSender = packetSender;
+            return;
         }
 
-        public void CreatureActionChanged(NitroxId id, CreatureAction newAction)
+        ErrorMessage.AddMessage($"[SEND] synced action: {newAction.GetType().FullName}");
```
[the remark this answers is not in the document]
dartasen: See my review message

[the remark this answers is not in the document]
tornac1234: Everything is "under control" right now so we now exactly what can and should be broadcasted so we don't need error messages at this layer

NitroxClient/Communication/Packets/Processors/CreatureActionProcessor.cs
```diff
@@ -0,0 +1,20 @@
...
+    private readonly AI ai;
+
+    public CreatureActionProcessor(AI ai)
+    {
+        this.ai = ai;
+    }
+
+    public override void Process(CreatureActionChanged packet)
+    {
+        ai.CreatureActionChanged(packet.CreatureId, packet.CreatureActionType);
```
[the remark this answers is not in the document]
tornac1234: For now yes this is the global catch, while every other specific action has its own stuff synced somewhere else

NitroxClient/GameLogic/RemotePlayer.cs
```diff
@@ -385,10 +394,20 @@ public void UpdateEquipmentVisibility(List<TechType> equippedItems)
...
+    /// Makes the RemotePlayer recognizable as an obstacle for buildings, and as a target for creatures
     /// </summary>
     private void SetupBody()
     {
+        // set as a target for reapers
+        EcoTarget sharkEcoTarget = Body.AddComponent<EcoTarget>();
+        sharkEcoTarget.SetTargetType(EcoTargetType.Shark);
+
+        EcoTarget playerEcoTarget = Body.AddComponent<EcoTarget>();
+        playerEcoTarget.SetTargetType(PLAYER_ECO_TARGET_TYPE);
```
Coding-Hen, line 406: I assume this sets a value that the reapers will go for. If two players have the same value does this cause us an issue? Will it only attack one person or just say the nearest one?

Coding-Hen (approved):
Other than the one comment looks good to me

NitroxClient/GameLogic/RemotePlayer.cs
dartasen: I think it's more like a category, that the repear will look after. Rather than an ID

dartasen (approved):

NitroxClient/GameLogic/RemotePlayer.cs
Coding-Hen: Perfect sounds good to go then

Nitrox.Test/Patcher/Patches/PatchesTranspilerTest.cs
```diff
@@ -122,6 +142,12 @@ public void AllPatchesTranspilerSanity(Type patchClassType, int ilDifference)
         }
 
         List<CodeInstruction> transformedIl = (transpilerMethod.Invoke(null, injectionParameters.ToArray()) as IEnumerable<CodeInstruction>)?.ToList();
+        
+        if (logInstructions)
+        {
+            Console.WriteLine(transformedIl.ToPrettyString());
+        }
```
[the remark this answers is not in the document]
tornac1234: It's way easier to just add a true/false on the line where you transpiler test is declared (which is the definitely a better place as it is a test and tests are supposed to let you know about the insights), than to modify the transpiler to be able to access the transformed IL before returning it (while in most cases we are returning directly the whole thing)

NitroxClient/GameLogic/RemotePlayer.cs
```diff
@@ -484,6 +503,41 @@ static void CopyEmitter(FMOD_CustomEmitter src, FMOD_CustomEmitter dst)
...
+        LiveMixin.health = 100;
+        // We set the remote player to invincible because we only want this component to be detectable but not to work
+        LiveMixin.invincible = true;
+    }
+
+    public void UpdateHealthAndInfection(float health, float infection)
+    {
+        if (LiveMixin)
+        {
+            LiveMixin.health = health;
```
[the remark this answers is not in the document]
tornac1234: No this is perfectly fine and relevant. If a player is dead, we don't want the leviathan to be able to target it before it respawns.

NitroxClient/Communication/Packets/Processors/SeaDragonAttackTargetProcessor.cs
```diff
@@ -0,0 +1,55 @@
...
+        {
+            collider = Player.mainCollider;
+        }
+        else
+        {
+            return;
+        }
+
+        seaDragonMeleeAttack.timeLastBite = Time.time;
+        if (seaDragonMeleeAttack.attackSound)
```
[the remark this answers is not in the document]
tornac1234: This edge case doesn't exist because at all times a collider will exist on the Player and on the remote players

NitroxPatcher/Patches/Dynamic/AttackCyclops_UpdateAggression_Patch.cs
```diff
@@ -0,0 +1,119 @@
...
+                                                new CodeInstruction(OpCodes.Ldarg_0),
+                                                new CodeInstruction(OpCodes.Call, Reflect.Method(() => FindClosestCyclopsNoiseManagerIfAny(default)))
+                                            ]).InstructionEnumeration();
+    }
+
+    public static bool Prefix(AttackCyclops __instance)
+    {
+        return !__instance.TryGetNitroxId(out NitroxId creatureId) ||
+               Resolve<SimulationOwnership>().HasAnyLockType(creatureId);
+    }
```
Jannify, line 57: Would move the prefix before the transpiler

NitroxPatcher/Patches/Dynamic/AttackCyclops_UpdateAggression_Patch.cs
```diff
@@ -0,0 +1,119 @@
...
+        CyclopsNoiseManager closest = null;
+        foreach (KeyValuePair<NitroxCyclops, float> cyclopsEntry in NitroxCyclops.ScaledNoiseByCyclops)
+        {
+            if (cyclopsEntry.Key.Pawns.Count == 0)
+            {
+                continue;
+            }
+
+            // Calculations from the "if (closestDecoy != null || cyclopsNoiseManager != null)" part
+            float distance = Vector3.Distance(cyclopsEntry.Key.transform.position, attackCyclops.transform.position);
```
Jannify, line 98: If this is really performance critical we could investigate if the usage of the squared distance (like sqrMagnitude vs magnitude but for distance, so the pc doesn't have to compute <99 distances) is possible.

NitroxPatcher/Patches/Dynamic/AttackCyclops_UpdateAggression_Patch.cs
```diff
@@ -0,0 +1,119 @@
...
+            
+            if (distance < cyclopsEntry.Value && distance < minDistance)
+            {
+                minDistance = distance;
+                closest = cyclopsEntry.Key.GetComponent<CyclopsNoiseManager>();
```
Jannify, line 103: I would only save  `cyclopsEntry.Key` and do the GetComponent part later so we don't do unnecessary heavy unity calls if `NitroxCyclops.ScaledNoiseByCyclops` has an unfortunate sorting.

Jannify (changes requested):

Nitrox.Test/Patcher/Patches/Dynamic/CyclopsSonarButton_Update_PatchTest.cs
```diff
@@ -1,26 +0,0 @@
-using System.Collections.ObjectModel;
-using System.Reflection.Emit;
-using HarmonyLib;
-using NitroxPatcher.Patches.Dynamic;
-using NitroxTest.Patcher;
-
-namespace Nitrox.Test.Patcher.Patches.Dynamic;
-
-[TestClass]
-public class CyclopsSonarButton_Update_PatchTest
```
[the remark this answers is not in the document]
tornac1234: At this point we shouldn't care anymore. We are writing transpilers for a game that will likely not get updated again. If an update comes in the future we'll have to do actual script comparisons to know if something happened, and not just hope nothing broke

NitroxPatcher/Patches/Dynamic/CyclopsDestructionEvent_DestroyCyclops_Patch.cs
```diff
@@ -1,27 +1,24 @@
...
 public sealed partial class CyclopsDestructionEvent_DestroyCyclops_Patch : NitroxPatch, IDynamicPatch
 {
     private static readonly MethodInfo TARGET_METHOD = Reflect.Method((CyclopsDestructionEvent t) => t.DestroyCyclops());
 
     public static void Prefix(CyclopsDestructionEvent __instance)
     {
-        if (__instance.TryGetNitroxId(out NitroxId nitroxId))
-        {
-            Resolve<SimulationOwnership>().StopSimulatingEntity(nitroxId);
-            EntityPositionBroadcaster.StopWatchingEntity(nitroxId);
```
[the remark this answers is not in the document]
tornac1234: Since we are still simulating the cyclops after it dies, it makes no sense to seamingly drop simulation of it

NitroxPatcher/TranspilerHelper.cs
```diff
@@ -134,4 +139,24 @@ public static CodeMatcher LogInstructions(this CodeMatcher matcher)
...
+    /// <summary>
+    /// Replaces first use of <see cref="Time.deltaTime"/> by <see cref="TimeManager.DeltaTime"/>.
+    /// Doesn't change <see cref="CodeMatcher.Pos"/>.
+    /// </summary>
+    public static CodeMatcher ReplaceDeltaTime(this CodeMatcher codeMatcher)
+    {
+        int pos = codeMatcher.Pos;
+        return codeMatcher.MatchStartForward(new CodeMatch(OpCodes.Call, DELTA_TIME_MATCHING_FIELD))
+                   .SetOperandAndAdvance(DELTA_TIME_INSERTED_METHOD)
+                   .Advance(pos - codeMatcher.Pos);
```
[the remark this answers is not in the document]
tornac1234: You're right + this is misleading

NitroxPatcher/Patches/Dynamic/Floater_FixedUpdate_Patch.cs
```diff
@@ -0,0 +1,23 @@
...
+    public static bool Prefix(Floater __instance)
+    {
+        if (!__instance.fixedJoint || !__instance.fixedJoint.connectedBody ||
+            !__instance.fixedJoint.connectedBody.TryGetNitroxId(out NitroxId jointId))
+        {
+            return true;
+        }
+
+        // If target has an id, we only apply the FixedUpdate when we have a lock type on it (true)
+        return Resolve<SimulationOwnership>().HasAnyLockType(jointId);
```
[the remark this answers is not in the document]
tornac1234: For now I'll restrict this check for vehicles only because it was really concerning but in the future we'll do something about it

NitroxPatcher/Patches/Dynamic/RangedAttackLastTarget_StartCharging_Patch.cs
```diff
@@ -0,0 +1,55 @@
...
+        BroadcastRangedAttack(__instance, RangedAttackLastTargetUpdate.ActionState.CHARGING);
+    }
+
+    /// <summary>
+    /// Broadcasts a range attack CHARGING or CASTING state if the attacking creature is whitelisted, valid and if the target is valid.
+    /// </summary>
+    /// <returns>
+    /// true if the broadcast was done (all conditions met)
+    /// </returns>
+    internal static bool BroadcastRangedAttack(RangedAttackLastTarget attackBehaviour, RangedAttackLastTargetUpdate.ActionState actionState)
```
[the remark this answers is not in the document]
tornac1234: It used to be useful (for logs) but it's no longer so I'll remove the bool

NitroxPatcher/Patches/Dynamic/AttackCyclops_UpdateAggression_Patch.cs
tornac1234: According to this article, https://dev.to/iamscottcab/exploring-the-myth-calculating-square-root-is-expensive-44ka, in Unity 2018, sqrt was way faster than power

NitroxServer/ConsoleCommands/PlayerCommand.cs
```diff
@@ -0,0 +1,61 @@
...
+using NitroxServer.ConsoleCommands.Abstract;
+using NitroxServer.ConsoleCommands.Abstract.Type;
+using NitroxServer.GameLogic;
+using NitroxServer.GameLogic.Entities;
+
+namespace NitroxServer.ConsoleCommands;
+
+internal class PlayerCommand : Command
+{
+    private readonly Lazy<PlayerManager> playerManager = new(NitroxServiceLocator.LocateService<PlayerManager>);
```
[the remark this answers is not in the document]
tornac1234: For some reason, due to how early those commands are initialized, the instance they use (e.g. PlayerManager) aren't initialized yet by the loading World so they'll give a wrong reference.

NitroxServer/Communication/Packets/Processors/Abstract/TransmitIfCanSeePacketProcessor.cs
```diff
@@ -0,0 +1,49 @@
...
+    {
+        this.playerManager = playerManager;
+        this.entityRegistry = entityRegistry;
+    }
+
+    /// <summary>
+    /// Transmits the provided <paramref name="packet"/> to all other players (excluding <paramref name="senderPlayer"/>)
+    /// who can see (<see cref="Player.CanSee"/>) entities corresponding to the provided <paramref name="entityIds"/> only if all those entities are registered.
+    /// </summary>
+    public void TransmitIfCanSeeEntities(Packet packet, Player senderPlayer, List<NitroxId> entityIds)
```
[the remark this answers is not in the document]
tornac1234: You are very right, the solution will be to let the server know about which "out of cell entities" every player can see, and to add a check in Player.CanSee to use the list of "out of cell entities" which the player currently sees.

NitroxServer/ConsoleCommands/PlayerCommand.cs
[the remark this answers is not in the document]
Measurity: Will be fixed here: https://github.com/SubnauticaNitrox/Nitrox/issues/2166

NitroxPatcher/Patches/Dynamic/AttackCyclops_UpdateAggression_Patch.cs
Measurity: The article doesn't test the implementation Unity uses but just `i * i` vs `Sqrt(i)`. But actual Unity does a mul always before the sqrt check so `Vector3.Distance` should be worse than `vector.sqrMagnitude`. 

Though with modern hardware I'm not sure how slow sqrt actually is. So I would be fine with keeping it as is.

### Unity distance code
```cs
public static float Distance(Vector3 a, Vector3 b)
{
    float num1 = a.x - b.x;
    float num2 = a.y - b.y;
    float num3 = a.z - b.z;
    return (float) Math.Sqrt((double) num1 * (double) num1 + (double) num2 * (double) num2 + (double) num3 * (double) num3);
}
```

Measurity: webhook test

dartasen: Sqrt is indeed highly optimized nowadays, since it's used for cryptography, IA, game dev, ...

Most of our processors have dedicated instructions to quickly compute square root, without speaking of GPU

tornac1234: In any way if we remove the sqrt of the Distance, we'll have to add a square to the distance it's compared to. So we better keep it this way

Jannify (approved):
LGTM CW

PR #2107, opened 2024-01-10
Add basic sync for PvP

tornac1234:
- [x] Basic sync for PvP (only allowed with knifes and heatblades for now)
- [x] ~~Maybe: add sync for propulsion and repulsion cannon~~

To be tested (and merged) upon #2106 (won't work without it)


ywgATustcbbs:
is there really one who likes pvp in Subnautica? if one wants pvp, why not go ahead to play pvp games? there are tons of such games. i don't think core players of Subnautica wants to pvp.

NinjaPedroX:
> is there really one who likes pvp in Subnautica? if one wants pvp, why not go ahead to play pvp games? there are tons of such games. i don't think core players of Subnautica are interested in pvp.

Most multiplayer games have ways for players to damage each other. Even if it's not the main focus of the game, it gives players more options to have fun with their friends. Having a last man standing battle within a large base or picking people up and repelling them far away with the propulsion cannon is the kind of thing people (like myself) like to do with our friends. I don't think the argument of "if you want to do PvP, play a PvP game" is a good reason to stop this.

Besides, you can just turn PvP off in the configs if you don't want it.

tornac1234:
Waiting for #2106 to be merged, then this will get some tests and will be ready to be merged

Jannify (changes requested):

NitroxClient/Communication/Packets/Processors/PvPAttackProcessor.cs
```diff
@@ -0,0 +1,15 @@
+using NitroxClient.Communication.Packets.Processors.Abstract;
+using NitroxModel.Packets;
+
+namespace NitroxClient.Communication.Packets.Processors;
+
+public class PvPAttackProcessor : ClientPacketProcessor<PvPAttack>
+{
+    public override void Process(PvPAttack packet)
+    {
+        if (Player.main && Player.main.liveMixin)
```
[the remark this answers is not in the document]
tornac1234: If local player isn't instantiated, or its live mixin isn't found, then it can only mean that the game is an early loading state/in an unloading state in whih case it's not necessary to log the failure

NitroxPatcher/Patches/Dynamic/LiveMixin_TakeDamage_Patch.cs
```diff
@@ -46,4 +48,43 @@ public static void Postfix(float __state, LiveMixin __instance, bool __runOrigin
...
+            return false;
+        }
+
+        // The health change check must happen after the PvP one
+        return healthChanged;
+    }
+
+    private static void HandlePvP(RemotePlayer remotePlayer, GameObject dealer, float damage)
+    {
+        if (dealer == Player.mainObject && Inventory.main.GetHeldObject())
```
Measurity, line 72: Would invert the if statement to reduce cognitive load.

NitroxServer/Communication/Packets/Processors/PvPAttackProcessor.cs
```diff
@@ -0,0 +1,37 @@
...
+
+    public override void Process(PvPAttack packet, Player player)
+    {
+        if (serverConfig.PvPEnabled &&
+            playerManager.TryGetPlayerById(packet.TargetPlayerId, out Player targetPlayer) &&
+            damageMultiplierByType.TryGetValue(packet.Type, out float multiplier))
+        {
+            packet.Damage *= multiplier;
+            targetPlayer.SendPacket(packet);
+        }
```
Measurity, line 35: The if statement is doing too much so would split it up.

```suggestion
        if (!serverConfig.PvPEnabled)
        {
            return;
        }
        if (!playerManager.TryGetPlayerById(packet.TargetPlayerId, out Player targetPlayer))
        {
            return;
        }
        if (!damageMultiplierByType.TryGetValue(packet.Type, out float multiplier))
        {
            return;
        }
        packet.Damage *= multiplier;
        targetPlayer.SendPacket(packet);
```

Measurity (changes requested):
CW review

Measurity:
Works for me so after code changes are done I'll approve.

Measurity (approved):
:+1:

Measurity (approved):
Tests still work so LGTM

NitroxServer/Communication/Packets/Processors/PvPAttackProcessor.cs
```diff
@@ -0,0 +1,45 @@
...
+
+public class PvPAttackProcessor : AuthenticatedPacketProcessor<PvPAttack>
+{
+    private readonly ServerConfig serverConfig;
+    private readonly PlayerManager playerManager;
+
+    // TODO: In the future, do a whole config for damage sources
+    private static readonly Dictionary<PvPAttack.AttackType, float> damageMultiplierByType = new()
+    {
+        { PvPAttack.AttackType.KnifeHit, 0.5f },
```
Coding-Hen, line 17: Why do we need a multiplier and not just use the base value of the weapon?

NitroxServer/Communication/Packets/Processors/PvPAttackProcessor.cs
```diff
@@ -0,0 +1,45 @@
...
+using NitroxServer.Serialization;
+
+namespace NitroxServer.Communication.Packets.Processors;
+
+public class PvPAttackProcessor : AuthenticatedPacketProcessor<PvPAttack>
+{
+    private readonly ServerConfig serverConfig;
+    private readonly PlayerManager playerManager;
+
+    // TODO: In the future, do a whole config for damage sources
```
Coding-Hen, line 14: Should we create a ticket to add in the other sources if we know what they are so they don't get forgotten?

Coding-Hen (changes requested):

NitroxServer/Communication/Packets/Processors/PvPAttackProcessor.cs
Measurity: I think it makes sense to reduce the damage of starter weapon so new players can ease into Pvp accidents.

Coding-Hen: Yeah makes sense. I think having that config Todo created then for future would be good for players to configure but for now this is a good start to the pvp sync

Coding-Hen (approved):

NinjaPedroX:
Would it be a good idea to maybe add a `/pvp on/off` console command so that it can be enabled or disabled while players are in game? (Instead of having to restart the server to change the setting)

tornac1234:
I believe a server config is not meant to be modified at play time and this kind of option should be thought about before starting a server. But I'd like to hear about others' opinions

NinjaPedroX:
> I believe a server config is not meant to be modified at play time and this kind of option should be thought about before starting a server. But I'd like to hear about others' opinions

Isn't the gamemode config setting modified during play time with the `/changeservergamemode mode` command? The possible keepInventory setting would also make sense to be changeable in-game when that gets implemented

Coding-Hen:
I could see an argument for both a command and not. I don't think if it's hard to add an option for it then we shouldn't but if it's going to be hard then maybe just a server setting in the config for now will do

Jannify (approved):

tornac1234:
Will add a "pvp on/off" command

NitroxServer/ConsoleCommands/PvpCommand.cs
```diff
@@ -0,0 +1,52 @@
...
+                break;
+            case "off":
+                break;
+            default:
+                SendMessage(args.Sender, "Parameter must be \"on\" or \"off\"");
+                return;
+        }
+
+
+        using (serverConfig.Update(Path.Combine(WorldManager.SavesFolderDir, serverConfig.SaveName)))
```
Coding-Hen, line 39: Probably not for this PR but I feel like we shouldn't have to specify this whole thing in the command and should just be able to call severconfig.update()

Measurity: Side effect of how bad the DI in server is setup right now, but issue is created for it: #2166

PR #2113, opened 2024-01-21
Sync spawning commands

tornac1234:
- [x] Fix "spawn" command (broken since Living Large)
- [x] [new] Sync "sub" command, we can now spawn cyclops with a command
~~- [ ] Maybe:  Move those to command callbacks ("/spawn [VEHICLE TYPE] [OPTIONAL COORDINATES]") with OP requirement~~
- [x] Require the moderator perms at least to use the those commands (only local check for now, but should be switched to the future command system once it's ready)

Adds the translation key: "Nitrox_MissingPermission" to be set like so
"You miss the permissions to execute this command [{PERMISSION}]"

### DIFF
--- a/NitroxClient/GameLogic/NitroxConsole.cs
+++ b/NitroxClient/GameLogic/NitroxConsole.cs
@@ -14,12 +14,12 @@ namespace NitroxClient.GameLogic
         public static bool DisableConsole { get; set; } = true;
 
         private readonly IPacketSender packetSender;
-        private readonly Items item;
+        private readonly Items items;
 
-        public NitroxConsole(IPacketSender packetSender, Items item)
+        public NitroxConsole(IPacketSender packetSender, Items items)
         {
             this.packetSender = packetSender;
-            this.item = item;
+            this.items = items;
         }
 
         //List of things that can be spawned : https://subnauticacommands.com/items
@@ -35,8 +35,7 @@ namespace NitroxClient.GameLogic
                 }
                 else
                 {
-                    SpawnItem(gameObject);
-                    //TODO: Add support for no AI creature that need to be spawned as well
+                    DefaultSpawn(gameObject);
                 }
             }
             catch (Exception ex)
@@ -46,29 +45,22 @@ namespace NitroxClient.GameLogic
         }
 
         /// <summary>
-        /// Spawns Seamoth, Exosuit and Cyclops
+        /// Spawns Seamoth, Exosuit or Cyclops
         /// </summary>
         private void SpawnVehicle(GameObject gameObject, TechType techType)
         {
             NitroxId id = NitroxEntity.GetIdOrGenerateNew(gameObject);
 
-            VehicleWorldEntity vehicleEntity = Vehicles.MakeVehicleEntity(gameObject, id, techType);
+            VehicleWorldEntity vehicleEntity = Vehicles.BuildVehicleWorldEntity(gameObject, id, techType);
             
             packetSender.Send(new EntitySpawnedByClient(vehicleEntity));
 
             Log.Debug($"Spawning vehicle {techType} with id {id} at {gameObject.transform.position}");
         }
 
-        /// <summary>
-        /// Spawns a Pickupable item
-        /// </summary>
-        private void SpawnItem(GameObject gameObject)
+        private void DefaultSpawn(GameObject gameObject)
         {
-            if (gameObject.TryGetComponent(out Pickupable pickupable))
-            {
-                Log.Debug($"Spawning item {pickupable.GetTechName()} at {gameObject.transform.position}");
-                item.Dropped(gameObject, pickupable.GetTechType());
-            }
+            items.Dropped(gameObject);
         }
 
         private static TechType GetObjectTechType(GameObject gameObject)
--- a/NitroxClient/GameLogic/NitroxConsole.cs
+++ b/NitroxClient/GameLogic/NitroxConsole.cs
@@ -1,11 +1,9 @@
 using System;
 using NitroxClient.Communication.Abstract;
-using NitroxClient.GameLogic.Helper;
 using NitroxClient.MonoBehaviours;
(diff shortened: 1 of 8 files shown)


NitroxPatcher/Patches/Dynamic/SubConsoleCommand_OnConsoleCommand_sub_Patch.cs
```diff
@@ -1,19 +1,41 @@
...
+                                                new CodeMatch(OpCodes.Ldloc_0),
+                                                new CodeMatch(OpCodes.Ldarg_0)
+                                            ])
+                                            .SetOpcodeAndAdvance(OpCodes.Ldnull)
+                                            .SetOperandAndAdvance(Reflect.Method(() => WrappedCallback(default)))
+                                            .InstructionEnumeration();
+    }
+
+    public static void WrappedCallback(GameObject prefab)
+    {
```
dartasen, line 35: What will happen if we spawn an aurora with the sub command ?

NitroxPatcher/Patches/Dynamic/SubConsoleCommand_OnConsoleCommand_sub_Patch.cs
```diff
@@ -1,19 +1,41 @@
...
+                                                new CodeMatch(OpCodes.Ldsfld),
+                                                new CodeMatch(OpCodes.Ldloc_0),
+                                                new CodeMatch(OpCodes.Ldarg_0)
+                                            ])
+                                            .SetOpcodeAndAdvance(OpCodes.Ldnull)
+                                            .SetOperandAndAdvance(Reflect.Method(() => WrappedCallback(default)))
+                                            .InstructionEnumeration();
+    }
+
+    public static void WrappedCallback(GameObject prefab)
```
dartasen, line 34: Method is named `WrappedCallback `in the patch and `Callback ` in SpawnConsole patch

NitroxPatcher/Patches/Dynamic/SubConsoleCommand_OnConsoleCommand_sub_Patch.cs
tornac1234: I tried and it just disappeared after some seconds for some reason (same for escapepod), but I'll probably restrict this to spawn cyclops

NitroxClient/GameLogic/Vehicles.cs
```diff
@@ -232,4 +234,11 @@ public static void RemoveNitroxEntitiesTagging(GameObject constructedObject)
             UnityEngine.Object.DestroyImmediate(nitroxEntity);
         }
     }
+
+    public static VehicleWorldEntity MakeVehicleEntity(GameObject constructedObject, NitroxId constructedObjectId, TechType techType, NitroxId constructorId = null)
```
dartasen, line 238:
```suggestion
    public static VehicleWorldEntity BuildVehicleWorldEntity(GameObject constructedObject, NitroxId constructedObjectId, TechType techType, NitroxId constructorId = null)
```
We usually use Build prefix for method that acts like a """factory"""

dartasen (approved):

NitroxClient/GameLogic/NitroxConsole.cs
```diff
@@ -48,32 +45,39 @@ public void Spawn(GameObject gameObject)
         }
 
         /// <summary>
-        /// Spawns a Seamoth or an Exosuit
+        /// Spawns Seamoth, Exosuit and Cyclops
```
Jannify, line 48: Minor:
```suggestion
        /// Spawns Seamoth, Exosuit or Cyclops
```

NitroxClient/GameLogic/NitroxConsole.cs
```diff
@@ -48,32 +45,39 @@ public void Spawn(GameObject gameObject)
...
+            Log.Debug($"Spawning vehicle {techType} with id {id} at {gameObject.transform.position}");
         }
 
-        /// <summary>
-        /// Spawns a Pickupable item
-        /// </summary>
-        private void SpawnItem(GameObject gameObject)
+        private void DefaultSpawn(GameObject gameObject)
+        {
+            item.Dropped(gameObject);
```
Jannify, line 63: Can we rename `item` to `items`? It's not really part of this pr but would be better for readability nonetheless.

NitroxPatcher/Patches/Dynamic/SubConsoleCommand_OnConsoleCommand_sub_Patch.cs
```diff
@@ -1,19 +1,63 @@
...
+     */
+    public static IEnumerable<CodeInstruction> Transpiler(IEnumerable<CodeInstruction> instructions)
+    {
+        return new CodeMatcher(instructions).MatchEndForward([
+                                                new CodeMatch(OpCodes.Stfld),
+                                                new CodeMatch(OpCodes.Ldsfld),
+                                                new CodeMatch(OpCodes.Ldloc_0),
+                                                new CodeMatch(OpCodes.Ldarg_0)
+                                            ])
+                                            .SetOpcodeAndAdvance(OpCodes.Ldnull)
```
Jannify, line 51: I am guessing that you replace the `Ldarg_0` here to null as it's no longer a local method being called. Would a `nop` also work? I am guessing that `nop` is more clear and can't produce any side effects.

tornac1234: Results of the discussion in the Discord:
"newobj constructor takes an object and a native int, the object being the instance (null in our case for  static) and the native int being the pointer given by ldftn.
ldftn will in either way (static or instance method) just give a pointer without the instance mentionned. It's about the instance we give to the constructor"

Jannify (approved):
LGTM CW

PR #2053, opened 2023-06-16
Game finders overhaul

dartasen:
Nitrox has always defined an order in the way it searches the Subnautica installation files. This works in most cases, since users generally have only one installation of the game. However, this approach raises questions when a user has the game on both Steam & Epic.

The aim of this PR is to do away with this automatic way of doing things, by providing an API that lets you search for one or more installations of a given game (GameInfo) via flags (GameLibraries).

i.e:
```csharp
IEnumerable<GameInstallation> installations = GameInstallerFinder.FindGame(GameInfo.SubnauticaZero, GameLibraries.EPIC | GameLibraries.STEAM);

var wantedInstallation = installations.Where(install => install.Origin == GameLibraries.Steam)
```

That will allow the launcher to display all the path, and let the user choose between them. Rather than imposing it.

### Progression

- [x] Reimplement finders
- [x] Rework option page of the launcher
- [x] Cleanup of the launcher (especially Styles + UI bugfixes)

Below zero will follow up in another Pull Request

PR #2121, opened 2024-02-25
Game finders overhaul 

dartasen:
### This is a spin-off of #2053, I splitted up this into several ones, so it will be easier to review. I applied the suggestions from the previous PR onto this one.

Nitrox has always defined an order in the way it searches the Subnautica installation files. This works in most cases, since users generally have only one installation of the game. However, this approach raises questions when a user has the game on both Steam & Epic.

The aim of this PR is to do away with this automatic way of doing things, by providing an API that lets you search for one or more installations of a given game (GameInfo) via flags (GameLibraries).

i.e:
```csharp
IEnumerable<GameInstallation> installations = GameInstallerFinder.FindGame(GameInfo.SubnauticaZero, GameLibraries.EPIC | GameLibraries.STEAM);

var wantedInstallation = installations.Where(install => install.Origin == GameLibraries.Steam)
```

That will allow the launcher to display all the path, and let the user choose between them. Rather than imposing it.

### What's changed

- Changed namespace to be file-scoped by default (Others are just automatic additions for available lines that were not in our .editorconfig)
- Reimplemented finders to use our GameInfo structure
- Added Microsoft finder (real game install is locked up behind OS secured folder, so for now we'll check for default path)
- Improved Steam finder to support Linux & OSX game path
- Removed CurrentDirectory finder


NitroxModel/Discovery/InstallationFinders/EnvironmentFinder.cs
```diff
@@ -0,0 +1,37 @@
...
+/// </summary>
+public sealed class EnvironmentFinder : IGameFinder
+{
+    public GameFinderResult FindGame(GameInfo gameInfo)
+    {
+        string path = Environment.GetEnvironmentVariable($"{gameInfo.FullName.ToUpper()}_INSTALLATION_PATH");
+        if (string.IsNullOrEmpty(path))
+        {
+            return Error($"Configured game path with environment variable {gameInfo.FullName.ToUpper()}_INSTALLATION_PATH was found empty");
+        }
```
Measurity, line 23: `gameInfo.FullName` returns "Subnautica: Below Zero", not "SubnauticaZero". Should probably be `gameInfo.Name` here.

Measurity (changes requested):

Measurity (approved):

Measurity (approved):
Approved again (I messed up a unit test)

NitroxModel/Discovery/InstallationFinders/SteamFinder.cs
```diff
@@ -0,0 +1,163 @@
...
+        {
+            return NotFound();
+        }
+
+        return Ok(new GameInstallation
+        {
+            Path = path,
+            GameInfo = gameInfo,
+            Origin = GameLibraries.STEAM
+        });
```
Jannify, line 49: Would validate the folder before the two returns

NitroxModel/Discovery/InstallationFinders/SteamFinder.cs
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    private static string GetSteamPath()
+    {
+        if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+        {
+            string steamPath = Platforms.OS.Windows.Internal.RegistryEx.Read<string>(@"Software\\Valve\\Steam\SteamPath");
```
Jannify, line 56: The mixture of double back quotes and one single seems wrong, is it legit?

NitroxModel/Helper/NitroxUser.cs
```diff
@@ -93,14 +95,21 @@ public static string GamePath
                     return gamePath;
                 }
 
-                List<string> errors = new();
-                string path = GameInstallationFinder.Instance.FindGame(errors);
-                if (!string.IsNullOrWhiteSpace(path) && Directory.Exists(path))
+                string path = null;
+                List<GameFinderResult> finderResults = GameInstallationFinder.Instance.FindGame(GameInfo.Subnautica).TakeUntilInclusive(r => r is { IsOk: false }).ToList();
```
Jannify, line 99: Shouldn't this be `isOk: true`?

NitroxModel/Helper/NitroxUser.cs
```diff
@@ -93,14 +95,21 @@ public static string GamePath
...
+                if (potentiallyValidResult?.IsOk == true)
                 {
+                    path = potentiallyValidResult.Installation.Path;
+                }
+
+                if (Directory.Exists(path))
+                {
+                    Log.Debug($"Game installation was found by {potentiallyValidResult.FinderName} at '{potentiallyValidResult.Installation.Path}'");
                     return gamePath = path;
                 }
```
Jannify, line 110: As the finders do extensive checking couldn't this be moved into the if above?

Jannify (changes requested):

NitroxModel/Helper/NitroxUser.cs
Measurity: No, but I agree it looks weird. It's purpose is to "take all the bad results and stop + 1". This way we can log the bad results instead of having to run the `FindGame` twice.

NitroxModel/Discovery/InstallationFinders/SteamFinder.cs
Measurity: Good catch. It still worked but no reason to keep it.

Jannify (approved):
LGTM CW

NitroxModel/Discovery/InstallationFinders/SteamFinder.cs
```diff
@@ -0,0 +1,165 @@
...
+        }
+
+        StreamReader file = new(libraryFolders);
+        char[] trimChars = [' ', '\t'];
+
+        while (file.ReadLine() is { } line)
+        {
+            line = Regex.Unescape(line.Trim(trimChars));
+            Match regMatch = Regex.Match(line, "\"(.*)\"\t*\"(.*)\"");
+            string key = regMatch.Groups[1].Value;
```
Coding-Hen, line 147: Do we need to check that a match has been found first or this will throw a null pointer potentially on the index

NitroxModel/Helper/NitroxUser.cs
```diff
@@ -93,14 +95,15 @@ public static string GamePath
                     return gamePath;
                 }
 
-                List<string> errors = new();
-                string path = GameInstallationFinder.Instance.FindGame(errors);
-                if (!string.IsNullOrWhiteSpace(path) && Directory.Exists(path))
+                List<GameFinderResult> finderResults = GameInstallationFinder.Instance.FindGame(GameInfo.Subnautica).TakeUntilInclusive(r => r is { IsOk: false }).ToList();
```
Coding-Hen, line 98: My only thought as we have now hardcoded as such in the GameInfo.Subnautica how we want this to search in future when we potentially have BZ if this is the default get and if it's worth addressing this now

Nitrox.Test/Model/ExtensionsTests.cs
```diff
@@ -1,9 +1,4 @@
-﻿using System;
-using System.Collections.Generic;
-using FluentAssertions;
-using Microsoft.VisualStudio.TestTools.UnitTesting;
```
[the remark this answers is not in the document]
dartasen: We have a GlobalUsing

NitroxModel/Helper/NitroxUser.cs
dartasen: As stated on the PR it's a first PR, others will come after. Previously everything was hardcoded for subnautica inside the finders, that's the aim of using GamInfo

PR #2137, opened 2024-05-09
Crops persistence

tornac1234:
- [x] Add crops persistence (grow progress and exact slot)
- [x] Make sure crops sync works
- [x] Sync trash can behaviour (related because you often need to throw items when playing with crops, ig)
- [x] Sync deletion of giver objects (which give you some item when you click on them like melons)
- [x] Sync fruit growing (from grown plants) and harvesting (picking up)


NitroxPatcher/Patches/Dynamic/PickPrefab_AddToContainerAsync_Patch.cs
```diff
@@ -0,0 +1,59 @@
...
+                                                new CodeInstruction(OpCodes.Ldarg_0),
+                                                new CodeInstruction(OpCodes.Call, Reflect.Method(() => BroadcastDeletion(default)))
+                                            ])
+                                            .MatchEndForward(new CodeMatch(OpCodes.Call, Reflect.Method(() => Object.Destroy(default))))
+                                            .Advance(1)
+                                            .InsertAndAdvance([
+                                                new CodeInstruction(OpCodes.Ldarg_0),
+                                                new CodeInstruction(OpCodes.Call, Reflect.Method(() => BroadcastDeletion(default)))
+                                            ])
+                                            .InstructionEnumeration();
```
Jannify, line 48: CodeMatcher has support for repeated operations, don't know if it's better here:
```
new CodeMatcher(instructions)
    .MatchForward(false, // false = move at the start of the match, true = move at the end of the match
        new CodeMatch(OpCodes.Stfld),
        new CodeMatch(OpCodes.Ldarg_0),
        new CodeMatch(i => i.opcode == OpCodes.Ldfld && ((FieldInfo)i.operand).Name == "test"))
    .Repeat( matcher => // Do the following for each match
                 matcher
                   .Advance(2) // Move cursor to before ldfld
                   .InsertAndAdvance(
                      new CodeInstruction(OpCodes.Dup),
                      new CodeInstruction(OpCodes.Call, AccessTools.Method(typeof(Foo), "Foo"))
                   )
    )
    .InstructionEnumerable(); // Finally, return the manipulated method instructions
```

NitroxPatcher/Patches/Dynamic/PickPrefab_AddToContainerAsync_Patch.cs
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    public static void BroadcastDeletion(PickPrefab pickPrefab)
+    {
+        if (pickPrefab.TryGetNitroxId(out NitroxId objectId) ||
+            (pickPrefab.TryGetComponent(out GrownPlant grownPlant) && grownPlant.seed && grownPlant.seed.TryGetNitroxId(out objectId)))
```
Jannify, line 54: Don't we need to check the simulation lock here?

Jannify (changes requested):

NitroxPatcher/Patches/Dynamic/PickPrefab_AddToContainerAsync_Patch.cs
tornac1234: I'll do it this time, but we should probably refrain from doing it everywhere because we could lose context in case Subnautica code gets updated. Even though tests are there to show us that some patch broke (e.g. now there are 3 matches and not 2), we might lose the exact intention of which matches exactly we wanted to patch, and which one we didn't want to patch.

NitroxPatcher/Patches/Dynamic/PickPrefab_AddToContainerAsync_Patch.cs
tornac1234: Picking up an item from those behaviours is a task executed locally, therefore only the local player knows about it, which is why they should not require simulation lock for it.

tornac1234:
New commit can be reviewed by itself

Jannify (changes requested):

tornac1234:
Note to fix something in StayAtLeashPositionMetadataProcessor (from leviathans PR)

kovadam69:
I merged this into the master branch on my local computer, and seems to work perfectly. Even restarting the server preserved the plant states, even the grow state. This seems to be OK.

NitroxModel/DataStructures/GameLogic/ItemData.cs
BrightenedDay, line 1: I believe ItemData was removed with #2196? So a rebase? I think? I'm not too familiar with the github workflow.

kovadam69: Me too. But I did it locally, since it was very disturbing, that crops were reset every time we restarted the game or rejoined it. So I cloned the master and merged with this pull request (and some others, like stalker sync and the latest pull request for box issues). There I simply deleted the ItemData.cs in Visual Studio after merge and also solved conflict in an other file. Then it compiled without any problems, and seems also to work flawlessly.

NitroxPatcher/Patches/Dynamic/FruitPlant_Update_Patch.cs
```diff
@@ -0,0 +1,73 @@
...
+        }
+
+        __state = (__instance.timeNextFruit, null, null);
+        return true;
+    }
+
+    public static void Postfix(FruitPlant __instance, (float, NitroxId, Plantable) __state)
+    {
+        // If no change was made
+        if (__state.Item1 == __instance.timeNextFruit)
```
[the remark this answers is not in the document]
tornac1234: float a = __instance.timeNextFruit;

float b = __instance.timeNextFruit;

is there a reason why "a" could be different than "b" if __instance.timeNextFruit didn't change in the meantime ?

tornac1234:
I'll be correcting this after all that happened since last year

NitroxClient/GameLogic/Spawning/Metadata/Processor/FruitPlantMetadataProcessor.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        // 2. The entity with an id has a Plantable (located in the plot's storage),
+        // we want to access the FruitPlant component which is on the spawned plant object
+        if (gameObject.TryGetComponent(out Plantable plantable))
+        {
+            if (plantable.linkedGrownPlant && plantable.linkedGrownPlant.TryGetComponent(out fruitPlant))
+            {
+                ProcessMetadata(fruitPlant, metadata);
+            }
+            return;
```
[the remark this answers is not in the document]
tornac1234: Iirc this processor is only ever called from places where we're sure that either case is the right one. I'll add an error log tho

tornac1234:
Corrected everything that was necessary, can I get more reviews and tests please ?

tornac1234:
fastgrow command is not synced and will just break any source of sync

tornac1234:
> This is really good, fixed both issues I was having before. Now for two more (minor) issues!
> 
>     * Sometimes when you harvest/grab a fully grown plant, and then immediately replant a seed of the same type, it will be instantly fully grown again. `fastgrow` was off. The max I was able to chain this was 5 times in a row, but it seems to "catch up" after a few seconds and realize the slot is no longer fully grown
> 
>     * growth state/percentage is not always consistent between players, but relogging will fix this.
> 
> 
> [game[Nike2]-20250225.log](https://github.com/user-attachments/files/18974797/game.Nike2.-20250225.log) [server[crops]-20250225.log](https://github.com/user-attachments/files/18974798/server.crops.-20250225.log) [game[Nike]-20250225.log](https://github.com/user-attachments/files/18974799/game.Nike.-20250225.log) [launcher-20250225.log](https://github.com/user-attachments/files/18974800/launcher-20250225.log)

Do you have any example of seed to replant fast ?

tornac1234:
I'm unfortunately unable to reproduce this bug... It might have been because of fastgrowth ? I'll try to sync the command but otherwise I'll just disable it

tornac1234:
That is a pretty bad error indeed. Can you explain to me the exact procedure you did to obtain these ? (please mention which commands you used to get the items, and when you picked up something/planted something, with the number of players)

tornac1234:
@NikeLaosClericus Thanks, I was able to reproduce the issues and fix them and I even synced both multiplayer-relevant fast commands (fastgrow and fasthatch)

tornac1234:
The bug I fixed requires you to use new mushrooms tho, it's about how metadata is set onto them when you pick them up

tornac1234:
Thanks a lot for the bug reports @NikeLaosClericus, I could fix both of them and fortunately, no one will have to come across those after release :)

tornac1234:
I am currently unable to reproduce the "ghost shroom" bug but I suspect it happens because of lag... And the second one happens because we never update the state of a plant when it finally grows up. I could do it but will give it a bit more thinking

NitroxPatcher/Patches/Dynamic/Inventory_Pickup_Patch.cs
```diff
@@ -0,0 +1,20 @@
...
+public sealed partial class Inventory_Pickup_Patch : NitroxPatch, IDynamicPatch
+{
+    private static readonly MethodInfo TARGET_METHOD = Reflect.Method((Inventory t) => t.Pickup(default, default));
+
+    public static void Prefix()
+    {
+        Resolve<Items>().InventoryPickingUp = true;
+    }
+
+    public static void Finalizer()
```
Jannify, line 16: Could you check if we have other patches where we set a global field. I fear we are using postfixes in them...

NitroxClient/GameLogic/InitialSync/PlayerInitialSyncProcessor.cs
```diff
@@ -137,8 +138,25 @@ private static void SetPlayerGameMode(NitroxGameMode gameMode)
...
+        StringBuilder cheatsEnabled = new("Cheats enabled:");
+        if (fastHatch)
+        {
+            cheatsEnabled.Append(" fastHatch");
+        }
+        if (fastGrow)
+        {
+            cheatsEnabled.Append(" fastGrow");
+        }
+        ErrorMessage.AddDebug(cheatsEnabled.ToString());
```
Jannify, line 160: Would use Log.InGame()

NitroxPatcher/Patches/Dynamic/NoCostConsoleCommand_OnConsoleCommand_bobthebuilder_Patch.cs
```diff
@@ -0,0 +1,24 @@
...
+/// Prevents "bobthebuilder" command from enabling "fasthatch" and "fastgrow", since both commands are managed by the server
+/// </summary>
+public sealed partial class NoCostConsoleCommand_OnConsoleCommand_bobthebuilder_Patch : NitroxPatch, IDynamicPatch
+{
+    public static readonly MethodInfo TARGET_METHOD = Reflect.Method((NoCostConsoleCommand t) => t.OnConsoleCommand_bobthebuilder(default));
+
+    public static void Prefix(out (bool, bool) __state)
+    {
+        __state = (NoCostConsoleCommand.main.fastHatchCheat, NoCostConsoleCommand.main.fastGrowCheat);
+        ErrorMessage.AddWarning("fastHatch and fastGrow cheats weren't enabled because they're shared cheat commands. Please enable them manually with their respective commands \"fastgrow\" and \"fasthatch\"");
```
Jannify, line 16: What about calling NoCostConsoleCommand_OnConsoleCommand_fastgrow_Patch.Prefix()? That way we at least try to enable them.

NitroxServer/ConsoleCommands/FastCommand.cs
```diff
@@ -0,0 +1,74 @@
...
+namespace NitroxServer.ConsoleCommands;
+
+public class FastCommand : Command
+{
+    private readonly PlayerManager playerManager;
+    private readonly SessionSettings sessionSettings;
+
+    public FastCommand(PlayerManager playerManager, SessionSettings sessionSettings) : base("fast", Perms.MODERATOR, "Enables/disables a fast cheat command, whether it be \"hatch\" or \"grow\"")
+    {
+        AddParameter(new TypeString("cheat", true, "The name of the fast cheat to change: \"hatch\" or \"grow\""));
```
Jannify, line 16: Would it make sense to make it an enum? If we rename FastCheatChanged.FastCheat we can even use it.

Jannify (changes requested):

NitroxPatcher/Patches/Dynamic/Inventory_Pickup_Patch.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace NitroxPatcher.Patches.Dynamic;
+
+public sealed partial class Inventory_Pickup_Patch : NitroxPatch, IDynamicPatch
+{
+    private static readonly MethodInfo TARGET_METHOD = Reflect.Method((Inventory t) => t.Pickup(default, default));
+
+    public static void Prefix()
+    {
+        Resolve<Items>().InventoryPickingUp = true;
```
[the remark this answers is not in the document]
tornac1234: I'll still need Inventory to be mentionned because this precise occurrence is for the Inventory class

NitroxClient/GameLogic/Items.cs
```diff
@@ -24,6 +24,11 @@ public class Items
     public static GameObject PickingUpObject { get; private set; }
     private readonly EntityMetadataManager entityMetadataManager;
 
+    /// <summary>
+    /// Whether or not <see cref="Inventory.Pickup"/> is running. It's useful to discriminate between Inventory.Pickup from a regular
```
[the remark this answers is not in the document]
tornac1234: I cut the comment again 💀 "a regular Pickupable.Pickup"

NitroxPatcher/Patches/Dynamic/Inventory_Pickup_Patch.cs
tornac1234: Couldn't find any other

Jannify (approved):

tornac1234:
I have to fix one more thing

tornac1234:
Two last commits needs review

NitroxClient/Debuggers/NetworkDebugger.cs
```diff
@@ -19,7 +19,7 @@ public class NetworkDebugger : BaseDebugger, INetworkDebugger
         {
             nameof(PlayerMovement), nameof(EntityTransformUpdates), nameof(PlayerStats), nameof(SpawnEntities), nameof(VehicleMovements), nameof(PlayerCinematicControllerCall),
             nameof(FMODAssetPacket), nameof(FMODEventInstancePacket), nameof(FMODCustomEmitterPacket), nameof(FMODStudioEmitterPacket), nameof(FMODCustomLoopingEmitterPacket), 
-            nameof(SimulationOwnershipChange), nameof(CellVisibilityChanged), nameof(PlayerInCyclopsMovement), nameof(CreatureActionChanged),
+            nameof(SimulationOwnershipChange), nameof(CellVisibilityChanged), nameof(PlayerInCyclopsMovement), nameof(CreatureActionChanged), nameof(FootstepPacket)
```
Coding-Hen, line 22: Do we want this now? I assume it will be broadcast a lot and so fill up the display? I know it's only debug so we might not be worried but thought I would query it

NitroxClient/GameLogic/Helper/InventoryContainerHelper.cs
```diff
@@ -55,8 +55,15 @@ public static bool TryGetOwnerId(Transform ownerTransform, out NitroxId ownerId)
                 ownerId = null;
                 return false;
             }
+            // TODO: in the future maybe use a switch on the PrefabId (it's always the same structure in a prefab)
+            // and then statically look for the right object because we'll know exactly which one it is
 
-            if (parent.GetComponent<Constructable>() || parent.GetComponent<IBaseModule>().AliveOrNull())
+            // This check must happen before the GetComponent<IBaseModule> which will be triggered (wrong result)
```
Coding-Hen, line 61: I'm unclear what this is trying to say. Is it saying that if the GetComponent call is done first it will produce an unwanted outcome? If so I think the end of the comment should be reworded it's a little unclear if being triggered is wrong or if this check prevents that

NitroxClient/GameLogic/ItemContainers.cs
```diff
@@ -13,86 +13,100 @@
...
 
-        public void BroadcastItemAdd(Pickupable pickupable, Transform containerTransform)
+        if (!InventoryContainerHelper.TryGetOwnerId(containerTransform, out NitroxId ownerId))
         {
-            // We don't want to broadcast that event if it's from another player's inventory
-            if (containerTransform.GetComponentInParent<RemotePlayerIdentifier>(true))
-            {
-                return;
-            }
+            // Error logging is done in the function
```
Coding-Hen, line 41: Minor: Slight reword maybe, In the try function?

NitroxClient/GameLogic/ItemContainers.cs
```diff
@@ -13,86 +13,100 @@
...
+            return;
+        }
 
-            if (!pickupable.TryGetIdOrWarn(out NitroxId itemId))
-            {
-                return;
-            }
+        // For planters, we'll always forcefully recreate the entity to ensure there's no desync
+        if (container.containerType == ItemsContainerType.LandPlants || container.containerType == ItemsContainerType.WaterPlants)
```
Coding-Hen, line 46: Is it worth us having Plants as a type then inherit? Then we can only have one check here and work for all plants

NitroxClient/GameLogic/Spawning/Bases/BuildingPostSpawner.cs
```diff
@@ -28,66 +31,75 @@ public static IEnumerator ApplyPostSpawner(GameObject gameObject, NitroxId objec
...
+        {
+            NitroxId leftId = waterParkId.Increment();
+            NitroxId rightId = leftId.Increment();
+
+            NitroxEntity.SetNewId(largeRoomWaterPark.planters.leftPlanter.gameObject, leftId);
+            NitroxEntity.SetNewId(largeRoomWaterPark.planters.rightPlanter.gameObject, rightId);
+            return;
+        }
+
+        NitroxId planterId = waterParkId.Increment();
```
Coding-Hen, line 101: I don't have the increment code in front of me but I wanted to check this gives not the same as the other waterpark increment call.

Coding-Hen (changes requested):

NitroxClient/Debuggers/NetworkDebugger.cs
tornac1234: we definitely don'tt need to see the spam of FootstepPackets in the network debugger, which is why I added it to the exclude filter

OhmV-IR: yeah I should've done it myself, but didn't

NitroxPatcher/Patches/Dynamic/Planter_AddItem_Patch.cs
```diff
@@ -7,15 +7,21 @@ namespace NitroxPatcher.Patches.Dynamic;
 
 public sealed partial class Planter_AddItem_Patch : NitroxPatch, IDynamicPatch
 {
-    public static readonly MethodInfo TARGET_METHOD = Reflect.Method((Planter p) => p.AddItem(default));
+    public static readonly MethodInfo TARGET_METHOD = Reflect.Method((Planter p) => p.AddItem(default, default));
 
-    public static void Prefix(InventoryItem item)
+    public static void Postfix(Plantable plantable, int slotID, Planter __instance)
```
Coding-Hen, line 12: I think we usually put instance first to be consistent
```suggestion
    public static void Postfix(Planter __instance, Plantable plantable, int slotID)
```

Coding-Hen (changes requested):
Finally found time to go through all the files. Looks good just a few more observations

Measurity (approved):

NitroxClient/GameLogic/ItemContainers.cs
tornac1234: It is not a valid general case because the items that need to go through this don't necessarily inherit from a "plant" related behaviour. Thus the easiest check is on the container type

NitroxClient/GameLogic/Spawning/Bases/BuildingPostSpawner.cs
tornac1234: This is a very precise calculation which always adds 1 to the hexadecimal in the NitroxId's GUID. It has to be this deterministic (and not less) so we can actually pretend to have sync here

NitroxClient/GameLogic/Spawning/Metadata/Processor/FruitPlantMetadataProcessor.cs
```diff
@@ -0,0 +1,63 @@
...
+        if (!gameObject.TryGetComponent(out Plantable plantable))
+        {
+            Log.Error($"[{nameof(FruitPlantMetadataProcessor)}] Could not find {nameof(FruitPlant)} related to {gameObject.name}");
+            return;
+        }
+
+        if (!plantable.linkedGrownPlant)
+        {
+            // This is an error which will happen quite often since this metadata
+            // is applied from PlantableMetadataProcessor even when linkedGrownPlant isn't available yet
```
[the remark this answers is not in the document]
tornac1234: This case is absolutely normal and we can do nothing about it:
In some cases, there will be 2 calls to this processor because in the first one, the base was inactive and thus the coroutines to spawn the plant (which is required in this processor) won't be able to run until the base is active. When the base is active, the coroutines will run, spawning the plant and allowing for the 2nd run of this processor.

NitroxClient/GameLogic/Spawning/WorldEntities/CreepvineEntitySpawner.cs
```diff
@@ -0,0 +1,52 @@
...
+    {
+        yield return defaultWorldEntitySpawner.SpawnAsync(entity, parent, cellRoot, result);
+        if (!result.value.HasValue)
+        {
+            yield break;
+        }
+
+        SetupObject(result.value.Value);
+
+        // result is already set
```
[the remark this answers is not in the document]
tornac1234: In any regular entity spawner, this line would be the one to set the result. To maintain this code structure, I let a comment here instead

NitroxPatcher/Patches/Dynamic/FruitPlant_Update_Patch.cs
```diff
@@ -0,0 +1,73 @@
...
+            FruitPlantMetadata metadata = Resolve<FruitPlantMetadataExtractor>().Extract(__instance);
+            Resolve<IPacketSender>().Send(new EntityMetadataUpdate(__state.Item2, metadata));
+        }
+        // If the NitroxEntity is on a distant Plantable object
+        else
+        {
+            // TODO: Refer to the TODO in PlantableMetadata.
+            // When TODO is done, change this to only update the FruitPlant metadata (like the above if)
+            PlantableMetadata metadata = Resolve<PlantableMetadataExtractor>().Extract(__state.Item3);
+            Resolve<IPacketSender>().Send(new EntityMetadataUpdate(__state.Item2, metadata));
```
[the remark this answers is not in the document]
tornac1234: Item2 is the id, item3 is the Plantable. We extract from Item3 and send the id (Item2)

tornac1234:
Test session with @NinjaPedroX was a success

dartasen (approved):

Coding-Hen (approved):
LGTM, will be good to get this in nice work!

PR #2145, opened 2024-06-16
Automatic entity resync

tornac1234:
- [x] Automatically request an entity resync when they were missing in a ReparentProcessor or in PlayerHeldItemChanged (client-side) **[needs to be tested]**
- [x] Fix old transpiler to now broadcast items drop on death
- [ ] Add a button to request resync for all vehicles **[children respawning is currently not correctly handled for this resync]**


Andbli (approved):
Looks good!

Coding-Hen:
Do we need this now or is it redundant after the changes to the entity system recently?

tornac1234:
This is all in an effort to avoid restarting the game every time you have some slight desync, but as it is right now, it needs to be thought about, so we either mostly fix the desync occurrences or we enable disconnecting without crashing game, and then if some bad issue persist, we'll implement this kind of resyncer

Issue #2097, opened 2023-12-03
System.OverflowException: Value was either too large or too small for an Int32.


Elmar8121:
### Are you on the latest version of Nitrox?

Yes, I'm on the latest release of Nitrox.

### Which OS are you playing on?

Windows

### On which platform did you buy Subnautica?

Steam

### Describe the issue.

I was playing with a friend. The game crashed and nitrox stopped working. I restarted nitrox, opened the game, and a new world began. It's good that there was a backup. I restored the copy. But when starting the nitrox server, it says that System.OverflowException: Value was either too large or too small for an Int32. Creation of a new world. The meaning of the error is clear. But how to fix it?

### Describe the steps you took to encounter the bug.

![изображение_2023-12-04_005041419](https://github.com/SubnauticaNitrox/Nitrox/assets/152815121/ae62411e-623e-49f2-aced-ac61b205b99d)

### Relevant log output

[game-20231203.log](https://github.com/SubnauticaNitrox/Nitrox/files/13540206/game-20231203.log)
[server-20231203.log](https://github.com/SubnauticaNitrox/Nitrox/files/13540207/server-20231203.log)

dartasen:
@Elmar8121 I think something went wrong with the save at some point, would you mind sending us the .zip please ?

SoZ0:
[worldBackup.zip](https://github.com/SubnauticaNitrox/Nitrox/files/14730521/worldBackup.zip)
Same issue with my world, narrowed it down the the entity json file, could not find anything though problematic with a few programs to check if anything invalided the INT32 range with no luck. Also did some manual searches with specific types of items like the SmallStorage an no luck. The file is huge (5.8 million lines when formatted) so going through by hand is just not possible

SleepyStew:
Had the same issue here, after a bit of debugging I was able to recover the world. Here's how I did it for anyone who might run into the same issue:

I figured out that the overflow was in EntityData although I couldn't find it after some digging so instead I grabbed the seed from my world (found in the backup zip, which I made a seperate backup of in case and would recommend) and put it in the sever config, then set the cache entities setting to True so it would generate the entities on first startup, I deleted the world folder (make sure you don't delete your only world backup) and let it generate a new uncorrupted EntityData, and then pasted in all the other files from my world.

This method has a few issues (lose all entity interactions) but for me was almost unnoticeable and is much better than losing the whole world.

PR #2165, opened 2024-08-14
Fixed server starting with errored entities (1.7.1.0)

uGuardian:
Fixed server starting with errored entities, most notably the issue where occasionally an entity will have a NaN location, causing the entire save file to be unloadable.
Refactored sorting into a single enumeration pass, in part to allow easy insertion of the try-catch.

I've noticed several people having this issue occasionally on the current version, and this is intended to be a small patch to allow the current public version to not occasionally end up losing entire saves due to "corrupted" data when the user has no idea of how to find or fix it.

### DIFF
--- a/NitroxServer/GameLogic/Entities/EntityManager.cs
+++ b/NitroxServer/GameLogic/Entities/EntityManager.cs
@@ -28,14 +28,46 @@ namespace NitroxServer.GameLogic.Entities
 
         public EntityManager(List<Entity> entities, BatchEntitySpawner batchEntitySpawner)
         {
-            entitiesById = entities.ToDictionary(entity => entity.Id);
+            entitiesById = new Dictionary<NitroxId, Entity>(entities.Count);
+            globalRootEntitiesById = new Dictionary<NitroxId, Entity>();
+            phasingEntitiesByAbsoluteCell = new Dictionary<AbsoluteEntityCell, List<Entity>>();
+            bool toWarn = false;
 
-            globalRootEntitiesById = entities.FindAll(entity => entity.ExistsInGlobalRoot)
-                                             .ToDictionary(entity => entity.Id);
-
-            phasingEntitiesByAbsoluteCell = entities.FindAll(entity => !entity.ExistsInGlobalRoot)
-                                                    .GroupBy(entity => entity.AbsoluteEntityCell)
-                                                    .ToDictionary(group => group.Key, group => group.ToList());
+            for (int i = 0; i < entities.Count; i++)
+            {
+                Entity entity = entities[i];
+                try
+                {
+                    if (entity.ExistsInGlobalRoot)
+                    {
+                        globalRootEntitiesById.Add(entity.Id, entity);
+                    }
+                    else
+                    {
+                        AbsoluteEntityCell cell = entity.AbsoluteEntityCell;
+                        if (phasingEntitiesByAbsoluteCell.TryGetValue(cell, out List<Entity> list))
+                        {
+                            list.Add(entity);
+                        }
+                        else
+                        {
+                            list = new List<Entity> {entity};
+                            phasingEntitiesByAbsoluteCell.Add(cell, list);
+                        }
+                    }
+                    // At the end since we don't want entities that cause an error
+                    entitiesById.Add(entity.Id, entity);
+                }
+                catch (System.Exception ex)
+                {
+                    toWarn = true;
+                    Log.Error(ex);
+                }
+            }
+            if (toWarn)
+            {
+                Log.Warn("One or more entities have failed to load");
+            }
 
             this.batchEntitySpawner = batchEntitySpawner;
         }
--- a/NitroxServer/GameLogic/Entities/EntityManager.cs
+++ b/NitroxServer/GameLogic/Entities/EntityManager.cs
@@ -31,7 +31,6 @@ namespace NitroxServer.GameLogic.Entities
             entitiesById = new Dictionary<NitroxId, Entity>(entities.Count);
             globalRootEntitiesById = new Dictionary<NitroxId, Entity>();
             phasingEntitiesByAbsoluteCell = new Dictionary<AbsoluteEntityCell, List<Entity>>();
-            bool toWarn = false;
 
             for (int i = 0; i < entities.Count; i++)
             {
@@ -60,13 +59,11 @@ namespace NitroxServer.GameLogic.Entities
                 }
                 catch (System.Exception ex)
                 {
-                    toWarn = true;
+                    Log.Warn($"Entity of type {entity.TechType.Name ?? "Unknown"} with GUID {entity.Id?.ToString() ?? "Unknown"} failed to load");
+#if DEBUG
                     Log.Error(ex);
+#endif
                 }
-            }
-            if (toWarn)
-            {
-                Log.Warn("One or more entities have failed to load");
             }
 
             this.batchEntitySpawner = batchEntitySpawner;
--- a/NitroxServer/GameLogic/Entities/EntityManager.cs
+++ b/NitroxServer/GameLogic/Entities/EntityManager.cs
@@ -1,5 +1,6 @@
 ﻿using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.InteropServices;
 using NitroxModel.Core;
 using NitroxModel.DataStructures;
 using NitroxModel.DataStructures.GameLogic;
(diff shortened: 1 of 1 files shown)


NitroxServer/GameLogic/Entities/EntityManager.cs
```diff
@@ -28,14 +28,46 @@ public class EntityManager
...
+                }
+                catch (System.Exception ex)
+                {
+                    toWarn = true;
+                    Log.Error(ex);
+                }
+            }
+            if (toWarn)
+            {
+                Log.Warn("One or more entities have failed to load");
```
Coding-Hen, line 69: If we are logging the error above without wrapping it in some context, would it not make more sense to just log "error loading entity: error" instead each time? Rather than wait till the end just have a lot of errors then a single log

NitroxServer/GameLogic/Entities/EntityManager.cs
```diff
@@ -28,14 +28,46 @@ public class EntityManager
...
 
-            globalRootEntitiesById = entities.FindAll(entity => entity.ExistsInGlobalRoot)
-                                             .ToDictionary(entity => entity.Id);
-
-            phasingEntitiesByAbsoluteCell = entities.FindAll(entity => !entity.ExistsInGlobalRoot)
-                                                    .GroupBy(entity => entity.AbsoluteEntityCell)
-                                                    .ToDictionary(group => group.Key, group => group.ToList());
+            for (int i = 0; i < entities.Count; i++)
+            {
+                Entity entity = entities[i];
```
Coding-Hen, line 38: Should we put this in the try catch incase something goes wrong with the index logic here and we out of index error?

NitroxServer/GameLogic/Entities/EntityManager.cs
```diff
@@ -28,14 +28,46 @@ public class EntityManager
...
+            phasingEntitiesByAbsoluteCell = new Dictionary<AbsoluteEntityCell, List<Entity>>();
+            bool toWarn = false;
 
-            globalRootEntitiesById = entities.FindAll(entity => entity.ExistsInGlobalRoot)
-                                             .ToDictionary(entity => entity.Id);
-
-            phasingEntitiesByAbsoluteCell = entities.FindAll(entity => !entity.ExistsInGlobalRoot)
-                                                    .GroupBy(entity => entity.AbsoluteEntityCell)
-                                                    .ToDictionary(group => group.Key, group => group.ToList());
+            for (int i = 0; i < entities.Count; i++)
```
Coding-Hen, line 36: Would a foreach not be better here as it would remove the index below and just loop without the counter?

NitroxServer/GameLogic/Entities/EntityManager.cs
```diff
@@ -28,14 +28,46 @@ public class EntityManager
...
+                    if (entity.ExistsInGlobalRoot)
+                    {
+                        globalRootEntitiesById.Add(entity.Id, entity);
+                    }
+                    else
+                    {
+                        AbsoluteEntityCell cell = entity.AbsoluteEntityCell;
+                        if (phasingEntitiesByAbsoluteCell.TryGetValue(cell, out List<Entity> list))
+                        {
+                            list.Add(entity);
```
Coding-Hen, line 50: Does adding to this list supplied with out mean it's the same reference and as such is updating the entity inside?

Coding-Hen (changes requested):

NitroxServer/GameLogic/Entities/EntityManager.cs
uGuardian: Fair enough, that's just my old habit of very verbose logging rearing its ugly head. I'll wrap it in an if debug preprocessor.

NitroxServer/GameLogic/Entities/EntityManager.cs
uGuardian: I considered that, and if it would be preferred that can be done, but generally for loop index access of List<T> is a bit faster than the enumeration of a foreach, and has near identical readability, so I tend to prefer it.

NitroxServer/GameLogic/Entities/EntityManager.cs
uGuardian: Yes.
List<T> is a reference type, so if you get the existing list object out of the dictionary and use it directly, you can avoid pointless reassignments to the dictionary for an identical result.

NitroxServer/GameLogic/Entities/EntityManager.cs
uGuardian: That's a valid option and would probably have an identical result.
Placing it directly next to the actual for loop is a strange habit I picked up because doing so with basic arrays actually causes a compiler optimization for accesses, although the impact of it is minimal and I don't think it works with List<T>.

As one additional comment that also applies to the foreach question:
List<T>.Count is a property that is always maintained, and the maximum index will always be 1 less than the Count property without fail, and the for loop will never iterate if it exceeds that number, even if it were to somehow be modified between loops.
The only circumstances such a thing would even happen is in cases where the equivalent foreach implementation would result in an immediate InvalidOperationException, due to an enumeration error.
Additionally, the only possible case for the Count to change between it being checked and the index being accessed would be if it were modified by multi-threaded code, which would cause far bigger problems since List<T> isn't thread safe in the first place.

Still, if it would be preferred to switch to a foreach, or to put the index access inside the try/catch, I'll change it.

uGuardian:
That's actually a very good point, and I was too preoccupied with making sure the entire save doesn't get discarded and thinking about it as a stopgap fix while waiting for the Living Large version that I didn't consider that it could be say... a player, a cyclops, or an immediately critical vehicle in deep water getting deleted, and implementing a location fix for it would be far better.

Want me to look into making a fix for that immediately? I'm not sure how fast I'd get it done around work and other IRL obligations, but it shouldn't be too difficult, especially if it's just correcting the data during the loading process.


As a side note I'm not exactly sure what causes the NaN location entities to come into existence, I'd assume it's something like a failed spawn or something, but I'm far from certain on that. In the case of the server I'm part of it was a random Stalker.

uGuardian:
I realized that the parent data is usually still present, and location can be approximated from that.

Also flagging previous comments as resolved as the index based access has to be where it is now with current design.

Coding-Hen:
Do we want to add a choice for the user in saying a non debug mode? I am just thinking that the normal users will at this point go to the help channel and start saying it doesn't work, I think we should come up with a default and then let advanced users fix it themselves, what do you think?

uGuardian:
Hmm, maybe?
Right now it just says to press a button to continue and if you press anything other than D it attempts to fix it. If you press D is deletes it, so even if it's someone who would just mash the enter key, it'd still "work" for them, and they'd at least get the warning. It basically defaults to automatic repair, it just makes sure you actually have the chance to override that.
I'm not sure, what do you think tornac?

tornac1234:
With that in mind, I'd just have a message at the beginning saying
"X entities were found to be broken, do you want to manually fix them (m) or to apply a default fix for all (a) ?"
Default fix would be to do the OutOfBoundsWarp I mentionned above for all entities

uGuardian:
You sure? The NaN error is very rare, and when it occurs the new warning tells you what kind of entity it happened on. If you ever get the error on more than one or two entities max there's probably something far more significant wrong with your save.
If you'd really prefer it the other way though I can do that.

Also for the default, I currently have it put it back in the parent it was already part of, the parent most likely being a cell or other type of container (I can't think of anything else it even could be, the one I've seen being an entity with no type that contains several fish entities).
It grabs the average position of all entities in the cell and drops it there. This is a lot more accurate than just plopping it in the middle of the map. If it fails to find any other entities, it'll drop it at 0,0,0 relative to the parent, which will still be more accurate on average than dropping it somewhere in the middle of the map.
It also admittedly doesn't make me have to figure out the parent's position relative to the center of the map, and then figure out the semi-random location based on that offset.
Also thinking about it now, I have no idea what would happen if it tried to load an entity parented to a chunk in a completely different location. If it was breaking, I'd also need to figure out what to reparent it to, and also alter both relevant children lists as well.

tornac1234:
Generally it'd happen to players which were launched into space because of some Subnautica wrong collision, but it can probably happen to any entity, for which the best fix would simply to put them back into the world.

When you say it's pretty rare of an error, what do you mean ? What other entity errors are you talking about ?

If you spawn an entity in it's original cell while messing up the positions, there are good chances that they'll just end up being under the ground. I even think no computation is better than any computation around other entities' positions

uGuardian:
Sorry that this took so long, work got in my way for several days straight.

Anyways, first of all getting launched into space should never cause the NaN error. Positional data is handled as a float, meaning that no matter how far they go it should only result in floating point rounding errors. The only circumstances I can think of that would ever result in a NaN are dividing the values by zero, or it never getting initialized or serialized properly in the first place.

As for saying it's a pretty rare error, this is specifically the error mentioned in #2097 and also happened to the server I'm part of, but does not seem to occur to many people and not very often. When it does occur though it's a pretty severe bug.
It's rare enough that it should only ever happen to one or two entities at most. If more than that have the same error, there's probably a much bigger problem with the server.

Here is the relevant world data for my server: [MP.zip](https://github.com/user-attachments/files/16682738/MP.zip)
The specific offending entity in my case is this one:
`{"SpawnedByServer":true,"Transform":{"LocalPosition":{"X":"NaN","Y":"NaN","Z":"NaN"},"LocalRotation":{"X":"NaN","Y":"NaN","Z":"NaN","W":"NaN"},"LocalScale":{"X":1.0,"Y":1.0,"Z":1.0}},"TechType":"Stalker","Id":"9ff6bab0-42be-862f-7262-cc29be1a10ee","Level":1,"ClassId":"cf8794a1-5cd6-492e-8acf-7da7c940ef70","WaterParkId":null,"SerializedGameObject":null,"ExistsInGlobalRoot":false,"ParentId":"8d5f104b-08ff-48bb-c8b6-07694a74bbc9","Metadata":null,"ExistingGameObjectChildIndex":null}`
The rotation data also being NaN and the scale data just being a flat default of 1.0 seems to support the idea of it never being initialized/serialized correctly.

As for the location setting back to an estimated position based on surrounding entities in parent, I created a test set of 100 theoretical repositions and manually visited each of them precisely in-game, which I recorded in test footage here: https://youtu.be/D-Jkua4cfG8?si=vSqnkAClfveR-md4
The very first tested reposition is the reposition of the stalker that actually caused the error on my server, the few after that were specifically random leviathans, then specifically creatures of a dozen or so creature types since creatures seems to be what it most likely happens to, and then the last 50 were completely random entities.
Any entity that's labelled as having no siblings is simply centered on the parent.
Also for quick review and/or scanning through the video, here's the results sheet: [EntityFixTest.txt](https://github.com/user-attachments/files/16682643/EntityFixTest.txt)

The biggest thing that concerns me about putting things back in the center of the world is... well, let's just say I have this image of this bug happening to a Leviathan, and it getting dropped right on top of the escape pod. Alternatively, this somehow happening to a vehicle during world save, loading back in, and finding out your vehicle is now all the way back up in the shallows while you're somewhere like the lost river. Lastly the slight chance of an object getting permanently stuck up in the air that isn't supposed to be seems almost as bad as ending up under the ground to me, although creatures would be more likely to find their way back to water than they would be to swim the correct direction to exit the ground, even if they're right next to it. (Although let's be honest, creatures clipping into surfaces is already something that happens quite a bit in Subnautica)
Also to be completely honest implementing an emulated OutOfBoundsWarp would take more work since it also would involve reparenting the entity and altering the child lists of the new and old entities just to make absolute certain there wouldn't be any weird bugs, so it also has a tiny bit of laziness based motivation, although that's far from the primary reason.

Comparatively, centering it in the parent does have a fairly high chance of ending up in the ground, but if there are other entities in the parent they'll be physics bound as well, meaning if they all cluster to a specific side or avoid a specific part of the area, then there's probably an obstacle there, and centering it on sibling positions will nudge it the correct direction to not end up in the ground most of the time. Also if it were something like a plant (which I don't think this bug really happens to but it's good to be sure) it'll generally put it back into the plant cluster around the right spot.

If the above tests still produce too many bad results in your opinion, I will try to implement an emulated OutOfBoundsWarp, I'm not the one in charge here after all and will defer to your judgement on the subject, I just genuinely think putting them back roughly where they belong is the better option.

killzoms:
The issue of the NaN numbers in positions is an issue with Nitrox's calculations regarding object transformations... its a bug that I myself struggle to wrap my brain around due to lack of enough math education regarding that particular topic.

uGuardian:
Do you know offhand where in the code the bugged transformation calculations take place? I'll also take a look at that.

killzoms:
[This](https://github.com/SubnauticaNitrox/Nitrox/blob/master/NitroxModel/DataStructures/Unity/NitroxTransform.cs) is the class that handles transformations in Nitrox, it could be data that is fed to it improperly or it could originate here.

killzoms:
spawning in Nitrox where the data originates from is [here](https://github.com/SubnauticaNitrox/Nitrox/blob/master/NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs) fyi

killzoms:
for the sources of the data if curious, [here](https://github.com/SubnauticaNitrox/Nitrox/blob/master/NitroxServer-Subnautica/Resources/ResourceAssetsParser.cs) and [here](https://github.com/SubnauticaNitrox/Nitrox/blob/master/NitroxServer/Serialization/BatchCellsParser.cs)

killzoms:
Perhaps targeting the main branch of Nitrox would be better for this particular issue? The reason is that Nitrox is getting ready to release a new version soon(TM), and the team doesn't plan on releasing any more patches for the 1.7 version.

   Releasing another patch on 1.7 for one fix could give the impression that progress has stalled. Plus, the team wants to prioritize getting Nitrox in top shape for its upcoming release, while also sinking any final regressions that surfaced after the Subnautica Living Large update.

uGuardian:
Fair enough, I just had no idea exactly how long it'd be until the 2.0 version of Nitrox came out, and the bug seemed severe enough to warrant a hotfix. (Being completely unable to load the server's world is… notable to say the least)
That and me not being sure if the underlying systems in the main branch had undergone major changes are why I'd submitted a pull request to this branch specifically. If you don't want to give false impressions though it's understandable.

dartasen:
@uGuardian Indeed we're not going to release another 1.7 version. Though help is always good to take and that would be a pleasure if you're willing to get this sorted out for 1.8. Eve, though we've reworked a lot of thing for 1.8, issue is still the same.

The current fix you gave looks like a help tool rather than a proper fix. What comes to my mind is to prevent such entities to be serialized into the save file with weird values like NaN or Infinite, before anything else. Rather than dealing with the issue when you're loading the server

uGuardian:
Yup, that'd be preferable, although it's always good to have error checking in the loader.
I will try looking into the core saving issue itself over the next few weeks. If I manage to find the source of the issue I'll submit fixes.
I'll also look into adding the error checking in the main branch when I get a chance, just because caution never hurts.

dartasen:
Closing it for now

PR #2055, opened 2023-06-24
Add MainMenu Controller support

Jannify:
- [x] Make server list controller friendly
- [x] Add delete confirmation for server list
- [x] Make server add window in SN based UI
- [x] Make server add window controller friendly
- [x] Make server join window controller friendly
- [x] Make server password window controller friendly
- [x] Add loading and (error-)message window

From Clement:
- [x] Unable to scroll with the mousewheel through the server listview when it, I need to hover or focus it with a click on the scrollbar to be able to use the mousewheel scroll
- [x] No visual hint that the server has been successfully added to the list using Add Server menu
- [x] No warning before clicking on delete on a serve rentry, which is a bit frustrating when using it accidentaly
- [x] Didn't look if there's a timeout in our code but in the Loading screen might need a Cancel button if it can be a long process
- [x] When joining a reachable server, there's no preview of the color we're choosing that we used to have
- [x] Unable to use tab to switch between entry in server create menu
- [x] Maybe we could add some margin in the server listview between items. The first item is overflowing a bit on the top, you can see if you force the scroll to offset the listview
- [x] Add server button looks like a server entry which could be confusing, maybe we could slightly restyle it or move it elsewhere
- [x] Unable to connect to server text, ip could be centered

grafik

Partial-Completes #1653
Closes #1816

PR #2167, opened 2024-08-21
Fix errors/warnings

dartasen:
* Fix a oopsie
* Reverts #2141  since it need to be merged after #2055 (Weblate has been changed way before the PR), and needs some rework inside Weblate
* Add Nowarn for Newtonsoft since we'll never be able to upgrade this package
* Remove spamming warnings from DI inside Tests
* Fix few random warnings

Those are the remaining warnings, but I guess it needs extra work since technically the analyzer is right, but in the whole context it makes absolutely no sense, or add too much complexity for a simple call that will happen 0.0000001% of the time
image


NitroxModel/NitroxModel.csproj
```diff
@@ -10,7 +10,7 @@
         <PackageReference Include="LiteNetLib" Version="1.2.0" />
         <PackageReference Include="Microsoft.Win32.Registry" Version="5.0.0" />
         <PackageReference Include="Mono.Nat" Version="3.0.2" />
-        <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
+        <PackageReference Include="Newtonsoft.Json" Version="12.0.3" NoWarn="NU1091;NU1092;NU1903;NU1904" />
```
Coding-Hen, line 13: What was the reason that we can't upgrade past this? I am going to assume it clashes with something else?

NitroxClient/GameLogic/Spawning/Metadata/Processor/Abstract/EntityMetadataProcessor.cs
```diff
@@ -4,13 +4,13 @@
...
-    public abstract void ProcessMetadata(GameObject gameObject, T metadata);
+    public abstract void ProcessMetadata(GameObject gameObject, TMetadata metadata);
 
     public void ProcessMetadata(GameObject gameObject, EntityMetadata metadata)
     {
-        ProcessMetadata(gameObject, (T)metadata);
+        ProcessMetadata(gameObject, (TMetadata)metadata);
     }
 
     protected T Resolve<T>() where T : class
```
Coding-Hen, line 16: Do we not want to change the resolve on this as well?

NitroxClient/GameLogic/Spawning/Metadata/Processor/Abstract/EntityMetadataProcessor.cs
```diff
@@ -4,13 +4,13 @@
 
 namespace NitroxClient.GameLogic.Spawning.Metadata.Processor.Abstract;
 
-public abstract class EntityMetadataProcessor<T> : IEntityMetadataProcessor where T : EntityMetadata
+public abstract class EntityMetadataProcessor<TMetadata> : IEntityMetadataProcessor where TMetadata : EntityMetadata
```
Coding-Hen, line 7: is this here to just enforce it having to be of the interface type rather than just anything?

Coding-Hen (commented):
I am glad that the strings are being replaced I noticed that the other day with a rebase I did and spent ages wondering if it was me that broke it

NitroxModel/NitroxModel.csproj
dartasen: We're using subnautica dll which is 12.0.3 to avoid conflict afaik, for some reasons UWE didn't want to upgrade their packages while upgrading their game

dartasen: Eventually we'll move everything to System.Text.Json when the linux branch will come

NitroxClient/GameLogic/Spawning/Metadata/Processor/Abstract/EntityMetadataProcessor.cs
dartasen: Don't think so, T can represent everything in that context

NitroxClient/GameLogic/Spawning/Metadata/Processor/Abstract/EntityMetadataProcessor.cs
dartasen: There's a constraint that requires TMetadata to be of type EntityMetadata if that's what you mean

Measurity (approved):

PR #2169, opened 2024-08-22
Removed Nitrox.BuildTool in favor of MSBuild tasks

Measurity:
Added custom MSBuild task "DiscoverGame"
Added reference to BepInEx.AssemblyPublicizer.MSBuild to publicize DLLs within MSBuild

- [x] Add [Nitrox.Discovery repository](https://github.com/SubnauticaNitrox/Nitrox.Discovery) and upload nuget package [Nitrox.Discovery.MSBuild](https://www.nuget.org/packages/Nitrox.Discovery.MSBuild)
- [x] Fix reference issue causing some unit tests to fail with type load error


Directory.Build.targets
```diff
@@ -1,301 +1,296 @@
...
+        <PropertyGroup>
+            <GameDir>$(GameDir)\</GameDir>
+            <GameManagedDir>$(GameDir)Subnautica_Data\Managed\</GameManagedDir>
+            <GameVersionFile>$(GameDir)Subnautica_Data\StreamingAssets\SNUnmanagedData\plastic_status.ignore</GameVersionFile>
+            <GameLegacyVersion>68598</GameLegacyVersion>
+        </PropertyGroup>
+        <ReadLinesFromFile File="$(GameVersionFile)">
+            <Output TaskParameter="Lines" ItemName="GameVersion" />
+        </ReadLinesFromFile>
+        <Error Condition="@(GameVersion) &lt;= $(GameLegacyVersion)" Text="Subnautica must not be on legacy branch. Expected version &gt;$(GameLegacyVersion) but was @(GameVersion)" />
```
NinjaPedroX, line 17: Shouldn't the "Expected Version" here be the latest version of Subnautica instead of Legacy? Or is the ">" symbol meant to say that its supposed to be greater than the legacy version?

Measurity:
This PR is ready for review

Coding-Hen (approved):
It feels a bit strange having another repo as a dependency but as a msbuild task I think this is actually quite well implemented and allows us to make changes easier for discovery

dartasen (approved):
So far so good, resolve the last problem we were having on linux branch by having BuildTool project being cross-referenced everywhere. Will also allow more flexbility for upcoming complexity in our build process.

Definetely a big yes 🥳

PR #2181, opened 2024-09-25
Standardises Discord IP Getters to match LogHowToConnectAsync

### LINKED ISSUE #2156: Refactor DiscordRequestIPProcessor to use GetIP functions like Server.LogHowToConnectAsync
### Describe the issue

Adapt DiscordRequestIPProcessor.cs to use the same GetIP function as the ones used in Server.LogHowToConnectAsync (currently we don't use the fallbacks ones)

Suggested by @tornac1234

MistaOmega:
Closes #2156

### DIFF
--- a/NitroxClient/Communication/Packets/Processors/DiscordRequestIPProcessor.cs
+++ b/NitroxClient/Communication/Packets/Processors/DiscordRequestIPProcessor.cs
@@ -12,10 +12,14 @@ public class DiscordRequestIPProcessor : ClientPacketProcessor<DiscordRequestIP>
     public override async void Process(DiscordRequestIP packet)
     {
         string ipPort = await GetIpPortAsync();
+        if (ipPort == "") return;
         DiscordClient.UpdateIpPort(ipPort);
     }
 
-    /** Get the WAN IP address or the Hamachi IP address if the WAN IP address is not available. */
+    /// <summary>
+    /// Get the WAN IP address or the Hamachi IP address if the WAN IP address is not available.
+    /// </summary>
+    /// <returns>Found IP or blank string if none found</returns>
     private static async Task<string> GetIpPortAsync()
     {
         Task<IPAddress> wanIp = NetHelper.GetWanIpAsync();
@@ -24,10 +28,11 @@ public class DiscordRequestIPProcessor : ClientPacketProcessor<DiscordRequestIP>
         {
             return wanIp.Result.ToString();
         }
+
         if (await hamachiIp != null)
         {
             return hamachiIp.Result.ToString();
         }
-        return "127.0.0.1";
+        return "";
     }
 }
--- a/NitroxClient/Communication/Packets/Processors/DiscordRequestIPProcessor.cs
+++ b/NitroxClient/Communication/Packets/Processors/DiscordRequestIPProcessor.cs
@@ -1,38 +1,13 @@
-﻿using System.Net;
-using System.Threading.Tasks;
-using NitroxClient.Communication.Packets.Processors.Abstract;
+﻿using NitroxClient.Communication.Packets.Processors.Abstract;
 using NitroxClient.MonoBehaviours.Discord;
-using NitroxModel.Helper;
 using NitroxModel.Packets;
 
 namespace NitroxClient.Communication.Packets.Processors;
 
 public class DiscordRequestIPProcessor : ClientPacketProcessor<DiscordRequestIP>
 {
-    public override async void Process(DiscordRequestIP packet)
+    public override void Process(DiscordRequestIP packet)
     {
-        string ipPort = await GetIpPortAsync();
-        if (ipPort == "") return;
-        DiscordClient.UpdateIpPort(ipPort);
-    }
-
-    /// <summary>
-    /// Get the WAN IP address or the Hamachi IP address if the WAN IP address is not available.
-    /// </summary>
-    /// <returns>Found IP or blank string if none found</returns>
-    private static async Task<string> GetIpPortAsync()
-    {
-        Task<IPAddress> wanIp = NetHelper.GetWanIpAsync();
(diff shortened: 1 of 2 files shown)


NitroxClient/Communication/Packets/Processors/DiscordRequestIPProcessor.cs
```diff
@@ -1,13 +1,33 @@
...
 {
-    public override void Process(DiscordRequestIP packet)
+    public override async void Process(DiscordRequestIP packet)
     {
-        DiscordClient.UpdateIpPort(packet.IpPort);
+        string ipPort = await GetIpPortAsync();
+        DiscordClient.UpdateIpPort(ipPort);
+    }
+
+    /** Get the WAN IP address or the Hamachi IP address if the WAN IP address is not available. */
```
Measurity, line 18: Pls no java style code docs, these aren't parsed in IDEs or doc export tools.

```suggestion
    /// <summary>
    ///     Get the WAN IP address or the Hamachi IP address if the WAN IP address is not available.
    /// </summary>
```

Measurity (changes requested):
The GetIp refactor should be done on server-side (not client side). Server-side also has `DiscordRequestIPProcessor` type which.

I think only hamachi IP should be added as fallback when the `wanIp` fails. It also shouldn't return localhost ip as the server can be running on another machine.

NitroxClient/Communication/Packets/Processors/DiscordRequestIPProcessor.cs
```diff
@@ -1,13 +1,33 @@
...
+        Task<IPAddress> hamachiIp = Task.Run(NetHelper.GetHamachiIp);
+        if (await wanIp != null)
+        {
+            return wanIp.Result.ToString();
+        }
+        if (await hamachiIp != null)
+        {
+            return hamachiIp.Result.ToString();
+        }
+        return "127.0.0.1";
```
Jannify, line 31: Don't think we should return the local IP here. Two different discord client on the same machine sound strange.

NitroxClient/Communication/Packets/Processors/DiscordRequestIPProcessor.cs
```diff
@@ -1,13 +1,33 @@
...
 using NitroxClient.MonoBehaviours.Discord;
+using NitroxModel.Helper;
 using NitroxModel.Packets;
 
 namespace NitroxClient.Communication.Packets.Processors;
 
 public class DiscordRequestIPProcessor : ClientPacketProcessor<DiscordRequestIP>
 {
-    public override void Process(DiscordRequestIP packet)
+    public override async void Process(DiscordRequestIP packet)
```
Jannify, line 12: Is this async method correctly run/handled from our framework? If not we could just Use Task.Run inside of it, no?

NitroxClient/Communication/Packets/Processors/DiscordRequestIPProcessor.cs
```diff
@@ -1,13 +1,33 @@
...
     {
-        DiscordClient.UpdateIpPort(packet.IpPort);
+        string ipPort = await GetIpPortAsync();
+        DiscordClient.UpdateIpPort(ipPort);
+    }
+
+    /** Get the WAN IP address or the Hamachi IP address if the WAN IP address is not available. */
+    private static async Task<string> GetIpPortAsync()
```
Jannify, line 19: If this is the same method as (used) in `LogHowToConnectAsync`, why not move it to NetHelper?

Jannify (changes requested):

NitroxClient/Communication/Packets/Processors/DiscordRequestIPProcessor.cs
Measurity: I think this would get messy from an API design point. You'd need to return a list of IPs back but you don't know which is wan and which is hamachi unless you return a key-value pair back. Which imo is too much code for just 2 usages.

NitroxClient/Communication/Packets/Processors/DiscordRequestIPProcessor.cs
Measurity: Yep. It makes sense if on client-side, but this code should be running server-side. So returning localhost is no good.

NitroxClient/Communication/Packets/Processors/DiscordRequestIPProcessor.cs
MistaOmega: It seems to run correctly and finds my WAN IP. So I *think* we can keep as is, if you're happy

Measurity (changes requested):
> The GetIp refactor should be done on server-side (not client side). Server-side also has DiscordRequestIPProcessor type.

Did you do this?

NitroxServer/Communication/Packets/Processors/DiscordRequestIPProcessor.cs
```diff
@@ -32,15 +33,34 @@ public override void Process(DiscordRequestIP packet, Player player)
...
-        packet.IpPort = ipPort = $"{wanIp.Result}:{serverConfig.ServerPort}";
+        packet.IpPort = ipPort = $"{result}:{serverConfig.ServerPort}";
         player.SendPacket(packet);
     }
+
+    /// <summary>
+    /// Get the WAN IP address or the Hamachi IP address if the WAN IP address is not available.
+    /// </summary>
+    /// <returns>Found IP or blank string if none found</returns>
+    private static async Task<string> GetIpPortAsync()
```
Measurity, line 51: I think this should be `GetIpAsync`

Measurity (approved):
Apart from function name, it looks good to me.

Coding-Hen (approved):

Measurity (approved):

Jannify (approved):
LGTM CW

Issue #2159, opened 2024-08-09
Defocus chat input field on enter and refocus on 'Y' press


Measurity:
### Describe the issue

Chat doesn't defocus when you press enter (if you hold enter and type you'll immediately submit the letter).
Make it so that:
- [x] `Enter` defocuses the input

This makes it more inline with other games like Counter Strike 2 or Deep Rock Galactic.

## Related code
https://github.com/SubnauticaNitrox/Nitrox/blob/cfc08f3cc1174e551bddfb2dabaec1f6f34cc1a3/NitroxClient/MonoBehaviours/Gui/Chat/PlayerChatInputField.cs#L101-L115

dartasen:
#2182 Did fullfill the first point

Risusama:
Just to clarify, what are the situations counted as "in-game cursor active" in which we would like the chat to refocus?
 - ESC menu?
 - Tab-menu (inventory, databank etc.)?
 - something else?

PR #2195, opened 2024-11-12
Allow chat to refocus from any view

Risusama:
Such as Tab menu (inventory, databank, etc.), ESC menu and such.
Fulfills the second point of #2159.

Could have done this in the previous PR, but being tired at the time I didn't realize there was a second point in the issue.

### DIFF
--- a/NitroxClient/MonoBehaviours/Gui/Input/KeyBindings/Actions/ChatKeyBindingAction.cs
+++ b/NitroxClient/MonoBehaviours/Gui/Input/KeyBindings/Actions/ChatKeyBindingAction.cs
@@ -7,11 +7,7 @@ namespace NitroxClient.MonoBehaviours.Gui.Input.KeyBindings.Actions
     {
         public override void Execute()
         {
-            // If no other UWE input field is currently active then allow chat to open.
-            if (FPSInputModule.current.lastGroup == null)
-            {
-                NitroxServiceLocator.LocateService<PlayerChatManager>().SelectChat();
-            }
+            NitroxServiceLocator.LocateService<PlayerChatManager>().SelectChat();
         }
     }
 }


NitroxClient/MonoBehaviours/Gui/Input/KeyBindings/Actions/ChatKeyBindingAction.cs
```diff
@@ -7,11 +7,7 @@ public class ChatKeyBindingAction : KeyBindingAction
     {
         public override void Execute()
         {
-            // If no other UWE input field is currently active then allow chat to open.
-            if (FPSInputModule.current.lastGroup == null)
```
dartasen, line 11: Does it work as attended when writing on a base sign for example ?

Risusama: Hm. Good point, I'll test on a different save, gimme a few minutes.

Risusama: Okay it'll take some time until I'm able to test it, my more advanced save is on the legacy branch and the master has only a save at the starting point :D

Jannify: You know that you can just cheat what you need for testing? Even spawn in creative if set up in the server.

dartasen: `/changeservergamemode creative` inside chat or `changeservergamemode creative`` console if you don't know how to use IG command. or you can edit server config file accordingly as Jannify said

Risusama: Ah, yeah I found out there was creative mode available while I was transferring save, then just had difficulties figuring out how to switch to creative on the fly. :D

But yeah, it seems it will also override sign text input, so no good like this. Hmm.

Risusama:
Ooh well, I better study C# some more. Not really understanding quite a lot of things. Closing, as I won't be able to solve this any time soon..

I thought what could work, was replacing the "null" with the "group" which signs, locker tags etc. belong to, but either I can't figure out what it is or I don't know how to work it. (uGUI_SignInput didn't work, or I'm doing it wrong. Says "'uGUI_SignInput' is a type, which is not valid in the given context").

I think this is some Unity specific thing? I don't really find much documentation on it either. Unless it's a Subnautica specific thing, then I guess there won't even be documentation for it :D 

I give up XD Sorry to have taken your time like this. 🙏

PR #2196, opened 2024-11-12
Fix items unequipping when rejoining the server

### LINKED ISSUE #2186: Equipped items are unequipped when rejoining the server

Yes, I'm on the latest release of Nitrox.

Windows

Steam

### Describe the issue.

When my friends and I are reconnecting to the server, our players are being unequipped every time.
Also the batteries in the charging station are disappearing

This happens on MSStore version, as well as the Steam version 

1. Join server
2. Equip stuff / place batteries in charger
3. Disconnect
4. Stop server
5. Start server
6. Connect again

### Relevant log output

Game: https://pastebin.com/RGf2Hpw5
Server: https://pastebin.com/shaiUdUX

SpaceMonkeyy86:
We may want to redo this implementation later if the server needs to handle more metadata updates itself. For now, this will work. This fix will help with testing so I think it's good if we merge it how it is.
Resolves #2186


dartasen:
Works fine IG

NitroxClient/GameLogic/InitialSync/EquippedItemInitialSyncProcessor.cs
```diff
@@ -32,44 +35,37 @@ public override IEnumerator Process(InitialPlayerSync packet, WaitScreen.ManualW
 
         using (PacketSuppressor<EntitySpawnedByClient>.Suppress())
         {
-            foreach (EquippedItemData equippedItem in packet.EquippedItems)
+            foreach (var pair in packet.EquippedItems)
```
tornac1234, line 38: We don't use var cause it's harder to directly see what's the type of the pair in question even if we have its decomposition below.

NitroxServer/Communication/Packets/Processors/EntityMetadataUpdateProcessor.cs
```diff
@@ -1,9 +1,11 @@
...
+using NitroxModel.DataStructures.GameLogic;
+using NitroxModel.DataStructures.GameLogic.Entities.Metadata;
 using NitroxModel.DataStructures.Util;
 using NitroxModel.Packets;
 using NitroxServer.Communication.Packets.Processors.Abstract;
 using NitroxServer.GameLogic;
 using NitroxServer.GameLogic.Entities;
+using System.Linq;
 
 namespace NitroxServer.Communication.Packets.Processors
```
tornac1234, line 10: I suggest a file full rewrite, improving what you did so we can already have that "serious implementation" you mentionned below.
First, convert to file-scoped namespace (namespace ``NitroxServer.Communication.Packets.Processors;``)

NitroxServer/Communication/Packets/Processors/EntityMetadataUpdateProcessor.cs
```diff
@@ -45,5 +49,22 @@ private void SendUpdateToVisiblePlayers(EntityMetadataUpdate packet, Player send
                 }
             }
```
tornac1234, line 50: Just replace the whole Process code by:
```cs
if (!entityRegistry.TryGetEntityById(packet.Id, out Entity entity))
{
    Log.Error($"Entity metadata updated on an entity unknown to the server {packet.Id} {packet.NewValue.GetType()} ");
    return;
}

if (ProcessMetadata(sendingPlayer, entity, packet.NewValue))
{
    entity.Metadata = packet.NewValue;
    SendUpdateToVisiblePlayers(packet, sendingPlayer, entity);
}
```
I like TryGetEntityById because it helps avoiding the optional layer (.Value). Then we can have ProcessMetadata be kind of a security filter for our data

NitroxServer/Communication/Packets/Processors/EntityMetadataUpdateProcessor.cs
```diff
@@ -45,5 +49,22 @@ private void SendUpdateToVisiblePlayers(EntityMetadataUpdate packet, Player send
...
+                if (player != null)
+                {
+                    player.EquippedItems.Clear();
+                    foreach (PlayerMetadata.EquippedItem item in playerMetadata.EquippedItems)
+                    {
+                        player.EquippedItems.Add(item.Slot, item.Id);
+                    }
+                }
+            }
+        }
```
tornac1234, line 68:
```suggestion
    private bool ProcessMetadata(Player sendingPlayer, Entity entity, EntityMetadata metadata)
    {
        return metadata switch
        {
            PlayerMetadata playerMetadata => ProcessPlayerMetadata(sendingPlayer, entity, playerMetadata),
            // By default, entities have no security layer
            _ => true,
        };
    }

    private static bool ProcessPlayerMetadata(Player sendingPlayer, Entity entity, PlayerMetadata playerMetadata)
    {
        // Security layer to ensure a player can only modify its own entity
        if (sendingPlayer.GameObjectId == entity.Id)
        {
            sendingPlayer.EquippedItems.Clear();
            foreach (PlayerMetadata.EquippedItem item in playerMetadata.EquippedItems)
            {
                sendingPlayer.EquippedItems.Add(item.Slot, item.Id);
            }
            return true;
        }
        
        Log.WarnOnce($"Player: {sendingPlayer.Name} tried updating metadata of another player's entity {entity.Id}");
        return false;
    }
```
A switch for ``ProcessMetadata`` will be required for the efficiency this method needs

tornac1234 (changes requested):
Pretty good fix, for which I'd like a lil improvement already (even though we could have as well "cleaned" this later on, I think the PR is is not too big so we can improve it already)

NitroxModel/Packets/InitialPlayerSync.cs
```diff
@@ -14,7 +14,7 @@ namespace NitroxModel.Packets
     public class InitialPlayerSync : Packet
     {
         public NitroxId AssignedEscapePodId { get; }
-        public List<EquippedItemData> EquippedItems { get; }
+        public Dictionary<string, NitroxId> EquippedItems { get; }
```
Jannify, line 17: `Nitrox.Test/Server/Serialization/WorldPersistenceTest.cs` is throwing errors with this change. The test needs to be addressed.

This should work:
```
            AssertHelper.IsDictionaryEqual(playerData.EquippedItems, playerDataAfter.EquippedItems, (keyValuePair, keyValuePairAfter) =>
            {
                Assert.AreEqual(keyValuePair.Key, keyValuePairAfter.Key);
                Assert.AreEqual(keyValuePair.Value, keyValuePairAfter.Value);
            });
            ```

File: NitroxServer/Communication/Packets/Processors/EntityMetadataUpdateProcessor.cs [hunk 6]
```diff
@@ -1,49 +1,76 @@
...
+            if (player != sendingPlayer && updateVisibleToPlayer)
             {
-                Log.Error($"Entity metadata updated on an entity unknown to the server {packet.Id} {packet.NewValue.GetType()} ");
+                player.SendPacket(packet);
             }
         }
+    }
 
-        private void SendUpdateToVisiblePlayers(EntityMetadataUpdate packet, Player sendingPlayer, Entity entity)
+    private bool ProcessMetadata(Player sendingPlayer, Entity entity, EntityMetadata metadata)
```
Jannify, line 49: Would rename to `ShouldProcessMetadata`

Jannify (changes requested):
From my investigation the following files can be removed with the changes:

- NitroxClient/GameLogic/Helper/SerializationHelper.cs
- NitroxPatcher/Patches/Dynamic/UniqueIdentifier_Id_Getter_Patch.cs

SpaceMonkeyy86:
@Jannify Changes have been addressed.

NitroxServer/GameLogic/Players/PersistedPlayerData.cs
```diff
@@ -23,36 +23,33 @@ public class PersistedPlayerData
...
+    public Dictionary<string, NitroxId> EquippedItems { get; set; } = new Dictionary<string, NitroxId>();
 
     [DataMember(Order = 5)]
-    public List<EquippedItemData> Modules { get; set; } = new List<EquippedItemData>();
-
-    [DataMember(Order = 6)]
     public ushort Id { get; set; }
 
-    [DataMember(Order = 7)]
+    [DataMember(Order = 6)]
```
Coding-Hen, line 31: I think in general we want to avoid having changes to his data structures but as we are going to breaking change this release I think it should be fine

NitroxServer/Serialization/SaveDataUpgrades/Upgrade_V1800.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace NitroxServer.Serialization.SaveDataUpgrades
+{
+    public sealed class Upgrade_V1800 : SaveDataUpgrade
+    {
+        public override Version TargetVersion { get; } = new Version(1, 8, 0, 0);
+
+        protected override void UpgradeWorldData(JObject data)
+        {
+            data["InventoryData"]["EquippedItems"] = new JObject(new Dictionary<string, NitroxId>());
```
Coding-Hen, line 15: Is this actually upgrading? Do we want to if we can move old to new to the other? It might not be possible with these changes but we should try if we can

NitroxServer/Communication/Packets/Processors/EntityMetadataUpdateProcessor.cs
```diff
@@ -1,49 +1,76 @@
...
+            Log.Error($"Entity metadata {packet.NewValue.GetType()} updated on an entity unknown to the server {packet.Id}");
+            return;
         }
 
-        public override void Process(EntityMetadataUpdate packet, Player sendingPlayer)
+        if (TryProcessMetadata(sendingPlayer, entity, packet.NewValue))
         {
-            Optional<Entity> entity = entityRegistry.GetEntityById(packet.Id);
+            entity.Metadata = packet.NewValue;
+            SendUpdateToVisiblePlayers(packet, sendingPlayer, entity);
```
Coding-Hen, line 32: Does only sending to visible players here mean anyone outside of view doesn't get the updates and so will be out of sync?

NitroxServer/Serialization/SaveDataUpgrades/Upgrade_V1800.cs
Jannify: Bc of the massive changes to the base piece data we unfortunately can not provide upgradability from V1.7.

NitroxServer/Communication/Packets/Processors/EntityMetadataUpdateProcessor.cs
Coding-Hen: Thought this was the case just wanted to check as haven't touched this part of the code base in a while

Jannify (approved):
LGTM CW

Needs in game testing to verify the dead code which was removed (PlantableItemData in particular) doesn't result in broken features

tornac1234 (approved):
Worked fine from IG test and CW

Coding-Hen (approved):

PR #2197, opened 2024-11-14
Improve vehicle destruction sync

SpaceMonkeyy86:
Properly moves vehicle destroying to the new entities system.


SpaceMonkeyy86:
Cyclops wrecks are also synced now

dartasen:
Works fine IG :

* We can see seamoth & exosuit being destroyed
* We can see cyclops exploding
* Wrecks are still here upon re-joining (with items, upgrades in it)

Bugs we noticed : 
* Wrecks will re-explode upon re-joining
* Decoys slots are gone upon re-joining

SpaceMonkeyy86:
> Works fine IG :
> 
>     * We can see seamoth & exosuit being destroyed
> 
>     * We can see cyclops exploding
> 
>     * Wrecks are still here upon re-joining (with items, upgrades in it)
> 
> 
> Bugs we noticed :
> 
>     * Wrecks will re-explode upon re-joining
> 
>     * Decoys slots are gone upon re-joining

Both of these should be fixed now.

Jannify (changes requested):

Jannify (approved):
In `SubRoot_SetCyclopsUpgrades_Patch.cs#L18` you have the return in the same line. Nitrox code style normally says that we do brackets for every if.

Besides that LGTM CW

tornac1234 (approved):
clear enough CW

PR #2205, opened 2024-12-07
Fix teleport commands not working correctly inside a subroot

SpaceMonkeyy86:


NitroxPatcher/Patches/Dynamic/Player_OnPlayerPositionCheat_Patch.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace NitroxPatcher.Patches.Dynamic;
+
+// This method sets the player's subroot to null, so send a packet accordingly
+public sealed partial class Player_OnPlayerPositionCheat_Patch : NitroxPatch, IDynamicPatch
+{
+    private static readonly MethodInfo TARGET_METHOD = Reflect.Method((Player t) => t.OnPlayerPositionCheat());
+
+    public static void Prefix(Player __instance)
+    {
+        if (__instance.GetCurrentSub() != null)
```
Jannify, line 15: `__instance.GetCurrentSub()` returns a MB and therefor should not be compared to null directly (it's a unity shenanigan).
You can use `if (__instance.GetCurrentSub())` instead of checking against null, kinda like in c++ with pointers.

Jannify (changes requested):

dartasen (approved):
Works fine IG

Tested with 2 local players : 
* In bases
* In cyclops
* In water
* Inside EscapePod

Coding-Hen (approved):
Tested and works fine for me

PR #2207, opened 2024-12-15
Log if stream.Read() reads not all requested data.

Jannify:


Coding-Hen (approved):

Jannify:
Is already fixed in #1848

PR #2211, opened 2024-12-22
Make sure all unit tests are passing

SpaceMonkeyy86:
There are still two unit tests (PacketSerializationTest and PacketProcessorTest) that pass when ran individually but fail when ran alongside all the other tests. This might be due to race conditions or problems with how assemblies get loaded during tests.


dartasen:
We've improved many unit tests in #1848 to be less flaky (OptionalTests, DI tests, ...)

dartasen (approved):

PR #2217, opened 2025-01-01
Last minute additions

tornac1234:
- [x] Keypad sync fix
- [x] Fix falling through the Aurora when spawning inside

I might want to add a things or two before release which I'll put in here

### DIFF
--- a/NitroxClient/GameLogic/Terrain.cs
+++ b/NitroxClient/GameLogic/Terrain.cs
@@ -2,10 +2,13 @@ using System.Collections;
 using System.Collections.Generic;
 using NitroxClient.Communication.Abstract;
 using NitroxClient.Map;
+using NitroxClient.Unity.Helper;
+using NitroxModel.Core;
 using NitroxModel.DataStructures.GameLogic;
 using NitroxModel.Packets;
 using NitroxModel_Subnautica.DataStructures;
 using UnityEngine;
+using UWE;
 using WorldStreaming;
 
 namespace NitroxClient.GameLogic;
@@ -77,11 +80,23 @@ public class Terrain
 
     }
 
+    public static void WaitForEntities()
+    {
+        // In case the player is spawned in the air, we need to hold them up while all the entities load around them
+        if (Player.main && !Player.main.IsUnderwater() && !Player.main.groundMotor.grounded)
+        {
+            Player.main.cinematicModeActive = true;
+            CoroutineHost.StartCoroutine(WaitForWorldLoad(Yielders.WaitFor2Seconds));
+        }
+    }
+
     /// <summary>
     /// Forces world streamer's to load the terrain around the MainCamera and waits until it's done to unfreeze the player.
     /// </summary>
-    public static IEnumerator WaitForWorldLoad()
+    public static IEnumerator WaitForWorldLoad(WaitForSeconds initialWait = null)
     {
+        Entities entities = NitroxServiceLocator.LocateService<Entities>();
+
         // In WorldStreamer.CreateStreamers() three coroutines are created to constantly call UpdateCenter() on the streamers
         // We force these updates so that the world streamer gets busy instantly
         WorldStreamer streamerV2 = LargeWorldStreamer.main.streamerV2;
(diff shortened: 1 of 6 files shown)


NitroxModel/DataStructures/GameLogic/Entities/Metadata/KeypadMetadata.cs
```diff
@@ -1,30 +1,33 @@
...
+    [DataMember(Order = 1)]
+    public bool Unlocked { get; }
 
-        [IgnoreConstructor]
-        protected KeypadMetadata()
-        {
-            // Constructor for serialization. Has to be "protected" for json serialization.
-        }
+    [DataMember(Order = 2)]
+    public string PathFromRoot { get; }
```
dartasen, line 15: Is this really a metadata ?

NitroxClient/GameLogic/Terrain.cs
```diff
@@ -77,11 +80,23 @@ public void UpdateVisibility()
...
+        }
+    }
+
     /// <summary>
     /// Forces world streamer's to load the terrain around the MainCamera and waits until it's done to unfreeze the player.
     /// </summary>
-    public static IEnumerator WaitForWorldLoad()
+    public static IEnumerator WaitForWorldLoad(WaitForSeconds initialWait = null)
     {
+        Entities entities = NitroxServiceLocator.LocateService<Entities>();
```
dartasen, line 98: You should be using dependency injection, this will generate a warning from our analyzers

dartasen (changes requested):
Shouldn't this be separate PR's ?

NitroxClient/GameLogic/Terrain.cs
tornac1234: It won't work in this static context

NitroxModel/DataStructures/GameLogic/Entities/Metadata/KeypadMetadata.cs
tornac1234: This metadata is applied to an object which doesn't necessarily hold the KeypadDoorConsole component. Thus we need a way to find the component, which varies from one prefab to another

NitroxClient/GameLogic/Terrain.cs
dartasen: Terrain is already being required inside Mutiplayer mb, so static can be removed

NitroxModel/DataStructures/GameLogic/Entities/Metadata/KeypadMetadata.cs
Coding-Hen: Can we not do this by extracting it from the object it's on rather than using the path?

tornac1234: GetComponentInChildren is quite costy and might pick the wrong door is there are two available in the children

NitroxClient/GameLogic/Terrain.cs
```diff
@@ -77,11 +80,23 @@ public void UpdateVisibility()
 
     }
 
+    public static void WaitForEntities()
+    {
+        // In case the player is spawned in the air, we need to hold them up while all the entities load around them
+        if (Player.main && !Player.main.IsUnderwater() && !Player.main.groundMotor.grounded)
```
tornac1234, line 86: Note to fix: this should not trigger when inside a subroot/escape pod

AtlantaDEV-ui:
// KeypadMetadata.cs
[DataContract]
public class KeypadMetadata
{
    // Constructor for serialization. Has to be "protected" for json serialization.
    protected KeypadMetadata() { }

    [DataMember(Order = 2)]
    public string PathFromRoot { get; }

    // Additional properties and methods as needed
}

// Terrain.cs
public class Terrain : MonoBehaviour
{
    private Entities _entities;

    // Constructor or initialization method to inject dependencies
    public Terrain(Entities entities)
    {
        _entities = entities;
    }

    public void WaitForEntities()
    {
        // In case the player is spawned in the air, we need to hold them up while all the entities load around them
        if (Player.main && !Player.main.IsUnderwater() && !Player.main.groundMotor.grounded)
        {
            // Check if the player is inside a subroot or escape pod
            if (!IsInsideSubrootOrEscapePod(Player.main))
            {
                // Logic to hold the player up
            }
        }
    }

    private bool IsInsideSubrootOrEscapePod(Player player)
    {
        // Implement logic to determine if the player is inside a subroot or escape pod
        return false; // Placeholder return value
    }
}

**Make sure to replace the placeholder logic in IsInsideSubrootOrEscapePod with the actual implementation that checks the player's position relative to the subroot or escape pod.**

CherrieTheShifter:

dartasen:
Closing for now

PR #2242, opened 2025-01-06
Fix item drop sync on death

tornac1234:
I fixed a lot of bad things concerning cell/entity spawning, within the leviathans PR which is now merged.
This patch is now useless.


Coding-Hen (approved):
LGTM short and sweet

PR #2254, opened 2025-01-12
Fix fake constructables being moved to GlobalRoot

tornac1234:
Fake constructables being the ones you encounter in the world, placed on the ground only for the player to deconstruct them and gain some materials from it


Jannify (approved):
LGTM CW

dartasen:
Tested InGame with Ninja
* go to wreck13
* deconstruct the table and the big box
* go to wreck 8
* go back to wreck13, the table and big box will respawn

tornac1234:
> Tested InGame with Ninja
> 
>     * go to wreck13
> 
>     * deconstruct the table and the big box
> 
>     * go to wreck 8
> 
>     * go back to wreck13, the table and big box will respawn and will break sync because it's not registered on the server

Seems weird to me because with this all the latest work merged (especially #2263), none of those are deconstructables. I am now unable to find constructable stuff in the wilderness.

Though in this PR, I changed the "constructable -> globalroot" behaviour's target from "all entities" to "only buildings" which is the accurate way of it, which should be reviewed as is.

tornac1234:
New tests showed it worked. My guess is that Clement didn't pull the latest modifications for the test before

PR #2255, opened 2025-01-12
Add sync for stalkers

### LINKED ISSUE #2243: Stalker teeth invisible?

Windows

Epic

### Describe the issue.

изображение
There no one stalker teeth in game. They can be scanned, but, like on screenshot, they doesn't appear on client, idk what is this

_No response_

### Relevant log output

_No response_

### LINKED ISSUE #2250: Add sync for Stalker
### Describe the issue

Currently some importants actions are not synced for the stalker, i.e: Tooth lose (used for base crafting, & crafting)

dartasen:
- [x] Stalker FindNest : [O] : leashPosition can be changed by FindNest
- [x] Stalker DropTooth : [O] : RemotePlayer cause DropTooth to duplicate items
- [x] AttackLastTarget (Done through current system)
- [x] MoveTowardsTarget (Done through current system)
- [x] FleeOnDamage : [O] StopPerform might change Creature.leashPosition
- [x] SwimToHeroPeeper : [CAW] 
- [x] AggressiveWhenSeeTarget (Done through current system)
- [x] MeleeAttack
- [x] PrisonPredatorSwimToPlayer : [O] Add support for RemotePlayer target
- [x] ~CreatureFlinch~ [X]
- [x] ~CreatureFear~ [X]
- [x] ~PrisonCreatureBehaviour~ [X]
- [x] ~AvoidObstacles~ [X]
- [x] ~StayAtLeashPosition~ [X]
- [x] AvoidEscapePod [O] : StopPerform might change Creature.leashPosition

Will need work after this PR : 
- MeleeAttack (Pickup held item from remote player)
- CreatureFriend System
- CollectShiny

Closes #2250
Closes #2243 
Contributes to #2627


NitroxPatcher/Patches/Dynamic/Stalker_CheckLoseTooth_Patch.cs
```diff
@@ -8,22 +9,48 @@ public sealed partial class Stalker_CheckLoseTooth_Patch : NitroxPatch, IDynamic
 {
     private static readonly MethodInfo TARGET_METHOD = Reflect.Method((Stalker t) => t.CheckLoseTooth(default(GameObject)));
 
-    //GetComponent<HardnessMixin> was returning null for everything instead of a HardnessMixin with a hardness value. Since this component
-    //isn't used for anything else than the stalker teeth drop, we hard-code the values and bingo.
+    // HardnessMixin seems to be a bit buggy (ie: undefined values for some scraps, which is a vanilla bug), so we'll just hard-code the values for now.
+    // Note that HardnessMixin is only used by Stalkers
     public static bool Prefix(Stalker __instance, GameObject target)
```
Coding-Hen, line 14: Is there a reason we can't trust the logic from the original method that does these checks for how the loose teeth works and we have to do the spawning etc for ourselves?

dartasen: Would need to transpile the LooseTooth method, because I need the gameObject to broadcast the item packet
But since I already have this patch that invalidate the whole original method, and LooseTooth is only use here. It can simplify a bit.

Coding-Hen: Sounds sensible to me. Just thought I would question it as to the reasoning behind it. Creating a transpiler without need doesn't seem sensible in this case then

AtlantaDEV-ui:
using UnityEngine;
using Nitrox.Patches;

public class StalkerSync : MonoBehaviour
{
    private Stalker stalker;
    private Vector3 lastPosition;
    private bool isFleeing;

    void Start()
    {
        stalker = GetComponent<Stalker>();
        lastPosition = stalker.transform.position;
        isFleeing = false;
    }

    void Update()
    {
        SyncStalkerPosition();
        SyncStalkerBehavior();
    }

    private void SyncStalkerPosition()
    {
        if (Vector3.Distance(stalker.transform.position, lastPosition) > 0.1f)
        {
            lastPosition = stalker.transform.position;
            // Send position update to other players
            SendPositionUpdate(lastPosition);
        }
    }

    private void SyncStalkerBehavior()
    {
        if (stalker.IsFleeing() != isFleeing)
        {
            isFleeing = stalker.IsFleeing();
            // Send behavior update to other players
            SendBehaviorUpdate(isFleeing);
        }
    }

    private void SendPositionUpdate(Vector3 position)
    {
        // Code to send position update over the network
        // For example, using RPC or another networking method
    }

    private void SendBehaviorUpdate(bool fleeing)
    {
        // Code to send behavior update over the network
        // For example, using RPC or another networking method
    }
}


Explanation of the Code:
Class StalkerSync: This class is responsible for synchronizing stalkers.
Start(): Initializes the stalker and sets the initial position.
Update(): This method checks for changes in the position and behavior of the stalker.
SyncStalkerPosition(): Checks if the stalker's position has changed and sends an update if it has.
SyncStalkerBehavior(): Checks if the stalker's behavior (e.g., whether it is fleeing) has changed and sends an update.
SendPositionUpdate() and SendBehaviorUpdate(): These methods should contain the code to send data over the network.
Notes:
You will need to integrate this code into the existing architecture of the Nitrox mod.
Make sure you have access to networking methods to send updates to other players.
You may need to adapt the code depending on the specific implementation of stalkers in your mod.
This code is a basic example and may require further refinement based on your needs and the architecture of the project.

NitroxPatcher/Patches/Dynamic/MeleeAttack_OnTouch_Patch.cs
```diff
@@ -0,0 +1,36 @@
...
+            Resolve<SimulationOwnership>().HasAnyLockType(creatureId))
+        {
+            return true;
+        }
+
+        return false;
+    }
+
+    // TODO: Add transpiler to add support for held item eat from remote players, we might need to add an equivalent of Inventory.GetHeldItem()
+    // For the MeleeAttack part , it'll natively work since RemotePlayer should have a LiveMixin
```
Coding-Hen, line 35:
```suggestion
    // For the MeleeAttack part, it'll natively work since RemotePlayer should have a LiveMixin
```
Sorry I know it's a todo but

NitroxPatcher/Patches/Dynamic/PrisonPredatorSwimToPlayer_Perform_Patch.cs
```diff
@@ -0,0 +1,49 @@
...
+        // Replace everything with
+        // return EvaluatePriority(__instance, creature, time);
+        yield return new CodeInstruction(OpCodes.Ldarg_0);
+        yield return new CodeInstruction(OpCodes.Ldarg_1);
+        yield return new CodeInstruction(OpCodes.Ldarg_2);
+        yield return new CodeInstruction(OpCodes.Call, Reflect.Method(() => Perform(default, default, default)));
+        yield return new CodeInstruction(OpCodes.Ret);
+    }
+
+    public static void Perform(PrisonPredatorSwimToPlayer instance, Creature creature, float time)
```
Coding-Hen, line 26: Is there a reason we have to implement these methods ourselves and can't hook into the current returns/logic?

Coding-Hen (changes requested):

NitroxPatcher/Patches/Dynamic/PrisonPredatorSwimToPlayer_Perform_Patch.cs
dartasen: I added the explanation along with the reviews you made, tell me if it's clear enough

Coding-Hen: Yeah that's a good explanation, nice work

NitroxPatcher/Patches/Dynamic/Stalker_CheckLoseTooth_Patch.cs
```diff
@@ -8,22 +10,58 @@ public sealed partial class Stalker_CheckLoseTooth_Patch : NitroxPatch, IDynamic
...
+            // Code from Stalker.LoseTooth()
+
+            GameObject toothGameObject = Object.Instantiate<GameObject>(__instance.toothPrefab);
+            toothGameObject.transform.position = __instance.loseToothDropLocation.transform.position;
+            toothGameObject.transform.rotation = __instance.loseToothDropLocation.transform.rotation;
+
+            if (toothGameObject.activeSelf && __instance.isActiveAndEnabled)
+            {
+                Collider[] componentsInChildren = toothGameObject.GetComponentsInChildren<Collider>();
+                for (int i = 0; i < componentsInChildren.Length; i++)
```
Jannify, line 54: Would make it a foreach

NitroxPatcher/Patches/Dynamic/Stalker_FindNest_Patch.cs
```diff
@@ -0,0 +1,39 @@
...
+namespace NitroxPatcher.Patches.Dynamic;
+
+public sealed partial class Stalker_FindNest_Patch : NitroxPatch, IDynamicPatch
+{
+    internal static readonly MethodInfo TARGET_METHOD = Reflect.Method((Stalker t) => t.FindNest());
+
+    public static bool Prefix(Stalker __instance)
+    {
+        if (__instance.TryGetNitroxId(out NitroxId creatureId)
+            && Resolve<SimulationOwnership>().HasAnyLockType(creatureId))
```
Jannify, line 17:
```suggestion
        if (__instance.TryGetNitroxId(out NitroxId creatureId) &&
            Resolve<SimulationOwnership>().HasAnyLockType(creatureId))
```

NitroxPatcher/Patches/Dynamic/Stalker_FindNest_Patch.cs
```diff
@@ -0,0 +1,39 @@
...
+            return true;
+        }
+
+        return false;
+    }
+
+    public static void Postfix(Stalker __instance, bool __result)
+    {
+        // If the original method returned false, we don't want to do anything since leashPosition didn't change
+        if (!__result)
```
Jannify, line 28: Does this also work when Prefix skips the original method?

NitroxPatcher/Patches/Dynamic/PrisonPredatorSwimToPlayer_Evaluate_Patch.cs
```diff
@@ -0,0 +1,116 @@
...
+{
+    internal static readonly MethodInfo TARGET_METHOD = Reflect.Method((PrisonPredatorSwimToPlayer t) => t.Evaluate(default(Creature), default(float)));
+
+    internal static readonly EcoRegion.TargetFilter isTargetValidFilter = new(IsTargetValid);
+
+    /// <summary>
+    /// Replace all the method with our custom <see cref="EvaluatePriority(PrisonPredatorSwimToPlayer, Creature, float)" />
+    /// 
+    /// Original method does hardcode Player.main usage and doesn't use any "GameObject" or Target abstraction as in other <see cref="CreatureAction"/>
+    /// So we need to rewrite <see cref="PrisonPredatorSwimToPlayer.Evaluate(Creature, float)"/> to use the same logic as <see cref="PrisonPredatorSwimToPlayer.Perform(Creature, float)"/>"/>
```
Jannify, line 23:
```suggestion
    /// So we need to rewrite <see cref="PrisonPredatorSwimToPlayer.Evaluate(Creature, float)"/> to use the same logic as <see cref="PrisonPredatorSwimToPlayer.Perform(Creature, float, float)"/>"/>
```

NitroxPatcher/Patches/Dynamic/PrisonPredatorSwimToPlayer_Evaluate_Patch.cs
```diff
@@ -0,0 +1,116 @@
...
+
+        if (target.TryGetComponent(out RemotePlayerIdentifier remotePlayerIdentifier) && !remotePlayerIdentifier.RemotePlayer.CanBeAttacked())
+        {
+            return false;
+        }
+
+        return true;
+    }
+
+    public static IEcoTarget GetNearestTarget(PrisonPredatorSwimToPlayer creatureAction)
```
Jannify, line 105:
```suggestion
    private static IEcoTarget GetNearestTarget(PrisonPredatorSwimToPlayer creatureAction)
```

NitroxPatcher/Patches/Dynamic/PrisonPredatorSwimToPlayer_Evaluate_Patch.cs
```diff
@@ -0,0 +1,116 @@
...
+
+        creature.TryGetNitroxId(out NitroxId nitroxId);
+
+        Log.InGame($"[PrisonPredatorSwimTo Eval] {creature.name} {nitroxId?.ToString()} {gameobject.name}");
+        lastTarget.SetTarget(gameobject);
+
+        return instance.GetEvaluatePriority();
+    }
+
+    public static bool IsTargetValid(IEcoTarget ecoTarget)
```
Jannify, line 84:
```suggestion
    private static bool IsTargetValid(IEcoTarget ecoTarget)
```

NitroxPatcher/Patches/Dynamic/PrisonPredatorSwimToPlayer_Evaluate_Patch.cs
```diff
@@ -0,0 +1,116 @@
...
+        LastTarget lastTarget = creature.GetComponent<LastTarget>();
+        if (!lastTarget)
+        {
+            Log.Error($"Creature {creature} does not have a LastTarget component.");
+            return 0f;
+        }
+
+        creature.TryGetNitroxId(out NitroxId nitroxId);
+
+        Log.InGame($"[PrisonPredatorSwimTo Eval] {creature.name} {nitroxId?.ToString()} {gameobject.name}");
```
Jannify, line 78: Leftover debug logging

NitroxPatcher/Patches/Dynamic/PrisonPredatorSwimToPlayer_Evaluate_Patch.cs
```diff
@@ -0,0 +1,116 @@
...
+    /// Original method does hardcode Player.main usage and doesn't use any "GameObject" or Target abstraction as in other <see cref="CreatureAction"/>
+    /// So we need to rewrite <see cref="PrisonPredatorSwimToPlayer.Evaluate(Creature, float)"/> to use the same logic as <see cref="PrisonPredatorSwimToPlayer.Perform(Creature, float)"/>"/>
+    /// </summary>
+    public static IEnumerable<CodeInstruction> Transpiler(IEnumerable<CodeInstruction> instructions)
+    {
+        yield return new CodeInstruction(OpCodes.Ldarg_0);
+        yield return new CodeInstruction(OpCodes.Ldarg_1);
+        yield return new CodeInstruction(OpCodes.Ldarg_2);
+        yield return new CodeInstruction(OpCodes.Call, Reflect.Method(() => EvaluatePriority(default, default, default)));
+        yield return new CodeInstruction(OpCodes.Ret);
```
Jannify, line 31: Like in `PrisonPredatorSwimToPlayer_Perform_Patch` I think we can manipulated the IL to get the same outcome.
Creating a new local variable, filling it with the ecoTarget and then replacing all usings of the Player local variable with the new one should work.

Jannify (changes requested):

OhmV-IR:
- Cameras are super desynced with stalkers grabbing them, I don't see the camera in the position that the other client does and they can pick it up after the stalker drops it to dupe cameras
- yellow goo attack VFX is not shown on the other client and neither is the SFX
- cuddlefish can escape alien containment through animation with player(probably out of scope for the PR)
- Stalkers hatched in alien containment but we did not see any change in behavior when we entered the room or the tank

rylaix:
Massive props to @dartasen for the work on stalker sync — creature AI, loot drop mechanics, and even behavioral patches for predator logic.

But damn, the amount of review churn and indecision around merging this makes me wonder:

Are we syncing stalkers, or becoming them? 😅
✅ Teeth drop sync
✅ AvoidEscapePod logic
✅ Transpiled behavior patches
✅ IL injection of PrisonPredator decision trees
✅ Copilot passive-aggressively reviewing 16 files
✅ Comment chains on function access modifiers

❌ Merge

Welcome to Nitrox.

NitroxPatcher/Patches/Dynamic/Stalker_FindNest_Patch.cs
dartasen: From [https://github.com/BepInEx/HarmonyX/wiki/Method-patches#postfix](https://github.com/BepInEx/HarmonyX/wiki/Method-patches#postfix)

> This method defines the code that is executed after the original method. All postfixes always get executed even if the original method code was skipped.

dartasen:
PrisonPredatorSwimToPlayer behavior is weird in singleplayer as well : 
* Will swim towards the player for 1s
* Will cancel and move somewhere else

Which has a weird look & feel when looking at the creature.
Multiple CreatureAction are evaluating different target independently : MoveTowardsTarget, PrisonPredatorSwimToPlayer, SwimToHeroPeeper

MoveTowardsTarget is very likely to run after PrisonPredatorSwimToPlayer, which is pretty quick since there's no delay between creatureActions

NitroxPatcher/Patches/Dynamic/MeleeAttack_OnTouch_Patch.cs
```diff
@@ -0,0 +1,31 @@
...
+/// 
+/// Adds RemotePlayer support for the simulating player
+/// </summary>
+public sealed partial class MeleeAttack_OnTouch_Patch : NitroxPatch, IDynamicPatch
+{
+    internal static readonly MethodInfo TARGET_METHOD = Reflect.Method((MeleeAttack t) => t.OnTouch(default));
+
+    public static bool Prefix(MeleeAttack __instance)
+    {
+        if (!__instance.TryGetNitroxId(out NitroxId creatureId) ||
```
Coding-Hen, line 18: Why do we need to check the Nitrox ID here? Surely we want the player to be able to attack a entity if we don't know about it but they still have lock it it?

NitroxPatcher/Patches/Dynamic/PrisonPredatorSwimToPlayer_Perform_Patch.cs
```diff
@@ -0,0 +1,46 @@
...
+    {
+        LastTarget lastTarget = creature.GetComponent<LastTarget>();
+        if (!lastTarget)
+        {
+            Log.Error($"Creature {creature} does not have a LastTarget component.");
```
Coding-Hen: Do we need to have this message and the other patch have slightly different errors so that we can distinguish them? Not sure if it would be helpful to have the patch or class to be able to debug more if they do arise

dartasen:
All systems online

NitroxPatcher/Patches/Dynamic/MeleeAttack_OnTouch_Patch.cs
dartasen: Will double check, not sure about it.

It was implemented with an older release of my brain software

dartasen: MeleeAttack is linked to a Creature, so we're executing this method only if we have the Lock on the creature.

NitroxPatcher/Patches/Dynamic/FleeOnDamage_StopPerform_Patch.cs
```diff
@@ -0,0 +1,42 @@
...
+    /**
+     * this.accumulatedDamage = 0f;
+     * if (this.breakLeash)
+     * {
+     *      creature.leashPosition = base.transform.position;
+     *      FleeOnDamage_StopPerform_Patch.BroadcastChange(creature);                          <======= [INSERTED LINE]
+     * }
+
+     **/
+    public static IEnumerable<CodeInstruction> Transpiler(IEnumerable<CodeInstruction> instructions)
```
tornac1234, line 25: could have done a postfix for simplicity but this is better performance wise (just sayin)

NitroxPatcher/Patches/Dynamic/PrisonPredatorSwimToPlayer_Evaluate_Patch.cs
```diff
@@ -0,0 +1,112 @@
...
+
+    internal static readonly EcoRegion.TargetFilter isTargetValidFilter = new(IsTargetValid);
+
+    /// <summary>
+    /// Replace all the method with our custom <see cref="EvaluatePriority(PrisonPredatorSwimToPlayer, Creature, float)" />
+    /// 
+    /// Original method does hardcode Player.main usage and doesn't use any "GameObject" or Target abstraction as in other <see cref="CreatureAction"/>
+    /// So we need to rewrite <see cref="PrisonPredatorSwimToPlayer.Evaluate(Creature, float)"/> to use the same logic as <see cref="PrisonPredatorSwimToPlayer.Perform(Creature, float, float)"/>"/>
+    /// </summary>
+    public static IEnumerable<CodeInstruction> Transpiler(IEnumerable<CodeInstruction> instructions)
```
tornac1234, line 23: I'd have probably just done a prefix calling our method and then returning false all the time for readability but it's still good

NitroxPatcher/Patches/Dynamic/PrisonPredatorSwimToPlayer_Evaluate_Patch.cs
```diff
@@ -0,0 +1,112 @@
...
+        yield return new CodeInstruction(OpCodes.Ldarg_0);
+        yield return new CodeInstruction(OpCodes.Ldarg_1);
+        yield return new CodeInstruction(OpCodes.Ldarg_2);
+        yield return new CodeInstruction(OpCodes.Call, Reflect.Method(() => EvaluatePriority(default, default, default)));
+        yield return new CodeInstruction(OpCodes.Ret);
+    }
+
+    public static float EvaluatePriority(PrisonPredatorSwimToPlayer instance, Creature creature, float time)
+    {
+        if (time < instance.timeStopSwim + instance.pauseInterval)
```
tornac1234, line 34: Can probably add comment that this is from the original code so we don't question it

NitroxPatcher/Patches/Dynamic/MeleeAttack_OnTouch_Patch.cs
```diff
@@ -0,0 +1,29 @@
...
+            Resolve<SimulationOwnership>().HasAnyLockType(creatureId))
+        {
+            return true;
+        }
+
+        return false;
+    }
+
+    // TODO: Add transpiler to add support for held item eat from remote players, we might need to add an equivalent of Inventory.GetHeldItem()
```
Coding-Hen, line 27: Maybe worth creating a issue for this so we don't forget and someone can pick it up later. Let's not delay this, just think worth tracking for something separate and specific in another ticket as another bit of work.

NitroxPatcher/Patches/Dynamic/PrisonPredatorSwimToPlayer_Evaluate_Patch.cs
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        GameObject gameobject = ecoTarget.GetGameObject();
+        if (!gameobject)
+        {
+            return 0f;
+        }
+
+        float num = Vector3.Distance(gameobject.transform.position, instance.transform.position);
+        if (num > instance.maxDistance || num < instance.minDistance)
```
tornac1234, line 57: if we wanted to keep that minDistance logic, we should probably move it to GetNearestTarget so it actually acts as a filter

NitroxPatcher/Patches/Dynamic/PrisonPredatorSwimToPlayer_Evaluate_Patch.cs
```diff
@@ -0,0 +1,112 @@
...
+        return true;
+    }
+
+    private static IEcoTarget GetNearestTarget(PrisonPredatorSwimToPlayer creatureAction)
+    {
+        IEcoTarget ecoTarget = EcoRegionManager.main.FindNearestTarget(
+            RemotePlayer.PLAYER_ECO_TARGET_TYPE,
+            creatureAction.transform.position,
+            isTargetValidFilter,
+            maxRings: 2
```
Coding-Hen, line 107: Is there a reason we are setting this to 2 and not another number?

Coding-Hen (commented):
LGTM CW just a few minor points

NitroxPatcher/Patches/Dynamic/PrisonPredatorSwimToPlayer_Evaluate_Patch.cs
```diff
@@ -0,0 +1,112 @@
...
+            return 0f;
+        }
+
+        float num = Vector3.Distance(gameobject.transform.position, instance.transform.position);
+        if (num > instance.maxDistance || num < instance.minDistance)
+        {
+            return 0f;
+        }
+
+        if (!creature.GetCanSeeObject(gameobject))
```
tornac1234, line 62: same here, GetCanSeeObject should be moved to IsTargetValid to be an actual filter
The current implementation could lead to situations like this one:
```
              player2 (distance = 50 => not target, GetCanSeeObject=true)
                             

 
               

Stalker |||wall||| player1 (distance = 30 => target, GetCanSeeObject=false)
```
player1 is closer to Stalker so they're picked as target, but then they're forgotten because they can't be seen, while player 2 might be in range and in sight range too (which means they're the only valid target here)

NitroxPatcher/Patches/Dynamic/PrisonPredatorSwimToPlayer_Evaluate_Patch.cs
```diff
@@ -0,0 +1,112 @@
...
+        {
+            return 0f;
+        }
+
+        LastTarget lastTarget = creature.GetComponent<LastTarget>();
```
tornac1234, line 67: Could you explain why this is important ? (this is not in the base game and it seems to me that this wouldn't have any feedback on the current behaviour which might end up in situations where a player is a locked target so the stalker tries to attack it, but then another player is closer so the stalker randomly starts swimming towards that other player) (also if we were too keep that, we'd probably have to adapt the whole behaviour to drop the player as locked target when they're no longer a valid target)

NitroxPatcher/Patches/Dynamic/PrisonPredatorSwimToPlayer_Evaluate_Patch.cs
```diff
@@ -0,0 +1,112 @@
...
+        if (!lastTarget)
+        {
+            Log.Error($"[{nameof(PrisonPredatorSwimToPlayer_Evaluate_Patch)}] Creature {creature} does not have a LastTarget component.");
```
tornac1234, line 70: Should probably only error once

NitroxPatcher/Patches/Dynamic/PrisonPredatorSwimToPlayer_Perform_Patch.cs
```diff
@@ -0,0 +1,46 @@
...
+        yield return new CodeInstruction(OpCodes.Call, Reflect.Method(() => Perform(default, default, default)));
+        yield return new CodeInstruction(OpCodes.Ret);
+    }
+
+    public static void Perform(PrisonPredatorSwimToPlayer instance, Creature creature, float time)
+    {
+        LastTarget lastTarget = creature.GetComponent<LastTarget>();
+        if (!lastTarget)
+        {
+            Log.Error($"[{nameof(PrisonPredatorSwimToPlayer_Perform_Patch)}]: Creature {creature} does not have a LastTarget component.");
```
tornac1234, line 33: Should probably only erroronce

NitroxPatcher/Patches/Dynamic/PrisonPredatorSwimToPlayer_Evaluate_Patch.cs
dartasen: It's the value I choose here but can be whatever.
The more rings you add, the heavier the algorithm will be to find a target.

If you wanna increase the range of detection you need to increase the maximum rings

tornac1234 (changes requested):
While you're at it, would you patch MeleeAttack.CanDealDamageTo with a prefix like to ensure remote players aren't attacked when they shouldn't be ?
```cs
public static bool Prefix(GameObject target, ref bool __result)
{
    if (target.TryGetComponent(out RemotePlayerIdentifier remotePlayerIdentifier) && !remotePlayerIdentifier.RemotePlayer.CanBeAttacked())
    {
        __result = false;
        return false;
    }
    return true;
}
```

NitroxPatcher/Patches/Dynamic/PrisonPredatorSwimToPlayer_Evaluate_Patch.cs
dartasen: > Could you explain why this is important ? (this is not in the base game and it seems to me that this wouldn't have any feedback on the current behaviour

Initial issue is that `PrisonPredatorSwimToPlayer` is using Player.main exclusively to decide behaviors of this Action. Other behaviors almost always have a local state (or another behavior like LastTarget that stores this information) to share the "target" between Evaluate and Perform methods.

So I do need a state to persist this information in whatever form it can take (static states in patch, Custom component, LastTarget, ...). With that being said it's not excluded that other behaviors are also using `LastTarget`, in fact it's the case for **Stalker**. Which is why I used `SetLockedTarget` instead of usual `SetTarget`.
However it does not prevent other behaviors to have another target state which can be different from LastTarget. But that's up to SN devs  that implemented behaviors that can conflict with each other, that's not really the result of my choices (see [https://github.com/SubnauticaNitrox/Nitrox/pull/2255#issuecomment-3740870634](https://github.com/SubnauticaNitrox/Nitrox/pull/2255#issuecomment-3740870634))

As a side note PrisonPredatorSwim behavior guarantee that the creature is inside the prison, thereferore they're friendly and all the agressive behaviors are disabled. Which can limit the weird behaviors between conflicting behaviors

> also if we were too keep that, we'd probably have to adapt the whole behaviour to drop the player as locked target when they're no longer a valid target

I can implement that in Perform if needed

```cs
    public static void Perform(PrisonPredatorSwimToPlayer instance, Creature creature, float time)
    {
        LastTarget lastTarget = creature.GetComponent<LastTarget>();
        if (!lastTarget)
        {
            Log.ErrorOnce($"[{nameof(PrisonPredatorSwimToPlayer_Perform_Patch)}]: Creature {creature} does not have a LastTarget component.");
            return;
        }

       if (!IsValidTarget(lastTarget.target))
       {
            lastTarget.UnlockTarget();
            return;
       }

        Transform targetTransform = lastTarget.target ? lastTarget.target.transform : null;
        if (targetTransform && time > instance.timeNextSwim)
        {
            instance.timeNextSwim = time + instance.swimInterval;
            instance.swimBehaviour.SwimTo(targetTransform.position, instance.swimVelocity);
        }

        lastTarget.UnlockTarget();
    }
````

NitroxPatcher/Patches/Dynamic/PrisonPredatorSwimToPlayer_Evaluate_Patch.cs
dartasen: I'll remove it, the rings behavior already implies to have a maxDistance

CherrieTheShifter:

PR #2257, opened 2025-01-13
Fix escape pod status on joining and repairing it

Jannify:


dartasen (commented):
Tested IG, seems to work overall

Upon relog, escape pod will still show the black smoke effect for 4-5s

Jannify:
Ready for full review. Tested heavily ingame.

NitroxClient/Debuggers/Drawer/Subnautica/UWEEventDrawer.cs
```diff
@@ -0,0 +1,76 @@
+﻿using System.Collections.Generic;
+using UnityEngine;
+using UWE;
+
+namespace NitroxClient.Debuggers.Drawer.Subnautica;
+
+public class UWEEventDrawer : IDrawer<Event<float>>, IDrawer<Event<PowerRelay>>
```
Measurity, line 7: Should be `UweEventDrawer` as it's 3+ letter acronym.

Measurity (dismissed):
Radio is now always repaired when joining a new world.

NitroxClient/Debuggers/Drawer/Subnautica/UWEEventDrawer.cs
Coding-Hen: I agree with this but I think we should be consistent. In this file we have GUI did we want to change that as well?

Measurity: If we could, yes I think so. But that's a Unity type.

NitroxClient/GameLogic/Spawning/Metadata/Processor/EscapePodMetadataProcessor.cs
```diff
@@ -0,0 +1,75 @@
...
+    /// <summary>
+    /// Applies repaired state without animations and minimal audio playback
+    /// </summary>
+    public static void ProcessInitialSyncMetadata(EscapePod pod, Radio radio, EscapePodMetadata metadata)
+    {
+        if (metadata.PodRepaired)
+        {
+            pod.liveMixin.health = pod.liveMixin.maxHealth;
+            pod.healthScalar = 1;
+            pod.damageEffectsShowing = true;
```
Coding-Hen, line 48: Do we not want this as false as not showing?

tornac1234 (dismissed):

dartasen:
Can you please rebase the branch

Noticed few things while testing In Game : 
* When joining server, I can see the "Press any button blabla" right before being able to move (Tested with debug build with a fresh survival world).
As it's not mentionned in the PR description, I don't know if it's intentional or not
image

* On a fresh new world, repairing the escape pod will also repair the radio for the server (will take effect upon relog). Or the radio spawns will full health already, that will cause the extractor to mention it's repaired
Before repairing escape pod only :
image

After repairing escape pod ONLY : 
image

tornac1234:
Any progress on this ?

Jannify:
Rebased and fixed issues

OhmV-IR:
image
image
1st IG test bug: repair %s get desynced, if I repair some of the damage to the secondary systems(radio or panel) while another player is joining, the change is not synced to them. relogging the other client does not fix the issue. This also has the effect of the fixing animation being desynced. 

2nd IG test bug: Seems that the repair kit in the pod also gets desynced, open on one client and closed on the other, I think that this is because at some point during the repair process, the kit closes itself but this didn't occur on the other client. relog does not fix. I grabbed the medkit also while the other player was connecting. 

3rd IG test bug: Other players' proximity to the status billboard does not affect whether or not it is lit up, although this is probably not merge-blocking, since the performance hit from checking might be more trouble than its worth

4th IG test bug: Opening the fabricator on one client does not display the animation for the other, although I'm not sure if this is an existing bug and within the scope of this PR to fix

5th IG test bug: Fixing only the radio with the welder does not change the radio status in the bad status display board, but I think this is just a thing in the vanilla game too

Otherwise worked fine for me, tested with 2 local clients on a localhost server, ty for the PR

Jannify:
> 1st IG test bug: repair %s get desynced, if I repair some of the damage to the secondary systems(radio or panel) while another player is joining, the change is not synced to them. relogging the other client does not fix the issue. This also has the effect of the fixing animation being desynced.

We currently lack a general HealthMixing syncing. Therefore this is out of scope for this pr.


> 2nd IG test bug: Seems that the repair kit in the pod also gets desynced, open on one client and closed on the other, I think that this is because at some point during the repair process, the kit closes itself but this didn't occur on the other client. relog does not fix. I grabbed the medkit also while the other player was connecting.

Are the door state synced at all? Or does it only breaks bc of repairing it?

> 3rd IG test bug: Other players' proximity to the status billboard does not affect whether or not it is lit up, although this is probably not merge-blocking, since the performance hit from checking might be more trouble than its worth

Yeah would also say that this "bug" is not that problematic.


> 4th IG test bug: Opening the fabricator on one client does not display the animation for the other, although I'm not sure if this is an existing bug and within the scope of this PR to fix

Yeah I think it also broken in bases and therefor out of scope for this pr.


> 5th IG test bug: Fixing only the radio with the welder does not change the radio status in the bad status display board, but I think this is just a thing in the vanilla game too

If it's the same like vanilla I would leave it that way.

dartasen:
Please re-review @Measurity @tornac1234
Will schedule a test session to get this PR merged ASAP.

dartasen (approved):
Works well

PR #2271, opened 2025-01-28
Fix engine speed menu being enabled when engine is off

OhmV-IR:
Seems to just be an issue of the state never being initially effected, or maybe some sort of race condition due to coroutines being used to effect the change in states? Likely will never know, I couldn't find the source myself. The CyclopsHelmHUDManager already sets the engine off text in its update loop, so I just added a bit to the existing Postfix patch for the method to also set the animator trigger that controls the engine speed menu.


OhmV-IR:
tested IG

tornac1234:
According to tests, this works

tornac1234 (approved):

PR #2274, opened 2025-02-02
UI improvements for player name input

### LINKED ISSUE #2224: Add hotkey to submit the player name when joining a server
### Describe the issue

Add a check for [`Input.GetKeyDown(KeyCode.Return)`](https://docs.unity3d.com/6000.0/Documentation/ScriptReference/Input.GetKeyDown.html) to `MainMenuJoinServerPanel` to call `OnJoinClick`.

image

### LINKED ISSUE #2225: Auto focus input on player name field when joining a server
### Describe the issue

image

SpaceMonkeyy86:
- When the "join server" menu is opened, the player name input field is automatically focused with the cursor at the end
- Pressing enter inside the input field joins the server

I also want to look into a system that automatically deduplicates names for when I open two clients at once. I don't know the best way to do that though. Maybe the client could have the server check if their username is taken, and if it is, the server responds with an alternative username that the client could use.

Closes #2224
Closes #2225

### DIFF
--- a/NitroxClient/MonoBehaviours/Gui/MainMenu/ServerJoin/MainMenuJoinServerPanel.cs
+++ b/NitroxClient/MonoBehaviours/Gui/MainMenu/ServerJoin/MainMenuJoinServerPanel.cs
@@ -138,7 +138,7 @@ public class MainMenuJoinServerPanel : MonoBehaviour, uGUI_INavigableIconGrid, u
 
     public void OnOpened() => StartCoroutine(OpenedRoutine());
 
-    public IEnumerator OpenedRoutine()
+    private IEnumerator OpenedRoutine()
     {
         SelectFirstItem();
         yield return new WaitForEndOfFrame();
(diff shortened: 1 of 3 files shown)


Measurity:
> I also want to look into a system that automatically deduplicates names for when I open two clients at once.

You can add the following to launcherSettings.json to start multiple clients. But the names will be static so might not be what you want.
```json
    "Instant launch \"world\" 2-players": {
      "commandName": "Project",
      "commandLineArgs": "instantlaunch --save \"world\" --players \"Player1\" \"Player2\""
    }
```

SpaceMonkeyy86:
Good to know I can use Instant Launch to make this faster, but there are cases when I can't use Instant Launch and would prefer to not have to manually edit the name every time (although it would be faster with this PR's changes). For example, when I'm testing the join queue PR I need both clients to click join right after each other, which Instant Launch can't guarantee. It is a pretty minor thing but still something to consider.

tornac1234:
I have a question for people able to test this with a controller. Does leaving the field trigger onSubmit ? (as if we were pressing "Enter" on keyboard)

NitroxClient/MonoBehaviours/Gui/MainMenu/ServerJoin/MainMenuJoinServerPanel.cs
```diff
@@ -133,6 +136,15 @@ public void UpdatePlayerPanelValues(string playerName, Vector3 hsb)
         colorPicker.SetHSB(hsb);
     }
 
+    public void OnOpened() => StartCoroutine(OpenedRoutine());
+
+    public IEnumerator OpenedRoutine()
```
Jannify, line 141:
```suggestion
    private IEnumerator OpenedRoutine()
```

Jannify (approved):
Besides the one comment LGTM CW

SpaceMonkeyy86:
> Will this also auto-focus the password field if the server requires one? Does it pass from password to name on enter?

It does now. Thanks for pointing this out.

Measurity (approved):

dartasen (approved):

PR #2277, opened 2025-02-02
Fix server closing when launcher closes (on Windows)

### LINKED ISSUE #2275: Server crashes attempting to access Console.BufferWidth
### Describe the issue

- Open launcher
- Start server (make sure it is in embedded view)
- Close launcher, click "No" to keep the server running
- Open launcher and start server again

```
[WRN] Port 11000 UDP is already in use. Please change the server port or close out any program that may be using it. Retrying for 30 seconds until it is available...
[ERR] The handle is invalid.
System.IO.IOException: The handle is invalid.
   at System.ConsolePal.GetBufferInfo(Boolean throwOnNoConsole, Boolean& succeeded)
   at System.Console.get_BufferWidth()
   at NitroxServer_Subnautica.Program.WaitForAvailablePortAsync(Int32 port, TimeSpan timeout, CancellationToken ct) in NitroxServer-Subnautica\Program.cs:line 461
   at NitroxServer_Subnautica.Program.StartServer(String[] args) in NitroxServer-Subnautica\Program.cs:line 108
   at NitroxServer_Subnautica.Program.Main(String[] args) in NitroxServer-Subnautica\Program.cs:line 42
   at NitroxServer_Subnautica.Program.<Main>(String[] args)

```

### LINKED ISSUE #2276: "Do you want to stop all online servers" modal has bad UX
### Describe the issue

This may just be personal preference but I find it very hard to use this modal.

- I find that most of the time you do want to stop the server (since otherwise you have to go into Task Manager and kill it or you cannot start another server), so the Yes button should be the accented/prominent one, not No. I guess if you are just a normal user and not a developer, it would save time not to have to restart it every time, but it is still problematic because:
- If you do click No and reopen the launcher, the server shows that it is not running even though it still is. If you try to start it again, it will crash because of #2275.
- Clicking the X in the top right is the same as clicking No, which is counterintuitive (intuition suggests it should close the modal and leave the launcher open).
- You cannot tab back over to the launcher window after this modal appears, since it is a "critical" popup and must be closed before the launcher becomes usable again. However, there is no way to close the modal without also closing the launcher, so this seems pointless and only serves to get in the way.

Measurity:
Closes #2275
Closes #2276

### DIFF
--- a/Nitrox.Launcher/ViewModels/LaunchGameViewModel.cs
+++ b/Nitrox.Launcher/ViewModels/LaunchGameViewModel.cs
@@ -251,9 +251,10 @@ public partial class LaunchGameViewModel : RoutableViewModelBase
         // Start game & gaming platform if needed.
         using ProcessEx game = platform switch
         {
-            Steam s => await s.StartGameAsync(subnauticaExe, GameInfo.Subnautica.SteamAppId, subnauticaLaunchArguments),
+            Steam s => await s.StartGameAsync(subnauticaExe, subnauticaLaunchArguments, GameInfo.Subnautica.SteamAppId),
             EpicGames e => await e.StartGameAsync(subnauticaExe, subnauticaLaunchArguments),
             MSStore m => await m.StartGameAsync(subnauticaExe),
+            Discord d => await d.StartGameAsync(subnauticaExe, subnauticaLaunchArguments),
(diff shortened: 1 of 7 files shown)


SpaceMonkeyy86 (approved):

dartasen (approved):

Measurity:
Now part of 
- #2280

PR #2283, opened 2025-02-14
patch out PDA freezetime

### LINKED ISSUE #2229: Disable PDA pause options
### Describe the issue

PDA options will cause a freezetime that is dangerous for multiplayer.

Behavior emerging from MiscSettings.pdaPause boolean in PDA::ManagedUpdate()

OhmV-IR:
Have a transpiler patching out the call, not tested at all yet though, will open for review when that gets done
Closes #2229

### DIFF
--- a/NitroxPatcher/Patches/Dynamic/PDA_ManagedUpdate_Patch.cs
+++ b/NitroxPatcher/Patches/Dynamic/PDA_ManagedUpdate_Patch.cs
@@ -23,13 +23,16 @@ namespace NitroxPatcher.Patches.Dynamic
     {
         private static readonly MethodInfo TARGET_METHOD = Reflect.Method((PDA t) => t.ManagedUpdate());
 
+        public static bool Prefix(PDA __instance)
+        {
+            FieldInfo initializedField = Player.main.playerAnimator.GetType().GetField("updateMode", BindingFlags.NonPublic | BindingFlags.Instance);
+            initializedField.SetValue(__instance, AnimatorUpdateMode.Normal);
+            return true;
+        }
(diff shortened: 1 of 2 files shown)


OhmV-IR:
Ready for CW review. Tested in-game

NitroxPatcher/Patches/Dynamic/PDA_ManagedUpdate_Patch.cs
```diff
@@ -0,0 +1,42 @@
...
+            PDA.UpdateTime(flag3);
+            Player.main.playerAnimator.updateMode = (flag3 ? AnimatorUpdateMode.UnscaledTime : AnimatorUpdateMode.Normal);
+        TO:
+            PDA.UpdateTime(false); [REPLACED]
+            Player.main.playerAnimator.updateMode = AnimatorUpdateMode.Normal; [REPLACED]
+        */
+
+        public static IEnumerable<CodeInstruction> Transpiler(IEnumerable<CodeInstruction> instructions)
+        {
+            return new CodeMatcher(instructions).MatchStartForward().RemoveInstructions(45).Insert(
```
Coding-Hen, line 29: Does this need to be a call here or can we just insert the lines directly? 

If we are doing a call then the description above needs to be changed to show we are calling out to our own method

OhmV-IR: We could insert the lines directly, but it's complicated to do in a transpiler so I moved it to a method. Updated the comment to reflect that.

dartasen:
Will close in favor of #2283

PR #2286, opened 2025-02-28
Disable PDA pause option

### LINKED ISSUE #2229: Disable PDA pause options

tornac1234:
- We were patching ``FreezeTime.Begin`` which internally uses ``FreezeTime.Set``. But PDA code used ``FreezeTime.Set``. So instead, we now patch ``FreezeTime.Set``
- disable MiscSettings.pdaPause at least once so PDA always consider that the option is false
- Remove the whole toggle in the options menu

Closes #2229


NitroxClient/MonoBehaviours/NitroxBootstrapper.cs
```diff
@@ -25,6 +25,8 @@ private void Awake()
         // This is very important, see Application_runInBackground_Patch.cs
         Application.runInBackground = true;
         Log.Info($"Unity run in background set to \"{Application.runInBackground}\"");
+        // Also very important for similar reasons
```
Coding-Hen, line 28: Would the similar reasons be in the same patch? If not much better good to specify where they can be found

Coding-Hen (changes requested):
LGTMCW just one minor comment

NitroxClient/MonoBehaviours/NitroxBootstrapper.cs
tornac1234: There's an XML comment in the said patch which explains the issue

Coding-Hen: I assumed it was referencing that one just wanted to make sure as it wasn't 100% clear

Coding-Hen (approved):

NitroxPatcher/Patches/Dynamic/FreezeTime_Set_Patch.cs
```diff
@@ -6,15 +6,15 @@
 namespace NitroxPatcher.Patches.Dynamic;
 
 /// <summary>
-/// Because we're in multiplayer mode, we generally don't want the game to freeze
+/// Because we're in multiplayer mode, we generally don't want the game to freeze.
+/// Thus we prevent any FreezeTime that is not for quit screen or (initial sync) wait screen.
```
Jannify, line 10: For what is `FreezeTime.Id.Quit` used in base sn? If it's for the esc menu isn't that against our intend?

tornac1234: Quit is only used by IngameMenu.QuitGameAsync (which means you're already leaving the world) (quit screen means the saving screen and then when going back to the menu).
Also we have modals like, KickedModal, ServerStoppedModal and LostConnectionModal (when we open those, the client is already disconnected from the server) which use Quit because it's the only FreezeTime id other than WaitScreen that is relevant for this purpose.

I believe Esc menu's freeze id is IngameMenu

Jannify (approved):
LGTM CW

PR #2289, opened 2025-03-03
macOS fixes

dartasen:
* Application name is overriden by informations we provide inside Info.plist upon macOS publish, but does not work inside IDE since this file does not exist
* Provide a basic native menu on macOS to avoid Avalonia emitting a default menu that displays a popup Nitrox is not aware and compatible of
* OnPlatform does not properly work with semi-binding, static providers, causing NRE on macOS

### DIFF
--- a/Nitrox.Launcher/App.axaml
+++ b/Nitrox.Launcher/App.axaml
@@ -1,7 +1,8 @@
-<Application
-    x:Class="Nitrox.Launcher.App"
-    xmlns="https://github.com/avaloniaui"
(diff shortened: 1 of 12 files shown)


oliver408i:
Can confirm works on my Mac with the correct menubar as well as fixing the server manage pagae

Measurity (approved):
:+1:

oliver408i:
tested new changes. All works

dartasen:
Works on my side too

Measurity (approved):

PR #2290, opened 2025-03-03
Fix SeaTreader_SetNextPathPoint not emitting metadata changes

dartasen:
Since some internal metadata of the SeaTreader are modified inside the method, we should extract and send metadata changes after this method is called

image


Coding-Hen (approved):
LGTM

tornac1234:
Modification looks good, are there any IG tests ?

Jannify:
Should this method be suppressed if it does not have any lock or does this not matter?

killzoms:
> Should this method be suppressed if it does not have any lock or does this not matter?

This is possible to pull off via 2 methods, our current suppression logic or some reflection shenanigans that targets all uses of that method and replaces them with a nop instruction or just removes it entirely, whichever is easiest

Jannify (approved):

Issue #2278, opened 2025-02-03
Prawn suit drilling doesn't save raw materials globally


leonidasjan:
### Are you on the latest version of Nitrox?

Yes, I'm on the latest commit of the repo (only for experts).

### Which OS are you playing on?

Windows

### On which platform did you buy Subnautica?

Steam

### Describe the issue.

The materials that i got from drilling are only local.
Other players cannot see the materials, regardless of where I move or store them.
After reconnect the materials are gone.
 Tested IG with a friend.

### Describe the steps you took to encounter the bug.

1. Spawn prawn suit
2. Put  drilling arm on prawn suit
3. Find large resource deposit and drill  it
4. Check prawns storage
5. Player 1 can see the resources but Player 2 cant

nickrbclark:
I'm seeing the same exact issue on f450264
Windows Clients, Linux Server.

It's like the client never tells the server that it obtained those resources, so they're lost when they rejoin.

Geldyy:
Had the exact same issue, i think it also has something todo with the not working animation/action sync of the exosuit.
Hope, this gets fixed, relatively fast, with this fixed, it would be nearly playable.

PR #2299, opened 2025-03-09
Fix prawn suit drilling doesn't save raw materials globally

Buivol16:
Fix issue link: https://github.com/SubnauticaNitrox/Nitrox/issues/2278


NitroxPatcher/Patches/Dynamic/ResourceTracker_OnPickedUp_Patch.cs
```diff
@@ -0,0 +1,27 @@
...
+    public static readonly MethodInfo TARGET_METHOD = Reflect.Method((ResourceTracker d) => d.OnPickedUp(default));
+    
+    public static bool Prefix(ResourceTracker __instance, Pickupable p)
+    {
+        TechType instanceTechType = __instance.techType;
+        GameObject instanceGameObject = __instance.gameObject;
+        if (instanceTechType != null && instanceGameObject != null)
+        {
+            var ITEM_SERVICE = NitroxServiceLocator.LocateService<Items>();
+            ITEM_SERVICE.PickedUp(instanceGameObject, instanceTechType);
```
Coding-Hen, line 23: I think you should be able to call this directly like other patches do without having to assign it first.

```suggestion
            Resolve<Items>().PickedUp(instanceGameObject, instanceTechType);
```

NitroxPatcher/Patches/Dynamic/ResourceTracker_OnPickedUp_Patch.cs
```diff
@@ -0,0 +1,27 @@
...
+namespace NitroxPatcher.Patches.Dynamic;
+
+/// <summary>
+/// Event when someone picks up an item.
+/// </summary>
+public sealed partial class ResourceTracker_OnPickedUp_Patch : NitroxPatch, IDynamicPatch
+{
+    public static readonly MethodInfo TARGET_METHOD = Reflect.Method((ResourceTracker d) => d.OnPickedUp(default));
+    
+    public static bool Prefix(ResourceTracker __instance, Pickupable p)
```
Coding-Hen, line 16: I don't think you need pickupable here as you don't make use of it
```suggestion
    public static bool Prefix(ResourceTracker __instance)
```

Coding-Hen (changes requested):
Just a few minor cleanup suggestions

Issue #2266, opened 2025-01-20
Oxygen tank is empty in inventory when reconnecting


kovadam69:
### Are you on the latest version of Nitrox?

Yes, I'm on the latest commit of the repo (only for experts).

### Which OS are you playing on?

Windows

### On which platform did you buy Subnautica?

Steam

### Describe the issue.

When you have an oxygen tank in the inventory when you reconnect it is empty.

### Describe the steps you took to encounter the bug.

Craft an additional oxygen tank and place it into your inventory. Fill first with oxygen. Then exit game, restart, log in, the oxygen tank in the inventory is empty (no air inside).

### Relevant log output

![Image](https://github.com/user-attachments/assets/9a9efe8e-6011-4ab4-9322-6d7979880024)

![Image](https://github.com/user-attachments/assets/7f42aadb-aaed-4197-92a0-ad2c12f5fe28)

dartasen:
*Note for devs*
Tanks have an `Oxygen` Monobehavior, that has a field :
```cs
	[ProtoMember(1)]
	[NonSerialized]
	public float oxygenAvailable;
```
That should be synced

PR #2302, opened 2025-03-09
fix oxygen tank is empty in inventory when reconnecting

Buivol16:
Fix for issue: https://github.com/SubnauticaNitrox/Nitrox/issues/2266

### DIFF
--- a/NitroxClient/GameLogic/ItemContainers.cs
+++ b/NitroxClient/GameLogic/ItemContainers.cs
@@ -3,6 +3,7 @@ using NitroxClient.Communication.Abstract;
 using NitroxClient.GameLogic.Helper;
 using NitroxClient.GameLogic.PlayerLogic;
 using NitroxClient.GameLogic.Spawning.Metadata;
+using NitroxClient.GameLogic.Spawning.Metadata.Extractor;
 using NitroxClient.MonoBehaviours;
 using NitroxClient.Unity.Helper;
 using NitroxModel.DataStructures;
@@ -26,7 +27,7 @@ namespace NitroxClient.GameLogic
             this.entityMetadataManager = entityMetadataManager;
         }
 
-        public void BroadcastItemAdd(Pickupable pickupable, Transform containerTransform)
+        public void BroadcastItemAddAndSendMetadata(Pickupable pickupable, Transform containerTransform)
         {
             // We don't want to broadcast that event if it's from another player's inventory
             if (containerTransform.GetComponentInParent<RemotePlayerIdentifier>(true))
@@ -47,6 +48,9 @@ namespace NitroxClient.GameLogic
             
             if (packetSender.Send(new EntityReparented(itemId, ownerId)))
             {
+                var iOxygenSource = pickupable.gameObject.GetComponent<IOxygenSource>();
+                var entityMetadataUpdate = GetEntityMetadataUpdateForIOxygenMetadata(iOxygenSource, itemId);
+                if (entityMetadataUpdate != null) packetSender.Send(entityMetadataUpdate);
                 Log.Debug($"Sent: Added item ({itemId}) of type {pickupable.GetTechType()} to container {containerTransform.gameObject.GetFullHierarchyPath()}");
             }
         }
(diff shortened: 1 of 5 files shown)


NitroxClient/GameLogic/ItemContainers.cs
```diff
@@ -47,6 +48,9 @@ public void BroadcastItemAdd(Pickupable pickupable, Transform containerTransform
             
             if (packetSender.Send(new EntityReparented(itemId, ownerId)))
             {
+                var iOxygenSource = pickupable.gameObject.GetComponent<IOxygenSource>();
+                var entityMetadataUpdate = GetEntityMetadataUpdateForIOxygenMetadata(iOxygenSource, itemId);
+                if (entityMetadataUpdate != null) packetSender.Send(entityMetadataUpdate);
```
Jannify, line 53: This seems like a very specific check when we have methods to extract metadata in a general fashion. Did you test any weird behaviour when sending every extractable metadata?
@tornac1234 what is your opinion on this?

NitroxClient/GameLogic/ItemContainers.cs
```diff
@@ -47,6 +48,9 @@ public void BroadcastItemAdd(Pickupable pickupable, Transform containerTransform
             
             if (packetSender.Send(new EntityReparented(itemId, ownerId)))
             {
+                var iOxygenSource = pickupable.gameObject.GetComponent<IOxygenSource>();
```
Jannify, line 51: We don't use var at all.
It seems you don't get visual feedback from your IDE for this. We have a `.editorconfig` file in the project which should setup out code style for you... :thinking:

NitroxClient/GameLogic/ItemContainers.cs
Jannify, line 53: Always use brackets for ifs.

NitroxModel/DataStructures/GameLogic/Entities/Metadata/EntityMetadata.cs
```diff
@@ -39,6 +39,7 @@ namespace NitroxModel.DataStructures.GameLogic.Entities.Metadata
     [ProtoInclude(79, typeof(SeaTreaderMetadata))]
     [ProtoInclude(80, typeof(StayAtLeashPositionMetadata))]
     [ProtoInclude(81, typeof(EggMetadata))]
+    [ProtoInclude(82, typeof(IOxygenSourceMetadata))]
```
Jannify, line 42: Should add tests for it in `WorldPersistenceTest`

NitroxClient/GameLogic/Spawning/Metadata/Processor/IOxygenSourceMetadataProcessor.cs
```diff
@@ -0,0 +1,22 @@
...
+
+namespace NitroxClient.GameLogic.Spawning.Metadata.Processor;
+
+public class IOxygenSourceMetadataProcessor : EntityMetadataProcessor<IOxygenSourceMetadata>
+{
+    public override void ProcessMetadata(GameObject gameObject, IOxygenSourceMetadata metadata)
+    {
+        IOxygenSource iOxygenSource = gameObject.GetComponent<IOxygenSource>();
+
+        if (iOxygenSource != null)
```
Jannify, line 13:
```suggestion
        if (gameObject.TryGetComponent(out IOxygenSource iOxygenSource))
```
Makes it a little bit clearer imo. Also you should never compare unity objects against null (see [here](https://discussions.unity.com/t/is-null-returns-false-on-null-objects/821933/11) for explaination), instead threat it like in c++ as a bool

NitroxClient/GameLogic/Spawning/Metadata/Processor/IOxygenSourceMetadataProcessor.cs
```diff
@@ -0,0 +1,22 @@
...
+        {
+            iOxygenSource.AddOxygen(metadata.AvailableOxygen);
```
Jannify, line 15: As this can also possibly be executed on existing tanks I would first calculate the delta of `GetOxygenAvailable()` and `metadata.AvailableOxygen`, to then use it for `AddOxygen()` or `RemoveOxygen()` depending on plus or minus.

NitroxClient/GameLogic/ItemContainers.cs
```diff
@@ -26,7 +27,7 @@ public ItemContainers(IPacketSender packetSender, EntityMetadataManager entityMe
             this.entityMetadataManager = entityMetadataManager;
         }
 
-        public void BroadcastItemAdd(Pickupable pickupable, Transform containerTransform)
+        public void BroadcastItemAddAndSendMetadata(Pickupable pickupable, Transform containerTransform)
```
Jannify, line 30: Also with this changes the value only gets updated when adding the tank to the inventory and not when we use breath the air in it or rise to the surface. Maybe we need some kind of system to send this data repentantly (like every 1-3 seconds or so).

Jannify (changes requested):

dartasen:
Closing for now, we'll explore another way of doing it

PR #2312, opened 2025-04-12
Improve Nitrox patching code

dartasen:
Original problem that I find, is that some nitrox dll's were missing inside Player.log causing infinite loop before subnautica's main menu

* Improved Nitrox Subnautica patching code to be more resilient and support more edge cases : 
    -  `AssemblyCSharp` gets **wiped** (Antivirus? or code that gets interrupted during "_critical_" section) but `AssemblyCSharp-Nitrox` does exist
    - Don't patch Assembly if it's already patched, shouldn't happen inside Nitrox code but can happen if someone modify files by hand
* Removed most of static dll files, causing headaches and weird behaviors during build
   - `LZ4`  is never imported during build process nor used inside our code ? If I remember it was used in our netcode code back in the days to compress packets before giving it to Litenetlib layer
   - `iTween` is never imported during build process nor used inside our code
   - `Serilogs.Sinks.Map` has a **_TODO_** that got added during rewrite of launcher, I don't find any dependencies between Serilog and `Sinks.Map` on nuget, that may explain the weird behavior. So I upgraded `Serilog `and added `Sinks.Map` nuget package, seems to fix this.
 - `discord_game_sdk.dll` is never imported, but I might see related warnings inside Player.log, not sure to understand why Nitrox would add a dependency SN needs to run (@Jannify any idea on this dll ?)
 ```txt
Fallback handler could not load library d:/programfiles/steam/steamapps/common/Subnautica/Subnautica_Data/Mono/data-0000023B573673E0.dll
Fallback handler could not load library d:/programfiles/steam/steamapps/common/Subnautica/Subnautica_Data/Mono/data-0000023B0F245920.dll
Fallback handler could not load library d:/programfiles/steam/steamapps/common/Subnautica/Subnautica_Data/Mono/data-0000023B575788A0.dll
Fallback handler could not load library d:/programfiles/steam/steamapps/common/Subnautica/Subnautica_Data/Mono/discord_game_sdk
Fallback handler could not load library d:/programfiles/steam/steamapps/common/Subnautica/Subnautica_Data/Mono/discord_game_sdk.dll
Fallback handler could not load library d:/programfiles/steam/steamapps/common/Subnautica/Subnautica_Data/Mono/discord_game_sdk
Fallback handler could not load library d:/programfiles/steam/steamapps/common/Subnautica/Subnautica_Data/Mono/libdiscord_game_sdk
Fallback handler could not load library d:/programfiles/steam/steamapps/common/Subnautica/Subnautica_Data/Mono/libdiscord_game_sdk.dll
Fallback handler could not load library d:/programfiles/steam/steamapps/common/Subnautica/Subnautica_Data/Mono/libdiscord_game_sdk
Fallback handler could not load library d:/programfiles/steam/steamapps/common/Subnautica/Subnautica_Data/Mono/discord_game_sdk
Fallback handler could not load library d:/programfiles/steam/steamapps/common/Subnautica/Subnautica_Data/Mono/discord_game_sdk.dll
Fallback handler could not load library d:/programfiles/steam/steamapps/common/Subnautica/Subnautica_Data/Mono/discord_game_sdk
Fallback handler could not load library d:/programfiles/steam/steamapps/common/Subnautica/Subnautica_Data/Mono/libdiscord_game_sdk
Fallback handler could not load library d:/programfiles/steam/steamapps/common/Subnautica/Subnautica_Data/Mono/libdiscord_game_sdk.dll
Fallback handler could not load library d:/programfiles/steam/steamapps/common/Subnautica/Subnautica_Data/Mono/libdiscord_game_sdk
```
* Bump some .NET dependencies

### CODE REVIEW

File: NitroxPatcher/Main.cs [hunk 1]
```diff
@@ -139,15 +139,20 @@ private static Assembly CurrentDomainOnAssemblyResolve(object sender, ResolveEve
...
         }
 
         if (!File.Exists(dllPath))
         {
             Console.WriteLine($"Nitrox dll missing: {dllPath}");
         }
+
+
+        Console.WriteLine($"Loading {dllFileName} at {dllPath}");
         return Assembly.LoadFile(dllPath);
```
dartasen, line 156: This part seems weird to me @Measurity 

We still Assembly.LoadFile even if we know the dll doesn't exist

Measurity: `Assembly.LoadFile` does a lot more than loading from a specific path, it lets Windows do its thing looking for the DLL name. It looks for DLLs in the GAC, or system32, or in `PATH` paths, etc. 

The function call just has a weird name but that's Windows API in general.

Measurity: https://learn.microsoft.com/en-us/dotnet/api/system.reflection.assembly.loadfrom?view=net-9.0

> Use the [LoadFile](https://learn.microsoft.com/en-us/dotnet/api/system.reflection.assembly.loadfile?view=net-9.0) method to load and examine assemblies **that have the same identity, but are located in different paths.** [LoadFile](https://learn.microsoft.com/en-us/dotnet/api/system.reflection.assembly.loadfile?view=net-9.0) does not load files into the load-from context, and does not resolve dependencies using the load path, as the [LoadFrom](https://learn.microsoft.com/en-us/dotnet/api/system.reflection.assembly.loadfrom?view=net-9.0) method does. [LoadFile](https://learn.microsoft.com/en-us/dotnet/api/system.reflection.assembly.loadfile?view=net-9.0) is useful in this limited scenario because [LoadFrom](https://learn.microsoft.com/en-us/dotnet/api/system.reflection.assembly.loadfrom?view=net-9.0) cannot be used to load assemblies that have the same identities but different paths; it will load only the first such assembly.

Measurity (approved):

PR #2314, opened 2025-04-17
Refactored server to use .NET Generic Host APIs

### LINKED ISSUE #2166: Refactor server project to use .NET Generic Host to remove code smells
### Describe the issue

Server code has the following [code smells](https://en.wikipedia.org/wiki/Code_smell):
1. Server starts and loads from files through DI container instead of what is conventional: loading after the DI container is fully resolved. Types registered in a DI container should not do IO until asked for by code that takes dependency on said types.
2. The dependency-flow on data is inverted from what it should be (_is_ parent->child but _should be_ child->parent). Types like `WorldPersistence` should take dependency on `Server` (which knows its SaveName) instead of recomputing the save location. The current situation has lead to awkward DI container setup where types need to run setup code to be able to register in the DI container.
3. Strongly related with 2: because file loading happens during DI container setup, cancellation of server is hard to implement (without first waiting for the server to fully load).

To assist with this refactor, the server should implement [.NET Generic Host API](https://learn.microsoft.com/dotnet/core/extensions/generic-host?tabs=appbuilder). 

Then:
1. Rename the `Server` class to `ServerService`, extending [`BackgroundService`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.hosting.backgroundservice?view=net-8.0). Have it figure out the save name from command line arguments (passed to the server executable).
4. Create a `ConfigService` which depends on `ServerService` and loads in the right server.cfg file based on the save name that `ServerService` knows. Any updates to the configuration should happen through this `ConfigService`.
5. Create a `WorldPersistenceService` which depends on `ServerService` to figure out where it should `Load`, `Save` and `Backup` to. If there is no world, have it create one in `WorldPersistenceService.StartupAsync` routine. Add events/dependencies to other services as needed to trigger world backups or saving.
6. Create a `CommandService` that will read from the command line and call the correct command on submit (i.e. enter pressed).
7. Continue creating APIs that depend on `ServerService` or any other as needed...

## PRs required before work can start
- #1848

Measurity:
- [x] Improved command handling
 - [ ] Use SQLite database for save state management instead of JSON files. Deprecate JSON/PROTOBUF save serializer.
 - [ ] Solve TODOs (marked with // TODO)

## Why move to SQLite?

Since server is using multiple threads, all the persisted state should be synchronized. This risks deadlocks or unnecessary lock contention if not applied carefully. SQLite would solve these issues for us.

## Issues to close
- Closes #2166


dartasen:
Seems like another huge piece of work and it will be hard to test/integrate.
I think we should probably slice it into multiple PR (.NET Host refactor, Encryption, Sqlite)

Measurity:
Ready for review :clown_face:

Nitrox.Model/Configuration/SubnauticaServerOptions.cs
```diff
@@ -0,0 +1,87 @@
...
+
+    [PropertyDescription("Set to true to Cache entities for the whole map on next run. \nWARNING! Will make server load take longer on the cache run but players will gain a performance boost when entering new areas.")]
+    public bool CreateFullEntityCache { get; set; } = false;
+
+    [Required]
+    [RegularExpression(@"\w+")]
+    public string Seed { get; set; } = "";
+
+    public bool DisableConsole { get; set; }
+    public string? AdminPassword { get; set; }
```
Jannify: Allowed to be null?

Nitrox.Model/Extensions/TaskExtensions.cs
```diff
@@ -5,6 +5,8 @@ namespace Nitrox.Model.Extensions;
 
 public static class TaskExtensions
 {
+    public static bool IsBusyOrDone(this Task? task) => task is { IsCompleted: true, IsFaulted: false, IsCanceled: false };
```
Jannify: Maybe I am misunderstanding this, but imo this returns true if the task is busy (working) or done. Wouldn't IsCompleted == true not reflect the busy part?

Nitrox.Model/Helper/HashTree.cs
```diff
@@ -0,0 +1,68 @@
...
+            next = new();
+            data.Add(node, next);
+        }
+        return next;
+    }
+
+    /// <summary>
+    ///     Returns true if the given tree is entirely within the current tree.
+    /// </summary>
+    public bool Contains(HashTree tree, HashTree? self = null)
```
Jannify: The `self` argument seems really wrong in a oop context

Nitrox.Server.Subnautica/Models/ConsoleUnhandledErrorHandler.cs
```diff
@@ -0,0 +1,39 @@
...
+using System.IO;
+
+namespace Nitrox.Server.Subnautica.Models;
+
+static class ConsoleUnhandledErrorHandler
+{
+    public static void Attach()
+    {
+        AppDomain.CurrentDomain.UnhandledException += CurrentDomainOnUnhandledException;
+    }
```
dartasen, line 13: No Detach ?

Nitrox.Server.Subnautica/Services/CommandService.cs
```diff
@@ -0,0 +1,50 @@
...
+/// </summary>
+internal sealed class CommandService(TextCommandProcessor textCommandProcessor, ILogger<CommandService> logger) : IHostedService
+{
+    private readonly TextCommandProcessor textCommandProcessor = textCommandProcessor;
+    private readonly Channel<Task> runningCommands = Channel.CreateUnbounded<Task>();
+    private Task commandWaiterTask;
+
+    public void ExecuteCommand(string command)
+    {
+        // TODO: Make ProcessCommand async
```
dartasen, line 20: 👁️

Measurity:
Will merge this PR in a week from now if no review comments are made. Resetting the count when all review comments are handled.

Nitrox.Launcher/Models/Design/ServerEntry.cs
```diff
@@ -71,6 +74,8 @@ public partial class ServerEntry : ObservableObject
     [ObservableProperty]
     private DateTime lastAccessedTime = DateTime.Now;
 
+    private int lastProcessId;
```
dartasen, line 80: Can be next his public Property for better readability

Nitrox.Launcher/Models/Design/ServerEntry.cs
```diff
@@ -99,39 +107,93 @@ public partial class ServerEntry : ObservableObject
     private Version version = NitroxEnvironment.Version;
 
     internal ServerProcess? Process { get; private set; }
```
dartasen, line 112: internal ? 👁️

Nitrox.Launcher/Models/Design/ServerEntry.cs
```diff
@@ -31,9 +33,10 @@ namespace Nitrox.Launcher.Models.Design;
 public partial class ServerEntry : ObservableObject
 {
     public const string DEFAULT_SERVER_ICON_NAME = "servericon.png";
-    public const string DEFAULT_SERVER_CONFIG_NAME = "server.cfg";
+    private static readonly Dictionary<string, ServerEntry> entriesByDirectory = [];
+    private static readonly Lock entriesByDirectoryLocker = new();
```
dartasen: Is there a reason to not use ConcurrentDictionnary ?

Nitrox.Launcher/Models/Services/PreventMultipleAppInstancesService.cs
```diff
@@ -0,0 +1,44 @@
...
+    public Task StartingAsync(CancellationToken cancellationToken)
+    {
+        if (App.allowInstances || RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+        {
+            return Task.CompletedTask;
+        }
+
+        try
+        {
+            using ProcessEx process = ProcessEx.GetFirstProcess("Nitrox.Launcher", process => process.Id != Environment.ProcessId && process.IsRunning);
```
dartasen: Hardcoded name

Nitrox.Launcher/App.axaml.cs
```diff
@@ -85,45 +100,61 @@ public static AppBuilder Create()
...
-        AppBuilder builder = AppBuilder.Configure<App>()
-                                       .UsePlatformDetect()
-                                       .LogToTrace()
-                                       .With(new SkiaOptions { UseOpacitySaveLayer = true });
-        builder = WithRenderingMode(builder, preferredRenderingMode);
-        return builder;
-
-        static AppBuilder WithRenderingMode(AppBuilder builder, X11RenderingMode? rendering)
+                EnvironmentName = Environment.GetEnvironmentVariable("DOTNET_ENVIRONMENT"),
+                ApplicationName = "Nitrox.Launcher",
```
dartasen: Same hardcoded name

Nitrox.Model/Extensions/RangeExtensions.cs
```diff
@@ -0,0 +1,10 @@
+#if NET9_0_OR_GREATER
```
dartasen: we can replace every NET X OR GREATER by NET or !NET depending on the context. That could avoid maintaining those preprocessor to be up to date with .NET version

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives#conditional-compilation

Nitrox.Launcher/Models/Design/ServerEntry.cs
```diff
@@ -193,14 +262,15 @@ public bool RefreshFromDirectory(string saveDir)
...
+        PortForward = config.PortForward;
+        AllowLanDiscovery = config.LanDiscovery;
         AllowCommands = !config.DisableConsole;
-        AllowPvP = config.PvPEnabled;
+        AllowPvP = config.PvpEnabled;
         AllowKeepInventory = config.KeepInventoryOnDeath;
         IsNewServer = !File.Exists(Path.Combine(saveDir, $"PlayerData{fileEnding}"));
         Version = serverVersion;
-        IsEmbedded = config.IsEmbedded || RuntimeInformation.IsOSPlatform(OSPlatform.OSX); // Force embedded on MacOS
+        // TODO: Store default option "IsEmbedded" in launcher cfg, not server.cfg
```
dartasen: Forgotten TODO or it's meant to be a follow up issue ?

Nitrox.Launcher/Models/Design/ServerEntry.cs
```diff
@@ -99,39 +108,94 @@ public partial class ServerEntry : ObservableObject
...
+    public int ProcessId
+    {
+        get
+        {
+            int processId = Process?.Id ?? 0;
+            if (processId > 0)
+            {
+                lastProcessId = processId;
+            }
+            return processId;
```
dartasen: Just to be sure, is it wanted to fallback to PID 0 which is a special PID on both windows & linux ?

Nitrox.Launcher/Models/Design/ServerEntry.cs
```diff
@@ -99,39 +108,94 @@ public partial class ServerEntry : ObservableObject
...
         string fileEnding = config.SerializerMode switch
         {
             ServerSerializerMode.JSON => ServerJsonSerializer.FILE_ENDING,
             ServerSerializerMode.PROTOBUF => ServerProtoBufSerializer.FILE_ENDING,
             _ => throw new NotImplementedException()
         };
 
-        File.WriteAllText(Path.Combine(saveDir, $"Version{fileEnding}"), null);
-        using (config.Update(saveDir))
+        await File.WriteAllTextAsync(Path.Combine(saveDir, $"Version{fileEnding}"), (string?)null);
```
dartasen, line 164: Nit: Just for clarity on which overload should be always selected
```suggestion
        await File.WriteAllTextAsync(Path.Combine(saveDir, $"Version{fileEnding}"), contents: null);
```

Nitrox.Launcher/Models/Design/ServerEntry.cs
```diff
@@ -177,14 +241,19 @@ public bool RefreshFromDirectory(string saveDir)
                     serverVersion = versionModel.Version;
                     break;
                 case ServerSerializerMode.PROTOBUF:
-                    serverVersion = new ServerProtoBufSerializer().Deserialize<SaveFileVersion>(stream)?.Version ?? NitroxEnvironment.Version;
+                    serverVersion = new ServerProtoBufSerializer(null).Deserialize<SaveFileVersion>(stream)?.Version ?? NitroxEnvironment.Version;
```
dartasen, line 244: null 👁️

Nitrox.Launcher/Models/Design/ServerEntry.cs
```diff
@@ -268,31 +337,61 @@ protected override void OnPropertyChanged(PropertyChangedEventArgs e)
...
         }
         base.OnPropertyChanged(e);
     }
 
+    [MemberNotNull(nameof(cts))]
+    internal void ResetCts()
+    {
+        cts?.Dispose();
+        cts = new();
+        cts.Token.Register(async void () =>
```
dartasen, line 355: it's async void by default, isn't it ?

Nitrox.Launcher/Models/Design/ServerEntry.cs
```diff
@@ -268,31 +337,61 @@ protected override void OnPropertyChanged(PropertyChangedEventArgs e)
...
+                        catch (ChannelClosedException)
+                        {
+                            await Task.Delay(500);
+                        }
+                    }
+                    CommandQueue = Channel.CreateUnbounded<string>();
+                    IsOnline = false;
+                    Output.Clear();
+                });
+                await Dispatcher.UIThread.InvokeAsync(() => IsServerClosing = false);
```
dartasen, line 378: Is this wanted to have 3 separate dispatch ?

Nitrox.Launcher/ViewModels/ManageServerViewModel.cs
```diff
@@ -363,7 +363,7 @@ private async Task RestoreBackup()
                     TryDeleteFile(file);
                 }
                 ZipFile.ExtractToDirectory(backupFile, SaveFolderDirectory, true);
-                Server!.RefreshFromDirectory(SaveFolderDirectory);
+                await Server!.RefreshFromDirectoryAsync(SaveFolderDirectory);
```
dartasen: Alexa Intruder mode, null forgiving operator detected

Nitrox.Model/Configuration/Validators/SaveNameAttribute.cs
```diff
@@ -0,0 +1,30 @@
...
+        int indexOfAny = str.IndexOfAny(invalidPathCharacters);
+        if (indexOfAny > -1)
+        {
+            return false;
+        }
+        if (str.EndsWith(".", StringComparison.OrdinalIgnoreCase))
+        {
+            return false;
+        }
+        return true;
```
dartasen, line 28: Should we add unit tests for those attributes ?

Nitrox.Model/Packets/Packet.cs
```diff
@@ -140,5 +143,11 @@ public Wrapper(Packet packet)
                 Packet = packet;
             }
         }
+
+        public enum UdpChannelId : byte
+        {
+            DEFAULT = 0,
+            MOVEMENTS = 1,
+        }
```
dartasen, line 151: Could be extracted to its own file

Nitrox.Model/Packets/Packet.cs
```diff
@@ -14,9 +14,19 @@ namespace Nitrox.Model.Packets
     public abstract class Packet
     {
         private static readonly Dictionary<Type, PropertyInfo[]> cachedPropertiesByType = new();
-        private static readonly StringBuilder toStringBuilder = new();
+        private static readonly object cachedPropertiesByTypeLocker = new();
+
+        [ThreadStatic]
+        private static StringBuilder? toStringBuilder;
```
dartasen, line 20: 👀

Nitrox.Server.Subnautica/Extensions/BufferWriterExtensions.cs
```diff
@@ -0,0 +1,33 @@
+using System.Buffers;
+using System.Text;
+
+namespace Nitrox.Server.Subnautica.Extensions;
+
+internal static class BufferWriterExtensions
+{
+    public static void Write(this IBufferWriter<byte> writer, ReadOnlySpan<char> utf16Value)
+    {
+        const int MAX_STACK_ALLOC = 64; // Should be divisible by 4
```
dartasen:
```suggestion
        const int MAX_STACK_ALLOC = 64;
        Debug.Assert(MAX_STACK_ALLOC % 4 == 0)
```

Nitrox.Server.Subnautica/Models/Commands/Abstract/Command.cs
```diff
@@ -35,13 +38,46 @@ protected Command(string name, Perms perms, string description)
...
+        /// <summary>
+        ///     Send a message to all connected players
+        /// </summary>
+        public static void SendMessageToAllPlayers(string message)
+        {
+            PlayerManager playerManager = NitroxServiceLocator.LocateService<PlayerManager>();
+            playerManager.SendPacketToAllPlayers(new ChatMessage(ChatMessage.SERVER_ID, message));
+            Logger.ZLogInformation($"[BROADCAST] {message}");
+        }
+
```
dartasen, line 80: I guess this could be deleted.

Shouldn't this moved to specialized classes, so they can inject PlayerManager properly

Nitrox.Server.Subnautica/Models/Commands/GameModeCommand.cs
```diff
@@ -9,23 +8,25 @@ namespace Nitrox.Server.Subnautica.Models.Commands;
 internal class GameModeCommand : Command
```
dartasen, line 8: sealed

Nitrox.Server.Subnautica/Models/Commands/DirectoryCommand.cs
```diff
@@ -1,33 +1,47 @@
 ﻿using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
-using System.Reflection;
 using Nitrox.Model.DataStructures.GameLogic;
 using Nitrox.Server.Subnautica.Models.Commands.Abstract;
 
 namespace Nitrox.Server.Subnautica.Models.Commands
 {
     internal class DirectoryCommand : Command
```
dartasen, line 9: sealed

Nitrox.Server.Subnautica/Models/Commands/MemoryCompactCommand.cs
```diff
@@ -0,0 +1,20 @@
...
+using Nitrox.Server.Subnautica.Models.Commands.Abstract;
+using Nitrox.Server.Subnautica.Services;
+
+namespace Nitrox.Server.Subnautica.Models.Commands;
+
+internal class MemoryCompactCommand : Command
+{
+    private readonly MemoryService memoryService;
+
+    public MemoryCompactCommand(MemoryService memoryService) : base("memorycompact", Perms.HOST, "Compacts memory")
```
dartasen: 👀

Nitrox.Server.Subnautica/Models/Commands/MemoryCompactCommand.cs
```diff
@@ -0,0 +1,20 @@
+﻿using Nitrox.Model.DataStructures.GameLogic;
+using Nitrox.Server.Subnautica.Models.Commands.Abstract;
+using Nitrox.Server.Subnautica.Services;
+
+namespace Nitrox.Server.Subnautica.Models.Commands;
+
+internal class MemoryCompactCommand : Command
```
dartasen: sealed

Nitrox.Server.Subnautica/Models/Commands/SaveCommand.cs
```diff
@@ -1,18 +1,22 @@
...
         {
+            this.saveService = saveService;
         }
 
         protected override void Execute(CallArgs args)
         {
-            Server.Instance.Save();
-            SendMessageToPlayer(args.Sender, "World saved");
+            saveService.QueueSave();
+            SendMessageToPlayer(args.Sender, "World is saving...");
```
dartasen:
```suggestion
            SendMessageToPlayer(args.Sender, "Saving world...");
```

Nitrox.Server.Subnautica/Models/Commands/StopCommand.cs
```diff
@@ -6,21 +6,17 @@ namespace Nitrox.Server.Subnautica.Models.Commands
 {
     internal class StopCommand : Command
     {
-        public override IEnumerable<string> Aliases { get; } = new[] { "exit", "halt", "quit", "close" };
+        private readonly IHostApplicationLifetime lifetime;
+        public override IEnumerable<string> Aliases { get; } = ["exit", "halt", "quit", "close"];
 
-        public StopCommand() : base("stop", Perms.ADMIN, "Stops the server")
+        public StopCommand(IHostApplicationLifetime lifetime) : base("stop", Perms.ADMIN, "Stops the server")
```
dartasen:
```suggestion
        public StopCommand(IHostApplicationLifetime lifetimeService) : base("stop", Perms.ADMIN, "Stops the server")
```

Nitrox.Server.Subnautica/Models/Commands/SummaryCommand.cs
```diff
@@ -1,23 +1,38 @@
...
             AllowedArgOverflow = true;
         }
 
         protected override void Execute(CallArgs args)
         {
-            Perms viewerPerms = args.Sender.OrNull()?.Permissions ?? Perms.PLAYER;
-            SendMessage(args.Sender, server.GetSaveSummary(viewerPerms));
+            Perms viewerPerms = args.Sender.OrNull()?.Permissions ?? Perms.HOST;
+            Player? sender = args.Sender.OrNull();
+            // TODO: Make command execute async
```
dartasen, line 25: 👁️

Nitrox.Server.Subnautica/Models/Configuration/Providers/NitroxConfigurationSource.cs
```diff
@@ -0,0 +1,32 @@
...
+using Microsoft.Extensions.FileProviders;
+
+namespace Nitrox.Server.Subnautica.Models.Configuration.Providers;
+
+internal sealed class NitroxConfigurationSource : FileConfigurationSource
+{
+    /// <summary>
+    ///     The config section to insert the parsed keys into.
+    /// </summary>
+    public string Section { get; init; }
```
dartasen:
```suggestion
    public required string Section { get; init; }
```
or
```suggestion
    public string? Section { get; init; }
```

Nitrox.Server.Subnautica/Models/Configuration/NitroxFormatterOptions.cs
```diff
@@ -0,0 +1,25 @@
...
+
+    public bool UseRedaction => Redactors.Length > 0;
+
+    /// <summary>
+    ///     If true, logs won't be processed if they have a <see cref="CaptureScope" />. Requires
+    ///     <see cref="ConsoleFormatterOptions.IncludeScopes" /> to be true.
+    /// </summary>
+    public bool OmitWhenCaptured { get; set; }
+
+    public IRedactor[] Redactors { get; set; } = [];
```
dartasen, line 24:
```suggestion
    public IRedactor[] Redactors { get; private set; } = [];
```

Nitrox.Server.Subnautica/Models/GameLogic/Entities/Spawning/BatchEntitySpawner.cs
```diff
@@ -1,29 +1,28 @@
...
 using Nitrox.Server.Subnautica.Models.Helper;
 using Nitrox.Server.Subnautica.Models.Resources;
+using Nitrox.Server.Subnautica.Models.Resources.Core;
+using Nitrox.Server.Subnautica.Models.Resources.Parsers;
 using Nitrox.Server.Subnautica.Models.Serialization;
 
 namespace Nitrox.Server.Subnautica.Models.GameLogic.Entities.Spawning;
 
-public class BatchEntitySpawner : IEntitySpawner
+sealed class BatchEntitySpawner : IEntitySpawner
```
dartasen, line 15: missing access modifier

Nitrox.Server.Subnautica/Models/GameLogic/Entities/EntityData.cs
```diff
@@ -15,73 +10,6 @@ public class EntityData
...
-        bool brokenEntitiesFound = false;
-
-        foreach (Entity entity in Entities)
-        {
-            if (entity is WorldEntity we)
-            {
-                NitroxVector3 pos = we.Transform.LocalPosition;
-                if (float.IsNaN(pos.X) || float.IsNaN(pos.Y) || float.IsNaN(pos.Z) ||
-                    float.IsInfinity(pos.X) || float.IsInfinity(pos.Y) || float.IsInfinity(pos.Z))
-                {
```
dartasen, line 34: It's unused ?

Nitrox.Server.Subnautica/Models/GameLogic/Entities/EntitySimulation.cs
```diff
@@ -7,23 +6,23 @@
 
 namespace Nitrox.Server.Subnautica.Models.GameLogic.Entities;
 
-public class EntitySimulation
+sealed class EntitySimulation
```
dartasen, line 9: missing access modifier

Nitrox.Server.Subnautica/Models/GameLogic/Entities/GlobalRootData.cs
```diff
@@ -12,57 +10,6 @@ public class GlobalRootData
...
-    [ProtoAfterDeserialization]
-    private void ProtoAfterDeserialization()
-    {
-        foreach (GlobalRootEntity entity in Entities)
-        {
-            EnsureChildrenTransformAreParented(entity);
-        }
-    }
-
-    [OnDeserialized]
```
dartasen, line 24: Unused ?

Nitrox.Server.Subnautica/Models/GameLogic/Entities/SubnauticaUweWorldEntityFactory.cs
```diff
@@ -0,0 +1,27 @@
...
+using Nitrox.Server.Subnautica.Models.Resources.Parsers;
+using UWE;
+
+namespace Nitrox.Server.Subnautica.Models.GameLogic.Entities;
+
+internal class SubnauticaUweWorldEntityFactory(WorldEntitiesResource resource) : IUweWorldEntityFactory
+{
+    private readonly WorldEntitiesResource resource = resource;
+
+    public bool TryFind(string classId, out UweWorldEntity? uweWorldEntity)
```
dartasen:
```suggestion
    public bool TryFind(string classId, [NotNullWhen(true)] out UweWorldEntity? uweWorldEntity)
```

Nitrox.Server.Subnautica/Models/GameLogic/Entities/SubnauticaUwePrefabFactory.cs
```diff
@@ -0,0 +1,44 @@
...
+using static LootDistributionData;
+
+namespace Nitrox.Server.Subnautica.Models.GameLogic.Entities;
+
+internal class SubnauticaUwePrefabFactory(EntityDistributionsResource distributionData) : IUwePrefabFactory
+{
+    private readonly EntityDistributionsResource resource = distributionData;
+    private readonly Dictionary<string, List<UwePrefab>> cache = new();
+
+    public bool TryGetPossiblePrefabs(string? biome, out List<UwePrefab>? prefabs)
```
dartasen:
```suggestion
    public bool TryGetPossiblePrefabs(string? biome, [NotNullWhen(true)] out List<UwePrefab>? prefabs)
```

Nitrox.Server.Subnautica/Models/GameLogic/Entities/WorldEntityManager.cs
```diff
@@ -11,44 +12,39 @@
...
 /// <remarks>
-/// Regular <see cref="WorldEntity"/> are held in cells and should be registered in <see cref="worldEntitiesByBatchId"/> and <see cref="worldEntitiesByCell"/>.
-/// But <see cref="GlobalRootEntity"/> are held in their own root object (GlobalRoot) so they should never be registered in cells (they're seeable at all times).
+///     Regular <see cref="WorldEntity" /> are held in cells and should be registered in <see cref="worldEntitiesByCell" />
+///     .
+///     But <see cref="GlobalRootEntity" /> are held in their own root object (GlobalRoot) so they should never be
+///     registered in cells (they're seeable at all times).
 /// </remarks>
-public class WorldEntityManager
+// TODO: Refactor to not use internal dictionaries but instead depend on EntityRegistry for data. Maybe add API to EntityRegistry to set custom filter+indexer on entity type (i.e. WorldEntity) for improved world entity lookups?
```
dartasen, line 20: 👁️

Nitrox.Server.Subnautica/Models/GameLogic/ConnectionAssets.cs
```diff
@@ -2,7 +2,7 @@
 {
     internal class ConnectionAssets
     {
-        public string ReservationKey { get; set; }
-        public Player Player { get; set; }
+        public string? ReservationKey { get; set; }
+        public Player? Player { get; set; }
     }
```
dartasen, line 7: record ?

Nitrox.Server.Subnautica/Models/GameLogic/PlayerManager.cs
```diff
@@ -205,17 +208,15 @@ public void PlayerDisconnected(INitroxConnection connection)
...
 
         assetsByConnection.Remove(connection);
 
-        PlayerCountChanged?.Invoke(connectedPlayersById.Count);
-
         if (!ConnectedPlayers().Any())
         {
-            Server.Instance.PauseServer();
-            Server.Instance.Save();
+            // TODO: Make this function async
```
dartasen, line 218: 👁️

Nitrox.Server.Subnautica/Models/Logging/Middleware/Core/ILoggerMiddleware.cs
```diff
@@ -0,0 +1,23 @@
...
+
+namespace Nitrox.Server.Subnautica.Models.Logging.Middleware.Core;
+
+internal interface ILoggerMiddleware
+{
+    public delegate void NextCall(ref Context context);
+
+    void ExecuteLogMiddleware(ref Context context, NextCall next);
+
+    public ref struct Context
```
dartasen, line 11: ref struct 😍

Nitrox.Server.Subnautica/Models/Packets/Processors/BuildingResyncRequestProcessor.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace Nitrox.Server.Subnautica.Models.Packets.Processors;
 
-public class BuildingResyncRequestProcessor : AuthenticatedPacketProcessor<BuildingResyncRequest>
+sealed class BuildingResyncRequestProcessor : AuthenticatedPacketProcessor<BuildingResyncRequest>
```
dartasen, line 10: missing access modifier

Nitrox.Server.Subnautica/Models/Packets/Processors/CellVisibilityChangedProcessor.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace Nitrox.Server.Subnautica.Models.Packets.Processors;
 
-public class CellVisibilityChangedProcessor : AuthenticatedPacketProcessor<CellVisibilityChanged>
+sealed class CellVisibilityChangedProcessor : AuthenticatedPacketProcessor<CellVisibilityChanged>
```
dartasen, line 10: missing access modifier

Nitrox.Server.Subnautica/Models/Packets/Processors/ChatMessageProcessor.cs
```diff
@@ -3,13 +3,15 @@
 
 namespace Nitrox.Server.Subnautica.Models.Packets.Processors
```
dartasen, line 4: namespace

Nitrox.Server.Subnautica/Models/Packets/Processors/ChatMessageProcessor.cs
```diff
@@ -3,13 +3,15 @@
...
 {
-    public class ChatMessageProcessor : AuthenticatedPacketProcessor<ChatMessage>
+    sealed class ChatMessageProcessor : AuthenticatedPacketProcessor<ChatMessage>
```
dartasen, line 6: missing access modifier

Nitrox.Server.Subnautica/Models/Packets/Processors/CyclopsDamageProcessor.cs
```diff
@@ -6,18 +6,20 @@ namespace Nitrox.Server.Subnautica.Models.Packets.Processors
     /// <summary>
     /// This is the absolute damage state. The current simulation owner is the only one who sends this packet to the server
     /// </summary>
-    public class CyclopsDamageProcessor : AuthenticatedPacketProcessor<CyclopsDamage>
+    sealed class CyclopsDamageProcessor : AuthenticatedPacketProcessor<CyclopsDamage>
```
dartasen, line 9: missing access modifier

Nitrox.Server.Subnautica/Models/Packets/Processors/DefaultServerPacketProcessor.cs
```diff
@@ -1,13 +1,13 @@
-using System;
 using System.Collections.Generic;
 using Nitrox.Server.Subnautica.Models.Packets.Processors.Core;
 using Nitrox.Server.Subnautica.Models.GameLogic;
 
 namespace Nitrox.Server.Subnautica.Models.Packets.Processors;
 
-public class DefaultServerPacketProcessor : AuthenticatedPacketProcessor<Packet>
+sealed class DefaultServerPacketProcessor : AuthenticatedPacketProcessor<Packet>
```
dartasen, line 7: missing access modifier

Nitrox.Server.Subnautica/Models/Packets/Processors/EntityReparentedProcessor.cs
```diff
@@ -5,27 +5,22 @@
 
 namespace Nitrox.Server.Subnautica.Models.Packets.Processors;
 
-public class EntityReparentedProcessor : AuthenticatedPacketProcessor<EntityReparented>
+sealed class EntityReparentedProcessor(EntityRegistry entityRegistry, PlayerManager playerManager, ILogger<EntityReparentedProcessor> logger) : AuthenticatedPacketProcessor<EntityReparented>
```
dartasen, line 8: missing access modifier

Nitrox.Server.Subnautica/Models/Packets/Processors/EscapePodChangedPacketProcessor.cs
```diff
@@ -3,18 +3,20 @@
...
+        public EscapePodChangedPacketProcessor(PlayerManager playerManager, ILogger<EscapePodChangedPacketProcessor> logger)
         {
             this.playerManager = playerManager;
+            this.logger = logger;
         }
 
         public override void Process(EscapePodChanged packet, Player player)
         {
-            Log.Debug(packet);
+            logger.ZLogDebug($"{packet}");
```
dartasen: We can remove this log, seems like a leftover

Nitrox.Server.Subnautica/Models/Packets/Processors/FMODAssetProcessor.cs
```diff
@@ -2,25 +2,21 @@
 using Nitrox.Model.GameLogic.FMOD;
 using Nitrox.Server.Subnautica.Models.Packets.Processors.Core;
 using Nitrox.Server.Subnautica.Models.GameLogic;
+using Nitrox.Server.Subnautica.Services;
 
 namespace Nitrox.Server.Subnautica.Models.Packets.Processors;
 
-public class FMODAssetProcessor : AuthenticatedPacketProcessor<FMODAssetPacket>
+sealed class FMODAssetProcessor(PlayerManager playerManager, FmodService fmodService, ILogger<FMODAssetProcessor> logger) : AuthenticatedPacketProcessor<FMODAssetPacket>
```
dartasen, line 9: missing access modifier

Nitrox.Server.Subnautica/Models/Packets/Processors/FootstepPacketProcessor.cs
```diff
@@ -2,23 +2,29 @@
 using Nitrox.Model.GameLogic.FMOD;
 using Nitrox.Server.Subnautica.Models.Packets.Processors.Core;
 using Nitrox.Server.Subnautica.Models.GameLogic;
+using Nitrox.Server.Subnautica.Services;
 
 namespace Nitrox.Server.Subnautica.Models.Packets.Processors;
 
-public class FootstepPacketProcessor : AuthenticatedPacketProcessor<FootstepPacket>
+sealed class FootstepPacketProcessor : AuthenticatedPacketProcessor<FootstepPacket>
```
dartasen, line 9: missing access modifier

Nitrox.Server.Subnautica/Models/Packets/Processors/FootstepPacketProcessor.cs
```diff
@@ -2,23 +2,29 @@
...
     {
         this.playerManager = playerManager;
-        whitelist.TryGetSoundData("event:/player/footstep_precursor_base", out SoundData soundData);
-        footstepAudioRange = soundData.Radius;
+        this.fmodService = fmodService;
     }
 
     public override void Process(FootstepPacket footstepPacket, Player sendingPlayer)
     {
+        if (footstepAudioRange == 0f && fmodService.TryGetSoundData("event:/player/footstep_precursor_base", out SoundData soundData))
```
dartasen, line 23: Since this packet is sent quite a lot, we should cache the value as it was done previously

Nitrox.Server.Subnautica/Models/Packets/Processors/KnownTechEntryAddProcessor.cs
```diff
@@ -1,29 +1,28 @@
 using Nitrox.Server.Subnautica.Models.Packets.Processors.Core;
 using Nitrox.Server.Subnautica.Models.GameLogic;
-using Nitrox.Server.Subnautica.Models.GameLogic.Unlockables;
 
 namespace Nitrox.Server.Subnautica.Models.Packets.Processors
```
dartasen, line 4: namespace

Nitrox.Server.Subnautica/Models/Packets/Processors/MultiplayerSessionReservationRequestProcessor.cs
```diff
@@ -6,18 +6,20 @@
 
 namespace Nitrox.Server.Subnautica.Models.Packets.Processors
 {
-    public class MultiplayerSessionReservationRequestProcessor : UnauthenticatedPacketProcessor<MultiplayerSessionReservationRequest>
+    sealed class MultiplayerSessionReservationRequestProcessor : UnauthenticatedPacketProcessor<MultiplayerSessionReservationRequest>
```
dartasen, line 9: namespace + missing accesser

If we can double check every file in this folder

Nitrox.Server.Subnautica/Models/Packets/Processors/PDAEncyclopediaEntryAddProcessor.cs
```diff
@@ -1,23 +1,22 @@
 ﻿using Nitrox.Server.Subnautica.Models.Packets.Processors.Core;
 using Nitrox.Server.Subnautica.Models.GameLogic;
-using Nitrox.Server.Subnautica.Models.GameLogic.Unlockables;
 
 namespace Nitrox.Server.Subnautica.Models.Packets.Processors
 {
-    public class PDAEncyclopediaEntryAddProcessor : AuthenticatedPacketProcessor<PDAEncyclopediaEntryAdd>
+    internal sealed class PDAEncyclopediaEntryAddProcessor : AuthenticatedPacketProcessor<PDAEncyclopediaEntryAdd>
```
dartasen, line 6: namespace

If we can double check every file in this folder

Nitrox.Server.Subnautica/Models/Packets/Processors/PieceDeconstructedProcessor.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace Nitrox.Server.Subnautica.Models.Packets.Processors;
 
-public class PieceDeconstructedProcessor : BuildingProcessor<PieceDeconstructed>
+sealed class PieceDeconstructedProcessor : BuildingProcessor<PieceDeconstructed>
```
dartasen, line 6: missing accesser

Nitrox.Server.Subnautica/Models/Resources/Parsers/EntityDistributionsResource.cs
```diff
@@ -0,0 +1,64 @@
...
+
+namespace Nitrox.Server.Subnautica.Models.Resources.Parsers;
+
+internal class EntityDistributionsResource(SubnauticaAssetsManager assetsManager, IOptions<ServerStartOptions> options) : IGameResource
+{
+    private readonly SubnauticaAssetsManager assetsManager = assetsManager;
+    private readonly IOptions<ServerStartOptions> options = options;
+
+    private ValueTask<LootDistributionData> lootDistribution;
+    public LootDistributionData LootDistribution => GetLootDistributionDataAsync().GetAwaiter().GetResult();
```
dartasen, line 16: GetAwaiter GetResult 🥲

Nitrox.Server.Subnautica/Models/Serialization/SaveDataUpgrades/NewtonsoftExtensions.cs
```diff
@@ -1,5 +1,4 @@
 ﻿#nullable enable
```
dartasen: I guess this could be removed

Nitrox.Server.Subnautica/Models/Serialization/World/PersistedWorldData.cs
```diff
@@ -1,40 +1,23 @@
...
-                    ParsedBatchCells = world.BatchEntitySpawner.SerializableParsedBatches,
-                    GameData = GameData.From(world.GameData.PDAState, world.GameData.StoryGoals, world.ScheduleKeeper, world.StoryManager, world.TimeKeeper),
-                    Seed = world.Seed,
-                },
-                PlayerData = PlayerData.From(world.PlayerManager.GetAllPlayers()),
-                GlobalRootData = GlobalRootData.From(world.WorldEntityManager.GetPersistentGlobalRootEntities()),
-                EntityData = EntityData.From(world.EntityRegistry.GetAllEntities(exceptGlobalRoot: true))
-            };
-        }
+        public EntityData? EntityData { get; set; }
```
dartasen, line 20: I don't think those should be flagged as nullable

Nitrox.Server.Subnautica/Models/Serialization/World/WorldData.cs
```diff
@@ -6,16 +6,16 @@
...
-        public List<NitroxInt3> ParsedBatchCells { get; set; } = [];
+        public List<NitroxInt3>? ParsedBatchCells { get; set; } = [];
 
         [DataMember(Order = 2)]
-        public GameData GameData { get; set; }
+        public GameData? GameData { get; set; }
 
         [DataMember(Order = 3)]
-        public string Seed { get; set; }
+        public string? Seed { get; set; }
```
dartasen, line 18: Same for nullable

Nitrox.Server.Subnautica/Models/Serialization/ServerJsonSerializer.cs
```diff
@@ -12,13 +12,13 @@ public class ServerJsonSerializer : IServerSerializer
...
 
-    public ServerJsonSerializer()
+    public ServerJsonSerializer(ILogger<ServerJsonSerializer> logger)
     {
         serializer = new JsonSerializer();
 
         serializer.Error += delegate (object _, Newtonsoft.Json.Serialization.ErrorEventArgs e)
         {
-            Log.Error(e.ErrorContext.Error, "Json serialization error: ");
+            logger.ZLogError(e.ErrorContext.Error, $"serialization error: ");
```
dartasen: string interpolation unused

Nitrox.Server.Subnautica/Models/AssemblyResolver.cs
```diff
@@ -0,0 +1,118 @@
...
+                cacheEntry.Attempts++;
+                if (cacheEntry.Attempts >= 5)
+                {
+                    throw new FileNotFoundException($"Failed to load DLL '{dllName}' at: {dllPath}");
+                }
+                return null;
+            }
+        }
+
+        Assembly assembly = ResolveFromLib(args.Name);
```
dartasen:
```suggestion
        Assembly? assembly = ResolveFromLib(args.Name);
```

Nitrox.Server.Subnautica/Models/AssemblyResolver.cs
```diff
@@ -0,0 +1,118 @@
...
+    private static readonly Dictionary<string, AssemblyCacheEntry> resolvedAssemblyCache = [];
+    private static string? gamePath;
+
+    /// <summary>
+    ///     The path to the game files so that game code can be loaded when needed.
+    /// </summary>
+    public static string? GamePath
+    {
+        get => Interlocked.Exchange(ref gamePath, gamePath);
+        set => Interlocked.Exchange(ref gamePath, value);
```
dartasen: I don't think Interlocked is useful here, doesn't seem to be a multi-threaded context

Nitrox.Server.Subnautica/Services/CommandService.cs
```diff
@@ -0,0 +1,50 @@
...
+namespace Nitrox.Server.Subnautica.Services;
+
+/// <summary>
+///     Enables processing of commands from a string.
+/// </summary>
+internal sealed class CommandService(TextCommandProcessor textCommandProcessor, ILogger<CommandService> logger) : IHostedService
+{
+    private readonly TextCommandProcessor textCommandProcessor = textCommandProcessor;
+    private readonly Channel<Task> runningCommands = Channel.CreateUnbounded<Task>();
+    private Task commandWaiterTask;
```
dartasen:
```suggestion
    private Task? commandWaiterTask;
```

Nitrox.Server.Subnautica/Services/FmodService.cs
```diff
@@ -0,0 +1,30 @@
...
+    public Task StartAsync(CancellationToken cancellationToken)
+    {
+        whitelist = FMODWhitelist.Load(gameInfo);
+        return Task.CompletedTask;
+    }
+
+    public Task StopAsync(CancellationToken cancellationToken)
+    {
+        return Task.CompletedTask;
+    }
```
dartasen, line 20: Just a thought

Should we clear the whitelist on stop ? 

I guess it's done indirectly because that would mean the server process is being killed, but what if added a reload command like minecraft servers

Nitrox.Server.Subnautica/Services/FmodService.cs
```diff
@@ -0,0 +1,30 @@
...
+
+    public bool TryGetSoundData(string assetPath, out SoundData soundData)
```
dartasen, line 22:
```suggestion
    public bool TryGetSoundData(string assetPath, [NotNullWhen(true)] out SoundData? soundData)
```

Nitrox.Server.Subnautica/Services/HibernateService.cs
```diff
@@ -0,0 +1,52 @@
...
+
+namespace Nitrox.Server.Subnautica.Services;
+
+internal sealed class HibernateService(IHibernate.SleepTrigger sleepTrigger, IHibernate.WakeTrigger wakeTrigger, ILogger<HibernateService> logger) : IHostedLifecycleService
+{
+    private readonly IHibernate.SleepTrigger sleepTrigger = sleepTrigger;
+    private readonly IHibernate.WakeTrigger wakeTrigger = wakeTrigger;
+    private readonly ILogger<HibernateService> logger = logger;
+    private bool isSleeping;
+    public bool IsSleeping => Interlocked.CompareExchange(ref isSleeping, true, true);
```
dartasen: Volatile.Read to avoid necessary operation maybe ? (like the write)

This behavior might be common when using service with lifecycle, and we might wanna factor this to avoid people having to deal with multi-thread shennanigans

Nitrox.Server.Subnautica/Services/MemoryService.cs
```diff
@@ -0,0 +1,57 @@
...
+            {
+                retries = STARTING_RETRIES;
+            }
+            if (--retries < 0)
+            {
+                break;
+            }
+        }
+        logger.ZLogTrace($"Stopped compacting memory");
+    }
```
dartasen, line 65: This is already what the GC is doing behind the scene ?

That's the main reason we can't give raw pointer to objects to unmanaged world (must use GCHandle) because GC might compact and move memory

Nitrox.Server.Subnautica/Services/NetworkPortAvailabilityService.cs
```diff
@@ -0,0 +1,75 @@
...
+    public Task StoppingAsync(CancellationToken cancellationToken) => Task.CompletedTask;
+
+    public Task StoppedAsync(CancellationToken cancellationToken) => Task.CompletedTask;
+
+    private async Task WaitForAvailablePortAsync(int port, CancellationToken ct = default)
+    {
+        bool first = true;
+        try
+        {
+            while (true)
```
dartasen, line 36: while true 👀

Nitrox.Server.Subnautica/Models/ConsoleUnhandledErrorHandler.cs
Measurity: This should be active the entire process. Detaching code might lead to new bugs so I left it out.

Nitrox.Server.Subnautica/Services/CommandService.cs
Measurity: Not gonna do this now because PR is already huge. Will do after this PR.

Nitrox.Launcher/Models/Design/ServerEntry.cs
Measurity: Our code format reorders fields automatically like this. Would have to change that but is out of scope of this PR imo.

Nitrox.Launcher/Models/Design/ServerEntry.cs
Measurity: Can't be changed to `private` or `public` because it references an inner type that's `internal`.

Nitrox.Launcher/Models/Design/ServerEntry.cs
Measurity: It's for the `ILogger` parameter that was added. Shared types shouldn't depend on `Log` directly but on `ILogger` so it can be swapped for something that does nothing.

Nitrox.Launcher/Models/Design/ServerEntry.cs
Measurity: yes, to give the UI a chance to update based on the new state of `IsServerClosing`. Otherwise it would be delayed until the next `await` to give the main thread time to compute stuff.

Nitrox.Model/Packets/Packet.cs
Measurity: ToString can be called on different threads at the same time. Had exceptions happening on that so added the ThreadStatic which solved it.

Nitrox.Server.Subnautica/Models/Commands/Abstract/Command.cs
Measurity: The entire command system is left as untouched as possible to prevent the 500+ files changed PR. Will do packets and command stuff after this PR.

Nitrox.Server.Subnautica/Services/MemoryService.cs
Measurity: Normally, yes. But because we use AssetsTools.NET which holds open file handlers without a way to close them (without reflection), this doesn't happen. That's why MemoryService exists, to force the GC to check if any file handler (pointers) are reachable and if not, deallocate them. Will add a comment.

Nitrox.Server.Subnautica/Models/GameLogic/Entities/GlobalRootData.cs
Measurity: It's now handled in WorldService. Depending on Protobug code is not desired.

Nitrox.Server.Subnautica/Models/GameLogic/Entities/EntityData.cs
Measurity: Handled in WorldService now.

Nitrox.Server.Subnautica/Models/GameLogic/Entities/WorldEntityManager.cs
Measurity: I didn't want to include the change here because it's critical to Nitrox and didn't want to sneak in bugs.

Nitrox.Server.Subnautica/Models/Logging/Middleware/Core/ILoggerMiddleware.cs
Measurity: Modern C# my beloved :heart: (lol)

Nitrox.Server.Subnautica/Models/Resources/Parsers/EntityDistributionsResource.cs
Measurity: Would require too much refactoring to do without GetAwaiter().GetResult() for this PR.

Measurity:
Will do the packet `internal sealed` + namespace block changes in another PR to avoid changing a lot of files and make this PR insanely huge.

Nitrox.Server.Subnautica/Models/Packets/Processors/FootstepPacketProcessor.cs
Measurity: This is still cached because `footstepAudioRange` will be not `0` so `TryGetSoundData` won't be called.

Nitrox.Server.Subnautica/Models/GameLogic/PlayerManager.cs
Measurity: Will handle TODO in another PR. Needs command system to be async as well so it requires too much refactoring.

Nitrox.Server.Subnautica/Models/Serialization/World/PersistedWorldData.cs
Measurity: The `IsValid` check does check for null. I followed how the code was currently written to apply nullability. Think any refactoring outside of .NET hosting API stuff is best left in another PR.

Nitrox.Server.Subnautica/Services/FmodService.cs
Measurity: SoundData is a struct so it can't be null. I like to keep the refactoring that doesn't need to happen in this PR, out of it.

Nitrox.Server.Subnautica/Models/GameLogic/ConnectionAssets.cs
Measurity: other code doesn't handle it like a record. changes are best suited in another PR.

Nitrox.Server.Subnautica/Models/Configuration/NitroxFormatterOptions.cs
Measurity: This options field must be a public settable or it has to be readonly using `with` statement. I'll keep it as is for now.

Nitrox.Server.Subnautica/Models/Configuration/Providers/NitroxConfigurationSource.cs
Measurity: Section should not be set to null. But I added the `required` keyword as suggested.

Nitrox.Server.Subnautica/Services/NetworkPortAvailabilityService.cs
Measurity: It's a while true but it can never loop forever. The cancellation token is immediately checked and will throw after 30 seconds.

Nitrox.Server.Subnautica/Services/HibernateService.cs
Measurity: `Volatile.Read` does not guarantee thread safe write/read, it reorders the instructions so the read will be "fresh" and not old value in a CPU cache. This means it would still be possible to have two writes at the same time messing up the value. I'll keep the `Interlocked` for now.

Added a setter to `IsSleeping` to avoid others having to write thread safe code.

Nitrox.Model/Packets/Packet.cs
Measurity: I like it as a nested type because it's specifically designed for `Packet`. If it was used as a field outside of Packet then it being in its own file would be better.

Nitrox.Model/Extensions/RangeExtensions.cs
Measurity: Added `MODERN_DOTNET` constant which define itself to match latest `NETX_X_OR_GREATER`.

Nitrox.Server.Subnautica/Services/FmodService.cs
Measurity: Don't really have an opinion either way on what's best. I'll add it just in case.

Nitrox.Launcher/Models/Design/ServerEntry.cs
Measurity: I have to specify `async void` to be able to do `await`. Without await I think all the calls inside will not wait on each other which could cause weird behavior.

Nitrox.Model/Configuration/Validators/SaveNameAttribute.cs
Measurity: Added unit test for save name validator. The other attributes are so simple (directly calling .NET APIs) that unit testing isn't worth it.

Measurity:
Handled all review comments now

Nitrox.Model/Serialization/NitroxConfig.cs
```diff
@@ -136,38 +156,38 @@ public void Serialize(string saveDir)
...
+        if (NitroxConfig<T>.TypeCache.Count == 0)
         {
+            Type type = typeof(T);
+            if (type.Name == "NitroxConfig`1")
+            {
+                type = type.GetGenericArguments().FirstOrDefault() ?? type;
+            }
+
             IEnumerable<MemberInfo> members = type.GetFields()
+                                                  .Where(f => f.Attributes != FieldAttributes.NotSerialized && !f.IsLiteral)
```
dartasen: FieldAttributes.NotSerialized seems obsolete in .NET9+

Nitrox.Server.Subnautica/Models/Communication/LiteNetLibServer.cs
```diff
@@ -40,91 +71,62 @@ public override bool Start(CancellationToken ct = default)
...
-            }
-        }
+        playerManager.SendPacketToAllPlayers(new ServerStopped());
+        // We want every player to receive this packet
+        await Task.Delay(500, cancellationToken);
+        server.Stop();
     }
 
-    public override void Stop()
+    protected void ClientDisconnected(INitroxConnection connection)
```
dartasen: class is sealed, it's generating a warning here
`'LiteNetLibServer.ClientDisconnected(INitroxConnection)': new protected member declared in sealed type`

Not new to that PR but there's another one : 
`'NitroxId.NitroxId(SerializationInfo, StreamingContext)': new protected member declared in sealed type`

Nitrox.Launcher/Models/Services/ServerService.cs
```diff
@@ -109,24 +105,12 @@ public async Task<bool> StartServerAsync(ServerEntry server)
...
-            server.Start(keyValueStore.GetSavesFolderDir(), onExited: () =>
-            {
-                lock (knownServerProcessIdsLock)
-                {
-                    knownServerProcessIds.Remove(server.Process?.Id ?? 0);
-                }
-            });
-            if (server.IsEmbedded)
+            bool preferEmbedded = keyValueStore.GetPreferEmbedded();
+            server.StartAsync(keyValueStore.GetSavesFolderDir(), preferEmbedded);
```
dartasen: Call isn't awaited

Nitrox.Model/Serialization/NitroxConfig.cs
```diff
@@ -176,27 +196,27 @@ private static Dictionary<string, MemberInfo> GetTypeCacheDictionary()
...
                 foreach (MemberInfo member in members)
                 {
-                    typeCache.Add(member.Name.ToLowerInvariant(), member);
+                    NitroxConfig<T>.TypeCache.Add(member.Name.ToLowerInvariant(), member);
                 }
             }
             catch (ArgumentException e)
             {
-                Log.Error(e, $"Type {type.FullName} has properties that require case-sensitivity to be unique which is unsuitable for .properties format.");
+                logger?.LogError(e, $"Type {type.FullName} has properties that require case-sensitivity to be unique which is unsuitable for .properties format.");
```
dartasen: Complaining we're using interpolated string in the template, instead of using template, parameters

dartasen (approved):

PR #2322, opened 2025-04-27
Fixes #2318, battery/power cell in chargers disappearing on resync

### LINKED ISSUE #2318: Batteries in Battery Charger disappear after server sync/restart and remain invisible to others

Windows

Steam

### Describe the issue.

When I place batteries into a Battery Charger, they disappear after either resynchronizing server buildings (using the relevant command) or restarting the Nitrox server entirely. The charger appears empty after the sync or restart.
Furthermore, even if I remove a battery from the charger before the sync/restart and place it somewhere else (e.g., drop it on the ground, put it in a storage locker), other players connected to the server cannot see this battery. It seems the battery becomes desynchronized and invisible to others once it has been interacted with via the Battery Charger.
Batteries placed in the Battery Charger should remain there and be visible to all players after a server building sync or a server restart. Batteries removed from the charger should also remain synchronized and visible/interactable for all players when placed elsewhere in the game world.

Build and place a Battery Charger inside a base.
Place one or more standard Batteries into the slots of the Battery Charger.
Do one of the following:
Execute the command to synchronize server buildings.
Shut down and restart the Nitrox dedicated server.
After the sync completes or the server restarts and you reconnect, observe the Battery Charger. The batteries previously placed inside are now gone.
(Testing visibility issue): Before performing step 3, take a battery out of the Battery Charger.
Place this removed battery either on the ground or inside a storage container (e.g., a Wall Locker).
Ask another player connected to the server to look at the location where you placed the battery.
Observe that the other player cannot see the battery you placed.

### Relevant log output

[server\[fishworld\]-20250427.log](https://github.com/user-attachments/files/19924406/server.fishworld.-20250427.log)
[game\[fish\]-20250427.log](https://github.com/user-attachments/files/19924405/game.fish.-20250427.log)
[game\[qwe\]-20250427.log](https://github.com/user-attachments/files/19924414/game.qwe.-20250427.log)

LeanHydra84:
Fixes issue #2318 where batteries and power cells would disappear after rejoining the game or using the building resync feature.

`ModuleEntitySpawner` now spawns `InstalledModuleEntity` child entities as well as `InventoryItemEntity`. Previously batteries/powercells would not be spawned because of this.

~When the charger was restored in preparation for the resync it would broadcast `ModuleRemoved` for each battery and cause them to be lost. `InstalledModuleEntitySpawner` now uses the `Equipment.AddItem()` method instead of inlining the same steps, which broadcasts `ModuleAdded`.~

### DIFF
--- a/NitroxPatcher/Patches/Dynamic/Pickupable_OnDestroy_Patch.cs
+++ b/NitroxPatcher/Patches/Dynamic/Pickupable_OnDestroy_Patch.cs
@@ -1,0 +1,24 @@
+using System.Reflection;
+using NitroxClient.GameLogic.Bases;
+using NitroxModel.Helper;
+
+namespace NitroxPatcher.Patches.Dynamic;
+
+public sealed partial class Pickupable_OnDestroy_Patch : NitroxPatch, IDynamicPatch
+{
+    private static readonly MethodInfo TARGET_METHOD = Reflect.Method((Pickupable t) => t.OnDestroy());
+
+    public static bool Prefix(Pickupable __instance)
+    {
+        // stops Pickupable.OnDestroy from triggering the OnDestroy function if BuildingHandler is Resyncing
+        // this stops the IItemsContainer.RemoveItem() method from being called which broadcasts a packet in some implementations 
+        if (BuildingHandler.Main.Resyncing)
+        {
+            __instance.isDestroyed = true;
+            return false;
+        }
+
+        return true;
+    }
+    
+}
--- a/NitroxPatcher/Patches/Dynamic/Pickupable_OnDestroy_Patch.cs
+++ b/NitroxPatcher/Patches/Dynamic/Pickupable_OnDestroy_Patch.cs
@@ -10,7 +10,7 @@ public sealed partial class Pickupable_OnDestroy_Patch : NitroxPatch, IDynamicPa
 
     public static bool Prefix(Pickupable __instance)
     {
-        // stops Pickupable.OnDestroy from triggering the OnDestroy function if BuildingHandler is Resyncing
+        // Stops Pickupable.OnDestroy from triggering the OnDestroy function if BuildingHandler is Resyncing
         // this stops the IItemsContainer.RemoveItem() method from being called which broadcasts a packet in some implementations 
         if (BuildingHandler.Main.Resyncing)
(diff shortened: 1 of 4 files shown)


dartasen:
@tornac1234 If you can please review that one, if that makes sense for you

LeanHydra84:
For brevity batteries and power cells will both be referred to as batteries. This applies to both.

Nitrox currently has an issue where, during the resync when current base objects are destroyed, it triggers the `Pickupable.OnDestroy()` function, which then calls `IItemsContainer.RemoveItem()`, which for batteries in chargers is `Equipment.RemoveItem()`

`Equipment.RemoveItem()` is patched (the only of the `IItemsContainer.RemoveItem()` implementation that is) to broadcast the removal (`EquipmentSlots.BroadcastUnequip()`). This caused batteries to disconnect from their chargers when the resync happens.

I originally fixed this by making the batteries broadcast that they were added whenever they spawned as child entities. This was a flawed solution.

The new behavior ensures no packets are sent by patching `Pickupable.OnDestroy()` to not call `RemoveItem()` if a resync is currently happening (`BuildingHandler.Main.Resyncing == true`)

NitroxPatcher/Patches/Dynamic/Pickupable_OnDestroy_Patch.cs
```diff
@@ -0,0 +1,24 @@
...
+
+namespace NitroxPatcher.Patches.Dynamic;
+
+public sealed partial class Pickupable_OnDestroy_Patch : NitroxPatch, IDynamicPatch
+{
+    private static readonly MethodInfo TARGET_METHOD = Reflect.Method((Pickupable t) => t.OnDestroy());
+
+    public static bool Prefix(Pickupable __instance)
+    {
+        // stops Pickupable.OnDestroy from triggering the OnDestroy function if BuildingHandler is Resyncing
```
Coding-Hen: Minor upcase S and lowercase r for resyncing as it is not an object/method name.
```suggestion
        // Stops Pickupable.OnDestroy from triggering the OnDestroy function if BuildingHandler is resyncing
```

tornac1234 (approved):

Coding-Hen (approved):
LGTM

PR #2331, opened 2025-05-10
Fix vehicle docking and undocking for standalone moonpools

### LINKED ISSUE #2327: Vehicle docking and undocking is not synced
### Describe the issue

- Create a new world and join with two players
- Build a moonpool with a solar panel on top, and build a seamoth (or exosuit) with a mobile vehicle bay
- Have the first player dock the vehicle at the moonpool

The other player will not see the vehicle docked. The vehicle will be in the right position, but will slowly sink down and teleport back to where it should be every so often. If the second player enters the vehicle and holds C to avoid docking, they will be able to control it normally while the first player sees it as docked. If the second player now docks the vehicle, everything is resynced. Similarly, when undocking a synced vehicle with the first player, the second player sees the vehicle as docked, but exosuit arms will still rotate and animate as if the vehicle were moving freely.

Possible cause: the `VehicleDocked` packet is sent and received correctly, but fails to apply because the second player has this error:
```
[15:00:24.045] [ERR] [VehicleDockingProcessor] could not find VehicleDockingBay component on 51d18e22-d930-40ad-9fee-0b8b460d6928
```
The first player can see an entity named "DockBottom" with the id `51d18e22-d930-40ad-9fee-0b8b460d6928`, but the second player cannot.

SpaceMonkeyy86:
Fixes #2327

### DIFF
--- a/NitroxClient/GameLogic/Entities.cs
+++ b/NitroxClient/GameLogic/Entities.cs
@@ -4,7 +4,6 @@ using System.Collections.Generic;
 using System.Linq;
 using NitroxClient.Communication;
 using NitroxClient.Communication.Abstract;
-using NitroxClient.GameLogic.PlayerLogic.PlayerModel.Abstract;
 using NitroxClient.GameLogic.Spawning;
 using NitroxClient.GameLogic.Spawning.Abstract;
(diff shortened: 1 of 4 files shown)


SpaceMonkeyy86:
Closing in favor of #2347

PR #2332, opened 2025-05-10
Fix static dlls not being copied over lib\net472 folder

### LINKED ISSUE #2324: MS-Store not working in RELEASE

Windows

MS-Store/Windows Store (currently not supported, don't file any reports if your are using this)

### Describe the issue.

Current commit version not work after compile in release mode. There is no "Network Game" option in menu. When compile in debug mode this option is present.
Month ago version have no this bug.

Click Multiplayer button in Launcher.

### Relevant log output

Launcher: https://pastebin.com/9UdzQCZ4
Game: no logs

dartasen:
Closes #2324 

* Inside a Player.log, I noticed that protobuf-net.dll was not resolved by our custom AssemblyResolve mecanism, because Reference isn't transitive, so that would mean we need to reference a .NET dll on every project, even if it doesn't need linking
* Added back discord_game_sdk.dll, until we ship the native libs per rid inside Nuget


Measurity (approved):
Not sure how this PR fixes MS Store issue but changes make sense.

PR #2347, opened 2025-06-26
Fix standalone moonpools not being registered on remote clients

tornac1234:
This way is more natural and fitting to the system than #2331 because the only issue was that when you built a moonpool on its own, remote players would never go through the code that creates the MoonpoolManager although it was created by every other way to have this standalone moonpool (when joining a world or when placing it on local client).

Important note: for bases composed of a standalone moonpool, the first id after the base id will be reserved for the said standalone moonpool


Coding-Hen (approved):
LGTM

PR #2348, opened 2025-06-27
Add Sync for Stalker Teeth Drop

### LINKED ISSUE #2243: Stalker teeth invisible?

dartasen:
Closes #2243 

I've extracted the #2255  commit to allow at least this synchronization for 1.8, which seems to be an important part of Gameplay for crafting and base building.

I've included a correction to the code that handles probabilities, I've extracted from the wiki some components that seem to influence the drop of a Stalker tooth, I've restricted myself to metals/minerals for the moment.

Note: The DropTooth code has been copied directly into the patch to avoid having to introduce a new patch for this method, which is only called by the method we're already patching

### DIFF
--- a/NitroxPatcher/Patches/Dynamic/Stalker_CheckLoseTooth_Patch.cs
+++ b/NitroxPatcher/Patches/Dynamic/Stalker_CheckLoseTooth_Patch.cs
@@ -1,4 +1,6 @@
-﻿using System.Reflection;
+using System.Reflection;
+using NitroxClient.GameLogic;
+using NitroxModel.DataStructures;
 using NitroxModel.Helper;
 using UnityEngine;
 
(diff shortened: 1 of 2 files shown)


NitroxPatcher/Patches/Dynamic/Stalker_CheckLoseTooth_Patch.cs
```diff
@@ -8,22 +10,55 @@ public sealed partial class Stalker_CheckLoseTooth_Patch : NitroxPatch, IDynamic
...
+
+        if (dropProbability == 0)
         {
-            dropProbability = 0.15f; // 15% probability
+            return false;
         }
 
-        if (UnityEngine.Random.value < dropProbability)
+        // Random.value returns a random float within[0.0..1.0] (range is inclusive)
+        if (Random.value < dropProbability && Random.value < 0.5f)
```
Jannify, line 40: Is the `< 0.5f` even needed as dropPeobability is always between 0 and 0.25f ?

NitroxPatcher/Patches/Dynamic/Stalker_CheckLoseTooth_Patch.cs
```diff
@@ -8,22 +10,55 @@ public sealed partial class Stalker_CheckLoseTooth_Patch : NitroxPatch, IDynamic
...
+
+            if (toothGameObject.activeSelf && __instance.isActiveAndEnabled)
+            {
+                foreach (Collider collider in toothGameObject.GetComponentsInChildren<Collider>())
+                {
+                    Physics.IgnoreCollision(__instance.stalkerBodyCollider, collider);
+                }
+            }
+
+            Utils.PlayFMODAsset(__instance.loseToothSound, toothGameObject.transform, 20f);
```
Jannify, line 56: Is the sound whitelisted for syncing?

NitroxPatcher/Patches/Dynamic/Stalker_CheckLoseTooth_Patch.cs
dartasen: Random.value would be different each time you call it so the code semantic is correct. 

That's a code extract from Subnautica's code, that how they artifically tear down the probabilities (i.e: for a probability of 0.25, it would produce a probability of 0.125), but one could argue that's a bit weird to to that

tornac1234:
LGTM CW && TAS
image

tornac1234 (approved):

NitroxPatcher/Patches/Dynamic/Stalker_CheckLoseTooth_Patch.cs
Jannify: Ah yes, mb.

Jannify (approved):

Coding-Hen (approved):
LGTM, will be nice to have this fixed and synced

PR #2355, opened 2025-07-26
Fix Vehicle sound issues

NinjaPedroX:
This PR fixes a few sound issues with the Cyclops, Seamoth, and PRAWN Suit.

**TODO:**

- [x] Fix global Cyclops engine sound.
- [x] Fix looping Cyclops engine sound when a player other than the client quits piloting the Cyclops while holding down a movement key at the same time.
  - [x] Do the same but for the Seamoth^.
  - [x] Do the same but for the PRAWN suit jet engines^
- [x] Fix unnecessary "Ahead (speed)" message from playing on remote players when doing the steps described in the second TODO above
- [x] Fix weird light sound bug for the Seamoth (toggle lights packet being sent even when just entering/leaving the seamoth).
- [x] Fix Seamoth engine sounds from playing despite being unpowered.
- [ ] ~Fix global Cyclops upgrade module install sound~ (leaving for a later PR).


NinjaPedroX:
This PR is now ready for testing and reviews.

NitroxClient/MonoBehaviours/Vehicles/SeaMothMovementReplicator.cs
```diff
@@ -30,7 +31,7 @@ public void Awake()
     {
         base.Update();
 
-        if (throttleApplied)
+        if (throttleApplied && seaMoth.IsPowered() && drivingPlayer != null)
         {
             seaMoth.engineSound.AccelerateInput(1);
         }
```
dartasen: Does not add any useful information to keep a reference to driving player, throttleApplied is set to false in the same method drivingPlayer become null

NitroxClient/MonoBehaviours/Vehicles/SeaMothMovementReplicator.cs
```diff
@@ -95,6 +96,7 @@ private void SetupSound()
 
     public override void Enter(RemotePlayer remotePlayer)
```
dartasen, line 97: Why remotePlayer isn't use in this method ?

NitroxClient/MonoBehaviours/Vehicles/SeaMothMovementReplicator.cs
NinjaPedroX: You're right, technically they should be the same. But when a remote player exits the seamoth while holding down a movement key, I think ThrottleApplied is changed back to true (within ApplyNewMovementData) which causes the infinite engine sound. Putting this here stops that issue.

Do you think it would be better to move this check into ApplyNewMovementData?

NinjaPedroX: Just double checked this now and I was correct. For both the Seamoth and Cyclops, throttleApplied is changed back to true after the player exits piloting while holding down a movement key.

image
image

I'll change the logic for this so that throttleApplied changes correctly instead of working around this bug.

NitroxClient/MonoBehaviours/Vehicles/SeaMothMovementReplicator.cs
Coding-Hen: If this is the override method would it not be required? Else it wouldn't match and be a valid override?

I agree though if we aren't using it then it doesn't seem worth having as a parameter though if we can drop it

NitroxClient/MonoBehaviours/Vehicles/CyclopsMovementReplicator.cs
```diff
@@ -75,7 +81,23 @@ public override void ApplyNewMovementData(MovementData newMovementData)
             }
         }
 
-        throttleApplied = vehicleMovementData.ThrottleApplied;
+        // Adjusting volume for the engine Sound
+        float distanceToPlayer = Vector3.Distance(Player.main.transform.position, transform.position);
+        float volumeRpmSound = SoundHelper.CalculateVolume(distanceToPlayer, radiusRpmSound, 1f);
+        rpmSound.GetEventInstance().setVolume(volumeRpmSound);
```
Jannify, line 87: Would cache `GetEventInstance()`if possible since this method is called quite often, right?

NitroxClient/MonoBehaviours/Vehicles/SeaMothMovementReplicator.cs
NinjaPedroX: > If this is the override method would it not be required? Else it wouldn't match and be a valid override?

That is correct
image

We could remove it if you don't think it's needed. However, the way my fixes work require knowing if there's a remote player in the vehicle via this parameter. Removing it would require us to find some other way of solving the infinite vehicle sound issues, plus `CyclopsMovementReplicator.cs` needs it for animation purposes.

```c#
if (drivingPlayer != null)
{
    drivingPlayer.AnimationController.SetFloat(CYCLOPS_YAW, subControl.steeringWheelYaw);
    drivingPlayer.AnimationController.SetFloat(CYCLOPS_PITCH, subControl.steeringWheelPitch);
}
```

NitroxClient/MonoBehaviours/Vehicles/SeaMothMovementReplicator.cs
NinjaPedroX: throttleApplied plus other related throttle variables now change correctly but my fixes still require the use `drivingPlayer`. Let me know if you prefer different way of fixing the issue, though I think that might require a bit more overhauling of the current movement replicator system.

NitroxClient/MonoBehaviours/Vehicles/ExosuitMovementReplicator.cs
```diff
@@ -119,6 +121,7 @@ private void SetupSound()
 
     public override void Enter(RemotePlayer remotePlayer)
     {
+        drivingPlayer = remotePlayer;
```
Jannify, line 124: Is this maybe making a problem if a remote player is already in a vehicle and **then** the player is joining?

NitroxClient/GameLogic/Spawning/Metadata/Processor/SeamothMetadataProcessor.cs
```diff
@@ -42,9 +42,17 @@ public override void ProcessMetadata(GameObject gameObject, SeamothMetadata meta
...
     {
+        ToggleLights toggleLights = seamoth.toggleLights;
+
+        if (lightsOn == toggleLights.GetLightsActive())
+        {
+            // Lights are already in the desired state, nothing to do.
+            return;
+        }
+
+        // In the future, maybe do a patch for all ToggleLights.SetLightsActive() to handle all cases of lights toggling (i.e. powercell being inserted, etc.)
```
tornac1234, line 53: in which case does this case happen ?

NitroxClient/MonoBehaviours/Vehicles/CyclopsMovementReplicator.cs
```diff
@@ -75,7 +81,24 @@ public override void ApplyNewMovementData(MovementData newMovementData)
...
+
+        throttleApplied = vehicleMovementData.ThrottleApplied && drivingPlayer != null;
+        //Log.Debug($"(Cyclops) Throttle applied: {throttleApplied}");
```
tornac1234: note to remove this debug here
and another one below

NitroxClient/MonoBehaviours/Vehicles/ExosuitMovementReplicator.cs
```diff
@@ -17,6 +17,8 @@ public class ExosuitMovementReplicator : VehicleMovementReplicator
     private bool jetsActive;
     private float timeJetsActiveChanged;
 
+    private RemotePlayer? drivingPlayer;
```
tornac1234, line 20:
```suggestion
    private RemotePlayer drivingPlayer;
```
I think you can remove the ? and it'll be fine too. The name of the variable should be enough to understand that it can be null (a vehicle is not always driven)

NitroxClient/MonoBehaviours/Vehicles/SeaMothMovementReplicator.cs
tornac1234: I agree this is required for ApplyNewMovementData to know if there's a driving player

NitroxClient/GameLogic/Spawning/Metadata/Processor/CyclopsMetadataProcessor.cs
```diff
@@ -65,6 +65,12 @@ private void SetEngineState(GameObject cyclops, bool isOn)
 
     private void SetEngineMode(GameObject cyclops, CyclopsMotorMode.CyclopsMotorModes mode)
     {
+        CyclopsMotorMode.CyclopsMotorModes oldMode = cyclops.GetComponentInChildren<SubControl>(true).cyclopsMotorMode.cyclopsMotorMode;
```
tornac1234: I think SubControl is directly on the cyclops so you should be able to use GetComponent instead, and maybe you can cache SubControl when initializing the class instead

tornac1234 (changes requested):
Sounds good code wise otherwise, haven't tested yet

NitroxClient/GameLogic/Spawning/Metadata/Processor/SeamothMetadataProcessor.cs
NinjaPedroX: The example I gave in that review comment was referring to #2358, but there could be more cases of this happening. I plan to try looking for a more universal fix to these global vehicle sounds in another PR, which would include this and other known global sounds like inserting/removing a cyclops upgrade.

NitroxClient/MonoBehaviours/Vehicles/CyclopsMovementReplicator.cs
NinjaPedroX: Oops 😶

NitroxClient/MonoBehaviours/Vehicles/ExosuitMovementReplicator.cs
tornac1234: right good catch, maybe there should be something calling enter when setpilotchair is called on remoteplayer ? or is it alteady belng called from there in a way ?

NinjaPedroX: That is a good point. Looks like `GlobalRootInitialSyncProcessor` handles that via the `RestoreDrivers()` function.

After doing some in-game testing however, it seems that it's only working for the Seamoth and Exosuit, and not the Cyclops (resulting in sound not playing for non-piloting players that just joined). I'll get this fixed and resolve this conversation once that's done.

NitroxClient/MonoBehaviours/Vehicles/CyclopsMovementReplicator.cs
NinjaPedroX: It is indeed called quite often. However, it doesn't look like caching `rpmSound.GetEventInstance()` as well as the Seamoth sounds into their own EventInstance variables works (unless I'm doing something wrong). I tried doing this now and those sounds became global again.

See the image below for the changes I made in my attempt to do this, and please do let me know if I'm not understanding this correctly.
image

NitroxClient/MonoBehaviours/Vehicles/ExosuitMovementReplicator.cs
dartasen: We started annotating codebase for NRT (C# 8) with @Measurity, so keeping the ? is correct. Expressing the intent that a variable is set to null at a given time is something valuable when we'll fully activate the feature

NitroxClient/GameLogic/Spawning/Metadata/Processor/SeamothMetadataProcessor.cs
tornac1234: Ok, this condition doesn't seem like it breaks anything so in any way I agree that we keep it. Even better if it temporarily fixes another issue. But do you feel like fixing that other issue in this PR since it seems to be in the scope ?

NinjaPedroX: True, I suppose it would be in the scope of this PR. But I'd have to learn how to do a transpiler first (to edit the `ToggleLights.SetLightsActive()` method) which might take more time for me to do and delay the 1.8 update further.

Additionally, those sounds that aren't fixed yet don't play very often so I think they are less of an issue to worry about for this update. I'll look into the transpiler once everything else in this PR is waiting for reviews in case it's easier to do than I think it is.

NitroxClient/MonoBehaviours/Vehicles/ExosuitMovementReplicator.cs
NinjaPedroX: Reverted this change back^

Coding-Hen (approved):
LGTM

dartasen:
Tested IG with Ninja

PR #2366, opened 2025-07-29
Add warning when constructing scanner rooms

### LINKED ISSUE #2365: Inform players about ScannerRoom being unsynced
### Describe the issue

While we don't have any sync on the scanner room, we should inform the brave players that are using that base piece about the horrible status of the sync inside Nitrox. (Modules not persisted, fake pings, pings about already picked up items, ...)

* Display an ingame log upon building a scanner room

SpaceMonkeyy86:
Closes #2365

### DIFF
--- a/Nitrox.Assets.Subnautica/LanguageFiles/en.json
+++ b/Nitrox.Assets.Subnautica/LanguageFiles/en.json
@@ -75,6 +75,7 @@
     "Nitrox_ResyncSettings": "Resynchronization",
     "Nitrox_SafeBuilding": "Safe building",
     "Nitrox_SafeBuildingLog": "Safe building log",
+    "Nitrox_ScannerRoomWarn": "Please note: scanner rooms do not work properly in this version of Nitrox. Expect many bugs.",
     "Nitrox_ServerEntry_DeleteWarning": "Are you sure you want to delete this server?",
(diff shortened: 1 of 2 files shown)


Nitrox.Assets.Subnautica/LanguageFiles/en.json
```diff
@@ -75,6 +75,7 @@
     "Nitrox_ResyncSettings": "Resynchronization",
     "Nitrox_SafeBuilding": "Safe building",
     "Nitrox_SafeBuildingLog": "Safe building log",
+    "Nitrox_ScannerRoomWarn": "Please note: scanner rooms do not work properly in this version of Nitrox. Expect many bugs.",
```
dartasen, line 78: Maybe we could create a more generic one like "This command is not available in Nitrox", so we can reuse that one in area we're not okay with the current sync

SpaceMonkeyy86: That would make it unclear what was causing the issue though. Maybe it could be a generic message with a place to insert what we are referring to: "This feature is not yet implemented in Nitrox, expect many bugs: Scanner Room". But that just feels lazy to me (and hard to translate properly). I would prefer individual messages.

dartasen (approved):

PR #2368, opened 2025-07-29
Send EntityDestroyed packet on time capsule destruction to server

### LINKED ISSUE #2357: Disable Time Capsules
### Describe the issue

Time capsules keep respawning at specific locations (go to 435 -103 50) when you swim away and go back to the area. Players can spam time capsules this way and get a ton of items.

For the 1.8 timeframe, I suggest disabling them.

cmedina-dev:
Certain resources, e.g. flares, notify the server via an `EntityDestroyed` packet once the resource is destroyed. Time capsules lacked an `EntityDestroyed` broadcast, resulting in clients seeing respawning capsules from the `BatchEntitySpawner`. 

This fix adds a patch which sends the `EntityDestroyed` packet to the server when a client opens a time capsule.

Fixes #2357.

### DIFF
--- a/NitroxPatcher/Patches/Dynamic/TimeCapsule_Open_Patch.cs
+++ b/NitroxPatcher/Patches/Dynamic/TimeCapsule_Open_Patch.cs
@@ -12,7 +12,7 @@ namespace NitroxPatcher.Patches.Dynamic;
 /// </summary>
 public sealed partial class TimeCapsule_Open_Patch : NitroxPatch, IDynamicPatch
 {
-    public static readonly MethodInfo TARGET_METHOD = Reflect.Method((TimeCapsule t) => t.Open());
+    internal static readonly MethodInfo TARGET_METHOD = Reflect.Method((TimeCapsule t) => t.Open());
     
     public static void Postfix(TimeCapsule __instance)
     {
@@ -21,7 +21,7 @@ public sealed partial class TimeCapsule_Open_Patch : NitroxPatch, IDynamicPatch
             return;
         }
         
-        var packet = new EntityDestroyed(timeCapsuleId);
+        EntityDestroyed packet = new(timeCapsuleId);
         Resolve<IPacketSender>().Send(packet);
     }
 }
--- a/NitroxPatcher/Patches/Dynamic/TimeCapsule_Open_Patch.cs
+++ b/NitroxPatcher/Patches/Dynamic/TimeCapsule_Open_Patch.cs
@@ -1,0 +1,27 @@
+﻿using NitroxModel.Packets;
+using System.Reflection;
+using NitroxClient.Communication.Abstract;
+using NitroxModel.DataStructures;
+using NitroxModel.Helper;
+
+namespace NitroxPatcher.Patches.Dynamic;
+
+/// <summary>
+/// Ensures that when a player opens a time capsule, it is permanently destroyed on the server,
+/// preventing it from respawning.
+/// </summary>
+public sealed partial class TimeCapsule_Open_Patch : NitroxPatch, IDynamicPatch
+{
+    public static readonly MethodInfo TARGET_METHOD = Reflect.Method((TimeCapsule t) => t.Open());
+    
+    public static void Postfix(TimeCapsule __instance)
+    {
+        if (!__instance.TryGetIdOrWarn(out NitroxId timeCapsuleId))
+        {
+            return;
+        }
+        
+        var packet = new EntityDestroyed(timeCapsuleId);
+        Resolve<IPacketSender>().Send(packet);
+    }
+}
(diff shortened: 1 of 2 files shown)


NitroxPatcher/Patches/Dynamic/TimeCapsule_Open_Patch.cs
```diff
@@ -0,0 +1,27 @@
...
+
+namespace NitroxPatcher.Patches.Dynamic;
+
+/// <summary>
+/// Ensures that when a player opens a time capsule, it is permanently destroyed on the server,
+/// preventing it from respawning.
+/// </summary>
+public sealed partial class TimeCapsule_Open_Patch : NitroxPatch, IDynamicPatch
+{
+    public static readonly MethodInfo TARGET_METHOD = Reflect.Method((TimeCapsule t) => t.Open());
```
dartasen:
```suggestion
    internal static readonly MethodInfo TARGET_METHOD = Reflect.Method((TimeCapsule t) => t.Open());
```

NitroxPatcher/Patches/Dynamic/TimeCapsule_Open_Patch.cs
```diff
@@ -0,0 +1,27 @@
...
+    
+    public static void Postfix(TimeCapsule __instance)
+    {
+        if (!__instance.TryGetIdOrWarn(out NitroxId timeCapsuleId))
+        {
+            return;
+        }
+        
+        var packet = new EntityDestroyed(timeCapsuleId);
```
dartasen:
```suggestion
        EntityDestroyed packet = new(timeCapsuleId);
```

Coding-Hen (approved):
LGTM for the code but should have a quick in game test if possible just to make sure it's working as intended. (I would unfortunately I'm away at the moment without access to my computer)

dartasen (commented):
few minor things

dartasen (changes requested):
image

Did not fully work on my side. It does send an EntityDestroyed to the server, but the other clients (those who didn't pick up the capsule) can still see the time capsule.

If I teleport back and forth, the capsule is indeed gone

cmedina-dev:
> image
> 
> Did not fully work on my side. It does send an EntityDestroyed to the server, but the other clients (those who didn't pick up the capsule) can still see the time capsule.
> 
> If I teleport back and forth, the capsule is indeed gone
> 
> 

Thanks for the testing. I'll try to get a fix out tonight or tomorrow at the latest.

NitroxServer/Communication/Packets/Processors/EntityDestroyedPacketProcessor.cs
```diff
@@ -32,14 +32,7 @@ public override void Process(EntityDestroyed packet, Player destroyingPlayer)
                 worldEntityManager.MovePlayerChildrenToRoot(vehicleWorldEntity);
             }
 
-            foreach (Player player in playerManager.GetConnectedPlayers())
-            {
-                bool isOtherPlayer = player != destroyingPlayer;
-                if (isOtherPlayer && player.CanSee(entity))
```
dartasen, line 38: If I understand correctly, does that mean that `CanSee` does not work for Time Capsules ?

cmedina-dev: More specifically, it seems like `CanSee` returns `false`  because the visibility state is stale, preventing the `EntityDestroyed` packet from being sent to other players.

This new fix sidesteps the `CanSee` check because permanent world entity destruction should be broadcast regardless of position of other players. After testing, this does seem to fix the previous issue.

dartasen: I'm a bit scared of changing entity generic logic code for a bugfix right before a release. I'd rather keep the code as it is now and understand why time capsule is filtered out, because the player has the WorldEntity given by the server and is in range

cmedina-dev: The only difference in `SendPacketToOtherPlayers` and the code in `EntityDestroyedPacketProcessor` is the `CanSee` check.

```cs
public void SendPacketToOtherPlayers(Packet packet, Player sendingPlayer)
{
    foreach (Player player in ConnectedPlayers())
    {
        if (player != sendingPlayer)
        {
            player.SendPacket(packet);
        }
    }
}
```

```cs
foreach (Player player in playerManager.GetConnectedPlayers())
{
    bool isOtherPlayer = player != destroyingPlayer;
    if (isOtherPlayer && player.CanSee(entity))
    {
        player.SendPacket(packet);
    }
}
```

The `CanSee` check for `EntityDestroyed` was most likely an attempt at optimization by not forwarding unnecessary packets. But `EntityDestroyed` should specifically not have this check because a player will _eventually_ travel to that location. Broadcasting this to every player, regardless of position, maintains proper world state sync, at effectively no cost as `EntityDestroyed` is rarely ever fired to begin with.

In short, Time Capsules don't appear to be the only entities being excluded. It seems every entity processed by `EntityDestroyedPacketProcessor` has this same sync issue.

PR #2370, opened 2025-07-30
Add null reference check when getting all player objects

### LINKED ISSUE #2360: NRE in Cyclops Sonar
### Describe the issue

CyclopsSonarCreatureDetector.ChekItemsOnHashSet spamming nullref from (CyclopsSonarCreatureDetector_CheckForCreaturesInRange_Patch.Postfix)

cmedina-dev:
I was not able to reproduce the issue in question since this particular error seems to arise from a narrow race condition. However, it seems fairly straightforward to see where the issue occurs.

The `ChekItemsOnHashSet()` signature is defined as:

```cs
  private void ChekItemsOnHashSet(
    HashSet<GameObject> checkList,
    CyclopsSonarDisplay.EntityType entityType)
```

Given that `entityType` is always defined in `CyclopsSonarCreatureDetector_CheckForCreaturesInRange_Patch.Postfix`, it's clear that `checkList` is receiving a null value in its HashSet.

The HashSet passed by the `Postfix` is `Resolve<PlayerManager>().GetAllPlayerObjects()`:

```cs
__instance.ChekItemsOnHashSet(Resolve<PlayerManager>().GetAllPlayerObjects(), PLAYER_TYPE);
```

Previously, `GetAllPlayerObjects()` retrieved every player object under the assumption that every `player.Body` would not be null. However, under precise circumstances, it seems there is a small window of time where the `Body` GameObject is destroyed without being removed from `playersById`, leading to a race condition where `GetAll()` retrieves a null value:

```cs
    public void RemovePlayer(ushort playerId)
    {
        if (playersById.TryGetValue(playerId, out RemotePlayer player))
        {
            player.Destroy(); // This is where the race condition occurs.
            playersById.Remove(playerId); // There is a time gap between the above Destroy() and this Remove().
            OnRemove(playerId, player);
            DiscordClient.UpdatePartySize(GetTotalPlayerCount());
        }
    }
```

During this race condition, `GetAll()` will pull a GameObject where the `Body` has already been destroyed, but not removed from the Dictionary of `playersById`:

```cs
    public IEnumerable<RemotePlayer> GetAll()
    {
        return playersById.Values;
    }
```

This fix adds a check to filter out `GameObjects` that have already had their `Body` destroyed but have not yet been removed from the `playersById` dictionary.

Fixes #2360.

### DIFF
--- a/NitroxClient/GameLogic/PlayerManager.cs
+++ b/NitroxClient/GameLogic/PlayerManager.cs
@@ -56,7 +56,10 @@ public class PlayerManager
                                                   .Select(player => player.Body)
                                                   .Where(body => body != null)
                                                   .ToSet();
-        remotePlayerObjects.Add(Player.mainObject);
+        if (Player.mainObject != null)
+        {
+            remotePlayerObjects.Add(Player.mainObject);    
+        }
         return remotePlayerObjects;
     }
 
--- a/NitroxClient/GameLogic/PlayerManager.cs
+++ b/NitroxClient/GameLogic/PlayerManager.cs
@@ -52,7 +52,10 @@ public class PlayerManager
 
     public HashSet<GameObject> GetAllPlayerObjects()
     {
-        HashSet<GameObject> remotePlayerObjects = GetAll().Select(player => player.Body).ToSet();
+        HashSet<GameObject> remotePlayerObjects = GetAll()
+                                                  .Select(player => player.Body)
+                                                  .Where(body => body != null)
+                                                  .ToSet();
         remotePlayerObjects.Add(Player.mainObject);
         return remotePlayerObjects;
     }


Measurity (approved):

NitroxClient/GameLogic/PlayerManager.cs
```diff
@@ -52,7 +52,10 @@ public IEnumerable<RemotePlayer> GetAll()
 
     public HashSet<GameObject> GetAllPlayerObjects()
     {
-        HashSet<GameObject> remotePlayerObjects = GetAll().Select(player => player.Body).ToSet();
+        HashSet<GameObject> remotePlayerObjects = GetAll()
+                                                  .Select(player => player.Body)
+                                                  .Where(body => body != null)
```
SpaceMonkeyy86, line 57: `body` is a GameObject, so the check should be `body => body` instead of `body => body != null`. `body != null` will be true even if the GameObject has been destroyed, but `body` will evaluate to false. Unity has a "fake null" for destroyed objects that throws a `NullReferenceException` when you try to access fields or use it in certain Unity-provided methods, and this kind of exception is the error we see in the logs.

NitroxClient/GameLogic/PlayerManager.cs
```diff
@@ -52,7 +52,10 @@ public IEnumerable<RemotePlayer> GetAll()
...
+                                                  .ToSet();
         remotePlayerObjects.Add(Player.mainObject);
```
SpaceMonkeyy86: Do you think it's possible for `Player.mainObject` to be destroyed/null? It may be worth moving the `Where()` call to below this line.

NitroxClient/GameLogic/PlayerManager.cs
Measurity: This is fine as Unity also overloads the != operator.
image

cmedina-dev:
> Looks good for the most part. Note that the log where we found this issue [game[spacemonkeyy]-20250706.log](https://github.com/user-attachments/files/21513397/game.spacemonkeyy.-20250706.log) has this error spammed seemingly every frame when the sonar was active, so I don't think the race condition you suggested could be the root cause. But the fix should still work. There is also another error in there about `CyclopsPawn.MaintainPosition()` but we can fix that in a different PR.

Unfortunately I can't see the log. However, the race condition will cause it to spam every half-second so long as the playerId is not removed while the Body is destroyed. The sonar uses `InvokeRepeating()` here:

```cs
  private void OnEnable()
  {
    if (this.IsInvoking("CheckForCreaturesInRange"))
      return;
    this.InvokeRepeating("CheckForCreaturesInRange", 0.5f, 0.5f);
  }
```

So this is likely the root cause.

NitroxClient/GameLogic/PlayerManager.cs
cmedina-dev: The race condition occurs only in the context of a remote player disconnecting with the sonar active, meaning `Player.mainObject` (the local player) is guaranteed to never be null in this scenario.

`Player.mainObject` should only ever be null when the local player quits the game. There is a potential but incredibly unlikely scenario where during teardown the `Player` object is destroyed, causing `mainObject` to become null, and that `CheckForCreaturesInRange()` fires one final time before the Cyclops is destroyed, which may result in a single error line occurring. However, the scene teardown should prevent this from occurring.

I'll add in an additional small if-statement to check the Player.mainObject though, just in case.

dartasen:
@SpaceMonkeyy86 Could you please create another issue about the MaintainPaw ?

dartasen:
Thank you for your contribution and all the details you provided !

PR #2374, opened 2025-07-30
Decouple pawn cleanup logic from RemotePlayer

### LINKED ISSUE #2372: NRE in CyclopsPawn
### Describe the issue

```
[10:34:39.988] [ERR-UNITY] NullReferenceException
NitroxClient.GameLogic.CyclopsPawn.MaintainPosition () (at <450532a80ed84539a5635e106379d255>:0)
NitroxClient.MonoBehaviours.Cyclops.NitroxCyclops.MaintainPawns () (at <450532a80ed84539a5635e106379d255>:0)
NitroxClient.MonoBehaviours.Cyclops.NitroxCyclops.Update () (at <450532a80ed84539a5635e106379d255>:0)
```

[game\[spacemonkeyy\]-20250706.log](https://github.com/user-attachments/files/21514115/game.spacemonkeyy.-20250706.log)

cmedina-dev:
As an initial note, I was unable to replicate this after about an hour of testing, so this is mostly speculation.

Currently, `InitializeGameObject` also adds a handler to `Body` for pawn cleanups with regards to the Cyclops:

```cs
PlayerDisconnectEvent.AddHandler(Body, _ =>
{
    Pawn?.Unregister();
    Pawn = null;
});
```

In normal circumstances, this would `Unregister()` the `GameObject` through `CyclopsPawn.cs`. However, there seems to be a race condition that can trigger a `NullReferenceException`.

In the event the `Body` is destroyed before the `PlayerDisconnectEvent` triggers, then the handler will never fire, and `Pawn?.Unregister()` is never called.

What this means is there is now a stale `CyclopsPawn` instance because it's never removed.

The `NullReferenceException` stems from `MaintainPosition()` in `CyclopsPawn.cs`, specifically the reference to `RealObject`. Notably, `RealObject` is set to `remotePlayer.Body`.

```cs
else if (player is RemotePlayer remotePlayer)
{
    RealObject = remotePlayer.Body;
    MaintainPredicate = () => !remotePlayer.PilotingChair;
}
```

And where RealObject, the null `Body`, is referenced:

```cs
public void MaintainPosition()
{
    RealObject.transform.localPosition = Handle.transform.localPosition; // This is where the NRE occurs.
    RealObject.transform.rotation = realCyclopsTransform.rotation;
    if (!isLocalPlayer)
    {
        RealObject.transform.localRotation = Handle.transform.localRotation;
    }
}
```

The fix decouples the handler logic from the `Body` and migrates it to the `Cyclops` instance. It also adds an additional defensive null check during `MaintainPosition()` to unregister stale pawns.

(Probably) Fixes #2372.

### DIFF
--- a/NitroxClient/GameLogic/CyclopsPawn.cs
+++ b/NitroxClient/GameLogic/CyclopsPawn.cs
@@ -104,6 +104,13 @@ public class CyclopsPawn
 
     public void MaintainPosition()
     {
+        if (RealObject == null)
+        {
+            Log.Warn($"CyclopsPawn for player '{player.PlayerName}' has a null RealObject. Unregistering stale pawn.");
+            Unregister();
+            return;
+        }
+        
         RealObject.transform.localPosition = Handle.transform.localPosition;
         RealObject.transform.rotation = realCyclopsTransform.rotation;
         if (!isLocalPlayer)
--- a/NitroxClient/GameLogic/RemotePlayer.cs
+++ b/NitroxClient/GameLogic/RemotePlayer.cs
@@ -107,12 +107,6 @@ public class RemotePlayer : INitroxPlayer
         vitals = playerVitalsManager.CreateOrFindForPlayer(this);
         RefreshVitalsVisibility();
 
-        PlayerDisconnectEvent.AddHandler(Body, _ =>
-        {
-            Pawn?.Unregister();
-            Pawn = null;
-        });
-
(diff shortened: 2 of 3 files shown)


dartasen:
Will keep this one on hold for now, problem seems to be bigger.
Looking at all players logs, there's a case somewhere that Nitrox doesn't deal with where RemotePlayer is in bad state for some reasons.

Likely related to cyclops destruction or bases.

PR #2375, opened 2025-07-31
Fix stuck bug for the Seamoth and Exosuit

### LINKED ISSUE #2373: Player can get stuck in Seamoth/Exosuit when a remote player quits game while in the vehicle
### Describe the issue

Steps to reproduce:

1. Join the game with two players.
2. Create a Seamoth and a moonpool (or Cyclops).
3. Enter Seamoth with Player 2, notice the enter animation play.
4. Leave server with Player 2 while still in Seamoth.
5. Enter Seamoth with Player 1, notice the enter animation doesn't play.
6. Press `Q` (or whatever your exit key is), notice that you can't leave Seamoth.
7. Dock in the moonpool (or Cyclops) to exit the Seamoth.

UPDATE: The same thing happens with the exosuit. You also can't use the arms while stuck inside

NinjaPedroX:
Fixes #2373

### DIFF
--- a/NitroxClient/GameLogic/RemotePlayer.cs
+++ b/NitroxClient/GameLogic/RemotePlayer.cs
@@ -357,7 +357,7 @@ public class RemotePlayer : INitroxPlayer
     {
         Log.Info($"{PlayerName} left the game");
         Log.InGame(Language.main.Get("Nitrox_PlayerLeft").Replace("{PLAYER}", PlayerName));
-        SetVehicle(null);
+        ResetStates();
         NitroxEntity.RemoveFrom(Body);
         Object.DestroyImmediate(Body);
     }


Measurity (approved):
Works for me. GJ!

NitroxClient/GameLogic/RemotePlayer.cs
```diff
@@ -357,6 +357,7 @@ public void Destroy()
     {
         Log.Info($"{PlayerName} left the game");
         Log.InGame(Language.main.Get("Nitrox_PlayerLeft").Replace("{PLAYER}", PlayerName));
+        SetVehicle(null);
```
dartasen: Maybe ResetStates() will do all the necessary cleanup for you ?

NinjaPedroX: I considered doing that at first, but ResetStates() does more than just reset the vehicle which I feel is unnecessary to do when the remote player is about to be destroyed anyway.

dartasen: Then It's unclear to me why SetVehicle have to be set to null, while the others can remain the same without causing any trouble (i.e: Piloting chair)

NinjaPedroX: Just double checked in-game now, but the stuck bug doesn't happen for the local player when a remote player leaves the game while piloting the Cyclops. It only happens with the Seamoth and Exosuit, so I don't think resetting the other states and animation controller via `ResetStates()` is necessary.

If you still think it's better to run that function though, I'm good with it.

Coding-Hen: I think I agree with Dartasen it is a bit unclear but seems sensible to not call the whole reset method if not needed. Maybe a small comment just to make it clear why we are setting to null would solve this for you Dartasen?

Coding-Hen (approved):
LGTM

NitroxClient/GameLogic/RemotePlayer.cs
tornac1234: I agree too with @dartasen, ResetStates has more global actions which are relevant to be executed, whether or not the body is destroyed later on. We better clear every states in case we put more there in the future, so that we don't let a single drop of remote player references in the world once the player has left

NinjaPedroX: The change has been made and tested, as requested

dartasen (approved):

PR #2378, opened 2025-08-01
Implement EntityDestroyed broadcast for StoryHandTarget

### LINKED ISSUE #2364: Add sync for StoryHandTarget
### Describe the issue

StoryHandTarget.OnHandClick should be synced : 
* PDA : Send an EntityDestroyed so it's picked up for everyone & the server won't resend those entities upon reloading a batch
* Precursor & GenericConsole : Update Entity Metadata to store the "Used" boolean

Image

cmedina-dev:
Implements entity destruction synchronization for StoryHandTarget objects like PDAs. These objects should now immediately disappear when picked up by any player.

Fixes #2364

### DIFF
--- a/NitroxPatcher/Patches/Dynamic/StoryHandTarget_OnHandClick_Patch.cs
+++ b/NitroxPatcher/Patches/Dynamic/StoryHandTarget_OnHandClick_Patch.cs
@@ -1,0 +1,58 @@
+﻿using System.Reflection;
+using NitroxClient.Communication.Abstract;
+using NitroxClient.GameLogic;
+using NitroxClient.GameLogic.Simulation;
+using NitroxClient.MonoBehaviours.Gui.HUD;
+using NitroxModel.Core;
+using NitroxModel.DataStructures;
+using NitroxModel.Helper;
+using NitroxModel.Packets;
+
+namespace NitroxPatcher.Patches.Dynamic
+{
+    /// <summary>
+    /// Synchronizes interactions with StoryHandTarget objects such as PDAs.
+    /// Ensures only one player can interact with an object at a time by requesting a simulation lock.
+    /// On a successful interaction, broadcasts the object's destruction to all clients.
+    /// </summary>
+    public sealed partial class StoryHandTarget_OnHandClick_Patch : NitroxPatch, IDynamicPatch
+    {
+        private static readonly MethodInfo TARGET_METHOD = Reflect.Method((StoryHandTarget t) => t.OnHandClick(default(GUIHand)));
+
+        // If true, then a lock is acquired and the original method is invoked.
+        private static bool skipPrefix;
+
+        public static bool Prefix(StoryHandTarget __instance, GUIHand hand)
+        {
+            if (!__instance.TryGetIdOrWarn(out NitroxId id) || Resolve<SimulationOwnership>().HasExclusiveLock(id) || skipPrefix)
+            {
+                return true;
+            }
+
+            HandInteraction<StoryHandTarget> context = new(__instance, hand);
+            LockRequest<HandInteraction<StoryHandTarget>> lockRequest = new(id, SimulationLockType.EXCLUSIVE, ReceivedSimulationLockResponse, context);
+            Resolve<SimulationOwnership>().RequestSimulationLock(lockRequest);
+
+            return false;
+        }
+
+        private static void ReceivedSimulationLockResponse(NitroxId id, bool lockAcquired, HandInteraction<StoryHandTarget> context)
+        {
+            StoryHandTarget storyHandTarget = context.Target;
+
+            if (lockAcquired)
+            {
+                skipPrefix = true;
+                storyHandTarget.OnHandClick(context.GuiHand);
+                skipPrefix = false;
+                
+                NitroxServiceLocator.LocateService<IPacketSender>().Send(new EntityDestroyed(id));
+            }
+            else
+            {
+                storyHandTarget.gameObject.AddComponent<DenyOwnershipHand>();
+                storyHandTarget.isValidHandTarget = false;
+            }
+        }
+    }
+}


NitroxPatcher/Patches/Dynamic/StoryHandTarget_OnHandClick_Patch.cs
```diff
@@ -0,0 +1,58 @@
...
+using NitroxClient.Communication.Abstract;
+using NitroxClient.GameLogic;
+using NitroxClient.GameLogic.Simulation;
+using NitroxClient.MonoBehaviours.Gui.HUD;
+using NitroxModel.Core;
+using NitroxModel.DataStructures;
+using NitroxModel.Helper;
+using NitroxModel.Packets;
+
+namespace NitroxPatcher.Patches.Dynamic
```
dartasen, line 11: you can use file scoped namespace, that allows to remove the extra brackets


NitroxPatcher/Patches/Dynamic/StoryHandTarget_OnHandClick_Patch.cs
```diff
@@ -0,0 +1,58 @@
...
+{
+    /// <summary>
+    /// Synchronizes interactions with StoryHandTarget objects such as PDAs.
+    /// Ensures only one player can interact with an object at a time by requesting a simulation lock.
+    /// On a successful interaction, broadcasts the object's destruction to all clients.
+    /// </summary>
+    public sealed partial class StoryHandTarget_OnHandClick_Patch : NitroxPatch, IDynamicPatch
+    {
+        private static readonly MethodInfo TARGET_METHOD = Reflect.Method((StoryHandTarget t) => t.OnHandClick(default(GUIHand)));
```
dartasen, line 20:
```suggestion
        internal static readonly MethodInfo TARGET_METHOD = Reflect.Method((StoryHandTarget t) => t.OnHandClick(default(GUIHand)));
```

NitroxPatcher/Patches/Dynamic/StoryHandTarget_OnHandClick_Patch.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        public static bool Prefix(StoryHandTarget __instance, GUIHand hand)
+        {
+            if (!__instance.TryGetIdOrWarn(out NitroxId id) || Resolve<SimulationOwnership>().HasExclusiveLock(id) || skipPrefix)
+            {
+                return true;
+            }
+
+            HandInteraction<StoryHandTarget> context = new(__instance, hand);
+            LockRequest<HandInteraction<StoryHandTarget>> lockRequest = new(id, SimulationLockType.EXCLUSIVE, ReceivedSimulationLockResponse, context);
```
dartasen, line 33: Does it need the whole simulationrequest process ?

dartasen (changes requested):
Image

Seems like GenericConsole & Terminal are somehow also connected to StoryHandTarget, how does that behave ?

NitroxPatcher/Patches/Dynamic/StoryHandTarget_OnHandClick_Patch.cs
cmedina-dev: There was an issue where neither player could interact with the object without it because it appeared reserved by both players. This fixes that particular issue.

tornac1234 (changes requested):
This PR is actually game breaking, as it doesn't take into account a lot of recurring objects which don't actually disappear, like precursor terminals. Although this is almost valid behaviour towards story PDAs (the little tablets on the ground).

The issue is:
- some objects get deleted entirely
- some objects only delete part of them and propagate some animation
- we want to sync that animation over clients watching this
- we want to reproduce that state when reconnecting to the server

Let's think about the right solution then:
- there's no concurrency issue, the story hand target is not a shared resource which could potentially be used by multiple people at the same time (like a vehicle for example)
- the state must be persisted for certain entities -> we have entity metadata system for it, please give it a try
- there's a way to know if the entire item is deleted: check if (destroyGameObject == current gameObject)

Thus your solution should be built like this:
- StoryHandTargetMetadata with no field in it (because having it on an entity would be enough to know that it was used already), with a processor which does the OnStoryHandTarget sendmessage part from OnHandClick
- StoryHandTarget.OnHandClick prefix patch:
  - if __instance.gameObject == __instance.destroyGameObject, then send entitydestroyed packet
  - else, broadcast metadata update with StoryHandTargetMetadata onto the object
- add a special case if destroyed object has StoryHandTarget in entitydestroyedprocessor (on client-side) and reproduce the OnStoryHandTarget sendmessage part from OnHandClick


Also it feels like you're using AI to redact your comments and replies in this repo which I don't like. Could you please write your own answers yourself so we can actually use logic to find the best solutions ? Or maybe am I just grumpy and this is a regular way to talk...

cmedina-dev:
> This PR is actually game breaking, as it doesn't take into account a lot of recurring objects which don't actually disappear, like precursor terminals. Although this is almost valid behaviour towards story PDAs (the little tablets on the ground).
> 
> The issue is:
> 
> * some objects get deleted entirely
> * some objects only delete part of them and propagate some animation
> * we want to sync that animation over clients watching this
> * we want to reproduce that state when reconnecting to the server
> 
> Let's think about the right solution then:
> 
> * there's no concurrency issue, the story hand target is not a shared resource which could potentially be used by multiple people at the same time (like a vehicle for example)
> * the state must be persisted for certain entities -> we have entity metadata system for it, please give it a try
> * there's a way to know if the entire item is deleted: check if (destroyGameObject == current gameObject)
> 
> Thus your solution should be built like this:
> 
> * StoryHandTargetMetadata with no field in it (because having it on an entity would be enough to know that it was used already), with a processor which does the OnStoryHandTarget sendmessage part from OnHandClick
> * StoryHandTarget.OnHandClick prefix patch:
>   
>   * if __instance.gameObject == __instance.destroyGameObject, then send entitydestroyed packet
>   * else, broadcast metadata update with StoryHandTargetMetadata onto the object
> * add a special case if destroyed object has StoryHandTarget in entitydestroyedprocessor (on client-side) and reproduce the OnStoryHandTarget sendmessage part from OnHandClick
> 
> Also it feels like you're using AI to redact your comments and replies in this repo which I don't like. Could you please write your own answers yourself so we can actually use logic to find the best solutions ? Or maybe am I just grumpy and this is a regular way to talk...

I don't use AI to talk, and any edits are my own. I've redacted a single comment because I accidentally posted it as a review on my own code since I had the wrong tab open. But this reply to me feels oddly aggressive so I'm not entirely comfortable continuing conversation.

tornac1234:
I apologize for being aggressive to you in my previous comment.

PR #2382, opened 2025-08-07
Destroy time capsules server-side when opened

### LINKED ISSUE #2357: Disable Time Capsules

SpaceMonkeyy86:
Adds a patch to send an `EntityDestroyed` packet when the player opens a time capsule. This removes the entity from the server's registry, so it won't reappear when that player or another player reenters the area.

The change to `EntityDestroyedPacketProcessor` in #2368 is not necessary, since time capsules are spawned by the server and not by the client. The client does not need to receive `EntityDestroyed` packets for entities it can't see, since the server will omit that entity from `SpawnEntities` if they load that cell, meaning the entity will not respawn. (Confirmed with testing)

Fixes #2357

### DIFF
--- a/NitroxPatcher/Patches/Dynamic/TimeCapsule_Open_Patch.cs
+++ b/NitroxPatcher/Patches/Dynamic/TimeCapsule_Open_Patch.cs
@@ -1,0 +1,20 @@
+using System.Reflection;
+using NitroxClient.Communication.Abstract;
+using NitroxModel.DataStructures;
+using NitroxModel.Helper;
+using NitroxModel.Packets;
+
+namespace NitroxPatcher.Patches.Dynamic;
+
+public sealed partial class TimeCapsule_Open_Patch : NitroxPatch, IDynamicPatch
+{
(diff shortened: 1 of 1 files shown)


SpaceMonkeyy86:
@Coding-Hen I chose to make it a prefix so multiple players couldn't open the time capsule at the same time. The capsule becomes invisible immediately when clicked so everything looks good for other players. I don't think story / PDA progression uses ids of entities (it just uses string keys for what was just done) so destroying it immediately shouldn't cause any problems.

Coding-Hen:
> @Coding-Hen I chose to make it a prefix so multiple players couldn't open the time capsule at the same time. The capsule becomes invisible immediately when clicked so everything looks good for other players. I don't think story / PDA progression uses ids of entities (it just uses string keys for what was just done) so destroying it immediately shouldn't cause any problems.

Nice, I thought that was the case but just wanted to make sure we hadn't overlooked it

PR #2385, opened 2025-08-09
Fix modules disappearing when swapping between slots

### LINKED ISSUE #2330: Switching the slot of a battery or power cell in a charger causes it to disappear

Windows

Epic

### Describe the issue.

If a battery/cell is already inside of a charger and you drag it into a different slot, it will disappear a majority of the time. This only seems to happen when at least 2 people are connected to the server.

The patched method Equipment.RemoveItem runs an unusual amount of times when this happens (3, when 1 would be expected), which broadcasts a packet each time.

I am confident that this issue is unrelated to #2318.

Make sure there are at least two people connected. I used two instances of Subnautica running on one computer.
Put a battery inside of a battery charger. 
Drag it from its current slot into any other slot. 

### Relevant log output

No relevant log data

SpaceMonkeyy86:
Fixes #2330

### DIFF
--- a/NitroxClient/GameLogic/Spawning/InventoryItemEntitySpawner.cs
+++ b/NitroxClient/GameLogic/Spawning/InventoryItemEntitySpawner.cs
@@ -25,7 +25,7 @@ public class InventoryItemEntitySpawner(EntityMetadataManager entityMetadataMana
     {        
         if (!CanSpawn(entity, out GameObject parentObject, out ItemsContainer container, out string errorLog))
         {
-            Log.Info(errorLog);
+            Log.Error(errorLog);
             result.Set(Optional.Empty);
             yield break;
         }
@@ -47,7 +47,7 @@ public class InventoryItemEntitySpawner(EntityMetadataManager entityMetadataMana
         }
         if (!CanSpawn(entity, out GameObject parentObject, out ItemsContainer container, out string errorLog))
         {
-            Log.Info(errorLog);
+            Log.Error(errorLog);
             return true;
         }
 


NitroxClient/GameLogic/Entities.cs
```diff
@@ -277,6 +269,14 @@ public static void DestroyObject(GameObject gameObject)
                 return;
             }
 
+            if (gameObject.TryGetComponent(out Pickupable pickupable))
```
tornac1234, line 272: Would add a comment to explain that we trigger this early so we can suppress it and that this code will not run twice if it's already null

NitroxClient/GameLogic/Spawning/InventoryItemEntitySpawner.cs
```diff
@@ -47,7 +47,7 @@ protected override bool SpawnSync(InventoryItemEntity entity, TaskResult<Optiona
         }
         if (!CanSpawn(entity, out GameObject parentObject, out ItemsContainer container, out string errorLog))
         {
-            Log.Error(errorLog);
+            Log.Info(errorLog);
```
tornac1234: Why would you change that

NitroxServer/Communication/Packets/Processors/ModuleRemovedProcessor.cs
```diff
@@ -31,6 +31,9 @@ public override void Process(ModuleRemoved packet, Player player)
 
             if (entity.Value is InstalledModuleEntity installedModule)
             {
+                // packet.NewParentId is actually the old container, so when the old equipment grid wasn't the player equipment,
```
Jannify, line 34:
```suggestion
                // TODO: packet.NewParentId is actually the old container, so when the old equipment grid wasn't the player equipment,
```

NitroxClient/GameLogic/Spawning/InventoryItemEntitySpawner.cs
SpaceMonkeyy86: The corresponding line in SpawnAsync uses Log.Info. Also, this line prints "Could not find container field on GameObject" when another player removes any module, so I wanted to clarify that this is known behavior and not a bug.

Coding-Hen: If this is known behaviour and not a bug or something we are worried about do we even need a log?

tornac1234: Both logs are actual errors happening, whether it's a known behaviour or not (something broke so we need to notify it as an error). It's my bad for writing Log.Info in SpawnAsync though 😅 both should be Log.Error

Jannify (commented):
:( :O

Issue #2389, opened 2025-08-09
Screenshots folder should be in AppData


dartasen:
### Describe the issue

Currently they seems to appear inside the launcher folder, this will be problematic for OSX.

And also to keep a consistency with what we've done with config and logs.

SpaceMonkeyy86:
Completed in #2399

PR #2394, opened 2025-08-09
Change OptionsView "Save File Location" to "Nitrox Logs Location"

NinjaPedroX:
This PR changes the open folder button in OptionsView to open the Nitrox logs folder instead.

I think this is better since ServersView already has buttons to open save folder locations, and users may need to see their log files (in case we ask for it in the Discord help channel) so making a button dedicated to it in OptionsView would make it easier for them to get to it.

image

### DIFF
--- a/Nitrox.Launcher/ViewModels/Designer/DesignOptionsViewModel.cs
+++ b/Nitrox.Launcher/ViewModels/Designer/DesignOptionsViewModel.cs
@@ -13,6 +13,6 @@ internal class DesignOptionsViewModel : OptionsViewModel
             Platform = Platform.STEAM
         };
         LaunchArgs = "-vrmode none";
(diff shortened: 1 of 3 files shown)


NinjaPedroX:
> I am happy with this if it would help with support more but was wondering if it might be better for the user to have the nitrox folder and then they can navigate to saves or logs as need? Though arguably they can go to the logs and go back up to the folder

That's true, because fixing #2389 would require us to make an easy way for users to open that folder.

I feel that just putting a button for the Nitrox AppData folder would make it unintuitive to find the logs/screenshots folders though. What if we added two buttons under it to open the Logs and Screenshots folder? And maybe also another button for the saves folder (We wouldn't need to show the file path for each one)

Measurity (approved):

PR #2397, opened 2025-08-10
Fix crash from OpenableStateChanged packet spamming

### LINKED ISSUE #2396: Tons of lag when docking vehicle in Cyclops
### Describe the issue

During a test session there were a few times where everyone's game crashed because a client was spamming OpenableStateChanged packets. It was caused by someone docking a seamoth inside a cyclops, but that alone is not enough to reproduce the bug.

SpaceMonkeyy86:
Fixes #2396


Measurity (approved):
LGCW

tornac1234 (approved):
lgtmcw lets hope this fully fixes the cyclops for next test session 🙏

PR #2399, opened 2025-08-10
Moved Screenshots Path to Nitrox AppData

dartasen:
image

Currently the folder look like this

"screenshots" is hardcoded inside Subnautica's code and would required multiple transpilers to get rid of. Since "saves" is already in lower, I decided to lowered the 2 others as well


Coding-Hen (approved):
Assuming that the screenshots go into a screenshot folder in the app data I think that makes sense to have the saves and screenshots in the same place

Coding-Hen (approved):
Agreed, makes sense to have both in the same place

PR #2402, opened 2025-08-10
Fix cyclops bridge and hull fragments not appearing

tornac1234:
Can be reviewed commit by commit for ease of read

## Refactoring
Three entity types were refactored, thus killing all previous savefiles (except if you build a python script to rename those in your EntityData and GlobalRootData files):
- PlayerWorldEntity -> PlayerEntity
- VehicleWorldEntity -> VehicleEntity
- EscapePodWorldEntity -> EscapePodEntity

## Why ?
WorldEntity is special type of entities which can be spawned and move in the world
the type is inherited by GlobalRootEntity which doesn't use the cell system
thus, it must be clear which entities built uppon WorldEntity and GlobalRootEntity

Also it was the occasion to make actual entity spawners for those entities instead of using the WorldEntitySpawner which is not suited for those as they don't need cells

## Cell level fix
I've found that world entities never had their cell level applied to their GameObject, so they'd spawn and then be moved to the wrong cell (which is bad), and then, you'd never be able to find them in the right cell because you'd be loading/unloading the actualy cell when going around.

## Cold start fix
I've just thought about it quickly and it seems reasonable to add one frame to wait for cleaned up entities to actually be destroyed before spawning their replacement. And this improves a lot the logging which is no longer spammed by "unregistering id" stuff.

## Occasional stuff
I came across oxygen pipes while fixing cell level so I just made a quick fix for it (and also added a relevant movement whitelist entry)

## Potentially breaking change
I set world entities to no longer spawn when they don't have a parent or when the cell they're supposed to spawn in is no longer loaded. Thus we need some testing to find out if this broke some spawning stuff here and there.

### DIFF
--- a/Nitrox.Test/Server/Serialization/WorldPersistenceTest.cs
+++ b/Nitrox.Test/Server/Serialization/WorldPersistenceTest.cs
@@ -430,9 +430,9 @@ public class WorldPersistenceTest
                                         break;
                                     case PlayerEntity when globalRootEntityAfter is PlayerEntity:
                                         break;
-                                    case VehicleEntity vehicleEntity when globalRootEntityAfter is VehicleEntity vehicleWorldEntityAfter:
-                                        Assert.AreEqual(vehicleEntity.SpawnerId, vehicleWorldEntityAfter.SpawnerId);
-                                        Assert.AreEqual(vehicleEntity.ConstructionTime, vehicleWorldEntityAfter.ConstructionTime);
+                                    case VehicleEntity vehicleEntity when globalRootEntityAfter is VehicleEntity vehicleEntityAfter:
+                                        Assert.AreEqual(vehicleEntity.SpawnerId, vehicleEntityAfter.SpawnerId);
+                                        Assert.AreEqual(vehicleEntity.ConstructionTime, vehicleEntityAfter.ConstructionTime);
                                         break;
                                     case RadiationLeakEntity radiationLeakEntity when globalRootEntityAfter is RadiationLeakEntity radiationLeakEntityAfter:
                                         Assert.AreEqual(radiationLeakEntity.ObjectIndex, radiationLeakEntityAfter.ObjectIndex);
--- a/Nitrox.Test/Server/Serialization/WorldPersistenceTest.cs
+++ b/Nitrox.Test/Server/Serialization/WorldPersistenceTest.cs
@@ -396,8 +396,8 @@ public class WorldPersistenceTest
                                     case BuildEntity buildEntity when globalRootEntityAfter is BuildEntity buildEntityAfter:
                                         Assert.AreEqual(buildEntity.BaseData, buildEntityAfter.BaseData);
                                         break;
-                                    case EscapePodWorldEntity escapePodWorldEntity when globalRootEntityAfter is EscapePodWorldEntity escapePodWorldEntityAfter:
-                                        Assert.IsTrue(escapePodWorldEntity.Players.SequenceEqual(escapePodWorldEntityAfter.Players));
+                                    case EscapePodEntity escapePodEntity when globalRootEntityAfter is EscapePodEntity escapePodEntityAfter:
+                                        Assert.IsTrue(escapePodEntity.Players.SequenceEqual(escapePodEntityAfter.Players));
                                         break;
                                     case InteriorPieceEntity interiorPieceEntity when globalRootEntityAfter is InteriorPieceEntity interiorPieceEntityAfter:
                                         Assert.AreEqual(interiorPieceEntity.BaseFace, interiorPieceEntityAfter.BaseFace);
@@ -430,9 +430,9 @@ public class WorldPersistenceTest
                                         break;
                                     case PlayerEntity when globalRootEntityAfter is PlayerEntity:
                                         break;
-                                    case VehicleWorldEntity vehicleWorldEntity when globalRootEntityAfter is VehicleWorldEntity vehicleWorldEntityAfter:
-                                        Assert.AreEqual(vehicleWorldEntity.SpawnerId, vehicleWorldEntityAfter.SpawnerId);
-                                        Assert.AreEqual(vehicleWorldEntity.ConstructionTime, vehicleWorldEntityAfter.ConstructionTime);
+                                    case VehicleEntity vehicleEntity when globalRootEntityAfter is VehicleEntity vehicleWorldEntityAfter:
+                                        Assert.AreEqual(vehicleEntity.SpawnerId, vehicleWorldEntityAfter.SpawnerId);
+                                        Assert.AreEqual(vehicleEntity.ConstructionTime, vehicleWorldEntityAfter.ConstructionTime);
                                         break;
                                     case RadiationLeakEntity radiationLeakEntity when globalRootEntityAfter is RadiationLeakEntity radiationLeakEntityAfter:
                                         Assert.AreEqual(radiationLeakEntity.ObjectIndex, radiationLeakEntityAfter.ObjectIndex);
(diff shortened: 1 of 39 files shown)


Nitrox.Test/Server/Serialization/WorldPersistenceTest.cs
```diff
@@ -428,11 +428,11 @@ private static void EntityTest(Entity entity, Entity entityAfter)
...
                                         break;
-                                    case PlayerWorldEntity when globalRootEntityAfter is PlayerWorldEntity:
+                                    case PlayerEntity when globalRootEntityAfter is PlayerEntity:
                                         break;
-                                    case VehicleWorldEntity vehicleWorldEntity when globalRootEntityAfter is VehicleWorldEntity vehicleWorldEntityAfter:
-                                        Assert.AreEqual(vehicleWorldEntity.SpawnerId, vehicleWorldEntityAfter.SpawnerId);
-                                        Assert.AreEqual(vehicleWorldEntity.ConstructionTime, vehicleWorldEntityAfter.ConstructionTime);
+                                    case VehicleEntity vehicleEntity when globalRootEntityAfter is VehicleEntity vehicleWorldEntityAfter:
+                                        Assert.AreEqual(vehicleEntity.SpawnerId, vehicleWorldEntityAfter.SpawnerId);
+                                        Assert.AreEqual(vehicleEntity.ConstructionTime, vehicleWorldEntityAfter.ConstructionTime);
```
Jannify:
```suggestion
                                    case VehicleEntity vehicleEntity when globalRootEntityAfter is VehicleEntity vehicleEntityAfter:
                                        Assert.AreEqual(vehicleEntity.SpawnerId, vehicleEntityAfter.SpawnerId);
                                        Assert.AreEqual(vehicleEntity.ConstructionTime, vehicleEntityAfter.ConstructionTime);
```

NitroxClient/GameLogic/Entities.cs
```diff
@@ -117,8 +117,13 @@ public void BroadcastEntitySpawnedByClient(Entity entity, bool requireRespawn =
             packetSender.Send(new EntitySpawnedByClient(entity, requireRespawn));
         }
 
-        private IEnumerator SpawnNewEntities()
+        private IEnumerator SpawnNewEntities(bool coldStart = false)
         {
+            if (coldStart)
+            {
+                yield return null;
```
Jannify:
```suggestion
                yield return null; // Skips a frame
```

NitroxClient/GameLogic/Spawning/WorldEntitySpawner.cs
```diff
@@ -31,24 +30,28 @@ public WorldEntitySpawner(EntityMetadataManager entityMetadataManager, PlayerMan
...
-            return null;
+            parentCell.EnsureRoot();
         }
-
         Optional<GameObject> parent = (entity.ParentId != null) ? NitroxEntity.GetObjectFrom(entity.ParentId) : Optional.Empty;
 
+        // No place to spawn the entity
+        if (!found && !parent.HasValue)
+        {
+            return null;
```
SpaceMonkeyy86, line 43: We should add a debug log here to see what entities this affects and if it has unintended consequences. Or do you think that spam the log too much?

NitroxClient/GameLogic/Spawning/WorldEntitySpawner.cs
```diff
@@ -59,17 +62,39 @@ protected override bool SpawnsOwnChildren(WorldEntity entity)
...
+        {
+            parentCell.EnsureRoot();
+        }
+        Optional<GameObject> parent = (entity.ParentId != null) ? NitroxEntity.GetObjectFrom(entity.ParentId) : Optional.Empty;
+
+        // No place to spawn the entity
+        if (!found && !parent.HasValue)
         {
-            // Error logging is done in EnsureCell
             return true;
```
SpaceMonkeyy86, line 75: Same thing here

NitroxClient/GameLogic/Spawning/WorldEntitySpawner.cs
tornac1234: It will definitely spam too much. I've tried looking at which entities actually go here and it was always entities out of range. My concern here is: if we have no parent nor cell, then we might break the cell system by forcing one entity to go in a cell that isn't awake at the time

NitroxClient/GameLogic/Spawning/WorldEntitySpawner.cs
```diff
@@ -31,24 +30,28 @@ public WorldEntitySpawner(EntityMetadataManager entityMetadataManager, PlayerMan
 
     protected override IEnumerator SpawnAsync(WorldEntity entity, TaskResult<Optional<GameObject>> result)
     {
-        EntityCell cellRoot = EnsureCell(entity);
-        if (cellRoot == null)
+        bool found = TryFindAwakeParentCell(entity, out EntityCell parentCell);
```
Jannify:
```suggestion
        bool foundParentCell = TryFindAwakeParentCell(entity, out EntityCell parentCell);
```

NitroxClient/GameLogic/Spawning/WorldEntitySpawner.cs
```diff
@@ -59,17 +62,39 @@ protected override bool SpawnsOwnChildren(WorldEntity entity)
 
     protected override bool SpawnSync(WorldEntity entity, TaskResult<Optional<GameObject>> result)
     {
-        EntityCell cellRoot = EnsureCell(entity);
-        if (cellRoot == null)
+        bool found = TryFindAwakeParentCell(entity, out EntityCell parentCell);
```
Jannify:
```suggestion
        bool foundParentCell = TryFindAwakeParentCell(entity, out EntityCell parentCell);
```

Jannify (changes requested):

NitroxModel/DataStructures/GameLogic/Entities/VehicleEntity.cs
```diff
@@ -18,20 +18,21 @@ public class VehicleWorldEntity : GlobalRootEntity
...
     [IgnoreConstructor]
-    protected VehicleWorldEntity()
+    protected VehicleEntity()
     {
         // Constructor for serialization. Has to be "protected" for json serialization.
     }
 
-    public VehicleWorldEntity(NitroxId spawnerId, float constructionTime, NitroxTransform transform, string classId, bool spawnedByServer, NitroxId id, NitroxTechType techType, EntityMetadata metadata) :
-        base(transform, 0, classId, spawnedByServer, id, techType, metadata, null, new List<Entity>())
+    // 100 is GlobalRoot level
```
Coding-Hen: Is it worth us having this in a constant somewhere so all things that need the globalroot can use the same and it's not magic numbers? Just a thought, I think the comment also helps

NitroxClient/GameLogic/Terrain.cs
```diff
@@ -34,6 +34,7 @@ public void CellLoaded(Int3 batchId, Int3 cellId, int level)
 
         if (!visibleCells.Contains(cell))
         {
+            removedCells.Remove(cell);
```
Coding-Hen, line 37: This seems more complete and how it should be but does this fix any bugs we have?

tornac1234: That is really for a theoretical edge case that I wasn't able to reproduce. It'd happen when:
- you load a cell
- you unload that cell
- you load that cell again

Then addedCells and removedCells contains the cell, but when processed by server-side, addedCells are computed first, and then removedCells are computed, which makes it so you will lack that cell which is loaded on your client.

Jannify (approved):
LGTM CW

Coding-Hen (approved):
LGTM

PR #2405, opened 2025-08-12
Fix global sounds of seamoth cell change and laser cutter

### LINKED ISSUE #2358: Seamoth light-on sound plays globally when installing powercell
### Describe the issue

1. Join a game with two players.
2. Build a Seamoth.
3. One player takes out the power cell while the other is far away.
4. The first player re-inserts the power cell, both players should hear the light on seamoth sound.

### LINKED ISSUE #2392: Laser cutter sound is global
### Describe the issue

We need to fix that.

Jannify:
Closes #2358
Closes #2392


Jannify:
This pr changes how many sounds are handled but I could only test the laser cutter time wise. Maybe is better to use the big test session for intensive testing of this change

NitroxModel/DotnetCompatibilityExtension.cs
```diff
@@ -0,0 +1,19 @@
+using System.Collections.Generic;
+
+namespace NitroxModel;
+
+public static class DotnetCompatibilityExtension
```
dartasen: What a lovely name xD

DictionnaryExtensions maybe

dartasen: Not sure about the real need of this by looking at the usage though

Jannify: `TryAdd()` is for some reason not in netFramework (but net6+). So this class is only for things that are not in framework. Therefor `DictionnaryExtensions` seems like the wrong name but I am open for suggestions.

SpaceMonkeyy86: I would rename this class to `PolyfillExtensions` and move it to `NitroxModel.Helper`.

dartasen (approved):

Jannify:
~~Should not be merged before #2416 and adapted to the changes in NitroxModel/Extensions~~ Cherry picked Meas commit

PR #2411, opened 2025-08-14
Overhaul input system

tornac1234:
- [x] Deprecate old system
- [x] Implement new one
- [x] Move Nitrox Entry Point to PlatformUtils.Awake so it's early enough to patch GameInputSystem

Please verify Subnautica files to use this (because entry point was moved)


Nitrox.Launcher/Models/Utils/NitroxEntryPatch.cs
```diff
@@ -19,7 +19,7 @@ public static class NitroxEntryPatch
     private const string NITROX_ENTRY_TYPE_NAME = "Main";
     private const string NITROX_ENTRY_METHOD_NAME = "Execute";
 
-    private const string TARGET_TYPE_NAME = "StartScreen";
```
dartasen, line 22: Can you please update the comment L71 accordingly

NitroxClient/MonoBehaviours/Gui/Input/KeyBindings/Actions/ChatKeyBindingAction.cs
```diff
@@ -1,17 +1,24 @@
...
 using NitroxModel.Core;
+using UnityEngine.InputSystem;
 
-namespace NitroxClient.MonoBehaviours.Gui.Input.KeyBindings.Actions
+namespace NitroxClient.MonoBehaviours.Gui.Input.KeyBindings.Actions;
+
+public class ChatKeyBindingAction : KeyBinding
 {
-    public class ChatKeyBindingAction : KeyBindingAction
+    private readonly Lazy<PlayerChatManager> playerChatManager = new(NitroxServiceLocator.LocateService<PlayerChatManager>);
```
dartasen: Couldn't that be injected by constructor ?

Right now the only purpose of KeyBindingManager is to host constant values

@Measurity

NitroxClient/MonoBehaviours/Gui/Input/KeyBindingManager.cs
```diff
@@ -1,40 +1,17 @@
...
-        // Returns highest custom key binding value. If no custom key bindings, returns 0. 
-        public int GetHighestKeyBindingValue()
-        {
-            return KeyboardKeyBindings.Select(keyBinding => (int)keyBinding.Button).DefaultIfEmpty(0).Max();
-        }
-    }
+public static class KeyBindingManager
+{
+    public const int NITROX_BASE_ID = 1000;
+    public static int CurrentBaseId = NITROX_BASE_ID;
```
dartasen: What's the need for a global variable (☹️) that is only used in GameInputSystem_Initialize_Patch ?

NitroxClient/MonoBehaviours/NitroxBootstrapper.cs
```diff
@@ -9,7 +8,8 @@ public class NitroxBootstrapper : MonoBehaviour
 {
     internal static NitroxBootstrapper Instance;
 
-    private void Awake()
+    // Awake is too early in Subnautica's lifecycle to access PlatformUtils
+    private void Start()
```
dartasen: Does that mean there's a race condition ?

NitroxPatcher/Patches/Persistent/Enum_GetValues_Patch.cs
```diff
@@ -0,0 +1,33 @@
...
+using NitroxModel.Helper;
+
+namespace NitroxPatcher.Patches.Persistent;
+
+/// <summary>
+/// Specific patch for GameInput.Button enum to return also our own values.
+/// </summary>
+public partial class Enum_GetValues_Patch : NitroxPatch, IPersistentPatch
+{
+    private static readonly MethodInfo TARGET_METHOD = Reflect.Method(() => Enum.GetValues(default));
```
dartasen, line 14: Not a huge fan of patching BCL methods, is this the only way ?

Coding-Hen: I think when we looked short of patching all instances which want to use the Button and switching it out for our patches would be the only way, though even that might not be possible

NitroxClient/MonoBehaviours/Gui/Input/KeyBindings/Actions/ChatKeyBindingAction.cs
Measurity: Yep this code is iffy. 

`PlayerChatManager` should not need to know about `IMultiplayerSession` because it's a manager and the session is once per game (or at least should be).

`PlayerChatManager` should be refactored as a static/singleton and not send packets directly. Instead should have an event-like API that multiplayer can listen to, to send chat packet. This event should be unsubscribed when multiplayer ends.

After this, `ChatKeyBindingAction` can call static `PlayerChatManager` without going through DI.

NitroxClient/MonoBehaviours/NitroxBootstrapper.cs
tornac1234: Not really since this is expected behaviour. What I mean here is just that we can't have this code happening in Awake so we run it into Start instead. I'll update the comment

NitroxPatcher/Patches/Persistent/Enum_GetValues_Patch.cs
tornac1234: I tried modifying the static readonly value with reflection and in this particular case, it wouldn't work because the value is already computed when we try to modify it and it's an array of fixed size which needs to be reallocated entirely to fit the extra values. And we do very specific case patching here as we only target a precise enum

NitroxClient/MonoBehaviours/Gui/Input/KeyBindings/Actions/ChatKeyBindingAction.cs
tornac1234: Should I do the PlayerChatManager refactoring now ? Because ChatKeyBindingAction is manually initialized in a static manager so it doesn't know about DI (and it should definitely not since Keybindings are working at all times since game start). As of now, if you try opening the chat in the main menu, the lazy will register a null reference (no scope yet) and then it won't work in game.

dartasen (approved):

Issue #1689, opened 2021-12-27
Expand nitrox command system to wrap subnautica one


dartasen:
Right now, we're starting to patch multiple vanilla command to either disable them or to add some nitrox behaviours to them : 
- time
- sub
- speed
- daynightspeed
- ...

Could be a great idea to patch `DevConsole.Submit` to redirect the flow of command to Nitrox server so we can : 
- Easily disable/filter commands
- Add some interactions with our internal data
- Add a permission system to subnautica command
- Allows server to enable subnautica console without allowing non admin people to use them
- ...

PR #2419, opened 2025-08-18
Add local permission check upon sending cheat commands

dartasen:
The main advantage of this temporary solution (until #1689 is done) is that it prevents people from using most cheat commands without any control and without the server host being aware of it or being able to do anything about it.

* Allows logging of all use of cheat commands in the logs => Can help analyze bug reports or complaints in the help channel
* Prevents bugs on unsynchronized commands for new players
* Gives leverage to hosts who wish to limit or even ban the use of cheat commands

IG:
image
Console:
image

### DIFF
--- a/NitroxPatcher/Patches/Dynamic/DevConsole_Submit_Patch.cs
+++ b/NitroxPatcher/Patches/Dynamic/DevConsole_Submit_Patch.cs
@@ -25,7 +25,7 @@ public sealed partial class DevConsole_Submit_Patch : NitroxPatch, IDynamicPatch
             return true; 
         }
 
-        Log.InGame(Language.main.Get("Nitrox_MissingPermission").Replace("{PERMISSION}", Perms.MODERATOR.ToString()));
+        Log.InGame(Language.main.Get("Nitrox_MissingPermission").Replace("{PERMISSION}", nameof(Perms.MODERATOR)));
         return false;
     }
 }
--- a/NitroxServer/Communication/Packets/Processors/CheatCommandProcessor.cs
+++ b/NitroxServer/Communication/Packets/Processors/CheatCommandProcessor.cs
@@ -1,3 +1,4 @@
+using NitroxModel.DataStructures.GameLogic;
 using NitroxModel.Packets;
 using NitroxServer.Communication.Packets.Processors.Abstract;
 
@@ -7,12 +8,12 @@ public class CheatCommandProcessor : AuthenticatedPacketProcessor<CheatCommand>
 {
     public override void Process(CheatCommand packet, Player player)
     {
-        if (player.Permissions < NitroxModel.DataStructures.GameLogic.Perms.MODERATOR)
+        if (player.Permissions < Perms.MODERATOR)
         {
             Log.Warn($"{player.Name} used cheat command: '{packet.Command}' without sufficient permissions.");
             return;
         }
 
-        Log.Info($"{player.Name} used cheat command : '{packet.Command}'");
+        Log.Info($"{player.Name} used cheat command: '{packet.Command}'");
     }
 }
--- a/NitroxModel/DataStructures/GameLogic/Perms.cs
+++ b/NitroxModel/DataStructures/GameLogic/Perms.cs
@@ -1,13 +1,28 @@
-﻿using System;
+using System;
 
 namespace NitroxModel.DataStructures.GameLogic
 {
     public enum Perms : byte
     {
+        /// <summary>
+        /// No permissions
+        /// </summary>
         NONE,
+        /// <summary>
+        /// Default player permission, cannot use cheat and have access to basic server commands (e.g: help, list, whisper, whois, ...)
+        /// </summary>
         PLAYER,
+        /// <summary>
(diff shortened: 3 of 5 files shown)


NitroxServer/Communication/Packets/Processors/CheatCommandProcessor.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    public override void Process(CheatCommand packet, Player player)
+    {
+        if (player.Permissions < NitroxModel.DataStructures.GameLogic.Perms.MODERATOR)
+        {
+            Log.Warn($"{player.Name} used cheat command: '{packet.Command}' without sufficient permissions.");
+            return;
+        }
+
+        Log.Info($"{player.Name} used cheat command : '{packet.Command}'");
```
NinjaPedroX:
```suggestion
        Log.Info($"{player.Name} used cheat command: '{packet.Command}'");
```
:)

NitroxPatcher/Patches/Dynamic/DevConsole_Submit_Patch.cs
```diff
@@ -0,0 +1,31 @@
...
+        Log.Info($"Used cheat command : '{value}'");
+        Resolve<IPacketSender>().Send(new CheatCommand(value));
+
+        // Allow submit if player has sufficient permissions
+        if (Resolve<LocalPlayer>().Permissions >= Perms.MODERATOR)
+        {
+            return true; 
+        }
+
+        Log.InGame(Language.main.Get("Nitrox_MissingPermission").Replace("{PERMISSION}", Perms.MODERATOR.ToString()));
```
Measurity:
```suggestion
        Log.InGame(Language.main.Get("Nitrox_MissingPermission").Replace("{PERMISSION}", nameof(Perms.MODERATOR)));
```

dartasen: Done

NitroxPatcher/Patches/Dynamic/DevConsole_Submit_Patch.cs
```diff
@@ -0,0 +1,31 @@
...
+{
+    internal static readonly MethodInfo TARGET_METHOD = Reflect.Method((DevConsole t) => t.Submit(default));
+
+    public static bool Prefix(string value)
+    {
+        Log.Info($"Used cheat command : '{value}'");
+        Resolve<IPacketSender>().Send(new CheatCommand(value));
+
+        // Allow submit if player has sufficient permissions
+        if (Resolve<LocalPlayer>().Permissions >= Perms.MODERATOR)
```
SpaceMonkeyy86, line 23: Seems like there is no way to set a player's permissions to moderator (only admin). You could add an argument to /op to set the permission level.

NitroxServer/Communication/Packets/Processors/CheatCommandProcessor.cs
```diff
@@ -0,0 +1,18 @@
+using NitroxModel.Packets;
+using NitroxServer.Communication.Packets.Processors.Abstract;
+
+namespace NitroxServer.Communication.Packets.Processors;
+
+public class CheatCommandProcessor : AuthenticatedPacketProcessor<CheatCommand>
+{
+    public override void Process(CheatCommand packet, Player player)
+    {
+        if (player.Permissions < NitroxModel.DataStructures.GameLogic.Perms.MODERATOR)
```
SpaceMonkeyy86: This should be documented in `Perms` along with explanations of PLAYER and ADMIN

NitroxPatcher/Patches/Dynamic/DevConsole_Submit_Patch.cs
dartasen: /promote [player] [permission]

NitroxServer/Communication/Packets/Processors/CheatCommandProcessor.cs
dartasen: Not sure to get what needs to be documented

SpaceMonkeyy86: Something like PLAYER: "Cannot use cheat or server commands", MODERATOR: "Can use vanilla cheat commands and some server commands", ADMIN: "Can use any command"

dartasen: Can do so for now, but I guess that should be up to the server owner to decide at some point

NitroxServer/Communication/Packets/Processors/CheatCommandProcessor.cs
```diff
@@ -0,0 +1,18 @@
...
+        {
+            Log.Warn($"{player.Name} used cheat command: '{packet.Command}' without sufficient permissions.");
```
SpaceMonkeyy86, line 13:
```suggestion
            Log.Warn($"{player.Name} tried to use cheat command: '{packet.Command}' without sufficient permissions.");
```

PR #2424, opened 2025-08-18
Fix crafting sync in vehicle upgrade station

### LINKED ISSUE #2421: Vehicle Upgrade Console Fabricator Crafting Not Synchronized

Windows

Steam

### Describe the issue.

The Vehicle Upgrade Console’s Fabricator crafting is not synchronized. I tested it with two clients when I crafted an upgrade, it didn’t appear on the other client. But when I crafted a different upgrade on the other client and picked it up, it recognized that a module was picked up, but it was only the module crafted on that client, even though they were different.This is why I assume that only the crafting itself is not synchronized, which may also be the reason why auto-pickup does not work for this fabricator.I tested all other fabricators in the game, and all of them work, except the Scanner Room fabricator but I assume this is expected behavior since it is not synced.

Image

1. Craft an upgrade module at the Vehicle Upgrade Console terminal.
2. Have another player nearby and let them look at it.
3. Craft a different upgrade on the other client and pick it up both upgrades will be deleted, even if they are different.

### Relevant log output

_No response_

tornac1234:
- [x] Change the NitroxEntity's target from GhostCrafter to CrafterLogic (it works for every kind of crafter except scanner room which will be worked on in its own PR in the future)
- [x] Sync the amount of items currently in a crafter (e.g. you can only pick up one of the 4 items you just crafted, then there'll be 3 of those in the crafters for all other players)

Fixes #2421


NitroxPatcher/Patches/Dynamic/GhostCrafter_OnCraftingBegin_Patch.cs
```diff
@@ -2,24 +2,30 @@
...
 {
     private static readonly MethodInfo TARGET_METHOD = Reflect.Method((GhostCrafter t) => t.OnCraftingBegin(default(TechType), default(float)));
 
     public static void Postfix(GhostCrafter __instance, TechType techType, float duration)
     {
-        if (__instance.TryGetIdOrWarn(out NitroxId crafterId))
+        // We favor targeting the CrafterLogic instead of the GhostCrafter because in the base upgrade console module, the NitroxEntity is
+        // on the CrafterLogic only. On every other crafter type, both CrafterLogic and GhostCrafter are on the same object.
+
+        // Also for base upgrade console module, crafterLogic is nullified and never updated, so use use _logic instead for every crafter
```
Jannify:
```suggestion
        // Also for base upgrade console module, crafterLogic is nullified and never updated, so we use _logic instead for every crafter
```

NitroxModel/DataStructures/GameLogic/Entities/Metadata/CrafterMetadata.cs
```diff
@@ -12,9 +12,12 @@ public class CrafterMetadata : EntityMetadata
     public NitroxTechType TechType { get; }
 
     [DataMember(Order = 2)]
-    public float StartTime { get; }
+    public int Amount { get; }
```
Coding-Hen: Is there any point to changing the ordering of these data members? Could you not just add the additional ones afterwards as to not break the current data model?

I know it's not a big thing as we are going to break saves but I think this slight change in the way we add new fields would save us breaking saved objects as new data members aren't breaking but changing the ordering or shifting the properties can

NitroxPatcher/Patches/Dynamic/CrafterLogic_NotifyPickup_Patch.cs
```diff
@@ -0,0 +1,30 @@
...
+public sealed partial class CrafterLogic_NotifyPickup_Patch : NitroxPatch, IDynamicPatch
+{
+    public static readonly MethodInfo TARGET_METHOD = Reflect.Method((CrafterLogic t) => t.NotifyPickup(default));
+
+    public static void Prefix(CrafterLogic __instance)
+    {
+        // See GhostCrafter_OnCraftingBegin_Patch.Postfix to know why we get NitroxId on CrafterLogic
+        if (__instance.TryGetIdOrWarn(out NitroxId crafterId))
+        {
+            // Certain crafts give you multiple items, so we need to know how many are lefts
```
Coding-Hen:
```suggestion
            // Certain crafts give you multiple items, so we need to know how many are left
```

NitroxPatcher/Patches/Dynamic/CrafterLogic_NotifyPickup_Patch.cs
```diff
@@ -0,0 +1,30 @@
...
+            // or in case there are none left, we send TechType.None
+            int amountLeft = __instance.numCrafted - 1;
+            TechType techType = amountLeft > 0 ? __instance.craftingTechType : TechType.None;
+            Resolve<Entities>().BroadcastMetadataUpdate(crafterId, new CrafterMetadata(techType.ToDto(), amountLeft, DayNightCycle.main.timePassedAsFloat, 0f));
```
Coding-Hen: Is -1 valid here as an amount left when handled at the other end? I assume numcrafted in theory could be 0 and then the -1 would be calculated, just want to make sure we won't crash out from it

tornac1234: The whole code is called under a direct "while(this.numCrafted > 0)" so amountLeft can never be -1. But just in case I will add some check

tornac1234:
Here you go, one more thing synced here: (example below)
when you craft a radiation suit you are also given 2 other pieces of radiation equipment)
if you have no place for the 2 other, they'll stay in the crafter

this is now synced

also I did what was asked in the review comments

NitroxPatcher/Patches/Dynamic/CrafterLogic_NotifyPickup_Patch.cs
```diff
@@ -0,0 +1,55 @@
...
+            TechType techType = __instance.craftingTechType;
+
+            // Sometimes, there are linked items appearing after the end of the regular craft results
+            // We implement this logic now to get the real values because it normally happens after this prefix
+            if (amountLeft == 0)
+            {
+                linkedIndex++;
+                if (linkedIndex < linkedItemsCount)
+                {
+                    amountLeft = 1;
```
Coding-Hen, line 42: Can the amount ever be more than one left? I would assume it would mean there is more than one linked item in the list and more than one item left for us to process?

tornac1234: No, it's like this in Subnautica code. If there are more than one linked items, amountLeft will go from 0 to 1 and each time add one to linkedIndex, which makes it possible to have many linked items given.

Coding-Hen: Nice, just wanted to check we weren't going to cause something to have an issue

Coding-Hen (approved):
LGTM

PR #2425, opened 2025-08-19
Fix falling through the map

### LINKED ISSUE #2305: Exosuit falls through the floor when rejoining inside an alien base
### Describe the issue

- Join server with one player
- In the vanilla console, run `goto antechamber` then `spawn exosuit`
- Leave and rejoin

The exosuit will fall through the floor while the terrain is loading in.

tornac1234:
- [x] Fix players not shown as swimming when on top of the ocean
- [x] Fix falling through the map when connecting as a player
- [x] Fix falling through the map for exosuits at all times

Two new states to be persisted for players:
- DisplaySurfaceWater: responsible for telling the local player to display or not the ocean water plane when reconnecting
- InPrecursor: responsive for telling the local player that it can still walk in a technically submerged area (precursor zones)

Some refactoring done too

Can be reviewed commit by commit for ease of understanding

Overall, big solutions (patching the entity awake detection) applied to seemingly simple problems (falling through the map), because Subnautica is never easy to fix

Fixes #2305


NitroxClient/MonoBehaviours/AnimationSender.cs
```diff
@@ -1,42 +1,52 @@
...
+    ON,
+    UNSET
+}
+
+public enum AnimChangeType
+{
+    UNDERWATER,
+    BENCH,
+    INFECTION_REVEAL
+}
```
Jannify: I don't like the placement of this here. The mb only uses one enum and in the packet it gets translated to int. If your idea is extensibility maybe we should add integer consts to check against.

NitroxClient/GameLogic/InitialSync/PlayerPositionInitialSyncProcessor.cs
```diff
@@ -44,10 +47,22 @@ public override IEnumerator Process(InitialPlayerSync packet, WaitScreen.ManualW
...
         Optional<NitroxId> subRootId = packet.PlayerSubRootId;
         if (!subRootId.HasValue)
         {
             yield return Terrain.WaitForWorldLoad();
+            Player.main.UpdateIsUnderwater();
+
+            // Check if Player might fall through the map
+            if (packet.InPrecursor || !Player.main.IsUnderwaterForSwimming())
+            {
+                // This coroutine must be started in a different thread to release the current one
```
Jannify:
```suggestion
                // This coroutine must be started fresh to release the current one
```
Thread is maybe misleading

NitroxClient/MonoBehaviours/AnimationSender.cs
```diff
@@ -1,42 +1,52 @@
 using NitroxClient.GameLogic;
-using NitroxModel.Core;
 using UnityEngine;
 
-namespace NitroxClient.MonoBehaviours
+namespace NitroxClient.MonoBehaviours;
+
+public class AnimationSender : MonoBehaviour
```
Jannify: Would rename to UnderwaterStateTracker

Jannify (changes requested):

NitroxClient/MonoBehaviours/AnimationSender.cs
tornac1234: This is legacy stuff here 😅 I can probably move both enums to a NitroxModel/GameLogic/PlayerAnimation/[enum].cs so it never needs to be converted

dartasen: legacy => yeet

Jannify (approved):
LGTM CW

NitroxModel/DataStructures/GameLogic/Entities/Metadata/VehicleMetadata.cs
```diff
@@ -14,19 +14,23 @@ public abstract class VehicleMetadata : NamedColoredMetadata
     [DataMember(Order = 1)]
     public float Health { get; init; }
 
+    [DataMember(Order = 3)]
```
dartasen: 3 ?
Shouldn't it be 2 ?

NitroxClient/Communication/Packets/Processors/EntityMetadataUpdateProcessor.cs
```diff
@@ -11,18 +12,24 @@ namespace NitroxClient.Communication.Packets.Processors;
...
     }
 
     public override void Process(EntityMetadataUpdate update)
     {
-        if (!NitroxEntity.TryGetObjectFrom(update.Id, out GameObject gameObject))
+        if (entities.SpawningEntities)
         {
             entityMetadataManager.RegisterNewerMetadata(update.Id, update.NewValue);
+        }
+
```
dartasen, line 30: what's the benefit of removing the check ?

NitroxClient/Communication/Packets/Processors/EntityDestroyedProcessor.cs
```diff
@@ -21,9 +21,14 @@ public EntityDestroyedProcessor(Entities entities)
     public override void Process(EntityDestroyed packet)
     {
         entities.RemoveEntity(packet.Id);
-        if (!NitroxEntity.TryGetObjectFrom(packet.Id, out GameObject gameObject))
+        
+        if (entities.SpawningEntities)
         {
             entities.MarkForDeletion(packet.Id);
+        }
+
```
dartasen, line 29: what's the benefit of removing the check ?

NitroxClient/Communication/Packets/Processors/EntityMetadataUpdateProcessor.cs
tornac1234: it helps avoid a race condition in which a metadata arriving after a spawnentities packet would apply to the current reference of the object which wasn't replaced yet (by the spawnentities packet) because spawning occurs on multiple frames. In that particular case, the entity reference is valid on the local client so the check is false.

NitroxClient/Communication/Packets/Processors/EntityDestroyedProcessor.cs
tornac1234: same reason as above

Coding-Hen (approved):
LGTM

PR #2429, opened 2025-08-23
Improvements on Windows Interop layer

dartasen:
Context:
image

* Cleaned up old Win32Native P/Invokes since we used them on pre 1.8 but seems no longer to be used
* Removed the need of manually opening Process on our Win32 implementation


tornac1234 (approved):
Works on my machine 👍

Measurity (approved):
Can remove all code using `Handle` as well and depend on Process entirely. But LGTM regardless.

Jannify (approved):
LGTM CW but not really familiar with Win32/Peocess stuff

PR #2439, opened 2025-08-25
Fix crabsnakes size

tornac1234:
Needs testing !!! (feedbacks from people building Nitrox on their own would be appreciated on this PR) (only fresh world is meaningful for this PR)

- I could check myself that the crabsnakes and the eggs had a regular case (warp 130 -95 -390)
- I also could check that spawned fragments and stuff at wreck13 were actually spawned under the cellroot and not under the group placeholder (which is the base game's correct behaviour)
- I imported @killzoms fix for floating stones

### DIFF
--- a/NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs
+++ b/NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs
@@ -136,7 +136,7 @@ public class BatchEntitySpawner : IEntitySpawner
     }
 
     /// <inheritdoc cref="CreateEntityWithChildren" />
-    private IEnumerable<Entity> SpawnEntitiesUsingRandomDistribution(EntitySpawnPoint entitySpawnPoint, List<UwePrefab> prefabs, DeterministicGenerator deterministicBatchGenerator, Entity parentEntity = null)
+    private List<Entity> SpawnEntitiesUsingRandomDistribution(EntitySpawnPoint entitySpawnPoint, List<UwePrefab> prefabs, DeterministicGenerator deterministicBatchGenerator, Entity parentEntity = null)
     {
         // See CSVEntitySpawner.GetPrefabForSlot for reference
         List<UwePrefab> allowedPrefabs = FilterAllowedPrefabs(prefabs, entitySpawnPoint, out float fragmentProbability, out float completeFragmentProbability);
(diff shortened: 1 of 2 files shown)


NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs
```diff
@@ -177,29 +177,37 @@ private IEnumerable<Entity> SpawnEntitiesUsingRandomDistribution(EntitySpawnPoin
...
-                                                                        i > 0);
-                foreach (Entity entity in entities)
+                                                                        i > 0)];
+
+                // Unparenting the new entity because it's directly under the cell root
+                entities[0].ParentId = null;
+                if (entities[0] is WorldEntity worldEntity)
                 {
-                    yield return entity;
+                    worldEntity.Transform.SetParent(null, true);
```
killzoms, line 203: Did the LocalPosition and LocalRotation get changed to be local to the CellRoot with this change? or at least swap Local transforms with global transforms?

NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs
```diff
@@ -177,29 +177,37 @@ private IEnumerable<Entity> SpawnEntitiesUsingRandomDistribution(EntitySpawnPoin
 
         if (chosenPrefab.Count == 0)
         {
-            yield break;
+            return [];
```
Jannify, line 180: Would change return type to `List<Entity>`

NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs
tornac1234: When first parented, the entity had their correct global position and rotation since the hierarchy is filled from the top, and thus, unparenting while locking the global position and rotation makes it so the entities are correctly placed

NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs
```diff
@@ -310,19 +318,35 @@ private IEnumerable<Entity> CreateEntityWithChildren(EntitySpawnPoint entitySpaw
...
-            foreach (Entity childEntity in AllChildren(spawnedEntity))
-            {
-                yield return childEntity;
-            }
+            spawnedEntities.AddRange(AllChildren(spawnedEntity));
         }
+
+        spawnedEntities.AddRange(newCellRootChildren);
+
+        return spawnedEntities;
```
Jannify, line 344: Same here. Change return type to List<Entity>

NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs
```diff
@@ -310,19 +318,35 @@ private IEnumerable<Entity> CreateEntityWithChildren(EntitySpawnPoint entitySpaw
...
+                newCellRootChildren.Add(spawnedChildEntity);
+            }
+        }
+
         entityBootstrapperManager.PrepareEntityIfRequired(ref spawnedEntity, deterministicBatchGenerator);
 
-        yield return spawnedEntity;
+        List<Entity> spawnedEntities = [];
+        // Order of addition is important, first, the created entity, then its children
+        spawnedEntities.Add(spawnedEntity);
```
Jannify:
```suggestion
        // Order of addition is important, first, the created entity, then its children
        List<Entity> spawnedEntities = { spawnedEntity };
```

Jannify: Would also add `and last newCellRootChildren bc ...`

NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs
```diff
@@ -310,19 +318,35 @@ private IEnumerable<Entity> CreateEntityWithChildren(EntitySpawnPoint entitySpaw
...
+        {
+            Entity spawnedChildEntity = spawnedEntity.ChildEntities[i];
+            // Regular case in which a spawned child would actually need to be spawned under the cell root instead of the current entity
+            if (spawnedChildEntity.ParentId != spawnedEntity.Id)
+            {
+                spawnedEntity.ChildEntities.RemoveAt(i);
+                newCellRootChildren.Add(spawnedChildEntity);
+            }
+        }
+
```
Jannify: Can this be moved after `spawnedEntities` init to remove the `newCellRootChildren` list?

Jannify (changes requested):

NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs
```diff
@@ -310,19 +318,35 @@ private IEnumerable<Entity> CreateEntityWithChildren(EntitySpawnPoint entitySpaw
...
         entityBootstrapperManager.PrepareEntityIfRequired(ref spawnedEntity, deterministicBatchGenerator);
 
-        yield return spawnedEntity;
+        List<Entity> spawnedEntities = [];
+        // Order of addition is important, first, the created entity, then its children
+        spawnedEntities.Add(spawnedEntity);
 
         if (parentEntity == null) // Ensures children are only returned at the top level
         {
             // Children are yielded as well so they can be indexed at the top level (for use by simulation
```
Coding-Hen: I assume this comment can change now to reflect not yeilding

Coding-Hen (changes requested):

NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs
```diff
@@ -312,33 +318,42 @@ private IEnumerable<Entity> CreateEntityWithChildren(EntitySpawnPoint entitySpaw
...
+        List<Entity> allChildren = [.. entity.ChildEntities];
 
-            if (child.ChildEntities.Count > 0)
-            {
-                foreach (Entity childOfChild in AllChildren(child))
-                {
-                    yield return childOfChild;
-                }
-            }
+        for (int i = 0; i < allChildren.Count; i++)
```
Coding-Hen, line 351: If this is iterating over all children would a for each not be better than getting the index and using that? Less chance of index expectation problems

tornac1234: You can't add items to the list you're using foreach on

NitroxServer/GameLogic/Entities/Spawning/BatchEntitySpawner.cs
```diff
@@ -327,42 +329,31 @@ private IEnumerable<Entity> CreateEntityWithChildren(EntitySpawnPoint entitySpaw
...
+        {
+            allChildren.AddRange(allChildren[i].ChildEntities);
```
Jannify, line 353: Won't this only go 1 or 2 layers deep?

NinjaPedroX:
Playtest with 3 others today, they all looked correctly sized.

A few of them still had the bug where they're stuck in place facing vertically upwards though, as if they were hiding in a mushroom tree despite not being inside one.

Jannify (approved):

tornac1234:
I encountered the same bug and didn't investigate it much. Though it might need some effort

killzoms:
ahhhh its an issue of desync between clients, the snake is swimming on one person's screen but on another persons screen the snake is still in the mushroom animation, and also a snake can be in a mushroom on the main simulator's screen but on another screen its swimming...

PR #2441, opened 2025-08-26
Show RadminVPN IP in server console

NinjaPedroX:
This PR adds the RadminVPN IP to the server IP list if the app is running. Also refactored the function to make it easier to add other VPNs in the future should we decide to do so, assuming that getting their IPs work the same way as they do for Hamachi and Radmin.

Tested it out on my machine and it works, though I am not sure if there would be any issues if both Hamachi and Radmin are running at the same time as I don't have Hamachi installed (but I don't think this will be the case for most users).

### DIFF
--- a/NitroxModel/Helper/NetHelper.cs
+++ b/NitroxModel/Helper/NetHelper.cs
@@ -121,11 +121,11 @@ public static class NetHelper
         }
     }
 
-    public static IPAddress? GetHamachiIp()
+    public static IPAddress? GetVPNIp(string vpnNetworkName)
     {
         foreach (NetworkInterface ni in NetworkInterface.GetAllNetworkInterfaces())
         {
-            if (ni.Name != "Hamachi")
+            if (ni.Name != vpnNetworkName)
             {
                 continue;
             }
@@ -139,6 +139,16 @@ public static class NetHelper
             }
         }
         return null;
+    }
+
+    public static IPAddress? GetHamachiIp()
+    {
+        return GetVPNIp("Hamachi");
+    }
+
+    public static IPAddress? GetRadminIp()
+    {
+        return GetVPNIp("Radmin VPN");
     }
 
     private static bool? hasInternet;
--- a/NitroxModel/Helper/NetHelper.cs
+++ b/NitroxModel/Helper/NetHelper.cs
@@ -46,7 +46,7 @@ public static class NetHelper
                         .OrderBy(n => n.NetworkInterfaceType is NetworkInterfaceType.Ethernet ? 1 : 0)
                         .ThenBy(n => n.Name);
 
-    public static IPAddress GetLanIp()
+    public static IPAddress? GetLanIp()
     {
         lock (lanIpLock)
         {
@@ -73,7 +73,7 @@ public static class NetHelper
         return null;
     }
 
-    public static async Task<IPAddress> GetWanIpAsync()
+    public static async Task<IPAddress?> GetWanIpAsync()
     {
         lock (wanIpLock)
         {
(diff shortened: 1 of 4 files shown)


NitroxServer/Communication/Packets/Processors/DiscordRequestIPProcessor.cs
```diff
@@ -52,6 +52,7 @@ private static async Task<string> GetIpAsync()
     {
         Task<IPAddress> wanIp = NetHelper.GetWanIpAsync();
         Task<IPAddress> hamachiIp = Task.Run(NetHelper.GetHamachiIp);
+        Task<IPAddress> radminIp = Task.Run(NetHelper.GetRadminIp);
```
Coding-Hen: The comment a few lines above this needs changing, probably just say VPN rather than just listing all the vpns might be better

NitroxServer/Server.cs
```diff
@@ -289,6 +290,10 @@ private async Task LogHowToConnectAsync()
         {
             options.Add($"{hamachiIp.Result} - Friends using Hamachi (VPN)");
         }
+        if (await radminIp != null)
+        {
+            options.Add($"{radminIp.Result} - Friends using Radmin (VPN)");
```
Coding-Hen: Assuming that both are installed I think we should test this, I don't know if adding multiple options will make discord unhappy. Though really their code should support that

NitroxServer/Communication/Packets/Processors/DiscordRequestIPProcessor.cs
Coding-Hen: From this also does the order here matter, say we are using local will that override if I also have hamachi running and it then falls through to the Lan IP by default?

NitroxModel/Helper/NetHelper.cs
```diff
@@ -141,6 +141,16 @@ public static async Task<IPAddress> GetWanIpAsync()
...
 
+    public static IPAddress? GetHamachiIp()
+    {
+        return GetVPNIp("Hamachi");
+    }
+
+    public static IPAddress? GetRadminIp()
+    {
+        return GetVPNIp("Radmin VPN");
+    }
```
dartasen: We're going to syscall network interfaces n time for n different VPN ?

NitroxServer/Server.cs
```diff
@@ -279,6 +279,7 @@ private async Task LogHowToConnectAsync()
         Task<IPAddress> localIp = Task.Run(NetHelper.GetLanIp);
         Task<IPAddress> wanIp = NetHelper.GetWanIpAsync();
         Task<IPAddress> hamachiIp = Task.Run(NetHelper.GetHamachiIp);
+        Task<IPAddress> radminIp = Task.Run(NetHelper.GetRadminIp);
```
dartasen: We should refactor this at some point, so NetHelper can return a list of something that can be displayed. (Similar refactor to what we did for Game Discovery)

Coding-Hen: I agree I think we should do this, I was thinking probably okay for now but then if we want to add another VPN we should do

NitroxModel/Helper/NetHelper.cs
NinjaPedroX: Are you thinking of doing something more like `List<string> vpnIps = GetVPNIps(["Hamachi", "Radmin VPN"])` and then have the server automatically list each VPN IP based on the name?

NitroxServer/Communication/Packets/Processors/DiscordRequestIPProcessor.cs
dartasen: Order does not matter, all the informations will be displayed. Up to the user to pick the one that makes sense

dartasen (approved):

NitroxServer/Server.cs
NinjaPedroX: Both with that code and the refactored code, the server was successfully able to show me the Radmin and Hamachi IPs at the same time (tested on a laptop that had Hamachi installed).

As for the DiscordRequestIPProcessor code, it seems to only send the first VPN's IP (Radmin for me) regardless of if a user joined with Local, Hamachi, or Radmin. It also seems to give the VPN IP even if the VPN isn't running on the server host's machine.

I think the Discord IP request system should be changed/fixed as the current one doesn't seem to work correctly (especially now that we can have multiple VPNs for it), but that should be done in a different PR.

PR #2447, opened 2025-09-03
Prevent bench deconstruction while remote player sitting

Jannify:
Adds `RemotePlayerBenchBlocker` to benches (and chairs) so sitting players block deconstruction.

Tested with two players (one sitting, one deconstructing).

### DIFF
--- a/NitroxPatcher/Patches/Dynamic/Bench_OnPlayerDeath_Patch.cs
+++ b/NitroxPatcher/Patches/Dynamic/Bench_OnPlayerDeath_Patch.cs
@@ -1,7 +1,7 @@
 using System.Reflection;
 using NitroxClient.GameLogic;
 using NitroxModel.DataStructures;
-using NitroxModel.GameLogic.PlayerAnimation;
+using NitroxModel.Packets;
 
 namespace NitroxPatcher.Patches.Dynamic;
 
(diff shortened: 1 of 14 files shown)


NitroxPatcher/Patches/Dynamic/Bench_OnPlayerDeath_Patch.cs
```diff
@@ -13,10 +13,10 @@ public static void Postfix(Bench __instance)
     {
         if (__instance.TryGetIdOrWarn(out NitroxId id))
         {
+            Resolve<LocalPlayer>().BroadcastBenchChanged(id, BenchChanged.BenchChangeState.UNSET);
```
Coding-Hen, line 16: Do we have an issue with the player getting stuck here if we can't find the id. Say if you are loading in on the bench and it's not spawned yet? Might not be an issue just thought I would question it as didn't know if it meant we might not unset depending on ordering

Jannify: An alternative would be to send without the bench id (so the remote player gets unstuck) but the server would still need to be restarted if they want to destruct the bench (bc the player wasn't removed from the list)

Coding-Hen: I thought that would probably be the case, not sure it would really do much then that would be useful. I think we should keep it as is then

PR #2454, opened 2025-09-20
Fix fruit plant growing in cyclops

### LINKED ISSUE #2445: Error in server console - packet processor

Windows

Epic

### Describe the issue.

I was playing with my 2 friends and I noticed this error in the server console

```
 [ERR] Error in packet processor NitroxServer.Communication.Packets.Processors.EntityMetadataUpdateProcessor
System.ArgumentNullException: Value cannot be null. (Parameter 'key')
   at System.ThrowHelper.ThrowArgumentNullException(String name)
   at NitroxServer.GameLogic.Entities.EntityRegistry.TryGetEntityById[T](NitroxId id, T& entity) in NitroxServer\GameLogic\Entities\EntityRegistry.cs:line 30
   at NitroxServer.Communication.Packets.Processors.EntityMetadataUpdateProcessor.Process(EntityMetadataUpdate packet, Player sendingPlayer) in NitroxServer\Communication\Packets\Processors\EntityMetadataUpdateProcessor.cs:line 23
   at NitroxServer.Communication.Packets.PacketHandler.ProcessAuthenticated(Packet packet, Player player) in NitroxServer\Communication\Packets\PacketHandler.cs:line 51
```

I'm not quite sure what caused it, but at one point it kept us from syncing to get into the game

progress: we have a base in the safe shallows, and we just got back from fixing the Aurora in our 3 cyclops when i noticed the error also our base keeps flooding when we log in, also there are parts of the world that are barren (no plants animals etc) 

### Relevant log output

server log []https://pastebin.com/fB164jFw
game log []https://pastebin.com/bpMqiqLL

tornac1234:
Some condition was too narrow, targetting only bases while cyclops could also be valid owners.

Fixes #2445

### DIFF
--- a/NitroxPatcher/Patches/Dynamic/FruitPlant_Update_Patch.cs
+++ b/NitroxPatcher/Patches/Dynamic/FruitPlant_Update_Patch.cs
@@ -42,6 +42,7 @@ public sealed partial class FruitPlant_Update_Patch : NitroxPatch, IDynamicPatch
             return Resolve<SimulationOwnership>().HasAnyLockType(ownerNitroxId);
         }
 
+        Log.Error($"[{nameof(FruitPlant_Update_Patch)}] Could not find either the plantable id [{entityId}] or the planter's owner id for {__instance.name}");
         __state = (__instance.timeNextFruit, null, null);
         return true;
     }
@@ -50,6 +51,11 @@ public sealed partial class FruitPlant_Update_Patch : NitroxPatch, IDynamicPatch
     {
         // If no change was made
         if (__state.Item1 == __instance.timeNextFruit)
+        {
+            return;
+        }
+
+        if (__state.Item2 == null)
         {
             return;
         }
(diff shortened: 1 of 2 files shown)


Coding-Hen (approved):
LGTM, classic case of not having a guard in place. As long as we are okay with the error log as we think it will be useful to us or the user and it's something we are actually saying is an error

NitroxPatcher/Patches/Dynamic/FruitPlant_Update_Patch.cs
```diff
@@ -42,6 +42,7 @@ public static bool Prefix(FruitPlant __instance, out (float, NitroxId, Plantable
             return Resolve<SimulationOwnership>().HasAnyLockType(ownerNitroxId);
         }
 
+        Log.Error($"[{nameof(FruitPlant_Update_Patch)}] Could not find either the plantable id [{entityId}] or the planter's owner id for {__instance.name}");
```
Jannify, line 45: Wouldn't it's place in Update() lead to a lot of log spam?

PR #2321, opened 2025-04-27
Fix warpme command not sending escape pod changed packet

LeanHydra84:
Fixes issue #2238:

"warpme" command now correctly broadcasts the EscapePodChange packet when teleporting the player into the lifepod.
Adds patch (EscapePod_RespawnPlayer_Patch) to hook the function that respawns the player into the lifepod, either after a death or a "warpme".

Now-redundant code removed from Player_MovePlayerToRespawnPoint_Patch.

PR #2455, opened 2025-09-20
Fix unauthenticated EscapePodChanged packet upon joining

dartasen:
Context: Regressed in #2321
Context : 
image

In #2321 we did change the entrypoint ofescapepod changed packet in order to sync more behavior. However it did show up some path where SN doing some init code for escapepod position and player position during Subnautica boot. Which Nitrox is already doing behind the hood.

So I did suppress this code that shouldn't be called by Subnautica
image

image


tornac1234 (approved):
Image

Jannify (approved):
LGTM CW

PR #2456, opened 2025-09-20
Broaden battery sync to any charge modification

### LINKED ISSUE #2451: Power cells not remaining charged when reloading the game, unless they are in vehicles

Windows

Epic

### Describe the issue.

Power cells don't stay charged when loading the game unless they are in vehicles

I have a few power cell chargers and I had the power cells fully charged, and I reloaded the game, and they were depleted and had to charge again. Also, I had a small locker full of charged power cells and I reloaded the game, and they were uncharged as well.

### Relevant log output

Game []https://pastebin.com/NKw2ngLJ PT1
          []https://pastebin.com/J9AJ2CnL PT2
Server []https://pastebin.com/MFEMGMyR

_PS—Pastebin wouldn't let me post the whole game log, so I had to break it up_

tornac1234:
We used to only sync EnergyMixin.ModifyCharge with is only used when uncharging a battery. But now we'll take into account any change (increase or decrease). Also there's still the check to only broadcast once every integer energy unit.

Fixes #2451

### DIFF
--- a/NitroxPatcher/Patches/Dynamic/Charger_Update_Patch.cs
+++ b/NitroxPatcher/Patches/Dynamic/Charger_Update_Patch.cs
@@ -1,0 +1,64 @@
+using System.Collections.Generic;
+using System.Reflection;
+using System.Reflection.Emit;
+using HarmonyLib;
+using NitroxClient.GameLogic;
+using NitroxModel.DataStructures;
+
+namespace NitroxPatcher.Patches.Dynamic;
+
+/// <summary>
+/// Prevents chargers from working on remote clients since battery charging is synced directly on charge property modification (<see cref="Battery_charge_set_Patch"/>).
+/// This works by postponing charge attempts again and again as long we don't have simulation ownership.
+/// </summary>
+public sealed partial class Charger_Update_Patch : NitroxPatch, IDynamicPatch
+{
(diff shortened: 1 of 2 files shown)


NitroxPatcher/Patches/Dynamic/Charger_Update_Patch.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        // In case we don't locally simulate the SubRoot responsible for this charger, we postpone the next charge attempt
+        if (charger.transform.parent.TryGetNitroxId(out NitroxId parentId) &&
+            !Resolve<SimulationOwnership>().HasAnyLockType(parentId))
+        {
+            // 5 seconds is the regular value between two attempts (see Update)
+            charger.nextChargeAttemptTimer = 5f;
+
+            // Copied from Update to ensure UI is still updated
+            charger.ToggleUIPowered(true);
```
Jannify, line 53: Is this something we need to toggle once we have base power sync? If so would like a TODO here

tornac1234: This has nothing to do with base power sync. It's just there because we cancel a big part of the code which contains UI effects too and those instructions need to be done manually if we cancel the whole thing.

Jannify (approved):
LGTM CW

PR #2469, opened 2025-10-06
Fix fragments wrongly placed

### LINKED ISSUE #2446: Floating Fragments

Windows

Steam

### Describe the issue.

Fragments at Lifepod 19 (ive only seen it here so far.) are out of place spawning higher than its intended place

Image

I went to lifepod 19
CreateFullEntityCache is set to true aswell if that changes anything.

### Relevant log output

_No response_

tornac1234:
**VERY IMPORTANT**
You'll have to delete the ``%appdata%/Nitrox/cache/PrefabPlaceholdersGroupAssetsCache.json`` so that the asset cache is correctly re-generated according to the newer calculations.

Issue was:
- PrefabPlaceholderGroup
  |- Slots
     |- EntitySlot 1 (is parented to CellRoot when being spawned)
  | - Entities
      |- Statically generated prop (is parented to "Entities" when being spawned)

Issue is EntitySlot 1 only had a local position relative to Slots, missing the Slots local position relative to PrefabPlaceholderGroup

This calculation will only be applied in case there's an intermediary parent like "Slots" or "Entities".
Also we'll just ignore the props which aren't entity slots because they're spawned directly under the right parent so we don't need to worry about their global positionning.

To test you can go to crabsnakes area:
- warp -50 -223 -211
A regular wreck:
- goto wreck13
Some random place where it's really easy to see the non-fixed issue:
- warp 316 -112 242

Also side note: it doesn't matter if we do more calculation during this process (the ones I added for example) because it's cached and only runs once on a computer

Fixes #2446


dartasen:
With #2470, you should be able to ask for a cache invalidation
It's not really reasonable to ask players to delete cache by hand when we're going to break stuff inside cache processing logic (Until we discover all UWE edge cases)

dartasen:
image

Fragments and wrecks that were floating during RC5 seems to work now

tornac1234:
Rebased to have the cache number stuff, and added a commit to have it bumped by one

tornac1234:
Guys approve pls

Coding-Hen (approved):
LGTM. Shows a valid use case of the cache version as well which will be useful

misterbubb:
image 
can confirm it works, amazing work you guys!!

PR #2470, opened 2025-10-06
Fixed server always reusing prefab placeholder cache even if version changes

Measurity:
Thanks @dartasen for highlighting this issue


Measurity:
Closing. We're going with version id in cache file. If no version id, should invalidate cache. If version id mismatch, should invalidate cache. Cache id should manually be incremented in source code when making changes to `PrefabPlaceholderGroupsParser`.

PR #2474, opened 2025-10-08
Update Bug Report template

dartasen:
🚀

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -25,7 +25,7 @@ body:
     id: nitrox_version_details
     attributes:
       label: Details about version
-      description: If you answered `Latest commit` or `Other`, please provide commit or links to that version.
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -12,23 +12,31 @@ body:
...
+        - 1.7.1.0
+        - Latest commit (Only for experts)
+        - Other
     validations:
       required: true
+  - type: input
+    id: nitrox_version_details
+    attributes:
+      label: Details about version
+      description: If you answered `Latest commit` or `Other`, please provide commit or links to that version.
```
Jannify:
```suggestion
      description: If you answered `Latest commit` or `Other`, please provide the commit hash or links to that version.
```

Measurity:
Maybe point users to use launcher to open log to game (not sure if we added that feature). So when we move paths we don't have to update template again.

image

PR #2487, opened 2025-10-10
Increased game version target that Nitrox works on

Measurity:
Lower game versions than this will cause Nitrox to crash

image

### DIFF
--- a/NitroxModel/Core/NitroxEnvironment.cs
+++ b/NitroxModel/Core/NitroxEnvironment.cs
@@ -54,15 +54,15 @@ public static class NitroxEnvironment
         }
     }
 
-    public static int GameHigherThanVersionTarget
+    public static int GameMinimumVersion
     {
         get
         {
-            if (!int.TryParse(Assembly.GetExecutingAssembly().GetMetaData("GameHigherThanVersionTarget"), out int gameHigherThanVersionTarget))
+            if (!int.TryParse(Assembly.GetExecutingAssembly().GetMetaData(nameof(GameMinimumVersion)), out int result))
             {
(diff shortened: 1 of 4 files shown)


NitroxModel/Core/NitroxEnvironment.cs
```diff
@@ -54,6 +54,18 @@ public static string GitShortHash
         }
     }
 
+    public static int GameHigherThanVersionTarget
```
dartasen:
```suggestion
    public static int MinimumGameVersion
```

NitroxModel/Core/NitroxEnvironment.cs
```diff
@@ -54,6 +54,18 @@ public static string GitShortHash
...
+    {
+        get
+        {
+            if (!int.TryParse(Assembly.GetExecutingAssembly().GetMetaData("GameHigherThanVersionTarget"), out int gameHigherThanVersionTarget))
+            {
+                throw new Exception("Failed to extract compatible game version number from embedded metadata");
```
dartasen, line 63: Maybe we can have a fallback instead of throwing an exception ?

NitroxModel/Core/NitroxEnvironment.cs
Measurity: This changes the meaning. Minimum would mean the exact version is valid, but it has to be higher than that.

NitroxModel/Core/NitroxEnvironment.cs
Measurity: This check is done on every game start. So it's pretty much impossible to miss if something goes wrong. Rather keep it like this.

PR #2502, opened 2025-10-11
Renamed projects and restructured namespaces in Nitrox.Server.Subnautica to reduce merge conflicts with future changes

Measurity:
## Explaining changes to Nitrox.Server.Subnautica

There are now 3 main folders inside this server: "Extensions", "Services" and "Models". All existing code went into Models. "Services" will be new .NET HostedService types but is empty right now. "Extensions" speaks for itself.

## Up next :eyes: 

- #2314 

"Services" will contain the bulk of the server complexity but with separation of concern in mind, unlike the spaghet we have now.


tornac1234:
How we supposed to review that though 💀

Measurity:
Here's a patch file of all useful changes only @tornac1234 
[changed-lines.patch](https://github.com/user-attachments/files/22868154/changed-lines.patch)
___
Above patch was generated with this command:
```sh
git diff origin/master --unified=0 --diff-filter=AM -I"(using|namespace) " -I"^\s+$"
```

tornac1234:
Testing this with worlds created from 1.8 will break them as they contain serialized types as
```
NitroxModel.DataStructures.GameLogic.Entities.RadiationLeakEntity, NitroxModel, Version=1.8.0.0, Culture=neutral, PublicKeyToken=null
```
So this PR needs to create a SaveDataUpgrade and to bump Nitrox version already

Measurity:
PR is ready for review again

[diff of useful changes](https://github.com/user-attachments/files/22981299/patched.patch)

Measurity:
> I do question some of the model.subnautica changes as I think things like WorldEntity I would have thought would be in both?

I want to be more restrictive in sharing code so we don't design ourself into a corner (future hacky code). `WorldEntity` and its base `Entity` contain UWE/Subnautica specific data that can't be easily applied to other projects. They _may_ work on below zero but if UWE keeps updating one game and not the other in the future we'll be screwed directly where it hurts to adapt.

The benefit for sharing code is code-size and not having to keep code in sync. But with UWE the latter is unpredictable.

PR #2504, opened 2025-10-12
Fix bases bugs

tornac1234:
Bugs are described in commit names, and both would end up in wrong build desync detection, which would end up breaking the base very quickly.

### DIFF
--- a/NitroxServer/Communication/Packets/Processors/BuildingProcessor.cs
+++ b/NitroxServer/Communication/Packets/Processors/BuildingProcessor.cs
@@ -33,17 +33,15 @@ public abstract class BuildingProcessor<T> : AuthenticatedPacketProcessor<T> whe
 
     /// <summary>
     /// Attempts to acquire simulation ownership on <paramref name="entity"/> for <paramref name="player"/>.
-    /// If the attempt is unsuccessful, nothing will happen.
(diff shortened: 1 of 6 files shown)


NitroxServer/Communication/Packets/Processors/BuildingProcessor.cs
```diff
@@ -30,10 +31,19 @@ public void SendToOtherPlayersWithOperationId(T packet, Player player, int opera
...
+    /// If the attempt is unsuccessful, nothing will happen.
+    /// </summary>
     public void ClaimBuildPiece(Entity entity, Player player)
     {
-        SimulatedEntity simulatedEntity = entitySimulation.AssignNewEntityToPlayer(entity, player, false);
-        SimulationOwnershipChange ownershipChangePacket = new(simulatedEntity);
-        playerManager.SendPacketToAllPlayers(ownershipChangePacket);
+        try
+        {
+            SimulatedEntity simulatedEntity = entitySimulation.AssignNewEntityToPlayer(entity, player, false);
```
Measurity: Should add try catch only on offending line ~~and also add debug log~~.

(Ideally, if it is expected that nothing can happen, we should change the code so that we can check for it without throwing exception.)

PR #2513, opened 2025-10-12
Fix certain world entities have global root level 100

tornac1234:
To test the bug:
- place a luggage/poster in your base (it's a GlobalRootEntity) and disconnect
- reconnect, it is now given the cell level 100 (global root)
- pick it up, and drop it in the water, it still has cell level 100

Fixes an issue where the server is not gonna start because of certain world entities with a level 100.


dartasen:
image
Tried with all of them

dartasen:
[Poster&Cie.zip](https://github.com/user-attachments/files/22888832/Poster.Cie.zip)

PR #2520, opened 2025-10-14
Add Heroic Games Launcher support

Jannify:
Adds finding and starting Subnautica through the Heroic Games Launcher. This is needed for Linux as Epic has no app for it but can also be used on Windows or macOS.

### DIFF
--- a/Nitrox.Launcher/ViewModels/LaunchGameViewModel.cs
+++ b/Nitrox.Launcher/ViewModels/LaunchGameViewModel.cs
@@ -237,7 +237,7 @@ internal partial class LaunchGameViewModel(DialogService dialogService, ServerSe
         string launchArguments = $"{keyValueStore.GetLaunchArguments(gameInfo)} {string.Join(" ", args ?? NitroxEnvironment.CommandLineArgs)}";
         ProcessEx game = NitroxUser.GamePlatform switch
         {
-            Steam => await Steam.StartGameAsync(gameExePath, launchArguments, gameInfo.SteamAppId, ShouldSkipSteam(launchArguments)),
+            Steam => await Steam.StartGameAsync(gameExePath, launchArguments, gameInfo.SteamAppId, ShouldSkipSteam(launchArguments), keyValueStore.GetUseBigPictureMode()),
             EpicGames => await EpicGames.StartGameAsync(gameExePath, launchArguments),
             HeroicGames => await HeroicGames.StartGameAsync(gameInfo.EgsNamespace, launchArguments),
             MSStore => await MSStore.StartGameAsync(gameExePath, launchArguments),
@@ -253,6 +253,12 @@ internal partial class LaunchGameViewModel(DialogService dialogService, ServerSe
 
     private bool ShouldSkipSteam(string args)
     {
+        // Check if Steam overlay is enabled by user setting
+        if (keyValueStore.GetUseBigPictureMode())
+        {
+            return false; // Use Steam if overlay is enabled
+        }
+
         if (App.InstantLaunch != null)
         {
             // Running through Steam is fine if single instance.
@@ -271,6 +277,6 @@ internal partial class LaunchGameViewModel(DialogService dialogService, ServerSe
             return true;
         }
 
-        return false;
+        return false; // Default: use Steam unless explicitly disabled for special cases
     }
 }
--- a/Nitrox.Launcher/ViewModels/OptionsViewModel.cs
+++ b/Nitrox.Launcher/ViewModels/OptionsViewModel.cs
@@ -55,6 +55,9 @@ internal partial class OptionsViewModel(IKeyValueStore keyValueStore, StorageSer
     private bool allowMultipleGameInstances;
 
     [ObservableProperty]
+    private bool useBigPictureMode;
+    
+    [ObservableProperty]
     private bool isInReleaseMode;
 
     private static string DefaultLaunchArg => "-vrmode none";
@@ -70,6 +73,7 @@ internal partial class OptionsViewModel(IKeyValueStore keyValueStore, StorageSer
         LogsFolderDir = Model.Logger.Log.LogDirectory;
         LightModeEnabled = keyValueStore.GetIsLightModeEnabled();
         AllowMultipleGameInstances = keyValueStore.GetIsMultipleGameInstancesAllowed();
+        UseBigPictureMode = keyValueStore.GetUseBigPictureMode();
         IsInReleaseMode = NitroxEnvironment.IsReleaseMode;
         await SetTargetedSubnauticaPathAsync(SelectedGame.PathToGame).ContinueWithHandleError(ex => LauncherNotifier.Error(ex.Message));
     }
@@ -197,6 +201,19 @@ internal partial class OptionsViewModel(IKeyValueStore keyValueStore, StorageSer
 
     partial void OnAllowMultipleGameInstancesChanged(bool value)
     {
+        if (value)
+        {
+            UseBigPictureMode = false;
+        }
         keyValueStore.SetIsMultipleGameInstancesAllowed(value);
+    }
+    
+    partial void OnUseBigPictureModeChanged(bool value)
+    {
+        if (value)
+        {
+            AllowMultipleGameInstances = false;
+        }
+        keyValueStore.SetBigPictureMode(value);
     }
 }
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -15,7 +15,8 @@ body:
       label: Which version are you playing on?
       description: If playing on older versions, you should try updating first.
       options:
-        - 1.8.0.0 (Latest release)
+        - 1.8.0.1 (Latest release)
+        - 1.8.0.0
         - 1.7.1.0
         - Latest commit (Only for experts)
         - Other
--- a/Nitrox.Launcher/Models/Controls/RecolorImage.cs
+++ b/Nitrox.Launcher/Models/Controls/RecolorImage.cs
@@ -144,6 +144,7 @@ public sealed class RecolorImage : Control
     {
         if (Source is not null)
         {
+            stream.Position = 0;
             Source.Save(stream);
         }
     }
--- a/Nitrox.Launcher/Models/Extensions/KeyValueStoreExtensions.cs
+++ b/Nitrox.Launcher/Models/Extensions/KeyValueStoreExtensions.cs
@@ -10,15 +10,13 @@ public static class KeyValueStoreExtensions
 
     public static bool GetIsLightModeEnabled(this IKeyValueStore store, bool defaultValue = false) => store.GetValue("IsLightModeEnabled", defaultValue);
 
-    public static void SetIsLightModeEnabled(this IKeyValueStore store, bool value)
-    {
-        store.SetValue("IsLightModeEnabled", value);
-    }
+    public static void SetIsLightModeEnabled(this IKeyValueStore store, bool value) => store.SetValue("IsLightModeEnabled", value);
 
     public static bool GetIsMultipleGameInstancesAllowed(this IKeyValueStore store, bool defaultValue = false) => store.GetValue("IsMultipleGameInstancesAllowed", defaultValue);
 
-    public static void SetIsMultipleGameInstancesAllowed(this IKeyValueStore store, bool value)
-    {
-        store.SetValue("IsMultipleGameInstancesAllowed", value);
-    }
+    public static void SetIsMultipleGameInstancesAllowed(this IKeyValueStore store, bool value) => store.SetValue("IsMultipleGameInstancesAllowed", value);
+
+    public static bool GetUseBigPictureMode(this IKeyValueStore store, bool defaultValue = false) => store.GetValue("UseBigPictureMode", defaultValue);
+
+    public static void SetBigPictureMode(this IKeyValueStore store, bool value) => store.SetValue("UseBigPictureMode", value);
 }
--- a/Nitrox.Launcher/Views/OptionsView.axaml
+++ b/Nitrox.Launcher/Views/OptionsView.axaml
@@ -11,7 +11,8 @@
     xmlns:vm="clr-namespace:Nitrox.Launcher.ViewModels"
     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
     xmlns:controls="clr-namespace:Nitrox.Launcher.Models.Controls"
-    xmlns:designer="clr-namespace:Nitrox.Launcher.ViewModels.Designer">
+    xmlns:designer="clr-namespace:Nitrox.Launcher.ViewModels.Designer"
+    xmlns:models="clr-namespace:Nitrox.Model.Platforms.Discovery.Models;assembly=Nitrox.Model">
     <Design.DataContext>
         <designer:DesignOptionsViewModel />
     </Design.DataContext>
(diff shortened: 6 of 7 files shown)


NitroxModel/GameInfo.cs
```diff
@@ -30,7 +32,8 @@ static GameInfo()
             DataFolder = "Subnautica_Data",
             ExeName = "Subnautica.exe",
             SteamAppId = 264710,
-            MsStoreStartUrl = @"ms-xbl-38616e6e:\\"
+            MsStoreStartUrl = @"ms-xbl-38616e6e:\\",
+            EpicGameName = "Jaguar"
```
Measurity: `EgsName` or `EpicGamesName` to match brand name better.

NitroxModel/Platforms/Store/HeroicGames.cs
```diff
@@ -0,0 +1,61 @@
...
+namespace NitroxModel.Platforms.Store;
+
+public sealed class HeroicGames : IGamePlatform
+{
+    public string Name => "Heroic Games Launcher";
+    public Platform Platform => Platform.HEROIC;
+
+    public bool OwnsGame(string gameDirectory)
+    {
+        return Directory.GetFiles(gameDirectory).Length >= 4; // Heroic Games has no files inside or in the parent of the game directory
```
Measurity: This is too generic. The order in which the platform is evaluated shouldn't matter. Maybe we should store the path from the search result and check for it here? Kind of an hack but at least more consistent.

Kiaos:
flatpak version of heroic games launcher does not work. 

appimage of heroic games launcher works.

Nitrox.Launcher/ViewModels/LaunchGameViewModel.cs
```diff
@@ -221,65 +222,34 @@ private void HandleInstantLaunchForDevelopment()
...
-        {
-            subnauticaExe = Path.Combine(subnauticaPath, "MacOS", GameInfo.Subnautica.ExeName);
-        }
-        else
-        {
-            subnauticaExe = Path.Combine(subnauticaPath, GameInfo.Subnautica.ExeName);
-        }
+
+        string gameExePathSuffix = RuntimeInformation.IsOSPlatform(OSPlatform.OSX) ? "MacOS" : string.Empty;
+        string subnauticaExe = Path.Combine(NitroxUser.GamePath, gameExePathSuffix, gameInfo.ExeName);
```
Coding-Hen: If we are removing the Subnatica reference below for I assume making bz easier do we want to make this more generic with launch path or something along those lines?

NitroxModel/Platforms/Store/HeroicGames.cs
```diff
@@ -0,0 +1,75 @@
...
+public sealed class HeroicGames : IGamePlatform
+{
+    public string Name => "Heroic Games Launcher";
+    public Platform Platform => Platform.HEROIC;
+
+    public bool OwnsGame(string gameDirectory)
+    {
+        // Heroic Games has no files inside or in the parent of the game directory
+        GameFinderResult? resultSubnautica = new HeroicGamesFinder().FindGame(GameInfo.Subnautica);
+        if (resultSubnautica.IsOk)
```
Coding-Hen: I think we either want this or the find for bz below to use the same format

NitroxModel/Platforms/Store/HeroicGames.cs
```diff
@@ -0,0 +1,75 @@
...
+        GameFinderResult? resultBelowZero = new HeroicGamesFinder().FindGame(GameInfo.SubnauticaBelowZero);
+        return resultBelowZero.IsOk && PathEquals(gameDirectory, resultBelowZero.Path);
+
+        static string PathNormalize(string path) => Path.GetFullPath(path).TrimEnd(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar);
+        static bool PathEquals(string path1, string path2) => string.Equals(PathNormalize(path1), PathNormalize(path2), StringComparison.Ordinal);
+    }
+
+    public async Task<ProcessEx> StartPlatformAsync()
+    {
+        await Task.CompletedTask; // Suppresses async-without-await warning - can be removed.
```
Coding-Hen: Can this be removed or do you mean it removes the wait on this?

Nitrox.Launcher/ViewModels/LaunchGameViewModel.cs
Jannify: Yes the intend is to move the const GameInfo.Subnautica higher in the call hierarchy. It's not ideal but a start. While refactoring I missed this name...

NitroxModel/Platforms/Store/HeroicGames.cs
Jannify: To be honest I copy pasted this from the epic file. Don't know what was meant with this... xD

Measurity: I left the `async` on the method so it's easier to extend later but then you need to hide the warning with an `await` inside.

Can also do the following to hide the method if you have direct instance to HeroicGames to indicate it should not be called:
```cs
Task<ProcessEx> IGamePlatform.StartPlatformAsync()
{
    throw new NotImplementedException();
}
```

Coding-Hen: I can see both arguments for it. I think let's go with whatever is more in line with the other files to keep consistency if we aren't bothered

NitroxModel/Platforms/Store/Steam.cs
```diff
@@ -134,7 +134,7 @@ await RegistryEx.CompareWaitAsync<int>(@"SOFTWARE\Valve\Steam\ActiveProcess\Acti
         return File.Exists(steamExecutable) ? Path.GetFullPath(steamExecutable) : null;
     }
 
-    public async Task<ProcessEx?> StartGameAsync(string pathToGameExe, string launchArguments, int steamAppId, bool skipSteam)
+    public async Task<ProcessEx?> StartGameAsync(GameInfo gameInfo, string pathToGameExe, string launchArguments, IKeyValueStore keyValueStore)
```
Measurity: I liked the platform-specific parameters better so you know when calling it exactly the inputs that are required instead of a catch-all container.

If we do want to go this way (to have a shared interface) then I suggest going with a `PlatformContext` object that can be platform specific. But I think that's over-engineering for our requirements.

Jannify: I don't see why custom platform-dependent logic is in a generic model file. We even have them as an interface so why do we try to implement it against the pattern? I thought about the PlatformContext but to be fair only steam is the special guest here. Every other launcher has nearly the same minimalist requirements for the data it needs. Also this PlatformContext would probably be provided by the IGamePlatform implementation so we would call the file to get some struct to then give it back to the file. Sounds not that good imo.

Measurity: > I don't see why custom platform-dependent logic is in a generic model file. 

Because each platform has specific requirements. Sometimes they match. But in real world Steam doesn't ask Epic Games how to design their API. So it's just luck and we shouldn't design around this.

> We even have them as an interface so why do we try to implement it against the pattern?

No, we had an interface for `OwnsGame` but wasn't made for starting the game, because it can't be generalized unless we aggregate all the inputs. As we can see now for MSStore for example. It means for every new platform we need to just keep adding parameters for all other platforms as well, which is bad design.

> Also this PlatformContext would probably be provided by the IGamePlatform implementation so we would call the file to get some struct to then give it back to the file. Sounds not that good imo.

I agree, but new code in this PR requires more refactoring in the future than what we had before.

Nitrox.Model/Platforms/Store/HeroicGames.cs
```diff
@@ -0,0 +1,75 @@
...
+            return true;
+        }
+
+        GameFinderResult? resultBelowZero = new HeroicGamesFinder().FindGame(GameInfo.SubnauticaBelowZero);
+        if (resultBelowZero.IsOk && PathEquals(gameDirectory, resultBelowZero.Path))
+        {
+            return true;
+        }
+
+        return false;
```
Measurity, line 40: We should pass the finder result to OwnsGame calls so we aren't doing redundant searches (First by `NitroxUser.GamePath` and then in each platform again).

Nitrox.Launcher/ViewModels/OptionsViewModel.cs
```diff
@@ -146,15 +146,15 @@ private void ResetArguments(IInputElement? focusTargetAfterReset = null)
...
+        keyValueStore.SetLaunchArguments(GameInfo.Subnautica, LaunchArgs);
         SetArgumentsCommand.NotifyCanExecuteChanged();
     }
 
     private bool CanSetArguments()
     {
         ShowResetArgsBtn = LaunchArgs != DefaultLaunchArg;
 
-        return LaunchArgs != keyValueStore.GetSubnauticaLaunchArguments(DefaultLaunchArg) && !isResettingArgs;
+        return LaunchArgs != keyValueStore.GetLaunchArguments(GameInfo.Subnautica, DefaultLaunchArg) && !isResettingArgs;
```
Coding-Hen, line 161: We are hard coding subnatica here

NitroxModel/Platforms/Discovery/GameInstallationFinder.cs
```diff
@@ -0,0 +1,86 @@
...
+        GameFinderResult potentiallyValidResult = finderResults.LastOrDefault();
+        if (potentiallyValidResult?.IsOk == true)
+        {
+            Log.Debug($"Game installation was found by {potentiallyValidResult.FinderName} at '{potentiallyValidResult.Path}'");
+            IGamePlatform platform = GamePlatforms.GetPlatformByFlag(potentiallyValidResult.Origin);
+            NitroxUser.SetGamePathAndPlatform(potentiallyValidResult.Path, platform);
+            return true;
+        }
+
+        Log.Error($"Could not locate Subnautica installation directory: {Environment.NewLine}{string.Join(Environment.NewLine, finderResults.Select(i => $"{i.FinderName}: {i.ErrorMessage}"))}");
```
Coding-Hen: Might want to change this slightly to game installation rather than just subnatica 
```suggestion
        Log.Error($"Could not locate game installation directory: {Environment.NewLine}{string.Join(Environment.NewLine, finderResults.Select(i => $"{i.FinderName}: {i.ErrorMessage}"))}");
```
Or use gameinfo.Name like you have in other places

Nitrox.Launcher/ViewModels/OptionsViewModel.cs
Jannify: I know but this is tied to the Launcher-UI-Logic, for this pr I only wanted to separate Game starting and files in Nitrox.Model

Nitrox.Launcher/Models/Converters/PlatformToIconConverter.cs
```diff
@@ -13,12 +13,13 @@ public override object Convert(object value, Type targetType, object parameter,
         return AssetHelper.GetAssetFromStream(GetIconPathForPlatform(value as Platform?), static stream => new Bitmap(stream));
     }
 
-    private static string GetIconPathForPlatform(Platform? platform) => platform switch
+    private static string GetIconPathForPlatform(Platform? platform) => $"/Assets/Images/store-icons/{platform switch
```
dartasen: I would rather don't use string interpolation here, besides the odd syntaxic look it's just allocating for free

NitroxModel/Extensions/DictionaryExtensions.cs
```diff
@@ -41,5 +41,20 @@ public static bool TryAdd<TKey, TValue>(this IDictionary<TKey, TValue> dict, TKe
...
+        {
+            throw new ArgumentNullException(nameof(dictionary));
+        }
+
+        if (key == null)
+        {
+            throw new ArgumentNullException(nameof(key));
+        }
+
+        return !dictionary.TryGetValue(key, out TValue obj) ? defaultValue : obj;
```
dartasen: Nit:

```suggestion
        return dictionary.TryGetValue(key, out TValue obj) ? obj : defaultValue;
```

Nitrox.Model/Platforms/Discovery/InstallationFinders/HeroicGamesFinder.cs
```diff
@@ -0,0 +1,61 @@
...
+                }
+            }
+        }
+
+        return NotFound();
+    }
+
+    private static bool IsGameInstalled(string installedJson, GameInfo gameInfo, out string installPath)
+    {
+        JsonNode? jsonRoot = JsonNode.Parse(installedJson);
```
dartasen, line 49: I guess you can early return if jsonRoot is null

NitroxModel/Platforms/Discovery/GameInstallationFinder.cs
```diff
@@ -0,0 +1,87 @@
...
+
+    /// <summary>
+    /// Searches the system for a valid installation of <param name="gameInfo"></param>.
+    /// If found saves it to <see cref="NitroxUser.GamePath"/> / <see cref="NitroxUser.GamePlatform"/> and returns true.
+    /// </summary>
+    public static bool FindPlatformAndGame(GameInfo gameInfo, GameLibraries gameLibraries = GameLibraries.ALL)
+    {
+        List<GameFinderResult> finderResults = FindGame(gameInfo, gameLibraries).TakeUntilInclusive(r => r is { IsOk: false }).ToList();
+        GameFinderResult potentiallyValidResult = finderResults.LastOrDefault();
+        if (potentiallyValidResult?.IsOk == true)
```
dartasen:
```suggestion
        if (potentiallyValidResult is { IsOk: true})
```

NitroxModel/Platforms/Discovery/GameInstallationFinder.cs
```diff
@@ -0,0 +1,87 @@
...
+    };
+
+    /// <summary>
+    /// Searches the system for a valid installation of <param name="gameInfo"></param>.
+    /// If found saves it to <see cref="NitroxUser.GamePath"/> / <see cref="NitroxUser.GamePlatform"/> and returns true.
+    /// </summary>
+    public static bool FindPlatformAndGame(GameInfo gameInfo, GameLibraries gameLibraries = GameLibraries.ALL)
+    {
+        List<GameFinderResult> finderResults = FindGame(gameInfo, gameLibraries).TakeUntilInclusive(r => r is { IsOk: false }).ToList();
+        GameFinderResult potentiallyValidResult = finderResults.LastOrDefault();
```
dartasen:
```suggestion
        GameFinderResult? potentiallyValidResult = finderResults.LastOrDefault();
```

Nitrox.Model/Platforms/Discovery/InstallationFinders/HeroicGamesFinder.cs
Jannify: It technically is. When I make it more clear with another if it's 4 very short lines added.

NitroxModel/Platforms/Discovery/GameInstallationFinder.cs
Jannify: GameFinderResult is a record and therefor can't be null. Does it make sense here to add nullability? I don't think LastOrDefault() returns null then.

Nitrox.Launcher/Models/Converters/PlatformToIconConverter.cs
Jannify: Fair. I was also unsure about this but wanted to get your opinion.

Jannify:
Did a rebase and tested that launcher, server and client starts and I can join the server.

Measurity (approved):

Coding-Hen (approved):
LGTM

Measurity (approved):

Measurity:
Any blockers still left on this PR?

Issue #2525, opened 2025-10-14
Charge on cyclops is negative after powering up engine without powercells present


Yorboot:
### Which version are you playing on?

1.8.0.0 (Latest release)

### Which OS are you playing on?

Windows

### On which platform did you buy Subnautica?

Steam

### Describe the issue.

When all power cells are removed from the Cyclops. the charge counter on the HUD displays a negative value instead of showing zero or indicating that there is no power available.

The displayed charge should just be 0%.

### Describe the steps you took to encounter the bug.

1. Make a Cyclops
2. Enter the cyclops
3. Withdraw all powercells
4. Look at charge count
clip https://medal.tv/games/subnautica/clips/liDCWeFp4dfQV0SfC?invite=cr-MSxzb2ksMjA3Mzk1ODQy&v=19

### Relevant log output

Game: https://pastebin.com/yneGtXAK
server: https://pastebin.com/kRczYGLP

Measurity:
Thanks for the issue.

I want to avoid patching UI text elements to hide (Subnautica bugs and) power issues with cyclops. We should first implement proper power sync and then redo cyclops power with it so values can't be negative anymore. See:

- #1811

I've attached this issue as sub-task to above.

PR #2526, opened 2025-10-14
Fix negative Cyclops power HUD display — clamp value to 0% when negative

Yorboot:
### **Issue Related**
https://github.com/SubnauticaNitrox/Nitrox/issues/2525
**📋 Problem Summary**

When all power cells are removed from the Cyclops (or under edge conditions), the HUD’s charge counter can show a negative percentage (e.g. “-5%”) instead of clamping to zero or indicating no power. This is confusing or misleading.

### **🛠 Solution Overview**

Intercept the HUD update logic when rendering the power percentage display.

Parse the numeric portion (strip the “%” sign), then check if the value is negative.

If negative, force it to show “0%” (or the zero-equivalent) instead.

Ensures no negative value is displayed even if internal logic briefly computes a negative number.

### **✅ Changes & Key Modifications**

Added a safe float.TryParse(...) wrapper around the string-to-number conversion from powerText.text.Replace("%", "").

Introduced a conditional check: value < 0 → value = 0 (clamp).

Updated the HUD text assignment to always produce a non-negative percentage.

### **🔍 Testing Done**

Removed all power cells from Cyclops, observed HUD displays “0%” instead of negative values.

Tested normal operation with power cells: continues to show correct positive percentages.

Verified that no new parsing errors or crashes occur when the HUD text is in unexpected formats.

### **📝 Notes for Reviewers / Considerations**

This patch does not alter the underlying power logic or distribution — only the UI display layer.

### DIFF
--- a/NitroxPatcher/Patches/Dynamic/CyclopsHelmHUDManager_Update_Patch.cs
+++ b/NitroxPatcher/Patches/Dynamic/CyclopsHelmHUDManager_Update_Patch.cs
@@ -17,6 +17,12 @@ public sealed partial class CyclopsHelmHUDManager_Update_Patch : NitroxPatch, ID
         {
             __instance.hudActive = true;
         }
+        if (__instance.hudActive && 
+            int.TryParse(__instance.powerText.text.Replace("%", ""), out int value) &&
+            value < 0)
+        {
+            __instance.powerText.text = "0%";
+        }
         if (__instance.subLiveMixin.IsAlive())
         {
             if (__instance.motorMode.engineOn)


Yorboot:
It doesn't cause any problems so thats why In my opinion it doesn't really mater if its just a patch. The core cause of this is that the devs don't do any checks before setting the value from what I could find. And I don't know of a way to alter this behavior hence why I haven chosen for this approach. (This is a base game subnautica bug I tested it)
image

dartasen:
Thank's for the work !
We'll consider another way of fixing this bug when we'll start working on energy sync for bases (if it needs fixes)

Closing for now

PR #2534, opened 2025-10-18
Hide server version mismatch message for compatible server versions

NinjaPedroX:
This PR hides the server version mismatch message in launcher when a user tries starting a server with a version that is greater than or equal to the latest compatible server version. This will allow for 1.8.0.0 servers to run without issues on 1.8.0.1 since no breaking changes to the save format have been made for this patch.

### DIFF
--- a/Nitrox.Launcher/Models/Services/ServerService.cs
+++ b/Nitrox.Launcher/Models/Services/ServerService.cs
@@ -137,9 +137,7 @@ internal sealed class ServerService : IMessageReceiver, INotifyPropertyChanged
 
     public async Task<bool> ConfirmServerVersionAsync(ServerEntry server)
     {
-        Version latestCompatibleVersion = new(1, 8, 0, 0); // Update this version with every release as necessary
-
-        if (server.Version >= latestCompatibleVersion)
+        if (server.Version >= new Version(NitroxEnvironment.Version.Major, NitroxEnvironment.Version.Minor))
         {
             return true;
         }


Nitrox.Launcher/Models/Services/ServerService.cs
```diff
@@ -136,13 +135,24 @@ public async Task<bool> StartServerAsync(ServerEntry server)
...
+    {
+        Version latestCompatibleVersion = new(1, 8, 0, 0); // Update this version with every release as necessary
+
+        if (server.Version >= latestCompatibleVersion)
+        {
+            return true;
+        }
+
+        // TODO: Add prompt for upgrade here if the version is incompatible but can be upgraded via upgrade script
+
```
dartasen, line 146: This doesn't seem right to me

I would rather limit the scope to major versions for now rather than hardcoding, and add a proper logic that keep tracks of upgrades at a later date

NinjaPedroX: This has been implemented

Measurity:
Not sure if we should do this. This change widens the burden of backwards compatibility beyond save file formats.

Server and client depend too strongly on each other. It'll be hard to know for sure there won't be bugs between slightly different versions. We'd have to add extra test sessions every release we allow (lower) version mixing (run older server on newer client and vice versa)

Besides that it is another dimension to check for with users in help-channel.

I think this PR is best left when Nitrox reaches 2.x, when we are more confident in maintaining backwards compatibility.

dartasen:
Closing in favor of #2544

PR #2536, opened 2025-10-18
Added log that will print other mods loaded with Nitrox

Measurity:
### game.log output
image

### DIFF
--- a/NitroxPatcher/Main.cs
+++ b/NitroxPatcher/Main.cs
@@ -132,7 +132,7 @@ public static class Main
             {
                 Log.Warn(otherModsSummary);
             }
-        });
+        }).ContinueWithHandleError();
         try
         {
             Patcher.Initialize();


NitroxPatcher/Helper/SupportHelper.cs
```diff
@@ -0,0 +1,73 @@
...
+    ///     Dependency names commonly used by Unity game mods.
+    /// </summary>
+    private static readonly string[] modDependencies = ["HarmonyX", "0Harmony", "Harmony", "MonoMod.RuntimeDetour"];
+
+    /// <summary>
+    ///     Logs a list of all other mods that are currently loaded.
+    /// </summary>
+    public static string? GetSummaryOfOtherMods()
+    {
+        Assembly[] otherMods = GetOtherMods().ToArray();
```
dartasen, line 22: Shouldn't we use a try catch since we're manipulating assembly stuff inside a fire and forget task ?

NitroxPatcher/Helper/SupportHelper.cs
```diff
@@ -0,0 +1,73 @@
...
+        StringBuilder sb = new();
+        sb.AppendLine("Nitrox has detect other mods! Please do not report Nitrox issues while mods are loaded!");
+        foreach (Assembly mod in otherMods)
+        {
+            AssemblyName modNameInfo = mod.GetName();
+            sb.Append(modNameInfo.Name)
+              .Append(" v")
+              .Append(modNameInfo.Version)
+              .Append(" path '")
+              .Append(mod.Location)
```
Jannify: Do we really want to log the full file path bc it can contain personal data?

NitroxPatcher/Helper/SupportHelper.cs
```diff
@@ -0,0 +1,73 @@
...
+    /// </summary>
+    public static string? GetSummaryOfOtherMods()
+    {
+        Assembly[] otherMods = GetOtherMods().ToArray();
+        if (otherMods.Length < 1)
+        {
+            return null;
+        }
+        StringBuilder sb = new();
+        sb.AppendLine("Nitrox has detect other mods! Please do not report Nitrox issues while mods are loaded!");
```
NinjaPedroX:
```suggestion
        sb.AppendLine("Nitrox has detected other mods! Please do not report Nitrox issues while mods are loaded!");
```
Minor spelling suggestion :)

PR #2544, opened 2025-10-20
Improve version mismatch message

dartasen:
Context: #2534
Context:
> Nitrox will always show a scary message to user when the current version of Nitrox is strictly different from the one specific in a world save

Alternative to #2534, until we'll able to figure out how to deal properly with save compatibility and warn the user only when needed.

|  Before:  | After: | 
|---|---|
| image | image  |

### DIFF
--- a/Nitrox.Launcher/Models/Services/ServerService.cs
+++ b/Nitrox.Launcher/Models/Services/ServerService.cs
@@ -140,7 +140,7 @@ internal sealed class ServerService : IMessageReceiver, INotifyPropertyChanged
         await dialogService.ShowAsync<DialogBoxViewModel>(model =>
         {
             model.Title = "Version Mismatch Detected";
-            model.Description = $"The version of '{server.Name}' is v{(server.Version != null ? server.Version.ToString() : "X.X.X.X")} and current Nitrox version is v{NitroxEnvironment.Version}{Environment.NewLine}{Environment.NewLine}It is advisable to find out about the latest changes and whether the backup is compatible.{Environment.NewLine}Note that it's always possible to use the world backup feature in case of failures.{Environment.NewLine}{Environment.NewLine}Would you still like to continue?";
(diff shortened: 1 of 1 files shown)


Nitrox.Launcher/Models/Services/ServerService.cs
```diff
@@ -141,7 +140,7 @@ public async Task<bool> ConfirmServerVersionAsync(ServerEntry server) =>
         await dialogService.ShowAsync<DialogBoxViewModel>(model =>
         {
             model.Title = "Version Mismatch Detected";
-            model.Description = $"The version of '{server.Name}' is v{(server.Version != null ? server.Version.ToString() : "X.X.X.X")}. It is highly recommended to NOT use this save file with Nitrox v{NitroxEnvironment.Version}. Would you still like to continue?";
+            model.Description = $"The version of '{server.Name}' is v{(server.Version != null ? server.Version.ToString() : "X.X.X.X")} and current Nitrox version is v{NitroxEnvironment.Version}{Environment.NewLine}{Environment.NewLine}It is advisable to find out about the latest changes and whether the backup is compatible.{Environment.NewLine}Note that it's always possible to use the world backup feature in case of failures.{Environment.NewLine}{Environment.NewLine}Would you still like to continue?";
```
Measurity:
```suggestion
            model.Description = $"The save '{server.Name}' is v{(server.Version != null ? server.Version.ToString() : "X.X.X.X")} but current Nitrox version is v{NitroxEnvironment.Version}{Environment.NewLine}{Environment.NewLine}It is advisable to find out about the latest changes and whether the backup is compatible.{Environment.NewLine}Note that it's always possible to use the world backup feature in case of failures.{Environment.NewLine}{Environment.NewLine}Would you still like to continue?";
```

Measurity (approved):

PR #2561, opened 2025-10-31
Fix tech lights being indestructible

tornac1234:
Bug report from Discord user "Matty"

```
Ahh no worries then. I did put a bug in for it with 100% repro rate steps. 
- Start a new server with access to Floodlights.
- Join the server.
- Equip the habitat tool and construct the Floodlight on a platform.
- Leave the server and then shut down the server.
- Start the server up again and re-join the server.
- Equip the habitat tool again and attempt to deconstruct the floodlight created in step 3.

Results: 
Players are unable to deconstruct the floodlight buildable. 

Repo rate: 
8/8. 100%.
```

The fix consists in forcing ``constructable.NotifyConstructedChanged(false)`` to happen when spawning a Constructable because they're in the constructed state by default.
The bug originates from TechLights having a placedByPlayer field which prevents players from destroying it if it's not true (and it's set to true by NotifyConstructedChanged).

### DIFF
--- a/NitroxClient/GameLogic/Spawning/Bases/ModuleEntitySpawner.cs
+++ b/NitroxClient/GameLogic/Spawning/Bases/ModuleEntitySpawner.cs
@@ -1,16 +1,15 @@
 using System.Collections;
 using System.Linq;
+using Nitrox.Model.DataStructures;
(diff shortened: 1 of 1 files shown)


Jannify (approved):
LGTM CW

dartasen (approved):

PR #2566, opened 2025-11-02
Basic IPV6 Support

oliver408i:
IPv6 connectivity is increasingly common (especially on modern home networks). This PR aims to add basic ipv6 support to Nitrox.
These changes let Nitrox host/join over IPv6 while preserving existing IPv4 behavior.

All changes were pretty much just in general networking, so I don't expect anything in game to have changed.
Some hardcoded IPv4 strings were changed to allow ipv6 to show up, but by default everything still shows IPv4. However, this allows IPv6-only networks (if somehow someone was using one of those) to function with nitrox.

A note:
LiteNetLib was still IPv4-only, helper utilities filtered to IPv4, and address handling broke when IPv6 literals or mapped addresses appeared (e.g., discovery showing `::ffff:`). This was fixed with changes to `NetHelper.cs`. These changes also added support for ipv6 to a few helper functions too.

Test: Open a server, try connecting to it with `::1` (IPv6 loopback). Confirm it works (it should also give you admin perms).

A note on authorship: Codex was used to review the code to see where code needed to be changed to use IPv6 and it also directly wrote a few of the helper functions in `NetHelper.cs` (mostly the IPv6 local ip detection). As this is pretty common code (not specific to Nitrox or Subnautica), I believe this is ok. I have reviewed and tested all changes.

A note on why add IPv6:
- Allows IPv6-only networks to use Nitrox
- Future proof (IPv6 connectivity continues to grow worldwide)
- Removes some hardcoded IPv4 logic
- No NAT on IPv6 (but firewalls still exist. In some situations, this can help a bit)
- It's nice to have I guess... There aren't any tradeoffs I can see.

IPv6 LAN discovery is NOT added in this PR, that could be future work.

### DIFF
--- a/Nitrox.Model/Extensions/IpAddressExtensions.cs
+++ b/Nitrox.Model/Extensions/IpAddressExtensions.cs
@@ -18,11 +18,11 @@ public static class IpAddressExtensions
         {
             return false;
         }
-        address = address.TryExtractMappedIPv4();
         if (IPAddress.IsLoopback(address))
         {
             return true;
         }
+        address = address.TryExtractMappedIPv4();
         foreach (NetworkInterface ni in NetHelper.GetInternetInterfaces())
         {
             foreach (UnicastIPAddressInformation ip in ni.GetIPProperties().UnicastAddresses)


Nitrox.Model/Helper/NetHelper.cs
```diff
@@ -56,17 +65,43 @@ public static IEnumerable<NetworkInterface> GetInternetInterfaces() =>
...
+                    {
+                        return lanIpCache = normalized;
+                    }
+                }
+                if (address.AddressFamily == AddressFamily.InterNetwork)
                 {
                     lock (lanIpLock)
                     {
-                        return lanIpCache = ip.Address;
+                        return lanIpCache = NormalizeAddress(address);
```
Coding-Hen: Rather than call the NormalizeAddress in all of these can we not do it first before the calls so it is normalized and then the address types checked? Saves having to call it at each step along the way

Nitrox.Model/Helper/NetHelper.cs
```diff
@@ -56,17 +65,43 @@ public static IEnumerable<NetworkInterface> GetInternetInterfaces() =>
...
                     }
                 }
+
+                if (ipv6Candidate == null
+                    && address.AddressFamily == AddressFamily.InterNetworkV6
```
Coding-Hen: You could also combine this if with the one above to reduce your code paths but it might be better to have them apart for readability

Nitrox.Model/Helper/NetHelper.cs
```diff
@@ -132,9 +167,20 @@ public static IEnumerable<NetworkInterface> GetInternetInterfaces() =>
 
             foreach (UnicastIPAddressInformation ip in ni.GetIPProperties().UnicastAddresses)
             {
-                if (ip.Address.AddressFamily == AddressFamily.InterNetwork)
+                IPAddress address = ip.Address;
+                if (address.AddressFamily is AddressFamily.InterNetwork or AddressFamily.InterNetworkV6)
                 {
-                    yield return (ip.Address, ni.Name.Replace("VPN", "").Trim());
+                    if (address.IsIPv4MappedToIPv6)
```
Coding-Hen: I believe if you restructure this as an if else and then have it drop out to the return below the ifs then you can move the normalise there and allow the inner works of that to mean you don't have to call it here

Nitrox.Model/Helper/NetHelper.cs
```diff
@@ -182,8 +228,14 @@ static long GetTotalInternetBytes() =>
     /// </summary>
     public static bool IsPrivate(this IPAddress address)
     {
+        address = NormalizeAddress(address);
```
Coding-Hen: I'm not sure what the benefit of doing this at a high scope is, can we not move this with the range method below as the other range method calls it itself anyway

Nitrox.Model/Serialization/ServerList.cs
```diff
@@ -153,6 +147,40 @@ public Entry(string name, IPAddress address, int port, bool persist = true) : th
 
             string name = parts[0].Trim();
             return new Entry(name, address, port);
+
+            static string ParseAddressWithOptionalPort(string value, out int parsedPort)
```
Coding-Hen: I'm not saying there is but I feel like there must be a better c# way of splitting an ipv6 and port off each other than having to do searches for indexes of : in the strings as well. I do know that ipv6 is not an easy thing to parse out with all the :: etc that can go on.

I assume we are then expecting people to put their ipv6 addresses in the server box in a particular format with the [ ] around it do we need to highlight this to them?

Nitrox.Model/Serialization/ServerList.cs
```diff
@@ -153,6 +147,40 @@ public Entry(string name, IPAddress address, int port, bool persist = true) : th
...
+                        if (!string.IsNullOrWhiteSpace(potentialPort))
+                        {
+                            int.TryParse(potentialPort, out parsedPort);
+                        }
+
+                        return trimmed.Substring(1, closingBracket - 1);
+                    }
+                }
+
+                int lastColonIndex = trimmed.LastIndexOf(':');
```
Coding-Hen: Do you not need to consider here what happens if someone doesn't have a port attached here, they put in say ipv6 without a port and no [] which will have the last index of :int and then will return the last octet as a port when it is actually part of the ip

Coding-Hen (changes requested):
Would be nice to have some ipv6 support but I have some comments on the code below

Nitrox.Model/Helper/NetHelper.cs
```diff
@@ -56,17 +65,43 @@ public static IEnumerable<NetworkInterface> GetInternetInterfaces() =>
...
         }
 
+        IPAddress? ipv6Candidate = null;
         foreach (NetworkInterface ni in GetInternetInterfaces())
         {
             foreach (UnicastIPAddressInformation ip in ni.GetIPProperties().UnicastAddresses)
             {
-                if (ip.Address.AddressFamily == AddressFamily.InterNetwork)
+                IPAddress address = ip.Address;
+                if (address.IsIPv4MappedToIPv6)
```
Coding-Hen: I think you either don't need this check here or you don't have it in the method you call to avoid doubling up the IsIPv4MappedToIPv6 call

Nitrox.Model/Helper/NetHelper.cs
```diff
@@ -56,17 +65,43 @@ public static IEnumerable<NetworkInterface> GetInternetInterfaces() =>
...
+                        return lanIpCache = NormalizeAddress(address);
                     }
                 }
+
+                if (ipv6Candidate == null
+                    && address.AddressFamily == AddressFamily.InterNetworkV6
+                    && !address.IsIPv6LinkLocal
+                    && !address.IsIPv6Multicast)
+                {
+                    ipv6Candidate = address;
```
Coding-Hen: Unless I have missed something why would you want to continue looping after assigning here rather than just returning early after assigning the cache

Nitrox.Model/Serialization/ServerList.cs
oliver408i: You only use the [] in urls and ssh (for example) when you need to specify the username or need to add `:port` after the ip. See [RFC3986](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.2) for more info.

I'll look into splitting the ipv6, but since the user is supposed to put the port in a different box in the ui anyway I don't think this is a big concern.

Nitrox.Model/Helper/NetHelper.cs
oliver408i: Changed in latest commit

Nitrox.Model/Serialization/ServerList.cs
oliver408i: I added a catch to prevent already correct address from going through the port parser, that should probably fix this problem.

Kiaos:
Did a play test and it works. Did the test on the test server.  Then we tested on my own server.

Kiaos:
After this is merged i have a update for server logs and what not here: https://github.com/Kiaos/Nitrox/compare/master...Kiaos:Nitrox:pr-2566 

you can check it out. this needs to be merged first.

Measurity (approved):

Nitrox.Model/Extensions/IpAddressExtensions.cs
```diff
@@ -0,0 +1,80 @@
...
+    public static bool IsPrivate(this IPAddress address)
+    {
+        address = address.TryGetAsIPv4();
+        switch (address.AddressFamily)
+        {
+            case AddressFamily.InterNetworkV6:
+            {
+                byte[] bytes = address.GetAddressBytes();
+                // Unique local addresses (fc00::/7)
+                if ((bytes[0] & 0xFE) == 0xFC)
```
Coding-Hen: Both the below checks do we want to ensure bytes is an appropriate length or will the address method call above always do this?

Nitrox.Model/Extensions/IpAddressExtensions.cs
```diff
@@ -0,0 +1,80 @@
...
+                    {
+                        return true;
+                    }
+                }
+                break;
+            }
+        }
+        return false;
+
+        static bool IsInRange(IPAddress ipAddress, string mask)
```
Coding-Hen: Is this a mask? I was under the assumption is the mask we are actually passing I'm here a mask? I believe for /24 say it would actually be 255.255.255.0 as the mask so it would be a little misleading

Nitrox.Model/Extensions/IpAddressExtensions.cs
```diff
@@ -0,0 +1,80 @@
...
+        {
+            string[] parts = mask.Split('/');
+            if (parts.Length != 2)
+            {
+                throw new ArgumentOutOfRangeException(nameof(mask), "Mask must contain 2 parts: IP and host bits");
+            }
+
+            int ipNum = BitConverter.ToInt32(ipAddress.GetAddressBytes(), 0);
+            int cidrAddress = BitConverter.ToInt32(IPAddress.Parse(parts[0]).GetAddressBytes(), 0);
+            int cidrMask = IPAddress.HostToNetworkOrder(-1 << (32 - int.Parse(parts[1])));
```
Coding-Hen: Is it worth maybe having a comment here to explain the bit shifting for clarity of those reading it

Nitrox.Server.Subnautica/Server.cs
```diff
@@ -281,22 +281,30 @@ private async Task LogHowToConnectAsync()
...
             options.Add("{ip:l} - Friends on another internet network (Port Forwarding)");
         }
         foreach ((IPAddress? vpnAddress, string? vpnName) in await vpnIps)
         {
-            options.Add($"{vpnAddress} - Friends using {vpnName} (VPN)");
+            if (vpnAddress == null)
+            {
+                continue;
+            }
+            IPAddress displayAddress = vpnAddress.TryGetAsIPv4();
```
Coding-Hen: Happy with this for now but do we want to always have it as ipv4? Say someone is expecting it to be ipv6 as they have entered that intentionally we would not display that here

NitroxClient/MonoBehaviours/Gui/MainMenu/ServersList/MainMenuCreateServerPanel.cs
```diff
@@ -113,6 +114,25 @@ IEnumerator DelayedScrollToNewEntry()
...
         }
+
+        static bool IsValidDnsOrIp(string value)
+        {
+            if (string.IsNullOrWhiteSpace(value))
+            {
+                return false;
+            }
+            // Valid IPv4 but port should be specified in designated field.
+            if (value.LastIndexOf(":", StringComparison.Ordinal) is var colonIndex and > 0 && IPAddress.TryParse(value.Substring(0, colonIndex), out _))
```
Coding-Hen, line 125: I assume it does but is "and" a special case here not come across that before, useful use case

Nitrox.Server.Subnautica/Server.cs
Measurity: This only happens for [special IPv6 addresses](https://en.wikipedia.org/wiki/IPv6_address#Special_addresses). Displaying IPv4 would be less redundant info to the user in this case.

Nitrox.Model/Extensions/IpAddressExtensions.cs
Measurity: [I improved the variable names](https://github.com/SubnauticaNitrox/Nitrox/pull/2566/files#diff-6804a160dc0c1eeb87301ad2d0af07c07c36adc31d283b24fc81a8d87b2a05acR143-R154). Should be clear what's going on now. The last `==` checks for equality of _only_ the network bits of the IP address. If they are equal, they are in same network.

PR #2567, opened 2025-11-03
Fixed unity object lifetime not being checked for IAnimParamReceiver

Measurity:


oliver408i:
Can confirm it works on my machine (fixes the previous problem)

Issue #2551, opened 2025-10-22
Add warning if mods are used when starting game through launcher


Measurity:
### Describe the issue

Would be best to add this feature to the launcher as the game can crash quite early due to mods.

dartasen:
Closed by #2569

PR #2569, opened 2025-11-04
feat(launcher): warn when BepInEx mods are detected before starting S…

mattwolf0:
…ubnautica

Detect .dlls under BepInEx/plugins and BepInEx/patchers and show a non-blocking warning before launching in both singleplayer and multiplayer flows.

Refs #2551

### DIFF
--- a/Nitrox.Launcher/Models/Utils/GameInspect.cs
+++ b/Nitrox.Launcher/Models/Utils/GameInspect.cs
@@ -64,36 +64,33 @@ internal static class GameInspect
 
     public static void WarnIfBepInExMods(string gameDir)
     {
-        if (GetBepInExModCount(gameDir) is var count and > 0)
+        if (string.IsNullOrWhiteSpace(gameDir) || !Directory.Exists(gameDir))
         {
-            Log.Warn($"BepInEx plugins detected: {count}");
-            LauncherNotifier.Warning($"BepInEx mod(s) were detected ({count}). Nitrox multiplayer does not support mods and they may cause instability.");
+            return;
+        }
+        string bepRoot = Path.Combine(gameDir, "BepInEx");
+        if (!Directory.Exists(bepRoot))
+        {
+            return;
         }
 
-        static int GetBepInExModCount(string gameDir)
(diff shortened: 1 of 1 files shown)


Nitrox.Launcher/ViewModels/LaunchGameViewModel.cs
```diff
@@ -292,4 +301,40 @@ private void UpdateGamePlatform()
         GamePlatform = NitroxUser.GamePlatform?.Platform ?? Platform.NONE;
         PlatformToolTip = GamePlatform.GetAttribute<DescriptionAttribute>()?.Description ?? "";
     }
+
+    // BepInEx mod detection
+    private static bool AreBepInExModsPresent(string gameDir, out int modCount)
+    {
+        modCount = 0;
+        if (string.IsNullOrWhiteSpace(gameDir) || !Directory.Exists(gameDir))
```
Coding-Hen: I understand why you have done this but I think for consistency could we have the if brackets here like we do in most of the code base

mattwolf0: Done, thanks!

Nitrox.Launcher/Models/Utils/GameInspect.cs
```diff
@@ -57,8 +58,43 @@ public static bool WarnIfGameProcessExists(GameInfo game)
...
+
+    public static void WarnIfBepInExMods(string gameDir)
+    {
+        if (GetBepInExModCount(gameDir) is var count and > 0)
+        {
+            Log.Warn($"BepInEx plugins detected: {count}");
+            LauncherNotifier.Warning($"BepInEx mod(s) were detected ({count}). Nitrox multiplayer does not support mods and they may cause instability.");
+        }
+
+        static int GetBepInExModCount(string gameDir)
```
Jannify: I don't like this usage of method in method. I prefer it when it encapsulates one algorithm inside a method especially when its used often. But here it's used once in a 3 liner. It makes the method more complicated to read as it needs to bc I need to have it on my mental stack the whole time in case it's used again or in a recursive way.

Measurity: Done

PR #2571, opened 2025-11-04
Change Enum.GetValues to allow other patches

### LINKED ISSUE #2531: Main Menu Breaks When Using any Submarine Mods

### Details about version

6a33dc7f8d3e0e24c01263ca116c1e7fcbf3e6d9

Windows

Steam

### Describe the issue.

Nitrox currently breaks the main menu (as seen in the screenshot) if any kind of submarine mod is present for example, the Hydra Shard mod.
I know the likely answer to this issue will be that Nitrox currently doesn’t support any mods, but before commit [#2411](https://github.com/SubnauticaNitrox/Nitrox/pull/2411), it worked fine. It didn’t work perfectly, but it did function and was even synced. The collision was weird and janky, but at least it worked.
After the mentioned commit, however, it broke every submarine mod. I’m fairly sure it’s that specific commit, because I tested it back when it was new. I didn’t report it at the time since I probably already knew what the answer would be.
Anyway, I just wanted to bring this up so you’re aware of the problem in case it comes up again in the future or maybe you’re able to fix it now.
I’ve attached the log from the Nitrox instance.

Image

1. Install a submarine mod with BepInEx
2. Launch the game with Nitrox

### Relevant log output

Line 70 in the attached file
Just so you know, I shortened the log because it keeps repeating the same line you see below.

[Nitrox.log](https://github.com/user-attachments/files/22980644/Nitrox.log)

killzoms:
With our current Enum.GetValues patch, we stop further execution of more patches thus causing mods to throw exceptions. This should fix that.

potentially Fixes #2531 

### Related issues
- https://github.com/SubnauticaModding/Nautilus/issues/644

### DIFF
--- a/Nitrox.Test/Patcher/Patches/PatchesTranspilerTest.cs
+++ b/Nitrox.Test/Patcher/Patches/PatchesTranspilerTest.cs
@@ -51,7 +51,7 @@ public class PatchesTranspilerTest
         [typeof(Flare_Update_Patch), 0],
         [typeof(FootstepSounds_OnStep_Patch), 6],
         [typeof(GrowingPlant_SpawnGrownModelAsync_Patch), -1],
-        [typeof(GameInputSystem_Patch), 2],
+        [typeof(GameInputSystem_Initialize_Patch), 2],
         [typeof(Player_TriggerInfectionRevealAsync_Patch), 1],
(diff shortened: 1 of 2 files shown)


Coding-Hen (approved):
LGTM

Issue #1811, opened 2022-05-25
Properly syncing Base Power 


Jannify:
We currently know about these bugs:

- [ ] Saved power is not synced (reset on rejoin)
- [ ] Chargers won't charge in creative mode
- [ ] Generated power only usable by builder of power source

Features we need:

- [ ] Simulation framework for bases and power management

Measurity:
I'm recording feedback from Discord here as well for whoever is going to implement this feature.

#### Feedback from @tornac1234  on implementation
> This will be really hard since sometimes items start the passive generation of a said amount of energy for a precise amount of time. And sometimes someone will enable the water filtration machine to function end then disable it some time after. There are way too many unpredictable ways to modify energy consumption of the base, and you don't necessarily know from where it's taken too

#### My thoughts on the feedback
>  For proper sync we need server to have the final say. If we just record when the player makes a change to power generation (providers/consumers). Then server can do the calculation no matter how short the time interval is between changes. Still leaves problem of desync with short intervals, but can be corrected with just a few power change packets because server knows the true state.
<sub>also for really fast switching of on/of states we can just ignore/merge that server side or throttle user action on client side, because actual gameplay impact will not be noticeable.</sub>
> 
> In cases where player can behave too radically (i.e. hard to rollback the client state) we can also have them wait on the server for confirmation first.
<sub>basically to ensure no power is used twice and improve accuracy.</sub>
> 
> If we do a naive implementation. For example, just send power change packets around constantly by whoever simulates the entity. Then it will be ripe for abuse and buggy and slow on network performance. But client will have immediate response like in singleplayer. 

#### @tornac1234 on current Nitrox implementation related to Subnautica power systems
> As of now, I'm doing everything so that the base energy is only used by the simulating player because he knows about everything that happens to a base (it's loaded at all times). Thus, remote clients only need to have an adequate power value by power generator so that when the simulating player leaves, one of them can take over from a valid point.
>
> Switching all of that to the server-side might require us to actually have the MonoBehaviours logic running there...

#### My reply
> True it works well enough for so far. But depends how high accuracy we want and if we can accept the network traffic for all the power changes.
> 
> Running monobehavior logic on server would suck indeed. It should run custom code to simulate the behavior we want, or not do it. For sure. Otherwise the server code will be extremely inflexible and crappy (processing/recording more information than needed). It would be better to change the game to fit our server.

PR #2575, opened 2025-11-05
Avoid negative Cyclops power percentage with no cells

mattwolf0:
Problem: Cyclops HUD can display a negative/NaN power percentage when no power cells are installed.
Change: In CyclopsHelmHUDManager_Update_Patch, guard against max==0, clamp the ratio, and update text/bar to 0%. Existing HUD behaviour is preserved.
Test: Spawn Cyclops → remove all 6 cells → HUD shows 0%; reinsert cells → percentage updates correctly.

Refs #2525


Measurity:
Thanks for working on the issue. 

I want to avoid patching UI text elements to hide (Subnautica bugs and) power issues with cyclops. We should first implement proper power sync and then redo cyclops power with it so values can't be negative anymore. See:
- #1811

Since above goes beyond the scope of this PR, I'll close it.

PR #2588, opened 2025-12-17
Sync Scanner room HUD by using position

Crapy:
Used last known position to remove and sync scanner room HUD when you pick up items, there is probably a better way to do it  but it's better than nothing 
tested and appears to work well


dartasen:
Thank you so much for contribution

While it may work, it does not fit the current roadmap/design we had for this base piece (which may one of the hardest)
if you wish to work on that feature with us we can discuss about it on Discord

PR #2594, opened 2025-12-19
Re-do the "restart" command

### LINKED ISSUE #2442: Re-do the "restart" command
### Describe the issue

The `restart` command will be removed with #2440 as it doesn't use the new world management system to restart the server, resulting in it creating and/or starting a server with the default "My World" name and making the launcher behave incorrectly.

This command should be remade to restart the server correctly and have the launcher handle it correctly. A couple of ideas for this include:

- Send an IPC message to the launcher asking it to restart the server using Nitrox.Launcher's ServerService.
- Utilize `ProcessEx.StartSelfCopyArgs` to restart the server.

misterbubb:
Fixes #2442

### DIFF
--- a/Nitrox.Launcher/Models/Design/ServerEntry.cs
+++ b/Nitrox.Launcher/Models/Design/ServerEntry.cs
@@ -412,7 +412,8 @@ internal sealed partial class ServerEntry : ObservableObject
                     {
                         "--save",
                         saveName,
-                        $"--game-path \"{NitroxUser.GamePath}\"",
+                        "--game-path",
+                        NitroxUser.GamePath
                     },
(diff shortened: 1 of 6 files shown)


Measurity:
Will review this after server refactor PR as it enables opportunities to better integrate this feature.

Measurity:
Think we'll go with a full restart instead of resetting server state. .NET IHostedService `StopAsync` is made with app shutdown in mind so we'd have to implemented a custom stop operation on all types that store state, which I don't think is worth it.

I have an implementation for this ready, but made for the new command system that's in the pipeline.

Measurity (approved):
PR is just like I would have written 🤡

dartasen (approved):

PR #2595, opened 2025-12-19
Add sync for StoryHandTarget

### LINKED ISSUE #2364: Add sync for StoryHandTarget

misterbubb:
Fixes #2364

- Sync PDA pickups via EntityDestroyed
- Sync PrecursorComputerTerminal used state
- Sync GenericConsole used state


Nitrox.Model.Subnautica/DataStructures/GameLogic/Entities/Metadata/GenericConsoleMetadata.cs
```diff
@@ -0,0 +1,28 @@
...
+using BinaryPack.Attributes;
+
+namespace Nitrox.Model.Subnautica.DataStructures.GameLogic.Entities.Metadata;
+
+[Serializable]
+[DataContract]
+public class GenericConsoleMetadata : EntityMetadata
+{
+    [DataMember(Order = 1)]
+    public bool GotUsed { get; }
```
dartasen, line 12: `GotUsed` but `Used` in PrecursorComputerTerminalMetadata, is it wanted ?

NitroxClient/GameLogic/Spawning/Metadata/Processor/GenericConsoleMetadataProcessor.cs
```diff
@@ -0,0 +1,17 @@
...
+namespace NitroxClient.GameLogic.Spawning.Metadata.Processor;
+
+public class GenericConsoleMetadataProcessor : EntityMetadataProcessor<GenericConsoleMetadata>
+{
+    public override void ProcessMetadata(GameObject gameObject, GenericConsoleMetadata metadata)
+    {
+        GenericConsole console = gameObject.GetComponent<GenericConsole>();
+        if (console && !console.gotUsed && metadata.GotUsed)
+        {
+            console.OnStoryHandTarget();
```
dartasen, line 16: Might wanna leave a quick comment to explain why OnStoryHandTarget is equivalent to deserialize field one by one, which is the case for GenericConsole but not true for every sync

Nitrox.Model.Subnautica/DataStructures/GameLogic/Entities/Metadata/GenericConsoleMetadata.cs
misterbubb: yes, the naming follows the original game field names.

dartasen (approved):

Nitrox.Model.Subnautica/DataStructures/GameLogic/Entities/Metadata/GenericConsoleMetadata.cs
Coding-Hen: I was thinking this is strange as well, I guess just another example of UWE not having consistency in the past code base. It does look strange but I checked as well before and does seem like the code in the classes with those

NitroxPatcher/Patches/Dynamic/StoryHandTarget_OnHandClick_Patch.cs
```diff
@@ -0,0 +1,37 @@
...
+            __instance.gameObject.TryGetNitroxId(out NitroxId destroyId))
+        {
+            Resolve<IPacketSender>().Send(new EntityDestroyed(destroyId));
+            return;
+        }
+
+        if (__instance.informGameObject)
+        {
+            if (__instance.informGameObject.TryGetComponent(out PrecursorComputerTerminal terminal) &&
+                __instance.informGameObject.TryGetNitroxId(out NitroxId terminalId))
```
Coding-Hen, line 26: I'm wondering if the extraction of the Nitrox I'd might be better moved to the if statement above (line23) and then just each if has the type extraction, then you don't have to have duplicated code, especially if in the future for some reason we decide to add more options here

dartasen (approved):
Works In Game

PR #2597, opened 2025-12-19
Implement proper sleep system

### LINKED ISSUE #2573: Sleeping on beds not passing time

1.8.0.1 (Latest release)

### Details about version

_No response_

Windows

Epic

### Describe the issue.

When sleeping on the beds, time does not pass for me (host) or my friends

Make a bed and sleep on it, then when you wake up the daylight/time is still the same

### Relevant log output

Game: [](https://pastebin.com/kua3eKE7
Server: [](https://pastebin.com/qAHReTbq 

https://github.com/user-attachments/assets/ef51ff22-957d-40b0-800a-a464f6480d24

misterbubb:
Fixes #2573


misterbubb:
still want to add more to this like waiting for all players to sleep and have prompts and such.

Nitrox.Model.Subnautica/Packets/SleepStatusUpdate.cs
```diff
@@ -0,0 +1,21 @@
...
+using Nitrox.Model.Packets;
+
+namespace Nitrox.Model.Subnautica.Packets;
+
+[Serializable]
+public class SleepStatusUpdate : Packet
+{
+    public int SleepingPlayers { get; }
+    public int TotalPlayers { get; }
+    public bool AllPlayersSleeping { get; }
```
Measurity: Could be computed property to reduce amount of data to send.
```cs
    public bool AllPlayersSleeping => SleepingPlayers >= TotalPlayers;
```

NitroxPatcher/Patches/Dynamic/Bed_Update_Patch.cs
```diff
@@ -0,0 +1,47 @@
...
+            {
+                HandReticle.main.SetText(HandReticle.TextType.Hand, "Get Up", true, GameInput.Button.Exit);
+                HandReticle.main.SetIcon(HandReticle.IconType.None, 1f);
+
+                if (Input.GetKeyDown(KeyCode.E))
+                {
+                    sleepManager.ExitBed();
+                }
+            }
+        }
```
Measurity: Use switch statement on `inUseMode`, move `currentPlayer` declaration closer to where it's used (inside `None` check.

NitroxPatcher/Patches/Dynamic/Bed_EnterInUseMode_Patch.cs
```diff
@@ -1,16 +1,31 @@
...
+        Bed.InUseMode currentMode = (Bed.InUseMode)inUseModeField.GetValue(__instance);
+        if (currentMode != Bed.InUseMode.None)
+        {
+            return false;
+        }
+
+        player.FreezeStats();
+        player.cinematicModeActive = true;
+        MainCameraControl.main.viewModel.localRotation = UnityEngine.Quaternion.identity;
+        inUseModeField.SetValue(__instance, Bed.InUseMode.Sleeping);
```
Measurity: Use direct field access: `__instance.inUseMode` and `__instance.inUseMode = Bed.InUseMode.Sleeping`

NitroxClient/Communication/Packets/Processors/SleepStatusUpdateProcessor.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        this.sleepManager = sleepManager;
+    }
+
+    public override void Process(SleepStatusUpdate packet)
+    {
+        if (packet.WasCancelled)
+        {
+            sleepManager.OnSleepCancelled();
+            ErrorMessage.AddMessage("Sleep was cancelled.");
```
Measurity: Use `Log.InGame("Sleep was cancelled.");`
similar for code below.

NitroxClient/GameLogic/SleepManager.cs
```diff
@@ -0,0 +1,105 @@
...
+        uGUI_PlayerSleep.main.StopSleepScreen();
+
+        currentBed = null;
+        isInBed = false;
+        isSleepInProgress = false;
+    }
+
+    public void OnSleepComplete()
+    {
+        if (currentBed != null && isInBed)
```
Measurity: Move `currentBed != null` check into property `IsInBed` and use that in if statements. The `isInBed` field could be removed too. Knowning that currentBed is not null has the same meaning.

NitroxPatcher/Patches/Dynamic/Bed_Update_Patch.cs
```diff
@@ -0,0 +1,47 @@
...
+using UnityEngine;
+
+namespace NitroxPatcher.Patches.Dynamic;
+
+public sealed partial class Bed_Update_Patch : NitroxPatch, IDynamicPatch
+{
+    public static readonly MethodInfo TARGET_METHOD = Reflect.Method((Bed t) => t.Update());
+    private static readonly FieldInfo inUseModeField = typeof(Bed).GetField("inUseMode", BindingFlags.Instance | BindingFlags.NonPublic);
+    private static readonly FieldInfo currentPlayerField = typeof(Bed).GetField("currentPlayer", BindingFlags.Instance | BindingFlags.NonPublic);
+    private static readonly MethodInfo subscribeMethod = typeof(Bed).GetMethod("Subscribe", BindingFlags.Instance | BindingFlags.NonPublic);
```
Measurity: Use direct field access, not reflection.

NitroxPatcher/Patches/Dynamic/Bed_Update_Patch.cs
```diff
@@ -0,0 +1,41 @@
...
+        switch (__instance.inUseMode)
+        {
+            case Bed.InUseMode.None:
+                if (__instance.currentPlayer != null)
+                {
+                    __instance.Subscribe(__instance.currentPlayer, false);
+                    __instance.currentPlayer = null;
+                }
+                break;
+
```
Measurity: The `None` case is same code as original game. We can `return true` by default and only `return false` for `Sleeping` to reuse game code and to show intent that we only handle what we need to.

NitroxClient/GameLogic/SleepManager.cs
```diff
@@ -0,0 +1,93 @@
...
+            {
+                currentBed.ExitInUseMode(player, false);
+            }
+        }
+
+        currentBed = null;
+        isSleepInProgress = false;
+    }
+
+    public bool IsInBed => currentBed != null;
```
Measurity: Can add following to this property so that code analysis knows `currentBed` is not null when this property is true.
```cs
[MemberNotNullWhen(true, nameof(currentBed))]
```

Measurity: Make sure your IDE is using the .editorconfig style. A code format should put the properties at the top.

misterbubb:
fixed. im pretty proud of this, looks and plays nice. (besides the animations ha)
Screenshot 2025-12-19 192400

misterbubb:
> Quite happy with the PR so far too. Did you see my other comment [#2597 (review)](https://github.com/SubnauticaNitrox/Nitrox/pull/2597#pullrequestreview-3600675712) ? I think these issues will be something people will experience in help channel so would be nice to fix them here, especially the first issue.

yes take a look now 🙂

Nitrox.Model.Subnautica/Packets/SleepStatusUpdate.cs
```diff
@@ -0,0 +1,24 @@
...
+[Serializable]
+public class SleepStatusUpdate : Packet
+{
+    public int SleepingPlayers { get; }
+    public int TotalPlayers { get; }
+
+    [IgnoredMember]
+    public bool AllPlayersSleeping => SleepingPlayers >= TotalPlayers;
+
+    public bool WasCancelled { get; }
```
Jannify: Minor:

```suggestion
    public bool WasCancelled { get; }

    [IgnoredMember]
    public bool AllPlayersSleeping => SleepingPlayers >= TotalPlayers;
```

Nitrox.Server.Subnautica/Models/Communication/NitroxServer.cs
```diff
@@ -53,6 +55,7 @@ protected void ClientDisconnected(INitroxConnection connection)
             return;
         }
 
+        sleepManager.PlayerDisconnected(player);
```
Jannify, line 58: I would move this to PlayerManager and add a subscribable event like `PlayerCountChanged` for disconnects.

Nitrox.Server.Subnautica/Models/GameLogic/SleepManager.cs
```diff
@@ -0,0 +1,126 @@
...
+namespace Nitrox.Server.Subnautica.Models.GameLogic;
+
+public class SleepManager
+{
+    private const float SLEEP_DURATION = 5f;
+    private const float SLEEP_TIME_SKIP = 396f;
+
+    private readonly PlayerManager playerManager;
+    private readonly TimeKeeper timeKeeper;
+    private readonly HashSet<string> sleepingPlayerIds = [];
```
Measurity: HashSet can store `ushort` as well for player id.

Jannify: Why is it string and not ushort?

Nitrox.Server.Subnautica/Models/GameLogic/SleepManager.cs
```diff
@@ -0,0 +1,126 @@
...
+using Nitrox.Model.Networking;
+using Nitrox.Model.Subnautica.Packets;
+using Timer = System.Timers.Timer;
+
+namespace Nitrox.Server.Subnautica.Models.GameLogic;
+
+public class SleepManager
+{
+    private const float SLEEP_DURATION = 5f;
+    private const float SLEEP_TIME_SKIP = 396f;
```
Jannify, line 12: Is this a constant from Subnautica? If so maybe add a comment from which place it comes.

NitroxPatcher/Patches/Dynamic/Bed_EnterInUseMode_Patch.cs
```diff
@@ -9,8 +9,21 @@ public sealed partial class Bed_EnterInUseMode_Patch : NitroxPatch, IDynamicPatc
...
+
+        player.FreezeStats();
+        player.cinematicModeActive = true;
+        MainCameraControl.main.viewModel.localRotation = UnityEngine.Quaternion.identity;
+        __instance.inUseMode = Bed.InUseMode.Sleeping;
+
         Resolve<IPacketSender>().Send(new BedEnter());
+        Resolve<SleepManager>().LocalPlayerEnteredBed(__instance);
+
+        return false;
```
Jannify, line 32: Is there a reason you copy the original method? If you want to remove lines we are using Transpilers to do it. They are a bit complicated, so ask in the discord for guidance.

NitroxPatcher/Patches/Dynamic/Bed_Update_Patch.cs
```diff
@@ -0,0 +1,33 @@
...
+using NitroxClient.GameLogic;
+using UnityEngine;
+
+namespace NitroxPatcher.Patches.Dynamic;
+
+public sealed partial class Bed_Update_Patch : NitroxPatch, IDynamicPatch
+{
+    public static readonly MethodInfo TARGET_METHOD = Reflect.Method((Bed t) => t.Update());
+
+    public static bool Prefix(Bed __instance)
```
Jannify, line 16: Same here

NitroxClient/Communication/Packets/Processors/SleepStatusUpdateProcessor.cs
```diff
@@ -0,0 +1,36 @@
...
+    {
+        this.sleepManager = sleepManager;
+    }
+
+    public override void Process(SleepStatusUpdate packet)
+    {
+        if (packet.WasCancelled)
+        {
+            sleepManager.OnSleepCancelled();
+            Log.InGame("Sleep was cancelled.");
```
Jannify: Should be a translated string. If you give me a key I can add it on our translation website.

NitroxClient/Communication/Packets/Processors/SleepStatusUpdateProcessor.cs
```diff
@@ -0,0 +1,36 @@
...
+            return;
+        }
+
+        if (packet.SleepingPlayers > 0)
+        {
+            Log.InGame($"{packet.SleepingPlayers}/{packet.TotalPlayers} players sleeping.");
```
Jannify: Same here with translated string.

NitroxClient/GameLogic/SleepManager.cs
```diff
@@ -0,0 +1,96 @@
...
+
+public class SleepManager
+{
+    private readonly IPacketSender packetSender;
+    private Bed? currentBed;
+    private bool isSleepInProgress;
+    private float timeLastSleepBeforeEntering;
+
+    [MemberNotNullWhen(true, nameof(currentBed))]
+    public bool IsInBed => currentBed != null;
```
Jannify: Monobehaviours should not be compared against null directly. Just use the overloaded bool conversion.

NitroxClient/GameLogic/SleepManager.cs
```diff
@@ -0,0 +1,96 @@
...
+    public SleepManager(IPacketSender packetSender)
+    {
+        this.packetSender = packetSender;
+    }
+
+    public void LocalPlayerEnteredBed(Bed bed)
+    {
+        currentBed = bed;
+        isSleepInProgress = false;
+        timeLastSleepBeforeEntering = Player.main != null ? Player.main.timeLastSleep : 0f;
```
Jannify: Same here with null comparison.

NitroxClient/GameLogic/SleepManager.cs
```diff
@@ -0,0 +1,96 @@
...
+
+    public void ExitBed()
+    {
+        if (!IsInBed || isSleepInProgress)
+        {
+            return;
+        }
+
+        Player player = Player.main;
+        if (player == null)
```
Jannify: Same here

Jannify: Also Player.main should not be null here right? Would break anything else in base SN.

NitroxClient/GameLogic/SleepManager.cs
```diff
@@ -0,0 +1,96 @@
...
+        packetSender.Send(new BedExit());
+
+        currentBed = null;
+    }
+
+    public void OnAllPlayersSleeping()
+    {
+        isSleepInProgress = true;
+        uGUI_PlayerSleep.main.StartSleepScreen();
+        DayNightCycle.main.SkipTime(396f, 5f);
```
Jannify: @tornac1234 Is a time skip outside of our TimeKeeper class here ok?

NitroxClient/GameLogic/SleepManager.cs
```diff
@@ -0,0 +1,96 @@
...
+    {
+        if (DayNightCycle.main.IsInSkipTimeMode())
+        {
+            DayNightCycle.main.StopSkipTimeMode();
+        }
+
+        if (IsInBed)
+        {
+            Player player = Player.main;
+            if (player != null)
```
Jannify: Same here

Nitrox.Server.Subnautica/Models/Communication/NitroxServer.cs
Coding-Hen: I agree this might be sensible as well. That way any future disconnect logic we can register without changes to this code

NitroxClient/GameLogic/SleepManager.cs
Measurity: Against null is fine as well. It's the `is null` kind that skips on the `==` overload that's the problem. Also problem are `?.`  and `??` operators.

If the analyzer is not complaining, then it's fine.

Nitrox.Server.Subnautica/Models/Communication/NitroxServer.cs
Measurity: We don't have a good way to dispose of the event. Recommend keeping it as is or use interface dispatch, but a server refactor to make this easier is still an open PR.

Jannify: But this is the server and not client. So we don't need to clean anything up, right?

Coding-Hen: This was my thoughts too it's server logic so having an event subscription here should be plausible

Measurity: I'm not a fan of delegates. They're not thread safe and have a lot of boilerplate. I prefer iterating interface implementations on a project or keeping it as is.

Mads Torgersen also dislikes events/delegates: https://youtu.be/T9UqIkuGnuo?si=VC2lO02aJuG2zrRp&t=4927 

Events are ancient .NET technology and we have better tools now like pub-sub messaging (like we do in Avalonia) or compile time interface binding source generators (see server refactor PR `IRedactor` or `ISummarize`). With these two tools you don't need to know where the delegate is, either send a pub-sub message or implement an interface. Which is way better for code discoverability.

Coding-Hen: Yeah okay understood. Pub sub is my go to in other languages so wouldn't be against that in this instance. Not used the interface approach before but if we don't like the event approach then I'm open to alternatives, my only thought would be insuring our overall code style so we don't have lots of different ways of doing things we should try and align around something

Nitrox.Server.Subnautica/Models/GameLogic/SleepManager.cs
misterbubb: yes, will do.

Nitrox.Server.Subnautica/Models/Communication/NitroxServer.cs
misterbubb: so whats it gonna be 🤔

Measurity: Keep it as is for now. After this PR and server refactor PR are merged we can look at best way forward.

NitroxPatcher/Patches/Dynamic/Bed_EnterInUseMode_Patch.cs
misterbubb: i do this so the blinking doesn't start until all players are sleeping. this prevents the prompt from being hidden and also just feels better to me.

NitroxClient/Communication/Packets/Processors/SleepStatusUpdateProcessor.cs
Coding-Hen: In light of a better replacement logic I am wondering if {0} and {1} might be better to avoid having to have named variables people will change with translations?

NitroxClient/GameLogic/SleepManager.cs
tornac1234: Any time skip must be orchestrated by the server only and then broadcasted to the players, thanks for noticing :)

Measurity: Resolved. All local time code removed from client SleepManager as it's handled by server.

Measurity (commented):
Not sure how to trigger the "Sleep Cancelled" behavior. It seems that _only_ one of the following can happen:
- Not sleep if not enough players
- Keeps going as normal even if a player leaves mid-sleep (which is good).

Could it be removed then?

misterbubb:
> Not sure how to trigger the "Sleep Cancelled" behavior. It seems that _only_ one of the following can happen:
> 
> * Not sleep if not enough players
> * Keeps going as normal even if a player leaves mid-sleep (which is good).
> 
> Could it be removed then?

yeah ill remove it lol

Measurity:
[rip my changes...](https://github.com/SubnauticaNitrox/Nitrox/pull/2597#discussion_r2641797465) please be careful with force pushing over other's work. sometimes instead of writing a review I commit stuff directly but redoing that is no fun for me.

Measurity (approved):
LGTM.

minor: It's possible to exit bed just when "all player sleep" procedure starts. You'll get a black screen that won't go away unless rejoining. I think using sync locks to request laying down/standing up would solve this but for now we can get this PR merged imo.

Coding-Hen (approved):
I think there are a few places somewhere if statements could be combined but it's personal style, this LGTMCW though.

If we are happy with the minor issue mentioned above then I see no reason for this to go in, it's not unrecoverable and a player shouldn't really be doing it.

Nitrox.Assets.Subnautica/LanguageFiles/en.json
```diff
@@ -85,6 +85,7 @@
     "Nitrox_ShowPing": "Show Ping",
     "Nitrox_SilenceChat": "Mute chat",
     "Nitrox_SilencedChatNotif": "Chat is now muted",
+    "Nitrox_SleepingPlayers": "{SLEEPING}/{TOTAL} players sleeping.",
```
Jannify: Because of huge merge conflicts in the past we only add translations directly on weblate and merge them in later. I added the string (https://hosted.weblate.org/translate/subnauticanitrox/subnautica/en/?checksum=7c2a2ba90f2c39f8). Can you now remove the changes to the file?

NitroxPatcher/Patches/Dynamic/Bed_Update_Patch.cs
```diff
@@ -0,0 +1,38 @@
...
+        // Let original code run for non-sleeping states
+        if (__instance.inUseMode != Bed.InUseMode.Sleeping)
+        {
+            return true;
+        }
+
+        SleepManager sleepManager = Resolve<SleepManager>();
+        if (sleepManager.CanExitBed)
+        {
+            HandReticle.main.SetText(HandReticle.TextType.Hand, "Get Up", true, GameInput.Button.Exit);
```
Jannify: This also needs to be translated.

NitroxPatcher/Patches/Dynamic/Bed_Update_Patch.cs
```diff
@@ -0,0 +1,38 @@
...
+            HandReticle.main.SetIcon(HandReticle.IconType.None, 1f);
+
+            if (Input.GetKeyDown(KeyCode.E))
```
Jannify: This should not be hard-coded. Instead use the binding from the "Exist/Holster" .

misterbubb:
@Measurity thanks 😊been a little busy here recently

PR #2598, opened 2025-12-20
Fix fabricator crafting sound not playing for remote players

### LINKED ISSUE #2434: Fabricator crafting sound doesn't seem to play
### Describe the issue

Context:  Regressed in #2424 

For remote players, the sound when someone is crafting something in the fabricator (base, cyclops, moonpool, upgrade station ...) does not play

misterbubb:
Fixes #2434

### DIFF
--- a/NitroxClient/GameLogic/Spawning/Metadata/Processor/CrafterMetadataProcessor.cs
+++ b/NitroxClient/GameLogic/Spawning/Metadata/Processor/CrafterMetadataProcessor.cs
@@ -74,19 +74,23 @@ public class CrafterMetadataProcessor : EntityMetadataProcessor<CrafterMetadata>
             return true;
         }
 
-        if (gameObject.TryGetComponent(out CrafterLogic crafterLogic))
+        if (!gameObject.TryGetComponent(out CrafterLogic crafterLogic))
         {
-            Base parentBase = gameObject.GetComponentInParent<Base>();
-            if (parentBase)
+            return false;
+        }
+
+        Base parentBase = gameObject.GetComponentInParent<Base>();
+        if (!parentBase)
+        {
+            return false;
+        }
+
+        foreach (GhostCrafter crafter in parentBase.GetComponentsInChildren<GhostCrafter>(true))
+        {
+            if (crafter._logic == crafterLogic)
             {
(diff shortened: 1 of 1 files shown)


NitroxClient/GameLogic/Spawning/Metadata/Processor/CrafterMetadataProcessor.cs
```diff
@@ -56,4 +58,38 @@ private static void SpawnItemInCrafter(CrafterLogic crafterLogic, CrafterMetadat
...
+        if (gameObject.TryGetComponentInChildren(out ghostCrafter, true))
+        {
+            return true;
+        }
+
+        if (gameObject.TryGetComponent(out CrafterLogic crafterLogic))
+        {
+            Base parentBase = gameObject.GetComponentInParent<Base>();
+            if (parentBase)
+            {
```
Jannify: Would use early return here for the two ifs to provide better readability.

NitroxClient/GameLogic/Spawning/Metadata/Processor/CrafterMetadataProcessor.cs
```diff
@@ -56,4 +58,38 @@ private static void SpawnItemInCrafter(CrafterLogic crafterLogic, CrafterMetadat
...
+                foreach (GhostCrafter crafter in parentBase.GetComponentsInChildren<GhostCrafter>(true))
```
Jannify: @tornac1234 Is there a better way to find the ghostcrafter in a base?

tornac1234: Supposedly there's a list of ghosts linked to a base (Base.ghosts) but I'm not sure if crafters are added to it

misterbubb: > Supposedly there's a list of ghosts linked to a base (Base.ghosts) but I'm not sure if crafters are added to it

checked Base.ghosts its a List<BaseGhost> for construction ghosts, not crafters GhostCrafter extends Crafter not BaseGhost so we need to use GetComponentsInChildren<GhostCrafter> here

PR #2599, opened 2025-12-21
Add auto-update feature to launcher

### LINKED ISSUE #2554: Allow upgrading to the latest version of Nitrox from the launcher
### Describe the issue

Currently, to update Nitrox, we need to be redirected from the launcher to GitHub for the download.

Is it possible for us to enable direct downloading from within the launcher and have it automatically install? This would be much more convenient.

misterbubb:
Fixes #2554 

image

### DIFF
--- a/Nitrox.Launcher/Models/Services/NitroxWebsiteApiService.cs
+++ b/Nitrox.Launcher/Models/Services/NitroxWebsiteApiService.cs
@@ -49,7 +49,7 @@ internal sealed class NitroxWebsiteApiService
             return null;
         }
 
-        return await httpFileService.GetFileStreamAsync(downloadInfo.DownloadUrl, cancellationToken);
+        return await httpFileService.GetFileAsync(downloadInfo.DownloadUrl, cancellationToken);
     }
 
     public sealed record ChangeLog
--- a/Nitrox.Launcher/Models/Services/NitroxWebsiteApiService.cs
+++ b/Nitrox.Launcher/Models/Services/NitroxWebsiteApiService.cs
@@ -1,4 +1,6 @@
 using System;
+using System.Collections.Generic;
+using System.IO;
 using System.Net.Http;
 using System.Net.Http.Json;
 using System.Text;
@@ -7,6 +9,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Nitrox.Launcher.Models.Attributes;
 using Nitrox.Launcher.Models.Design;
+using Nitrox.Model.Core;
 
 namespace Nitrox.Launcher.Models.Services;
 
@@ -35,6 +38,35 @@ internal sealed class NitroxWebsiteApiService
 
     public async Task<NitroxRelease?> GetNitroxLatestVersionAsync() => await httpClient.GetFromJsonAsync<NitroxRelease>("version/latest");
 
+    public async Task DownloadFileAsync(string url, string destinationPath, IProgress<(long bytesRead, long? totalBytes)>? progress = null, CancellationToken cancellationToken = default)
+    {
+        // Ensure we have an absolute URL
+        string absoluteUrl = url;
+        if (!Uri.IsWellFormedUriString(url, UriKind.Absolute))
+        {
+            absoluteUrl = $"https://nitrox.rux.gg/api/{url.TrimStart('/')}";
+        }
+
+        using HttpRequestMessage request = new(HttpMethod.Get, absoluteUrl);
+        using HttpResponseMessage response = await httpClient.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, cancellationToken);
+        response.EnsureSuccessStatusCode();
+
+        long? totalBytes = response.Content.Headers.ContentLength;
+        await using Stream contentStream = await response.Content.ReadAsStreamAsync(cancellationToken);
+        await using FileStream fileStream = new(destinationPath, FileMode.Create, FileAccess.Write, FileShare.None, 8192, true);
+
+        byte[] buffer = new byte[8192];
+        long totalBytesRead = 0;
+        int bytesRead;
+
+        while ((bytesRead = await contentStream.ReadAsync(buffer, cancellationToken)) > 0)
+        {
+            await fileStream.WriteAsync(buffer.AsMemory(0, bytesRead), cancellationToken);
+            totalBytesRead += bytesRead;
+            progress?.Report((totalBytesRead, totalBytes));
+        }
+    }
+
     public sealed record ChangeLog
     {
         /// <summary>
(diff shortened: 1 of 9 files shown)


Nitrox.Launcher/ViewModels/UpdatesViewModel.cs
```diff
@@ -94,6 +124,258 @@ await Dispatcher.UIThread.InvokeAsync(async () =>
...
+        // Check if running from a development build
+        // Dev builds have a "runtimes" folder that official releases don't have
+        string currentDir = NitroxUser.LauncherPath ?? AppDomain.CurrentDomain.BaseDirectory;
+        bool isDevBuild = Directory.Exists(Path.Combine(currentDir, "runtimes"));
+
+        if (isDevBuild)
+        {
+            LauncherNotifier.Info("Development build detected. Please use git pull to update your local repository.");
+            return;
+        }
```
Measurity: Can use `!NitroxEnvironment.IsReleaseMode` to remove need for file system access.

Nitrox.Launcher/Models/Services/NitroxWebsiteApiService.cs
```diff
@@ -52,22 +54,61 @@ public sealed record ChangeLog
...
-        [JsonPropertyName("url")]
-        public required string DownloadUrl { get; init; }
+        public ArchitectureInfo? GetCurrentPlatformDownload()
+        {
+            if (Platforms == null)
+            {
+                return null;
+            }
 
+            string platform = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? "windows" : "linux";
```
Measurity: I think getting platform name should be a property on `NitroxEnvironment` and return macos as well. Same for `arch` below.

Nitrox.Launcher/ViewModels/UpdatesViewModel.cs
```diff
@@ -94,6 +124,258 @@ await Dispatcher.UIThread.InvokeAsync(async () =>
...
+            await fileStream.WriteAsync(buffer.AsMemory(0, bytesRead), cancellationToken);
+            totalBytesRead += bytesRead;
+
+            if (totalBytes.HasValue)
+            {
+                DownloadProgress = (double)totalBytesRead / totalBytes.Value * 100;
+                DownloadStatus = $"Downloading... {totalBytesRead / 1024.0 / 1024.0:F1} / {totalBytes.Value / 1024.0 / 1024.0:F1} MB";
+            }
+        }
+    }
```
Measurity: This http code should be part of `NitroxWebsiteApiService`. (It already uses nitrox.rux.gg./api/ as base url and provides access to a `httpClient`).

Nitrox.Launcher/ViewModels/UpdatesViewModel.cs
```diff
@@ -94,6 +124,258 @@ await Dispatcher.UIThread.InvokeAsync(async () =>
...
+
+    private static async Task<string> ComputeMd5HashAsync(string filePath)
+    {
+        using MD5 md5 = MD5.Create();
+        await using FileStream stream = File.OpenRead(filePath);
+        byte[] hash = await md5.ComputeHashAsync(stream);
+        return BitConverter.ToString(hash).Replace("-", "").ToLowerInvariant();
+    }
```
Measurity: Should be part of the `Hashing` class.

Nitrox.Launcher/ViewModels/UpdatesViewModel.cs
```diff
@@ -94,6 +124,258 @@ await Dispatcher.UIThread.InvokeAsync(async () =>
...
+
+    private static string CreateUpdaterScript(string sourcePath, string destinationPath, string tempDir)
+    {
+        string scriptPath;
```
Measurity: Since we're doing `rm -f` should check that this path is rooted: `Path.IsPathRooted(destinationPath)` to prevent deleting files we don't want to if nitrox is installed in weird places and we mess up anywhere by using relative paths.

Nitrox.Launcher/ViewModels/UpdatesViewModel.cs
```diff
@@ -94,6 +124,258 @@ await Dispatcher.UIThread.InvokeAsync(async () =>
...
+
+    private bool CanDownloadUpdate() => !IsDownloading;
+
+    [RelayCommand(CanExecute = nameof(CanCancelDownload))]
+    private void CancelDownload()
+    {
+        downloadCancellationTokenSource?.Cancel();
+    }
+
+    private bool CanCancelDownload() => IsDownloading;
```
Measurity: Can you check the members of a type are ordered and grouped consistently (i.e. like rest of the code)?

```
all fields
all properties
all constructors
all methods
all inner types
all enums
```

Measurity (changes requested):

Measurity (approved):
Works well on Linux. Can someone test this on Windows?
___
Really like how snappy download & install is, good work again @misterbubb!

misterbubb:
> Works well on Linux. Can someone test this on Windows?
> ___
> Really like how snappy download & install is, good work again @misterbubb!

thanks alot!!

Coding-Hen:
Do we need to remove the old configs? Will this not have the unintentional effects of deleting custom configs people have setup? I am wondering if we should keep the configs and then out code should be able to handle as it currently does and set any new values over the top of it

Measurity:
Bump. Anyone available to test this on Windows?

Measurity:
@Kiaos has tested this on Windows, and it works :sunglasses:

We'll have more people test this PR before we release it. And might consider a backup/revert system in case installation goes bad (due to anti virus interfering or otherwise).

Coding-Hen (approved):
Tested on Windows 10, seems to work well. Patch notes and read me are a nice touch to download

Nitrox.Model/Helper/NitroxUser.cs
```diff
@@ -187,6 +187,10 @@ public static string? ExecutableFilePath
             {
                 path = new Uri(path).LocalPath;
             }
+            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                path = Path.GetFileNameWithoutExtension(path);
+            }
```
dartasen, line 193:
```suggestion
}
            else
            {
                path = Path.GetFileNameWithoutExtension(path);
            }
```

Measurity: Didn't do this because I think it would imply the `else` does something to handle the `if` condition, if it fails. But, the logic of both blocks is entirely unrelated, they just happen to check on platform. So I opted to keep the `if` separate.

Measurity:
We'll add a revert system in another PR. Will make issue for it with 1.9 deadline.

PR #2600, opened 2025-12-21
Fix Exosuit global sounds for walking and drilling

misterbubb:


NitroxClient/MonoBehaviours/Vehicles/ExosuitMovementReplicator.cs
```diff
@@ -133,6 +145,28 @@ private void SetupSound()
...
     }
 
+    private void UpdateDrillArmSoundVolume(IExosuitArm arm)
+    {
+        if (arm is not ExosuitDrillArm drillArm)
+        {
+            return;
+        }
+
+        float drillVolume = FMODSystem.CalculateVolume(transform.position, Player.main.transform.position, drillLoopSoundDistance, 1f);
```
Jannify: As `FMODSystem.CalculateVolume()` uses square root we should limit it's calculations if it's happening in the update loop. A solution would be to reuse the jetsound calculation (so do distance calculation manually and then do `SoundHelper.CalculateVolume(distance, radius, volume)` for jet and drill.
It's also possible that I am to picky and this would not be that performance critical, what do the others think about this?

dartasen: Fine for me if we can avoid it, probably won't make a difference for medium/high-end computer but still

misterbubb: good point

dartasen (approved):

Measurity (changes requested):
I managed to get infinite looping drill sound with this PR still.

1. Player 1 start drilling with Player 2 close by.
2. Player 2 swims away for 100 meters (still hear it)
3. Player 1 stops drilling
4. Player 2 stills hears looping sound and is now stuck even if restarting drills on Player 1.

PR #2601, opened 2025-12-22
Added "State to clipboard" button to GameObject debugger

Measurity:
## Purpose

Comparing two different instances of same game entities to more easily see differences. The data is sorted on name & type so it matches line-by-line.

## Example
[Prawnsuit State.txt](https://github.com/user-attachments/files/24294471/Prawnsuit.State.txt)

image

___
Feedback would be appreciated if this is something useful for debugging and/or if the format should be different. Eg changing to JSON for `jq` tool compatibility. Although for multi line values (like matrices) it's gonna look ugly in JSON...

### DIFF
--- a/Nitrox.Launcher/ViewModels/LaunchGameViewModel.cs
+++ b/Nitrox.Launcher/ViewModels/LaunchGameViewModel.cs
@@ -128,7 +128,7 @@ internal partial class LaunchGameViewModel(DialogService dialogService, ServerSe
                 {
                     const string PATCHER_DLL_NAME = "NitroxPatcher.dll";
 
-                    string patcherDllPath = Path.Combine(NitroxUser.ExecutableRootPath ?? "", "lib", "net472", PATCHER_DLL_NAME);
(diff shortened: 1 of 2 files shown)


Coding-Hen:
You have a if (str.Contains("\n")) and the followed by a replace for \r\n to be \n, would it not make sense to have this contains look for this as you might find \n but no occurrences of \r\n exist

Measurity:
> You have a if (str.Contains("\n")) and the followed by a replace for \r\n to be \n, would it not make sense to have this contains look for this as you might find \n but no occurrences of \r\n exist

If it's a `\n` in the value, I take it to mean a new line should be there. But `\r\n` is used on Windows based text, so I flatten to `\n` to ensure no left over `\r` characters. If I did the contains on `\r\n` I'll miss Unix based new lines.

Coding-Hen:
Understood, you are just trying to handle both cases. My mistake I assumed you were just trying to handle windows only

PR #2605, opened 2025-12-23
Disabled drill_loop and drill_hit_loop auto sync as it's handled manually

Measurity:
Previously the FMOD sync for these 2 looping sounds conflicted with the prawnsuit action sync, causing sound to never stop in some cases.
Also makes "volume by distance" work done in this PR function properly:
- https://github.com/SubnauticaNitrox/Nitrox/pull/2600

Only the slight rumble sound on resource node hit (drill_hit_loop) will not play for remote player. In future, we should sync this up manually through the remotely controlled prawnsuit, which would proper match visuals with sound too.

On client that controls prawn suit, drill audio is as expected.

### DIFF
--- a/NitroxClient/MonoBehaviours/Vehicles/ExosuitMovementReplicator.cs
+++ b/NitroxClient/MonoBehaviours/Vehicles/ExosuitMovementReplicator.cs
@@ -1,7 +1,5 @@
 using FMOD.Studio;
 using NitroxClient.GameLogic;
-using NitroxClient.GameLogic.FMOD;
-using NitroxClient.GameLogic.Helper;
 using Nitrox.Model.GameLogic.FMOD;
 using Nitrox.Model.Subnautica.Packets;
 using UnityEngine;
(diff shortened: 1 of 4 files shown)


NitroxClient/MonoBehaviours/Vehicles/ExosuitMovementReplicator.cs
```diff
@@ -12,7 +10,7 @@ public class ExosuitMovementReplicator : VehicleMovementReplicator
 {
     private const string DRILL_LOOP_SOUND_PATH = "event:/sub/exo/drill_loop";
 
-    private Exosuit exosuit;
+    private Exosuit exosuit = null!;
```
Jannify, line 13:
```suggestion
    private Exosuit? exosuit = null!;
```
or not?

Measurity: `exosuit` shouldn't be null anywhere in the code after `.Awake()` which is done first thing, like a C# constructor normally would. But C# nullability analyzer doesn't know about this so have to "help" with `null!` to tell analyzer this really will be "non-null" ASAP.

Changing to `Exosuit?` would indicate this can be null, even after "object init", which is not the intent.

dartasen (approved):

PR #2612, opened 2025-12-29
Show ipv6 in server console

oliver408i:
On #2566 , @Kiaos said we should show ipv6 in the server console after that pr is merged, and I totally forgot about that.
So that's what pr is for.

The format I'm using may be a bit opinionated, feel free to suggestion others.

## Output
image


Measurity (approved):

Coding-Hen (approved):
LGTM

PR #2613, opened 2025-12-29
Fix keybind translation

Jannify:
Changes key-bind translation string keys to better reflect their place and changed default key for `Nitrox_Settings_Keybind_FocusDiscord` to `i` (bc `f` is already used by SN).


Coding-Hen (commented):
LGTM

PR #2629, opened 2026-01-09
Show path to loaded DLL on unhandled type-related exceptions

### LINKED ISSUE #2556: Show path to loaded DLL on unhandled type-related exceptions
## Example of error that should show path
Image

misterbubb:
Fixes #2556
when a TypeLoadException occurs, it now logs the path and version of the loaded assembly to help diagnose version conflicts

## Testing

Can trigger a TypeLoadException for testing using this code and calling `new Error()` somewhere:

```cs
using System;
using System.Runtime.InteropServices;

namespace Nitrox.Server.Subnautica;

public class Error
{
    public Error()
    {
        foreach (var field in typeof(Container).GetFields())
        {
            Console.WriteLine($"{field.Name}: {field.FieldType}");
        }
    }

    private abstract class Abstract
    {
        public int a;
    }

    [StructLayout(LayoutKind.Sequential)]
    private sealed class TestClass : Abstract
    {
        public int x;
    }

    private sealed class Container
    {
        public TestClass tc;
    }
}
```

### DIFF
--- a/Nitrox.Server.Subnautica/Program.cs
+++ b/Nitrox.Server.Subnautica/Program.cs
@@ -489,6 +489,7 @@ public class Program
         if (e.ExceptionObject is Exception ex)
         {
             Log.Error(ex);
+            LogAssemblyPathFromException(ex);
         }
         if (!Environment.UserInteractive || Console.IsInputRedirected || Console.In == StreamReader.Null)
         {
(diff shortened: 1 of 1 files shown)


Nitrox.Server.Subnautica/Program.cs
```diff
@@ -510,4 +511,52 @@ private static void CurrentDomainOnUnhandledException(object sender, UnhandledEx
 
         Environment.Exit(1);
     }
+
+    /// <summary>
+    ///     Logs the file path of assemblies referenced in type-related exceptions to help diagnose version conflicts.
+    /// </summary>
+    private static void LogAssemblyPathFromException(Exception ex)
+    {
+        string assemblyName = ex switch
```
dartasen:
```suggestion
        string? assemblyName = ex switch
```

We're slowly updating the codebase to support NRT (annotations only for now)

Nitrox.Server.Subnautica/Program.cs
```diff
@@ -510,4 +511,52 @@ private static void CurrentDomainOnUnhandledException(object sender, UnhandledEx
...
+        };
+
+        if (string.IsNullOrWhiteSpace(assemblyName))
+        {
+            return;
+        }
+
+        // Extract just the assembly name without version info
+        string shortName = assemblyName.Split(',')[0].Trim();
+        Assembly loadedAssembly = AppDomain.CurrentDomain.GetAssemblies()
```
dartasen:
```suggestion
        Assembly? loadedAssembly = AppDomain.CurrentDomain.GetAssemblies()
```

Nitrox.Server.Subnautica/Program.cs
```diff
@@ -489,6 +489,7 @@ private static void CurrentDomainOnUnhandledException(object sender, UnhandledEx
         if (e.ExceptionObject is Exception ex)
         {
             Log.Error(ex);
+            LogAssemblyPathFromException(ex);
```
dartasen, line 492: So we would log two times the error ?

@Measurity Thoughts on that ?

Measurity: LogAssemblyPathFromException doesn't log the stacktrace so LGTM.

Measurity (approved):

Nitrox.Server.Subnautica/Program.cs
```diff
@@ -510,4 +511,56 @@ private static void CurrentDomainOnUnhandledException(object sender, UnhandledEx
...
+        }
+
+        try
+        {
+            // Extract just the assembly name without version info
+            AssemblyName name = new(assemblyName);
+            Assembly? loadedAssembly = AppDomain
+                                       .CurrentDomain
+                                       .GetAssemblies()
+                                       .FirstOrDefault(a => a.GetName().Name?.Equals(name.Name, StringComparison.OrdinalIgnoreCase) == true);
```
Coding-Hen: Can we not just compare Name to assemblyName and allow us to not have to create it to a type and then call .Name on it?

Measurity: The `FullName` of an assembly might not match with the loaded assembly (if type to load contains different metadata than any known type, it will still cause typeload but not match any loaded assembly).

For above reason, and to try be more accurate as you suggested, I changed the logic to sort by full name match first. This way we get at least some information that could be useful for debugging.

Coding-Hen (approved):

Issue #2625, opened 2026-01-08
Add backup & revert system for updating Nitrox in Launcher


Measurity:
## Things to check

- Works on both Windows and Linux
- Also backs-up save files and config
- To improve UX in future versions: provides a nice view in launcher to apply a backup.

PR #2630, opened 2026-01-10
Add backup and restore system for Nitrox updates

misterbubb:
Adds a backup and restore system for Nitrox updates, addressing #2625.

Features:
- Automatic backup creation before each update (includes installation files and save files)
- Backups section in the Updates view showing all available backups
- Restore functionality - creates a script that waits for launcher to close, extracts the backup, and restarts
- Delete backups directly from the UI
- Open backups folder button for manual access

Testing:
Windows - Backup creation, restore, and delete all work correctly
Linux - Needs testing (restore script uses unzip and bash, should work but untested)

Note: Linux support is implemented but has not been tested. The restore script uses standard Linux commands (unzip, pgrep, chmod) which should be available on most distributions, but verification on an actual Linux system would be appreciated.

image

### DIFF
--- a/Nitrox.Launcher/Models/Services/BackupService.cs
+++ b/Nitrox.Launcher/Models/Services/BackupService.cs
@@ -59,10 +59,7 @@ public class BackupService(IKeyValueStore keyValueStore)
             {
                 progress?.Report((50, "Backing up save files..."));
                 string savesDir = keyValueStore.GetSavesFolderDir();
-                if (Directory.Exists(savesDir))
-                {
-                    await AddDirectoryToArchiveAsync(archive, savesDir, "saves", cancellationToken);
-                }
+                await AddDirectoryToArchiveAsync(archive, savesDir, "saves", cancellationToken);
             }
 
             // Add metadata file


Nitrox.Launcher/Models/Services/BackupService.cs
```diff
@@ -0,0 +1,355 @@
...
+            progress?.Report((90, "Finalizing backup..."));
+            ZipArchiveEntry metadataEntry = archive.CreateEntry("backup_info.txt");
+            await using (StreamWriter writer = new(metadataEntry.Open()))
+            {
+                await writer.WriteLineAsync($"Nitrox Version: {NitroxEnvironment.Version}");
+                await writer.WriteLineAsync($"Backup Date: {DateTime.Now:F}");
+                await writer.WriteLineAsync($"Includes Saves: {includeSaves}");
+                await writer.WriteLineAsync($"Installation Path: {launcherPath}");
+                await writer.WriteLineAsync($"Saves Path: {keyValueStore.GetSavesFolderDir()}");
+            }
```
Measurity, line 79: I think going with JSON here would be easier to maintain.

Nitrox.Launcher/ViewModels/UpdatesViewModel.cs
```diff
@@ -312,4 +352,100 @@ private void CancelDownload()
...
+    }
+
+    [RelayCommand]
+    private void OpenBackupsFolder()
+    {
+        string backupsDir = BackupService.BackupsDirectory;
+        if (!Directory.Exists(backupsDir))
+        {
+            Directory.CreateDirectory(backupsDir);
+        }
```
Measurity: `Directory.CreateDirectory(backupsDir);` already checks if it exists. The if can be removed.

Nitrox.Launcher/Models/Services/BackupService.cs
```diff
@@ -0,0 +1,388 @@
...
+            // Backup installation files
+            progress?.Report((10, "Backing up installation files..."));
+            await AddDirectoryToArchiveAsync(archive, launcherPath, "installation", cancellationToken);
+
+            // Backup save files if requested
+            if (includeSaves)
+            {
+                progress?.Report((50, "Backing up save files..."));
+                string savesDir = keyValueStore.GetSavesFolderDir();
+                if (Directory.Exists(savesDir))
```
Coding-Hen: Do we want to do something here if it doesn't? Else this will silently fail and maybe not give the behaviour the person wants?

Nitrox.Launcher/Models/Services/BackupService.cs
```diff
@@ -0,0 +1,388 @@
...
+        string backupPath = Path.Combine(BackupsDirectory, $"{backupName}.zip");
+
+        try
+        {
+            Directory.CreateDirectory(BackupsDirectory);
+
+            string? launcherPath = NitroxUser.LauncherPath;
+            if (string.IsNullOrEmpty(launcherPath) || !Directory.Exists(launcherPath))
+            {
+                Log.Error("Cannot create backup: Launcher path not found");
```
Coding-Hen, line 44: Probably not for this pr but do we want to think about having the launcher allow for these to be multi language and we can pull from a dictionary like we do in the game. Just thinking we have a lot of logs here

Nitrox.Launcher/Models/Services/BackupService.cs
```diff
@@ -0,0 +1,388 @@
...
+            }
+            throw;
+        }
+        catch (Exception ex)
+        {
+            Log.Error(ex, "Failed to create backup");
+            // Clean up partial backup
+            if (File.Exists(backupPath))
+            {
+                File.Delete(backupPath);
```
Coding-Hen: Will this crash and not be handled if it can't delete?

Nitrox.Launcher/Models/Services/BackupService.cs
Measurity: Think we'll revisit this after moving to Microsoft `ILogger` API so we can more easily switch out with something  that does language translation.

PR #2631, opened 2026-01-10
Bump assembly version

dartasen:


Coding-Hen:
I will caveat this with unless we want to wait as we might decide to do another minor release but given the feedback on this release I don't think it likely

PR #2633, opened 2026-01-11
Improve chat window experience

### LINKED ISSUE #2632: Improve chat window experience
### Describe the issue

- Add slider to increase/decrease time that chat window stays open. At `0` the chat window should be off entirely (not show up when other people are chatting)
- Fix chat window not being movable to far corners if using widescreen aspect ratios.

misterbubb:
Closes #2632

- Add slider to control chat visibility duration (0-15 seconds)
- Setting to 0 disables chat completely
- Fix chat window not movable to far corners on widescreen monitors
- Use canvas RectTransform size instead of hardcoded resolution

## Keys to add in Weblate

- `Nitrox_Settings_ChatVisibilityDuration`
- `Nitrox_Settings_ChatVisibilityDuration_Tooltip`


Nitrox.Assets.Subnautica/LanguageFiles/en.json
```diff
@@ -87,8 +87,9 @@
     "Nitrox_Settings_OfflineClockSyncDuration": "Clock sync procedure duration (s)",
     "Nitrox_Settings_SafetyLatencyMargin": "Safety latency margin (ms)",
     "Nitrox_ShowPing": "Show Ping",
-    "Nitrox_SilenceChat": "Mute chat",
-    "Nitrox_SilencedChatNotif": "Chat is now muted",
+    "Nitrox_ChatVisibilityDuration": "Chat visibility duration (s)",
+    "Nitrox_ChatVisibilityDuration_Tooltip": "How long the chat stays visible. Set to 0 to disable chat popups.",
```
Coding-Hen: I do like this but do we still want a button to disable chat?

misterbubb: setting it to 0 should do that.

Coding-Hen: Yes indeed. I just meant if that was something we would rather over an explicit button which was disabled. Happy either way, 0 is fine with me

dartasen:
In Game Test : 
* When slider is at 0, and I open the chat : The chat will never disappear and there's no way to close it. We should be able to click outside so the chat can fade out

dartasen:
@Kiaos I remember you have a wide screen ? If you could please test that 🙏

NitroxClient/MonoBehaviours/Gui/Chat/PlayerChat.cs
```diff
@@ -52,7 +52,7 @@ public IEnumerator SetupChatComponents()
 
             yield return new WaitForEndOfFrame(); //Needed so Select() works on initialization
             IsReady = true;
-            if (NitroxPrefs.SilenceChat.Value)
+            if (NitroxPrefs.ChatVisibilityDuration.Value <= 0f)
             {
                 Log.InGame(Language.main.Get("Nitrox_SilencedChatNotif"));
```
Coding-Hen, line 58: I believe this might cause issues here, it's refering to a language line that you have removed in the file above. Maybe the reason that is failed in game?

Coding-Hen (changes requested):

Kiaos:
> @Kiaos I remember you have a wide screen ? If you could please test that 🙏

Yea Open a subchannel and invite to it and send me the build. I could build it myself but i haven't setup visual studio on windows completely yet.

misterbubb:
i mean i have a widescreen 2560x1080 and it works fine for me
working on "When slider is at 0, and I open the chat : The chat will never disappear and there's no way to close it. We should be able to click outside so the chat can fade out" right now

NitroxClient/MonoBehaviours/Gui/Chat/PlayerChatInputField.cs
```diff
@@ -68,12 +69,63 @@ public void OnDeselect(BaseEventData eventData)
...
         {
-            timeLeftUntilAutoClose = PlayerChat.CHAT_VISIBILITY_TIME_LENGTH;
-            FreezeTime = false;
+            float duration = NitroxPrefs.ChatVisibilityDuration.Value;
+            timeLeftUntilAutoClose = duration;
+            // If duration is 0 (chat disabled), freeze time so it doesn't immediately hide
+            FreezeTime = duration <= 0f;
         }
 
         private void Update()
```
Coding-Hen, line 78: It might be unavoidable but this seems a lot of checks to be doing on each frame update. If we do need to do it then if we can find something that will early return that would be good

Measurity:
for me, setting delay to 0 still opens chat when I press Y.

NitroxClient/MonoBehaviours/Gui/Chat/PlayerChatInputField.cs
```diff
@@ -219,5 +177,43 @@ private IEnumerator MoveToEndOfText()
...
+            {
+                // Clicked outside the chat area
+                if (selected)
+                {
+                    playerChatManager.DeselectChat();
+                }
+                playerChatManager.HideChat();
+                FreezeTime = true;
+            }
+        }
```
Measurity: Would move the `GetMouseButtonDown` inside this function and do a rect transform bounds check instead of raycast:

```suggestion
        private void HideIfClickOutsideChat()
        {
            if (!GetMouseButtonDown(0))
            {
                return;
            }
            if (RectTransformUtility.RectangleContainsScreenPoint(playerChatManager.PlayerChatTransform.GetComponent<RectTransform>(), mousePosition))
            {
                return;
            }

            // Clicked outside the chat area
            if (selected)
            {
                playerChatManager.DeselectChat();
            }
            playerChatManager.HideChat();
            FreezeTime = true;
        }
```

misterbubb:
> for me, setting delay to 0 still opens chat when I press Y.

it was discussed that it should be like
"When slider is at 0, and I open the chat : The chat will never disappear and there's no way to close it. We should be able to click outside so the chat can fade out"
should i just add an option to mute chat completely then?

NitroxClient/MonoBehaviours/Gui/Chat/PlayerChatInputField.cs
```diff
@@ -68,12 +69,21 @@ public void OnDeselect(BaseEventData eventData)
...
         private void Update()
         {
+            // When slider is at 0, chat won't auto-hide, so allow click-outside to close it
+            if (FreezeTime && UnityEngine.Input.GetMouseButtonDown(0) && PlayerChat.IsReady)
+            {
+                CheckClickOutsideChat();
+            }
+
+            // Early return if time is frozen (chat hidden or slider at 0)
             if (FreezeTime)
```
Measurity: Can be wrapped in an if to not check `FreezeTime` twice.

NinjaPedroX:
> > for me, setting delay to 0 still opens chat when I press Y.
> 
> it was discussed that it should be like
> "When slider is at 0, and I open the chat : The chat will never disappear and there's no way to close it. We should be able to click outside so the chat can fade out"
> should i just add an option to mute chat completely then?

I don't think that makes sense to do since setting the slider to 0 won't open chat unless you open it yourself. I don't see a point to adding an option to disable it completely when nothing else can open it.

Measurity:
> should i just add an option to mute chat completely then?

yes I think some people don't want to see chat at all. if they use steam chat instead, for example. Disabling chat when it's `0` is fine with me. No need for extra option though I suppose it's subjective what's best UX.

Measurity:
I'm fine with Ninja's suggestion. But then we need to ensure the chat won't receive focus (minimize annoyances) when opening chat. As you cannot send messages when it's at `0` anyway.

Basically go with what's easiest to implement and has good UX (similar to chat of other games).

Measurity:
I noticed that the window fade logic has a race condition when set to 1 second. Which will cause the chat to stop fading out on the other end.

also, could you add numpad-enter key support to submit messages?

misterbubb:
> I noticed that the window fade logic has a race condition when set to 1 second. Which will cause the chat to stop fading out on the other end.
> 
> also, could you add numpad-enter key support to submit messages?

I dont have a numpad so ill need you to try it lol

NitroxClient/MonoBehaviours/Gui/Chat/PlayerChatInputField.cs
```diff
@@ -85,9 +85,11 @@ private void Update()
 
             if (selected)
             {
+                bool enterPressed = UnityEngine.Input.GetKey(KeyCode.Return) || UnityEngine.Input.GetKey(KeyCode.KeypadEnter);
```
Measurity: `enterPressed` bool can be an if statement, wrapping the if below as both blocks require `enterPressed` to do stuff anyway.

Measurity (commented):
Minor thing and then LGTM

Measurity (approved):

Coding-Hen (approved):
LGTM CW

Nitrox.Assets.Subnautica/LanguageFiles/en.json
dartasen: Those should follow the same pattern as above Nitrox_Settings, and we should add them on weblate straight away

Measurity:
Removed the changes to en.json and handled review comments.

dartasen (approved):

PR #2635, opened 2026-01-11
Fix remote player equipment visibility initialization

### LINKED ISSUE #2359: Recheck player armors mapping
### Describe the issue

We noticed that some armor pieces like rebreather may not appear.

We should have a look and recheck in this code area if everything is alright, and while at it check if all armor pieces are correctly bind

misterbubb:
Closes #2359 

Initialize equipment visibility to empty when spawning remote players.

Previously, equipment visibility depended on what the local player had equipped when the body prototype was created, which could cause armor pieces like the rebreather to incorrectly appear or not appear on remote players.


dartasen:
Related to #2359

dartasen:
Not sure to understand the fix properly

misterbubb:
> Not sure to understand the fix properly

The fix ensures remote players start with a clean "no equipment" visibility state, rather than inheriting whatever the local player happened to be wearing when the body prototype was created.

dartasen:
In Game Test : 
Whole equipment visibility system seems to be flakey. Sometimes show pieces of equipment the player don't have, sometimes it shows nothing

PR #2641, opened 2026-01-11
Fix OpenDirectory returning false on Windows

### LINKED ISSUE #2636: UI shows incorrect “Directory does not exist” error, but folder opens successfully

### Details about version

_No response_

Windows

Steam

### Describe the issue.

even with a valid directory still pops up an error saying "Can't Open. Directory does not exist." upon clicking "Open"
seems to work just fine though.

Image

- Go to options inside the launcher
- at the Nitrox Program Data Location section click "Open"
- pops up and error saying "Can't Open. Directory does not exist."

### Relevant log output

_No response_

misterbubb:
Closes #2636 

on Windows, `Process.Start` with `UseShellExecute = true` returns `null` when opening directories because Explorer is already running and handles the request without spawning a new process. the fix is to just return `true` after the `Process.Start` call since we already validated the directory exists beforehand.

### DIFF
--- a/Nitrox.Model/Platforms/OS/Shared/ProcessEx.cs
+++ b/Nitrox.Model/Platforms/OS/Shared/ProcessEx.cs
@@ -188,7 +188,7 @@ public class ProcessEx : IDisposable
             Verb = "open",
             UseShellExecute = true
         });
-        return proc != null;
+        return true;
     }
 
     public static ProcessEx? GetFirstProcess(string procName, Func<ProcessEx, bool>? predicate = null)


Nitrox.Model/Platforms/OS/Shared/ProcessEx.cs
```diff
@@ -188,7 +188,7 @@ public static bool OpenDirectory(string? directory)
             Verb = "open",
             UseShellExecute = true
         });
-        return proc != null;
+        return true;
```
Coding-Hen, line 191: If we are assigning proc above and then not using it here do we need to assign it at all?

dartasen: `using Process? proc = Process.Start` will be lowered to something like :

```cs

Process proc;
try {
    proc = Process.Start()
}
finally
{
   proc.Dispose();
}
```

So yes it's neccesary

dartasen:
How does that behave on linux, is it still correct or we need to branch per operating system ?

Measurity (approved):
Works on Linux as well. LGTM.

Issue #2640, opened 2026-01-11
Investigate sync for PrecursorGun


dartasen:
### Describe the issue

Currently, we should have things that are synced or partially synced for precursors : 

* Teleporters opening
* Door opening
* Terminal sync : will be done by #2364 

Investigate if there's anything else to sync.

Maybe some info that can be useful that we saw during 1.8 test sessions : 
Currently PrecursorGun seems broken (animations are weird) and the gun deactivation isn't synced across players (causing problems for rocket lift-off).

misterbubb:
- Gun terminal interaction - no patch exists yet #2595 (Add sync for StoryHandTarget)
- Gun visual state - story goal syncs but PrecursorGunStoryEvents visuals may not update for other players
will take a peek

misterbubb:
> ### Describe the issue
> Currently, we should have things that are synced or partially synced for precursors :
> 
> * Teleporters opening
> * Door opening
> * Terminal sync : will be done by [Add sync for StoryHandTarget #2364](https://github.com/SubnauticaNitrox/Nitrox/issues/2364)
> 
> Investigate if there's anything else to sync.
> 
> Maybe some info that can be useful that we saw during 1.8 test sessions : Currently PrecursorGun seems broken (animations are weird) and the gun deactivation isn't synced across players (causing problems for rocket lift-off).

could you provide more details about "PrecursorGun seems broken (animations are weird)"?

dartasen:
@misterbubb During 1.8 playtests, we did notice that the gun was "drinking in the water" during the sunbeam event for all players.

That could come from the fact that some internal states of the gun aren't synced : isUsed, missing some story goals, ... that may cause the gun to misbehave.

PR #2647, opened 2026-01-15
Sync PrecursorDisableGunTerminal firstUse state

misterbubb:
Related to #2640

Adds metadata sync for the Precursor Gun disable terminal's firstUse flag.

Also adds handling for entities that already exist before the entity spawning phase (e.g., PlayerEntity set up by earlier sync processors), applying metadata to existing objects instead of trying to respawn.

## Testing

1. Teleport to:
```
warp 448 -91 1168
```
2. Then teleport to:
```
warp 363 -70 1081
```
3. Player A: Use gun disable terminal
4. Player B: Use gun disable terminal, see shorter animation is playing on both ends (animation of terminal arm & syringe not synced yet)

### DIFF
--- a/NitroxClient/GameLogic/Entities.cs
+++ b/NitroxClient/GameLogic/Entities.cs
@@ -189,6 +189,17 @@ namespace NitroxClient.GameLogic
                     UpdateEntity(entity);
                     continue;
                 }
+                // World entities can be streamed by the game before Nitrox spawns them.
+                // In this case, we just need to apply metadata and mark as spawned.
+                else if (entity is WorldEntity && NitroxEntity.TryGetObjectFrom(entity.Id, out GameObject existingObject))
+                {
+                    Log.Debug($"[Entities] World entity already exists, applying metadata: {entity.Id} ({entity.GetType().Name})");
+                    entityMetadataManager.ApplyMetadata(existingObject, entity.Metadata);
+                    simulationOwnership.ApplyNewerSimulation(entity.Id);
+                    MarkAsSpawned(entity);
+                    batch.AddRange(entity.ChildEntities);
+                    continue;
+                }
                 else if (entity.ParentId != null && !IsParentReady(entity.ParentId))
                 {
                     AddPendingParentEntity(entity);
--- a/NitroxClient/GameLogic/Entities.cs
+++ b/NitroxClient/GameLogic/Entities.cs
@@ -189,17 +189,6 @@ namespace NitroxClient.GameLogic
                     UpdateEntity(entity);
                     continue;
                 }
-                // Some entities (like PlayerEntity) may already exist because they were set up by earlier
-                // sync processors before the entity spawning phase. In this case, just apply metadata and mark as spawned.
-                else if (entity is WorldEntity && NitroxEntity.TryGetObjectFrom(entity.Id, out GameObject existingObject))
-                {
-                    Log.Debug($"[Entities] Entity already exists, applying metadata: {entity.Id} ({entity.GetType().Name})");
-                    entityMetadataManager.ApplyMetadata(existingObject, entity.Metadata);
-                    simulationOwnership.ApplyNewerSimulation(entity.Id);
-                    MarkAsSpawned(entity);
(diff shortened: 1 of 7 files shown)


dartasen:
Related to #2640

dartasen:
> Also adds handling for world entities that are already streamed by the game before Nitrox spawns them, applying metadata to existing objects instead of trying to respawn.

Which entities are streamed before Nitrox spawns them ? 👀

NitroxClient/GameLogic/Entities.cs
```diff
@@ -189,6 +189,17 @@ public IEnumerator SpawnBatchAsync(List<Entity> batch, bool forceRespawn = false
                     UpdateEntity(entity);
                     continue;
                 }
+                // World entities can be streamed by the game before Nitrox spawns them.
```
killzoms: Except the only possible Entities that could be streamed by the game shouldnt exist in the first place or this is the result of a random NitroxId collision(not likely but plausible)

My reason for saying they shouldnt exist in the first place is that we intercept the games attempt to load any files that contain entities Nitrox would care about.

Plus NitroxIds shouldnt be applied to GameObjects before they are spawned and recognized by the server, if you know of a spot where something like this is happening that would be an ideal spot to check

NitroxPatcher/Patches/Dynamic/PrecursorDisableGunTerminal_OnPlayerCinematicModeEnd_Patch.cs
```diff
@@ -0,0 +1,31 @@
...
+        // The terminal component is on a child object, so we need to look up the hierarchy for NitroxEntity
+        if (__instance.TryGetComponentInParent(out NitroxEntity entity, true))
+        {
+            Log.Debug($"[PrecursorDisableGunTerminal] Broadcasting metadata update: firstUse={__instance.firstUse}, id={entity.Id}");
+            Resolve<Entities>().BroadcastMetadataUpdate(entity.Id, new PrecursorDisableGunTerminalMetadata(__instance.firstUse));
+        }
+        else
+        {
+            Log.Warn($"[PrecursorDisableGunTerminal] No NitroxEntity found in hierarchy for {__instance.gameObject.GetFullHierarchyPath()}");
+        }
```
Measurity: Invert if to early log & return.

NitroxClient/GameLogic/Spawning/Metadata/Processor/PrecursorDisableGunTerminalMetadataProcessor.cs
```diff
@@ -0,0 +1,27 @@
...
+        
+        if (terminal)
+        {
+            Log.Debug($"[PrecursorDisableGunTerminalMetadataProcessor] Applying metadata: firstUse={metadata.FirstUse} to {terminal.gameObject.name}");
+            terminal.firstUse = metadata.FirstUse;
+        }
+        else
+        {
+            Log.Warn($"[PrecursorDisableGunTerminalMetadataProcessor] No PrecursorDisableGunTerminal component found on {gameObject.name} or its children");
+        }
```
Measurity: Invert if to early log & return.

Measurity (commented):
only some nits

Coding-Hen (commented):
Is there any code we have had previously we need to remove now to not oversync?

misterbubb:
> Is there any code we have had previously we need to remove now to not oversync?

there's no prior sync code for PrecursorDisableGunTerminal.firstUse that needs to be removed. The story goal system handles the gun-disabled story event, but that's a different concern it tracks whether the gun has been disabled as a narrative milestone, while this PR syncs the firstUse flag which only controls the terminal's animation behavior (first interaction plays a longer cinematic). They don't overlap.

Jannify (approved):
LGTM CW

Coding-Hen (approved):
LGTM

Measurity (approved):
LGTM

works in game - animation of terminal arm / syringe not synced yet but this PR does sync state so that shorter "animation" is playing.

dartasen (approved):

PR #2661, opened 2026-01-30
Enabled CRC for packets and use native sockets (latter only server-side)

Measurity:
Should protect against data corruption (like when using unstable VPN connections)


dartasen (approved):

dartasen:
I guess it's fine to keep it in DEBUG as well the overhead is minimal, but fine to me

Coding-Hen (approved):
LGTM

PR #2663, opened 2026-02-05
Nautilus compatibility for GameInput keybinds

JackNytely:
#### Problem

When Nitrox and Nautilus run together, Nitrox buttons may not get `InputAction` entries. That can cause `RegisterKeybindsActions` to fail because `gameInputSystem.actions[NitroxButton]` does not exist.

When Nitrox runs **without** Nautilus, the game was showing duplicate keybinds in the input settings ("Open chat" and "Focus on the Discord invite window" twice) because Nitrox buttons were added in two places: `GameInput.AllActions` and `Enum.GetValues(GameInput.Button)`. The game builds the binding list from both, so each Nitrox keybind appeared twice.

#### Solution

- **Extend `GameInput.AllActions`**: In the `Initialize` prefix, extend `AllActions` with Nitrox buttons so the game builds the action map for them. A duplicate check prevents adding them again when `AllActions` was already initialized from the patched `Enum.GetValues`.
- **Cleanup on deinitialize**: Add a prefix on `GameInputSystem.Deinitialize()` to restore the original `AllActions` array.
- **Fallback for missing actions**: If the game doesn't create an action (e.g. some builds don't use `AllActions`), `RegisterKeybindsActions` creates and registers `InputAction`s for Nitrox buttons in the same way as Nautilus's `InitializePostfix`.
- **Duplicate-safe enum extension**: `Enum_GetValues_Patch` still extends `Enum.GetValues(GameInput.Button)` with Nitrox buttons (required for the rebinding UI to recognize them as valid), but now checks for duplicates so Nautilus or repeated calls don't add them twice. Combined with the `AllActions` duplicate check, each keybind appears once in the settings UI.

#### Changes

- **GameInputSystem_Initialize_Patch**:
  - Add transpiler/patch logic for `Deinitialize` to restore `AllActions`.
  - Add duplicate check before extending `AllActions` (prevents double entries when already initialized from patched `Enum.GetValues`).
  - Add `TryGetValue` check in `RegisterKeybindsActions` with manual `InputAction` creation when an entry is missing.
  - Wire Nitrox actions to `gameInputSystem.OnActionStarted`, add bindings, and enable them to mirror Nautilus behavior.
  - Register `"Option{buttonId}"` localization entries in `Language.main` so the binding conflict dialog shows translated labels (e.g. "Open chat") instead of "Option1000".
- **Enum_GetValues_Patch**: Add duplicate check so Nitrox buttons are only appended when not already present (Nautilus compatibility).
- **uGUI_TabbedControlsPanel_AddBindingOption_Patch**: Fix infinite recursion that prevented Nitrox keybinds from being rebindable. The old code checked button ID, translated the label, then recursively called `AddBindingOption` — which triggered the prefix again on the same button, recursing infinitely. Changed to a `void` prefix that modifies the `label` parameter by reference, so the original method runs exactly once with the correct display text.
- **ChatKeyBindingAction / DiscordFocusBindingAction**: Add default controller bindings (`dpad/up` and `dpad/down`) so the game creates proper controller binding slots during initialization, enabling controller rebinding.

---

#### Problem

When stopping a server through the launcher, the shutdown could hang indefinitely if the server process didn't respond to the `quit` command. Additionally, embedded servers were left running when the launcher was closed.

#### Solution

Add timeout-based shutdown with force-kill fallback, and stop all embedded servers on launcher dispose.

#### Changes

- **ServerEntry.cs**: Add a 10-second `CancellationTokenSource` timeout to the shutdown loop. If the server doesn't respond to `quit` within the timeout, force-terminate the process. Also handle `OperationCanceledException` in `StopAsync` by force-killing the process when the graceful shutdown times out.
- **ServerService.cs**: On `Dispose`, iterate all online embedded servers and call `StopAsync`, waiting up to 30 seconds for all to finish. Non-embedded servers (with their own console window) are left running.

---

#### Problem

In Release, dependencies were **moved** into `lib/` (Debug only **copied**). Some code paths load assemblies before or outside the custom `AssemblyResolve` (e.g. Release vs Debug), so when DLLs existed only in `lib/`, the launcher could fail to load them. That led to empty server list, missing types, or other assembly-load issues in Release only.

#### Solution

Use **Copy** for both Debug and Release instead of Move in Release, so dependencies stay in the output root as well as in `lib/`.

#### Changes

- **Nitrox.Shared.targets**: Always copy `FilesToMove` into `lib/` (remove the Release-only Move). Update comment to explain that keeping DLLs in the output root avoids empty UI / missing types when they would otherwise be only in `lib/`.

---

### 4. Minor improvements

- **PrefabPlaceholderGroupsResource.cs**: Combine the cache version check into the `if (cache.HasValue)` condition so an outdated cache falls through to the rebuild path directly. Log message now distinguishes between outdated cache rebuild and fresh build.
- **Reflect.cs**: Add nullable annotation to `implementingType` parameter.

---

### Testing

**Keybinds**

- [x] Nitrox alone — no duplicate keybinds, rebinding works for both keyboard and controller
- [x] Nitrox + Nautilus * Nitrox bindings cant be rebound with nautilus enabled *
- [x] Nitrox + Nautilus + (Epic Weather Mod + Prototype Sub + Deathrun Remade) * Nitrox bindings cant be rebound with nautilus enabled *
- [x] Binding conflict dialog shows correct translated labels

**Server shutdown**

- [x] Embedded server stops gracefully via quit command
- [x] Embedded server is force-killed after timeout
- [x] Embedded servers stop when launcher is closed

**Release build**

- [x] Release build runs without empty server list or assembly-load errors
- [x] Server list displays correctly; create server works

### DIFF
--- a/Nitrox.Launcher/Models/Design/ServerEntry.cs
+++ b/Nitrox.Launcher/Models/Design/ServerEntry.cs
@@ -322,7 +322,21 @@ internal sealed partial class ServerEntry : ObservableObject
         }
         catch (OperationCanceledException)
         {
-            // ignored
+            // Graceful shutdown timed out, force kill the process
+            Log.Warn($"Server process {LastProcessId} did not stop gracefully within timeout, force killing...");
+            try
+            {
+                using ProcessEx? proc = ProcessEx.GetFirstProcess(GetServerExeName(), ex => ex.Id == LastProcessId);
+                if (proc != null)
+                {
+                    proc.Terminate();
+                    Log.Info($"Server process {LastProcessId} forcefully terminated");
+                }
+            }
+            catch (Exception ex)
+            {
+                Log.Error(ex, $"Failed to force kill server process {LastProcessId}");
+            }
         }
     }
 
@@ -359,17 +373,33 @@ internal sealed partial class ServerEntry : ObservableObject
                 await Dispatcher.UIThread.InvokeAsync(() => IsServerClosing = true);
                 await Dispatcher.UIThread.InvokeAsync(async () =>
                 {
-                    while (ProcessEx.ProcessExists(GetServerExeName(), ex => ex.Id == LastProcessId))
+                    using CancellationTokenSource shutdownCts = new(TimeSpan.FromSeconds(10));
+                    try
                     {
-                        try
+                        while (ProcessEx.ProcessExists(GetServerExeName(), ex => ex.Id == LastProcessId))
                         {
-                            await CommandQueue.Writer.WriteAsync("quit");
-                            CommandQueue.Writer.TryComplete();
+                            try
+                            {
+                                await CommandQueue.Writer.WriteAsync("quit", shutdownCts.Token);
+                                CommandQueue.Writer.TryComplete();
+                            }
+                            catch (ChannelClosedException)
+                            {
+                                await Task.Delay(500, shutdownCts.Token);
+                            }
+                            catch (OperationCanceledException)
+                            {
+                                // Timeout reached, force terminate
+                                Log.Warn($"Server {Name} did not respond to quit command, attempting force terminate");
+                                using ProcessEx? proc = ProcessEx.GetFirstProcess(GetServerExeName(), ex => ex.Id == LastProcessId);
+                                proc?.Terminate();
+                                break;
+                            }
                         }
-                        catch (ChannelClosedException)
-                        {
-                            await Task.Delay(500);
-                        }
+                    }
+                    catch (Exception ex)
+                    {
+                        Log.Error(ex, $"Error during server shutdown for {Name}");
                     }
                     CommandQueue = Channel.CreateUnbounded<string>();
                     IsOnline = false;
--- a/Nitrox.Launcher/Models/Services/ServerService.cs
+++ b/Nitrox.Launcher/Models/Services/ServerService.cs
@@ -247,6 +247,41 @@ internal sealed class ServerService : IMessageReceiver, INotifyPropertyChanged
         serverRefreshCts.Dispose();
         WeakReferenceMessenger.Default.UnregisterAll(this);
         watcher?.Dispose();
+        
+        // Stop all embedded servers when launcher closes
+        // Non-embedded servers are left running as they have their own console window
+        lock (serversLock)
+        {
+            List<Task> stopTasks = [];
+            foreach (ServerEntry server in servers)
+            {
+                if (server.IsOnline && server.IsEmbedded && server.Process != null)
+                {
+                    try
+                    {
+                        Log.Info($"Stopping embedded server {server.Name} during launcher shutdown");
+                        stopTasks.Add(server.StopAsync());
+                    }
+                    catch (Exception ex)
+                    {
+                        Log.Error(ex, $"Error stopping server {server.Name} during shutdown");
+                    }
+                }
+            }
+            
+            // Wait for all servers to stop with a timeout
+            if (stopTasks.Count > 0)
+            {
+                try
+                {
+                    Task.WaitAll([.. stopTasks], TimeSpan.FromSeconds(30));
+                }
+                catch (Exception ex)
+                {
+                    Log.Error(ex, "Error waiting for servers to stop during shutdown");
+                }
+            }
+        }
     }
 
     public ServerEntry[] Servers
--- a/Nitrox.Model/Helper/Reflect.cs
+++ b/Nitrox.Model/Helper/Reflect.cs
@@ -77,7 +77,7 @@ public static class Reflect
         return (PropertyInfo)GetMemberInfo(expression);
     }
 
-    private static MemberInfo GetMemberInfo(LambdaExpression expression, Type implementingType = null)
+    private static MemberInfo GetMemberInfo(LambdaExpression expression, Type? implementingType = null)
     {
         Expression currentExpression = expression.Body;
         while (true)
--- a/Nitrox.Server.Subnautica/Models/Resources/Parsers/PrefabPlaceholderGroupsResource.cs
+++ b/Nitrox.Server.Subnautica/Models/Resources/Parsers/PrefabPlaceholderGroupsResource.cs
@@ -147,12 +147,8 @@ internal sealed class PrefabPlaceholderGroupsResource(SubnauticaAssetsManager as
         {
             logger.ZLogWarning($"An error occurred while deserializing the prefab cache. Re-creating it: {ex.Message:@Error}");
         }
-        if (cache.HasValue)
+        if (cache.HasValue && cache.Value.Version == CACHE_VERSION)
         {
-            if (cache.Value.Version != CACHE_VERSION)
-            {
-                logger.ZLogInformation($"Found outdated cache ({cache.Value.Version}, expected {CACHE_VERSION})");
-            }
             prefabPlaceholdersGroupPaths = cache.Value.PrefabPlaceholdersGroupPaths;
             randomPossibilitiesByClassId = cache.Value.RandomPossibilitiesByClassId;
             groupsByClassId = cache.Value.GroupsByClassId;
@@ -162,7 +158,14 @@ internal sealed class PrefabPlaceholderGroupsResource(SubnauticaAssetsManager as
         // Fallback solution
         else
         {
-            logger.ZLogInformation($"Building cache, this may take a while...");
+            if (cache.HasValue)
+            {
+                logger.ZLogInformation($"Found outdated cache ({cache.Value.Version}, expected {CACHE_VERSION}). Rebuilding cache, this may take a while...");
+            }
+            else
+            {
+                logger.ZLogInformation($"Building cache, this may take a while...");
+            }
             // Get all prefab-classIds linked to the (partial) bundle path
             string prefabDatabasePath = Path.Combine(options.Value.GetSubnauticaResourcesPath(), "StreamingAssets", "SNUnmanagedData", "prefabs.db");
             Dictionary<string, string> prefabDatabase = LoadPrefabDatabase(prefabDatabasePath);
--- a/Nitrox.Shared.targets
+++ b/Nitrox.Shared.targets
@@ -87,21 +87,13 @@
                  Condition="!Exists('$(OutputDirectory)')"
                  ContinueOnError="WarnAndContinue" />
 
-        <!-- Copy in debug because some systems break when they don't use our assembly resolve to look inside lib folder -->
+        <!-- Copy (don't Move) so dependencies stay in output root. Some code paths load assemblies before/when AssemblyResolve runs or from contexts that don't use it (e.g. Release vs Debug), causing empty UI or missing types when DLLs are only in lib/. -->
         <Copy SourceFiles="@(FilesToMove)"
               DestinationFolder="$(OutputDirectory)"
               OverwriteReadOnlyFiles="True"
               SkipUnchangedFiles="True"
               Retries="3"
               RetryDelayMilliseconds="100"
-              Condition="'$(Configuration)' == 'Debug'"
-              ContinueOnError="ErrorAndContinue" />
-
-        <!-- Move every matching files to OutputDirectory -->
-        <Move SourceFiles="@(FilesToMove)"
-              DestinationFolder="$(OutputDirectory)"
-              OverwriteReadOnlyFiles="True" 
-              Condition="'$(Configuration)' == 'Release'"
               ContinueOnError="ErrorAndContinue" />
     </Target>
 
--- a/NitroxClient/MonoBehaviours/Gui/Input/KeyBindings/Actions/ChatKeyBindingAction.cs
+++ b/NitroxClient/MonoBehaviours/Gui/Input/KeyBindings/Actions/ChatKeyBindingAction.cs
@@ -5,7 +5,7 @@ namespace NitroxClient.MonoBehaviours.Gui.Input.KeyBindings.Actions;
 
 public class ChatKeyBindingAction : KeyBinding
 {
-    public ChatKeyBindingAction() : base("Nitrox_Settings_Keybind_OpenChat", "y") { }
+    public ChatKeyBindingAction() : base("Nitrox_Settings_Keybind_OpenChat", "y", "dpad/up") { }
 
     public override void Execute(InputAction.CallbackContext _)
     {
--- a/NitroxPatcher/NitroxPatcher.csproj
+++ b/NitroxPatcher/NitroxPatcher.csproj
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
 
     <PropertyGroup>
         <TargetFrameworks>net472;net10.0</TargetFrameworks>
--- a/NitroxPatcher/Patches/Persistent/GameInputSystem_Initialize_Patch.cs
+++ b/NitroxPatcher/Patches/Persistent/GameInputSystem_Initialize_Patch.cs
@@ -63,21 +63,40 @@ public class GameInputSystem_Initialize_Patch : NitroxPatch, IPersistentPatch
 
     private static void Prefix(GameInputSystem __instance)
     {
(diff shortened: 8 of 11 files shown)


Measurity:
~~I'm not sure what this PR fixes since the input already works after PR #2571~~

~~Can you take a screenshot or paste output of the error you're seeing?~~

I see now thx to @misterbubb - the new update to Nautilus broke the compatibility:
```cs
NullReferenceException: Object reference not set to an instance of an object
  at GameInput.GetBinding (GameInput+Device device, GameInput+Button action, GameInput+BindingSet bindingSet) [0x00000] in <c055c80802d14a9db359fbdc91eb371c>:0 
  at GameInput.AppendDisplayText (GameInput+Button action, System.Text.StringBuilder sb, System.Boolean allBindingSets) [0x0001c] in <c055c80802d14a9db359fbdc91eb371c>:0 
  at GameInput.FormatButton (GameInput+Button action, System.Boolean allBindingSets) [0x0000d] in <c055c80802d14a9db359fbdc91eb371c>:0 
  at uGUI_CameraCyclops.UpdateBindings () [0x00000] in <c055c80802d14a9db359fbdc91eb371c>:0 
  at uGUI_CameraCyclops.UpdateTexts () [0x00000] in <c055c80802d14a9db359fbdc91eb371c>:0 
  at uGUI_CameraCyclops.OnEnable () [0x00000] in <c055c80802d14a9db359fbdc91eb371c>:0 
UnityEngine.Object:Internal_InstantiateSingle_Injected(Object, Vector3&, Quaternion&)
UnityEngine.Object:Internal_InstantiateSingle(Object, Vector3, Quaternion)
UnityEngine.Object:Instantiate(Object, Vector3, Quaternion)
UnityEngine.Object:Instantiate(GameObject, Vector3, Quaternion)
UWE.Utils:InstantiateWrap(GameObject, Vector3, Quaternion)
<InitializeAsync>d__43:MoveNext()
UnityEngine.SetupCoroutine:DMD<UnityEngine.SetupCoroutine::InvokeMoveNext>(IEnumerator, IntPtr)
```

NitroxPatcher/Patches/Persistent/GameInputSystem_Initialize_Patch.cs
```diff
@@ -38,6 +61,12 @@ public static void Prefix(GameInputSystem __instance)
         }
     }
 
+    public static void DeinitializePrefix()
+    {
+        FieldInfo allActionsField = typeof(GameInput).GetField(nameof(GameInput.AllActions), BindingFlags.Public | BindingFlags.Static);
+        allActionsField?.SetValue(null, oldAllActions);
```
Measurity: Direct access to `GameInput.AllActions` is available, which has higher perf and easier to maintain than reflection.

NitroxPatcher/Patches/Persistent/GameInputSystem_Initialize_Patch.cs
```diff
@@ -9,17 +11,38 @@
...
+        PatchPrefix(harmony, DEINITIALIZE_METHOD, ((Action)DeinitializePrefix).Method);
+    }
 
+    public static void Prefix(GameInputSystem __instance)
+    {
+        // Extend AllActions so the game creates InputAction entries for Nitrox buttons when building the action map.
+        // Without this, gameInputSystem.actions[NitroxButton] would not exist and RegisterKeybindsActions would throw.
         int buttonId = KeyBindingManager.NITROX_BASE_ID;
+        oldAllActions = GameInput.AllActions;
+        FieldInfo allActionsField = typeof(GameInput).GetField(nameof(GameInput.AllActions), BindingFlags.Public | BindingFlags.Static);
```
Measurity: Same here. Reflection not needed to access `GameInput.AllActions`.

Measurity (changes requested):
Nitrox keybindings appears twice with these changes:

Image 

Minor: when BepInEx mods are loaded, I'm unable to change Nitrox keybindings as well.

JackNytely:
> Nitrox keybindings appears twice with these changes:
> Image
> 
> Minor: when BepInEx mods are loaded, I'm unable to change Nitrox keybindings as well.

Will look into fixing this, not sure why it appears as double, will investigate asap

Nitrox.Launcher/Models/Design/ServerEntry.cs
```diff
@@ -416,6 +419,9 @@ private ServerProcess(string saveDir, CancellationTokenSource cts, bool isEmbedd
                     WindowStyle = isEmbeddedMode ? ProcessWindowStyle.Hidden : ProcessWindowStyle.Normal,
                     CreateNoWindow = isEmbeddedMode
                 };
+                // Ensure server uses same data path as launcher (saves, logs, cache)
+                startInfo.ArgumentList.Add("--data-path");
+                startInfo.ArgumentList.Add(nitroxAppDataPath);
```
Measurity: Supplying `--data-path` is good add.

Would change to `NitroxUser.AppDataPath` directly, AppDataPath should not be same as savesDir.

Nitrox.Launcher/ViewModels/CreateServerViewModel.cs
```diff
@@ -27,7 +28,7 @@ public partial class CreateServerViewModel : ModalViewModelBase
     [ObservableProperty]
     private SubnauticaGameMode selectedGameMode = SubnauticaGameMode.SURVIVAL;
 
-    private string SavesFolderDir => keyValueStore.GetSavesFolderDir();
+    private string SavesFolderDir => (keyValueStore ?? KeyValueStore.Instance).GetSavesFolderDir();
```
Measurity: KeyValueStore should be supplied as it was before. Optionally can throw exception if it turns out to be null:

```cs
private string SavesFolderDir => keyValueStore?.GetSavesFolderDir() ?? throw new Exception($"{nameof(keyValueStore)} must not be null");
```

Nitrox.Launcher/Nitrox.Launcher.csproj
```diff
@@ -102,11 +102,38 @@
...
+    <!-- In Release, Nitrox.Shared.targets moves Server DLLs to lib/, so we must check both output root and lib subfolder. -->
     <Target Name="_NitroxFixMissingMicrosoftExtensionsDlls" AfterTargets="$(_NitroxMoveDependenciesToLibTaskName)">
+        <PropertyGroup>
+            <_ServerOutputPath>$(MSBuildProjectDirectory)\..\Nitrox.Server.Subnautica\bin\$(Configuration)\$(TargetFramework)</_ServerOutputPath>
+        </PropertyGroup>
         <ItemGroup>
-            <_FilesToCopy Include="$(MSBuildProjectDirectory)\..\Nitrox.Server.Subnautica\bin\$(Configuration)\$(TargetFramework)\Microsoft.Extensions.*.dll" />
+            <_FilesToCopy Include="$(_ServerOutputPath)\Microsoft.Extensions.*.dll" />
+            <_FilesToCopy Include="$(_ServerOutputPath)\lib\Microsoft.Extensions.*.dll" />
         </ItemGroup>
```
Measurity: Same here.

Nitrox.Server.Subnautica/AssemblyResolver.cs
```diff
@@ -21,7 +21,13 @@ internal static class AssemblyResolver
...
         if (assembly == null && !args.Name.Contains(".resources"))
         {
-            assembly = Assembly.Load(args.Name);
+            // Don't fall through to default loader for Nitrox.Model* - it might load net472 from lib/net472/
+            // which lacks types like AsyncBarrier (#if NET). We need the net10.0 version from lib/.
+            string simpleName = args.Name.Split(',')[0].Trim();
+            if (!simpleName.StartsWith("Nitrox.Model", StringComparison.Ordinal))
+            {
+                assembly = Assembly.Load(args.Name);
+            }
```
Measurity: Same here.

Nitrox.Model/Helper/AsyncBarrier.cs
```diff
@@ -11,26 +11,57 @@ namespace Nitrox.Model.Helper;
 public sealed class AsyncBarrier
 {
```
Measurity, line 12: Same here.

Measurity (changes requested):
The changes to support `AsyncBarrier` in net472 code is quite involved. But `AsyncBarrier` is not used right now except for (net10) server. 

Would you mind reverting changes made for AsyncBarrier in an effort to keep complexity to the minimum?

JackNytely:
> The changes to support `AsyncBarrier` in net472 code is quite involved. But `AsyncBarrier` is not used right now except for (net10) server.
> 
> Would you mind reverting changes made for AsyncBarrier in an effort to keep complexity to the minimum?

I was attempting to fix an issue where the server list in the launcher itself was empty and crashing if you tried creating a server, however, I found out that the issue was in the way the app builds a release, publishing works, but if you run from direct release, it moves the dlls over instead of copying which causes it to break in some areas.

I removed my changes and fixed the issue in the build script instead

Kept the fixes for nautilus to work though.

JackNytely:
> > Nitrox keybindings appears twice with these changes:
> > Image
> > Minor: when BepInEx mods are loaded, I'm unable to change Nitrox keybindings as well.
> 
> Will look into fixing this, not sure why it appears as double, will investigate asap

I am not getting duplicates on my side and keybinds seem to work now

Measurity (approved):
Thanks ❤️

Measurity (changes requested):
Oh still duplicated input when starting normally without mods... :(

I think you've handled Nautilus happy path, but without Nautilus the logic now duplicates the input.

JackNytely:
> Oh still duplicated input when starting normally without mods... :(
> 
> I think you've handled Nautilus happy path, but without Nautilus the logic now duplicates the input.

The changes I made caused the UI to build the input bindings from both Enum.GetValues and GameInput.AllActions, by adding nitrox in both it caused the duplication when running without nautilus enabled

I have fixed this issue now, thanks for pointing it out <3

Measurity:
Thanks for working on this.

I noticed the input bindings cannot be rebound when starting without mods. Please verify behavior doesn't regress when playing Nitrox without BepInEx / Nautilus.

JackNytely:
> Thanks for working on this.
> 
> I noticed the input bindings cannot be rebound when starting without mods. Please verify behavior doesn't regress when playing Nitrox without BepInEx / Nautilus.

I fixed the issue with rebinds not working on nitrox without mods, found the issue in my code and fixed it ( nautilus with bepinex still has the binding issues, I will need to investigate this in the future)

I also ran into issues with nitrox servers sometimes not starting, and the stop buttong sometimes does nothing, forcing me to close the game and nitrox and task manager the server (I also fixed this issue in my PR)

NitroxPatcher/NitroxPatcher.csproj
```diff
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
```
Measurity, line 1: Please revert file changes.

Nitrox.Launcher/Models/Services/ServerService.cs
```diff
@@ -247,6 +247,41 @@ public void Dispose()
...
+                {
+                    Task.WaitAll([.. stopTasks], TimeSpan.FromSeconds(30));
+                }
+                catch (Exception ex)
+                {
+                    Log.Error(ex, "Error waiting for servers to stop during shutdown");
+                }
+            }
+        }
     }
```
Measurity, line 285: Is this code ever called? No debugger breakpoint hits on launcher exit. I think the code can be removed.

MainWindowViewModel also already handles server exit.

Nitrox.Launcher/Models/Design/ServerEntry.cs
```diff
@@ -359,18 +373,34 @@ internal async Task ResetCtsAsync(bool cancelPreexisting = false)
...
+                            {
+                                await Task.Delay(500, shutdownCts.Token);
+                            }
+                            catch (OperationCanceledException)
+                            {
+                                // Timeout reached, force terminate
+                                Log.Warn($"Server {Name} did not respond to quit command, attempting force terminate");
+                                using ProcessEx? proc = ProcessEx.GetFirstProcess(GetServerExeName(), ex => ex.Id == LastProcessId);
+                                proc?.Terminate();
+                                break;
```
Measurity, line 396: Terminating the server on timeout can break server saving. I'd prefer we investigate why the server doesn't exit and fix instead of terminating it.

If you know how to trigger the bug, please create a new issue with steps. Then we fix this in a separate PR.

Nitrox.Server.Subnautica/Models/Resources/Parsers/PrefabPlaceholderGroupsResource.cs
```diff
@@ -147,12 +147,8 @@ private async Task CreateOrLoadPrefabCacheAsync(string nitroxCachePath)
         {
             logger.ZLogWarning($"An error occurred while deserializing the prefab cache. Re-creating it: {ex.Message:@Error}");
         }
-        if (cache.HasValue)
+        if (cache.HasValue && cache.Value.Version == CACHE_VERSION)
```
Measurity, line 150: Thanks for noticing this bug. Fixed in #2674 in case your PR stays unmerged for a while.

Measurity (changes requested):
Job well done, thanks. 

Setting input bindings works with Nitrox. And mods load now at least.

That said, I'm not getting the "server close not working" issue, but would like the GameInput keybinds fixes to be merged.

Can the server related bug fixing be split in a separate PR?

Nitrox.Launcher/Models/Design/ServerEntry.cs
Coding-Hen: Yeah I agree, I think it could also help with speed of merging the prs. Here it feels like three prs in one

NitroxClient/MonoBehaviours/Gui/Input/KeyBindings/Actions/ChatKeyBindingAction.cs
```diff
@@ -5,7 +5,7 @@ namespace NitroxClient.MonoBehaviours.Gui.Input.KeyBindings.Actions;
 
 public class ChatKeyBindingAction : KeyBinding
 {
-    public ChatKeyBindingAction() : base("Nitrox_Settings_Keybind_OpenChat", "y") { }
+    public ChatKeyBindingAction() : base("Nitrox_Settings_Keybind_OpenChat", "y", "dpad/up") { }
```
Coding-Hen, line 8: Are these universal across all unity controllers? Will it only work for say playstation but not Xbox?

Nitrox.Shared.targets
```diff
@@ -87,21 +87,13 @@
...
               RetryDelayMilliseconds="100"
-              Condition="'$(Configuration)' == 'Debug'"
-              ContinueOnError="ErrorAndContinue" />
-
-        <!-- Move every matching files to OutputDirectory -->
-        <Move SourceFiles="@(FilesToMove)"
-              DestinationFolder="$(OutputDirectory)"
-              OverwriteReadOnlyFiles="True" 
-              Condition="'$(Configuration)' == 'Release'"
               ContinueOnError="ErrorAndContinue" />
```
dartasen, line 97: This is intended behavior. Avalonia doesn't use our `AppDomain.CurrentDomain.AssemblyResolve`. In Debug we're fine with duplicating the files so Avalonia Previewer works.

Nitrox.Model/Helper/Reflect.cs
```diff
@@ -77,7 +77,7 @@ public static PropertyInfo Property<T>(Expression<Func<T, object>> expression)
         return (PropertyInfo)GetMemberInfo(expression);
     }
 
-    private static MemberInfo GetMemberInfo(LambdaExpression expression, Type implementingType = null)
+    private static MemberInfo GetMemberInfo(LambdaExpression expression, Type? implementingType = null)
```
Measurity, line 80: Thanks for improving this code. Added in #2674 in case your PR stays unmerged for a while.

CherrieTheShifter:

dartasen:
Closing in favor of #2740

PR #2664, opened 2026-02-06
Fixed Steam exe not found for some setups on Linux

Measurity:
This caused proton to be launched without a proper Steam exe.

### DIFF
--- a/Nitrox.Model/Platforms/Discovery/InstallationFinders/SteamFinder.cs
+++ b/Nitrox.Model/Platforms/Discovery/InstallationFinders/SteamFinder.cs
@@ -3,7 +3,7 @@ using System.IO;
 using System.Runtime.InteropServices;
 using System.Text.RegularExpressions;
 using Nitrox.Model.Platforms.Discovery.InstallationFinders.Core;
-using Nitrox.Model.Platforms.OS.Windows;
+using Nitrox.Model.Platforms.Store;
 using static Nitrox.Model.Platforms.Discovery.InstallationFinders.Core.GameFinderResult;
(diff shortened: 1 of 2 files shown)


Nitrox.Model/Platforms/Discovery/InstallationFinders/SteamFinder.cs
```diff
@@ -50,81 +50,7 @@ public GameFinderResult FindGame(GameInfo gameInfo)
...
-            string steamPath = Path.Combine(homePath, "Library", "Application Support", "Steam");
-            if (Directory.Exists(steamPath))
-            {
-                return steamPath;
-            }
-        }
-
-        return null;
-    }
+    private static string GetSteamPath() => Path.GetDirectoryName(Steam.GetExeFile()) ?? "";
```
dartasen: Maybe the naming was not appropriate previously and can be misleading. But the equivalence between steam exe directory and steam game library directory is not true for macos.

Since you can't (or not supposed to) write anything inside application bundle, the dynamic data can't be nowhere near steam exe

as per line 69, ~/Library/Application Support/Steam/... should contain common, appcache, logs, config, appcache, ...

dartasen (changes requested):

Nitrox.Model/Platforms/Discovery/InstallationFinders/SteamFinder.cs
Measurity: Thx, reverted the change.

dartasen (approved):

PR #2665, opened 2026-02-07
Fixed prefab resource cache not storing all required data

Measurity:
This caused wrecks to break when loading prefab cache. First time load worked fine.

Also fixed possible race condition in `EntityDistributionsResource`.

### Situation before

1. Start new world and join
2. `warp -48 -20 -426`
3. See wreck is fine
4. Create *new* world (do NOT load existing save!)
5. `warp -48 -20 -426`
6. See wreck is **broken** when using prefab JSON cache.

### Situation after

1. Start new world and join
2. `warp -48 -20 -426`
3. See wreck is fine
4. Create *new* world  (do NOT load existing save!)
5. `warp -48 -20 -426`
6. See wreck is **still fine** when using prefab JSON cache.

### DIFF
--- a/Nitrox.Server.Subnautica/Models/Resources/Parsers/PrefabPlaceholderGroupsResource.cs
+++ b/Nitrox.Server.Subnautica/Models/Resources/Parsers/PrefabPlaceholderGroupsResource.cs
@@ -10,6 +10,8 @@ using Nitrox.Server.Subnautica.Models.GameLogic.Entities;
 using Nitrox.Server.Subnautica.Models.Helper;
 using Nitrox.Server.Subnautica.Models.Resources.AddressablesTools.Catalog;
 using Nitrox.Server.Subnautica.Models.Resources.Core;
+using ClassIdByRuntimeKeyDictionary = System.Collections.Generic.Dictionary<string, string>;
+using AddressableCatalogDictionary = System.Collections.Generic.Dictionary<string, string[]>;
 
 namespace Nitrox.Server.Subnautica.Models.Resources.Parsers;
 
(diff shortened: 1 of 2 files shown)


Nitrox.Server.Subnautica/Models/Resources/Parsers/PrefabPlaceholderGroupsResource.cs
```diff
@@ -169,24 +155,33 @@ private Dictionary<string, PrefabPlaceholdersGroupAsset> CreateOrLoadPrefabCache
...
         }
         // Fallback solution
-        if (prefabPlaceholdersGroupPaths is null)
+        else
         {
             logger.ZLogInformation($"Building cache, this may take a while...");
-            prefabPlaceholdersGroupPaths = new(GetPrefabPlaceholderGroupAssetsByGroupClassId(assetsManager, GetAllPrefabPlaceholdersGroupsFast(assetsManager)));
-            Cache.Serialize(serializer, new Cache(CACHE_VERSION, prefabPlaceholdersGroupPaths, randomPossibilitiesByClassId), cacheFilePath);
+            // Get all prefab-classIds linked to the (partial) bundle path
+            string prefabDatabasePath = Path.Combine(options.Value.GetSubnauticaResourcesPath(), "StreamingAssets", "SNUnmanagedData", "prefabs.db");
```
Coding-Hen, line 167: Do we want this to come out of the game values to avoid a rewrite for bz with hard coded paths? If we can't avoid that's fine

Measurity: If we do that we'll end up overgeneralizing I think. For now, copy & specialize this in the BZ server until we have a better picture on what we can reuse.

Coding-Hen (approved):
LGTM CW

PR #2670, opened 2026-02-16
Upgrade LiteNetLib to 1.3.5

### LINKED ISSUE #2317: Bump LiteNetLib to 1.3.1
### Describe the issue

https://github.com/RevenantX/LiteNetLib/releases/tag/1.3.1
Will require some tests and maybe some settings to be adjusted so we can keep this for 1.9

dartasen:
Closes #2317 

Unfortunately 2.x drops support for netstandard2.0, so I pinned the new version. As it will probably be the last version we'll use.

# What's new for us

* Disable MTU discovery by default because some routers can break detection and connection in result. Remove UseSafeMtu (for this just use MtuOverride). To enable MTU discovery set NetManager.MtuDiscovery to true
* Fix `ObjectDisposedException` with Native Sockets (we're now using NativeSockets as per #2661)
* Added multiple Span/ReadonlySpan overloads for various methods
* Lock unreliable channel only if there is pending packets
* Unreliable packets optimization
* Prevent unnecessary allocations in NetManager.ProcessNtpRequests

[https://github.com/RevenantX/LiteNetLib/compare/v1.2.0...1.3.5](https://github.com/RevenantX/LiteNetLib/compare/v1.2.0...1.3.5)

### DIFF
--- a/Nitrox.Server.Subnautica/Models/GameLogic/Entities/Spawning/BatchEntitySpawner.cs
+++ b/Nitrox.Server.Subnautica/Models/GameLogic/Entities/Spawning/BatchEntitySpawner.cs
@@ -39,7 +39,7 @@ internal sealed class BatchEntitySpawner(
 
     private readonly Lock parsedBatchesLock = new();
     private readonly Lock emptyBatchesLock = new();
-    private HashSet<NitroxInt3> parsedBatches;
+    private HashSet<NitroxInt3> parsedBatches = [];
 
     public List<NitroxInt3> SerializableParsedBatches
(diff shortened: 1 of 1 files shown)


Measurity:
Can you add a `= [];` to `BatchEntitySpawner.parsedBatches`? 
While testing this PR with 2 players it crashes on a NRE there (weird that it didn't happen before, maybe LNL so fast it triggers this bug).

LNL upgrade works for me otherwise. Seamoth driving, wreck spawning, bases building. LGTM :)

PR #2671, opened 2026-02-16
Handle entities that exist before spawning phase

misterbubb:
During initial sync, some entities (like PlayerEntity) are set up by earlier sync processors before GlobalRootInitialSyncProcessor spawns entities. This causes the entity spawning phase to encounter entities that already have NitroxIds and GameObjects.

This fix detects such cases and applies metadata to the existing object instead of attempting to respawn it, preventing errors during player join.

### DIFF
--- a/NitroxClient/GameLogic/Entities.cs
+++ b/NitroxClient/GameLogic/Entities.cs
@@ -189,6 +189,17 @@ namespace NitroxClient.GameLogic
                     UpdateEntity(entity);
                     continue;
                 }
+                // Some entities (like PlayerEntity) may already exist because they were set up by earlier
+                // sync processors before the entity spawning phase. In this case, just apply metadata and mark as spawned.
+                else if (entity is WorldEntity && NitroxEntity.TryGetObjectFrom(entity.Id, out GameObject existingObject))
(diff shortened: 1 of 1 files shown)


NitroxClient/GameLogic/Entities.cs
```diff
@@ -191,6 +189,17 @@ public IEnumerator SpawnBatchAsync(List<Entity> batch, bool forceRespawn = false
                     UpdateEntity(entity);
                     continue;
                 }
+                // Some entities (like PlayerEntity) may already exist because they were set up by earlier
+                // sync processors before the entity spawning phase. In this case, just apply metadata and mark as spawned.
+                else if (entity is WorldEntity && NitroxEntity.TryGetObjectFrom(entity.Id, out GameObject existingObject))
```
tornac1234, line 194: This doesn't seem right to me since this is really a sensitive place for entity spawning. Treating any entity spawning here would be short circuiting the actual entity spawners which are responsible for treating bad cases as the one you describe here. I'd rather have some special case in the PlayerEntitySpawner, and leave this very generic place here without workarounds for specific entities.

PR #2677, opened 2026-03-04
Improved debuggers

dartasen:
* Assigned unique **window ID** to each debugger (including main one). `GUIUtility.GetControlID(FocusType.Keyboard)` seems to misbehave on my side causing strange behaviors
* Added `Ctrl+R` button on main debugger to allow resetting debugger's position (I can drag them out my screen by mistake or sometimes they disappear because of `GUIUtility.GetControlID(FocusType.Keyboard)` which is quite annoying
* Debugger list is now opened by default when pressing F7
* Removed `MoveOverlappingSceneDebugger` and `ResetWindowPosition` from `SceneExtraDebugger` : It's causing more problems than it's actually solving. I rather drag & drop the window by hand like the other debuggers
* `SceneExtraDebugger` : It's now possible to click on the Ray tracing button to toggle it
* Removed `Update` logic inside `AbstractDebugger` since it's unused and we're iterating on the list each frame for nothing.
* Some Cleanup


Measurity (changes requested):
Debuggers don't remember window position when enabling / disabling (default hotkey is `F7`).

I think it would be good to have so that developers don't need to reposition windows.

dartasen:
Funnily enough, it was behavior that seems to never have worked in the past : 
image

Measurity (approved):

PR #2678, opened 2026-03-08
Add basic VS Code config

dartasen:


Measurity (approved):
Starting server through VSCode doesn't work well on Linux for me but if it works for you then LGTM.
___
I get occasional exception messages in output and server input doesn't work, neither ctrl+C to close server.
```cs
Exception thrown: 'System.AggregateException' in System.Private.CoreLib.dll
```

dartasen:
dotnet task doesn't support providing `externalTerminal` option, I switched to coreclr task to fix that and prevent vscode to redirect input/output which can break the server, nice catch👍

Measurity:
This window pops-up after trying to launch server but no server process is running:
image

On Linux I think it needs to start a new shell window using the shell GUI program and run an executable in order for "external window" to work. Like so on KDE systems:
```
konsole --hold -e bash -c '${workspaceFolder}/Nitrox.Server.Subnautica/bin/Debug/net10.0/Nitrox.Server.Subnautica'
```

Setting above command as "program" on Linux still shows the error message window.

I'm fine with merging the PR as it was before, since it works for the most part.

Measurity (approved):
Works now after setting terminal to `konsole` for me (GUI shell on Fedora). Thanks for looking into the problem!

Image

PR #2680, opened 2026-03-12
Fix change detection on server config files

Measurity:
`FileConfigurationSource` first argument, `Path`, should be relative to the directory given to `IFileProvider`.

With this fix, change detect should work.

**Needs testing on Windows.**

## Testing

1. Start server
2. Change `server.Development.json` by setting `PortForward` to true.
3. Notice: after a few seconds that the server will port forward the default port.
4. Change `ServerPort` in server.cfg to `11001` (is in save folder of server)
5. Notice: after a few seconds that the port forward be will removed from `11000` and set to `11001`.

### DIFF
--- a/Nitrox.Model/Configuration/SubnauticaServerOptions.cs
+++ b/Nitrox.Model/Configuration/SubnauticaServerOptions.cs
@@ -1,6 +1,5 @@
 using System;
 using System.ComponentModel.DataAnnotations;
-using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Options;
 using Nitrox.Model.Constants;
 using Nitrox.Model.DataStructures.GameLogic;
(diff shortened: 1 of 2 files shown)


Nitrox.Server.Subnautica/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -35,7 +35,13 @@ internal static partial class ServiceCollectionExtensions
     [GenerateServiceRegistrations(AssignableTo = typeof(IRedactor), Lifetime = ServiceLifetime.Singleton)]
     internal static partial IServiceCollection AddRedactors(this IServiceCollection services);
 
-    [GenerateServiceRegistrations(AssignableTo = typeof(IAdminFeature<>), CustomHandler = nameof(AddImplementedAdminFeatures))]
+    /// <summary>
+    ///     Adds an interface -> service mapping that for handling administrative actions.
+    /// </summary>
+    /// <remarks>
+    ///     If a multiple of the same interface type are registered then the last registered implementation will be used.
```
Jannify: I think there is something wrong in the sentence.

PR #2682, opened 2026-03-17
fix: server-side entity spawn race conditions

Jacqueb-1337:
Fix 3 bugs caused by concurrent CellVisibilityChanged packet processing:

- EntityDestroyedPacketProcessor: ReplyAsync sends destruction back to the sender instead of other players. Switch to SendAsync with explicit player loop.

- BatchEntitySpawner: concurrent callers for the same batch race on parsedBatches  second caller sees batch done and returns empty list before entities are registered. Replace HashSet+lock with ConcurrentDictionary<NitroxInt3, Lazy<Task<List<Entity>>>>.

- WorldEntityManager: same deduplication problem at the registration layer. Add ConcurrentDictionary<NitroxInt3, Lazy<Task<int>>> batchRegistrationTasks.


Measurity (approved):
Impressive detective work on finding these bugs! Thank you very much!

PR #2684, opened 2026-03-20
Migrate to .slnx

dartasen:
Context: [https://devblogs.microsoft.com/dotnet/introducing-slnx-support-dotnet-cli/](https://devblogs.microsoft.com/dotnet/introducing-slnx-support-dotnet-cli/)

Supported by all the IDE since .NET 9 (slnx was opt-in back then, default on new project since .NET 10)


Coding-Hen (approved):
As long as it doesn't cause lots of issues for current developers who might use older VS Studio instances then I think this makes sense unless it gives us a big advantage switching to this?

dartasen:
.NET 10 isn't supported by older Visual Studios (minimum version is Visual 2026 18.0), We were using .NET 9 for 1.8 version which is minimum VS 2022 17.12. If there's active developers that do not meet those requirements, I'll be glad to hear from them (as there's no apparent good reasons to not be up to date)

Measurity (approved):

Issue #2352, opened 2025-07-15
pressure-vessel launch fails on some systems


anewusername:
### Are you on the latest version of Nitrox?

Yes, I'm on the latest commit of the repo (only for experts).

### Which OS are you playing on?

Linux (currently not supported, don't file any reports if your are using this)

### On which platform did you buy Subnautica?

Steam

### Describe the issue.

Trying to start on my Linux machine, I get
`pressure-vessel-adverb[957142]: E: Failed to execute child process “/jan_ssd/SteamLibrary/steamapps/common/SteamLinuxRuntime_sniper/pressure-vessel/bin/steam-runtime-launcher-interface-0” (No such file or directory)`

The file does exist, pressure-vessel just doesn't see it. Possibly related to https://github.com/ValveSoftware/steam-runtime/issues/288 .

I mentioned this over on https://github.com/Measurity/Nitrox/pull/19#issuecomment-2395588993 but wanted to make sure it didn't get forgotten.

As a workaround, I am launching without pressure-vessel, with the following edits in`NitroxModel/Platforms/Store/Steam.cs`; probably you could fall back to this if the normal way fails:
```
//FileName = Path.Combine(sniperruntimepath, "_v2-entry-point"),
//Arguments = $" --verb=run -- \"{Path.Combine(protonPath, "proton")}\" run \"{pathToGameExe}\" {launchArguments}",
FileName = Path.Combine(protonPath, "proton"),
Arguments = $"waitforexitandrun \"{pathToGameExe}\" {launchArguments}",
```

### Relevant log output

Console output:
```
$ ./Nitrox.Launcher 
[18:52:54.678] [DBG] Game installation was found by NitroxModel/Discovery/InstallationFinders/ConfigFinder at '/jan_ssd/SteamLibrary/steamapps/common/Subnautica'
[18:52:54.926] [INF] Trying to request data from https://nitrox.rux.gg/api/version/latest
[18:53:06.757] [INF] Launching Subnautica in multiplayer mode
[18:53:06.764] [DBG] Removing Nitrox entry point from Subnautica
[18:53:08.664] [DBG] Removed Nitrox entry point from Subnautica
[18:53:08.665] [DBG] Adding Nitrox entry point to Subnautica
[18:53:09.708] [DBG] Writing assembly to Assembly-CSharp-Nitrox.dll
[18:53:09.709] [DBG] Deleting Assembly-CSharp.dll
[...]

Issue #2687, opened 2026-03-21
Nitrox only works in some directories on Linux


CodesInChaos:
### Which version are you playing on?

1.8.1.0 (Latest release) and the current `master` branch

### Which OS are you playing on?

Linux (Manjaro)

### On which platform did you buy Subnautica?

Steam

### Describe the issue.

Launching the game from the Nitrox launcher in "multiplayer mode" via steam (i.e. "allow multiple" is disabled) will only work in some directories. Such as `.../steamapps/...` or `/home/<user>/...`. When Nitrox is installed in other directories, the `Directory.Exists` check fails, and the game starts in single player mode, without any visible error on screen or in logs.

Apparently this is caused by the steam runtime sandboxing the game, so it can only see some directories. This appears to be the same issue as [this reddit post](https://www.reddit.com/r/linux_gaming/comments/1h69qzn/wineproton_unable_to_see_directories_other_than/).

Some ways Nitrox could improve the user experience:

1. Have the launcher detect when it's installed in an inaccessible directory, and log a warning. Or if it's reliable enough, show an error. Unfortunately it's not obvious which directories these are. One could use the paths listed in the FAQ below.
2. Log errors during initialization to a file, instead of just the invisible console output. Which seems to be tricky, since dependencies are unavailable at that point. Plus it requires the log directory to be accessible (it likely is, since it defaults to being in the user's home directory)
3. Pass the directories required by Nitrox to  `PRESSURE_VESSEL_FILESYSTEMS_RW` and/or `PRESSURE_VESSEL_FILESYSTEMS_RO` to make them accessible inside the sandbox. But setting these environment variables when invoking steam doesn't seem to work (at least if it's already running).

So it looks like improving this will be tricky.

----

> [Steam Workshop outside the home directory](https://github.com/ValveSoftware/steam-runtime/blob/master/doc/steamlinuxruntime-known-issues.md#steam-workshop-outside-the-home-directory)
> -----------------------------------------
> 
> If a game has Steam Workshop support and is installed outside your
> home directory, it will not necessarily find the Steam Workshop content.
> 
> Workaround: Move it to your home directory, as above.
> 
> ([#257](https://github.com/ValveSoftware/steam-runtime/issues/257))

> [Sharing directories with the container](https://github.com/ValveSoftware/steam-runtime/blob/master/doc/steamlinuxruntime-known-issues.md#sharing-directories-with-the-container)
> --------------------------------------
> 
[...]

PR #2689, opened 2026-03-21
Grant access to steam libraries in sandbox

### LINKED ISSUE #2688: Starting game installed in unusual location doesn't work (Linux, Steam)

1.8.1.0 (Latest release) and `master`

Linux

Steam

### Describe the issue.

My steam games library is installed in an unusual location: `/drives/apps/Games/steamapps/`. This seems to cause the steam sandbox to prevent it from starting via the "allow multiple instances" option which starts the game executable directly.

### Relevant log output

> pv-adverb[1062822]: E: Failed to execute child process “/drives/apps/Games/steamapps/common/Proton 9.0 (Beta)/proton” (No such file or directory)

CodesInChaos:
The Steam Runtime (Pressure Vessel) sandboxes applications and doesn't grant it access to the whole file system. So if `steamapps` is installed in an unusual location, starting proton will fail. This change adds all steam libraries to the locations the game can access.

Fixes #2688

### DIFF
--- a/NitroxPatcher/Main.cs
+++ b/NitroxPatcher/Main.cs
@@ -34,7 +34,7 @@ public static class Main
         {
             string path = (args[i], args[i + 1]) switch
             {
-                ("--nitrox", { } value) when Directory.Exists(value) => Path.GetFullPath(value),
+                ("--nitrox", { } value) when !string.IsNullOrEmpty(value) => Path.GetFullPath(value),
                 _ => null
             };
             if (!string.IsNullOrEmpty(path))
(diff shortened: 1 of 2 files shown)


NitroxPatcher/Main.cs
```diff
@@ -69,9 +69,14 @@ public static void Execute()
         AppDomain.CurrentDomain.ReflectionOnlyAssemblyResolve += CurrentDomainOnAssemblyResolve;
 
         Console.WriteLine("Checking if Nitrox should run...");
+        if (string.IsNullOrEmpty(nitroxLauncherDir.Value))
+        {
+            Console.WriteLine($"Nitrox will not load because launcher path was not provided");
+            return;
+        }
```
dartasen, line 76: This check is redundant, Directory.Exists is already looking for null or empty path and returns false in those cases

Measurity (approved):
Code looks good. Still works on my end so I'll trust it's working also for steam libraries now. Thank you for the PR!

CodesInChaos:
What I don't understand is why it works without including the launcher path in the whitelist, while the same launcher directory causes problems when starting via steam (#2687).

dartasen (commented):
Thank's for the PR !

_If you could write down few words about the fix in the PR description, that would be great 😄_ 
EDIT: nvm It's detailed in the issue

NitroxPatcher/Main.cs
Measurity: I think this is fine because the error message becomes more detailed. This one it's not provided at all. Other one it's "does not exist or is inaccessible".

CodesInChaos: Having a separate check results in a clearer and more specific error message.

dartasen (approved):

CodesInChaos:
While the error isn't exactly the same it might fix #2352 as well.

PR #2699, opened 2026-04-03
Added player list tooltip to Nitrox Launcher

NinjaPedroX:
This PR adds a tooltip to the online player count on ServersView and ManageServerView which lists the names of all online players in that server.

image


Measurity (approved):
Works for me, thanks!

PR #2703, opened 2026-04-06
Fix windows not being draggable

dartasen:
The actual fix is this : 
```cs
window.ExtendClientAreaToDecorationsHint = false;
window.ExtendClientAreaTitleBarHeightHint = -1;
```

This avoid launcher overlapping system bar, which prevents to drag the window. 

---

I also noticed that Modals were not being applied all the fixes we made to windows on both Linux & MacOS. So I might need a linux folk to test those things as well.

### DIFF
--- a/Nitrox.Launcher/Models/Extensions/VisualExtensions.cs
+++ b/Nitrox.Launcher/Models/Extensions/VisualExtensions.cs
@@ -1,6 +1,8 @@
+using System;
 using System.Runtime.InteropServices;
 using Avalonia;
 using Avalonia.Controls;
+using Nitrox.Launcher.Models.Design;
 using Nitrox.Model.Platforms.OS.Windows;
 
 namespace Nitrox.Launcher.Models.Extensions;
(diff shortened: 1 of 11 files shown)


Measurity (changes requested):
Every use of `CustomTitlebar` now has `IsVisible="{Binding $parent[Window].(design:NitroxAttached.UseCustomTitleBar)}"` so it can be part of the control theme

Measurity (approved):
Is the PR also tested on Windows?

Nitrox.Launcher/Models/Extensions/VisualExtensions.cs
```diff
@@ -1,33 +1,42 @@
...
-    public static void ApplyOsWindowStyling(this Visual visual)
+    public static void ApplyPlatformWindowChrome(this Window window)
     {
         if (IsDesignMode)
         {
             return;
         }
-        if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+
+        if (OperatingSystem.IsLinux())
```
Jannify, line 18: Can we have the comments from [WindowEx.cs](https://github.com/SubnauticaNitrox/Nitrox/pull/2703/changes#diff-93dd9ade60eb44b625733c031dd8987a6b548597337207a2c3efc2860f2608ff) here?

NinjaPedroX:
> Is the PR also tested on Windows?

Window dragging seems to work fine on Windows 11 with this PR

Jannify (dismissed):

Nitrox.Launcher/Views/Abstract/ModalBase.cs
```diff
@@ -4,5 +4,8 @@ namespace Nitrox.Launcher.Views.Abstract;
 
 public abstract class ModalBase : Window
 {
-    protected override void OnInitialized() => this.ApplyOsWindowStyling();
+    protected override void OnInitialized()
+    {
+        this.ApplyPlatformWindowChrome();
```
Measurity: WindowChrome is a Windows only thing afaik. Maybe change name to `ApplyPlatformWindowStyle` ?

Nitrox.Launcher/Models/Extensions/VisualExtensions.cs
dartasen: Done

Measurity (approved):

PR #2704, opened 2026-04-06
Improve PrefabPlaceHolder cache handling

dartasen:
Context : 
image
---

We had many reports of users in help-channel that are showing this error "Successfully loaded 0 x".
It may be possible to have transient errors during cache building that could cause a cache created with 0 values inside.

However, I did not spot the precise area where it could fail exactly, though this fix is necessary to avoid flagging empty cache as valid.


Jannify (approved):
LGTM CW

Measurity (approved):

PR #2710, opened 2026-04-15
Add Library View to the launcher

NinjaPedroX:
This PR adds a new tab to the launcher, named: **Library**

In this tab, you can:
1. View all instances of the game installed on your computer, all of which are saved into the Nitrox long-term cache.
2. Manually add/remove game installations to/from the launcher.
3. Easily manage which instance of Subnautica you would like to use Nitrox with.
4. View the status of any server you joined/added in the main menu (data such as online status and player counts/names).

Also, Options view was changed to support multiple game installations instead of only one, featuring the same game installation management features as Library View.

**TODO:**
- [x] Add communication between the launcher and remote servers to allow for the retrieval of its online status, host server name, online player count, max player count, online player names, and server image (not anymore).

Images:
image
image
image
(no installations)
image
image


Nitrox.Launcher/Models/Design/RecentServerEntry.cs
```diff
@@ -0,0 +1,34 @@
...
+using Nitrox.Launcher.Models.Utils;
+
+namespace Nitrox.Launcher.Models.Design;
+
+public partial class RecentServerEntry : ObservableObject
+{
+
+    public string ServerName { get; set; } = string.Empty;
+    public string ServerIP { get; set; } = string.Empty;
+    public int ServerPort { get; set; } = 11000;
```
Measurity: Use constant value from `SubnauticaServerConstants` in Nitrox.Model

Nitrox.Launcher/Models/Design/RecentServerEntry.cs
```diff
@@ -0,0 +1,34 @@
...
+    public partial Bitmap ServerIcon { get; set; } = AssetHelper.GetAssetFromStream("/Assets/Images/subnautica-icon.png", static stream => new Bitmap(stream));
+
+    [ObservableProperty]
+    public partial bool IsOnline { get; set; } = false;
+
+    [ObservableProperty]
+    public partial int PlayerCount { get; set; } = 0;
+
+    [ObservableProperty]
+    public partial int MaxPlayers { get; set; } = 100;
```
Measurity: Time to add this as a constant to `SubnauticaServerConstants`. And then use this constant in `SubnauticaServerOptions` too.

Nitrox.Launcher/Models/Services/GameInstallationService.cs
```diff
@@ -0,0 +1,412 @@
...
+    private static readonly JsonSerializerOptions jsonSerializerOptions = new(JsonSerializerDefaults.Web);
+    private bool hasLoggedInitialCacheSnapshot;
+    private readonly HashSet<string> ignoredGamePaths = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+
+    [ObservableProperty]
+    public partial KnownGame SelectedGame { get; set; } = new() { PathToGame = string.Empty, Platform = Platform.NONE };
+
+    public AvaloniaList<KnownGame> InstalledGames { get; } = [];
+
+    public List<KnownGame> RefreshInstalledGames(GameInfo gameInfo)
```
Measurity: All code that's accessing files should be `async` so that they don't cause UI hangs / stutter.

Measurity: This method is getting chunky. Move logic into private methods that do one "thing". For example, logging should be moved into a method call. And `savedGames` code can be a separate method as well which also allowed use of IEnumerable<T> and `yield return` to simplify processing.

CherrieTheShifter:

NinjaPedroX:
This PR is now ready for testing/reviews.

dartasen:
Yay 🎶

When I saw the library design on Figma, I was thinking of doing a curse forge like experience. Where you can start the game with your profile (which would be game installation for our need)
i.e: 
image
image

And I feel like it's missing UX from this feature. Currently the feature is +- the same as option page. I think we should be able to start the game from this view, but also be able to switch profile from the LaunchGame view as well, maybe something like that : 

image

The question that remains for me is why we would keep the UI of the options page if the new “Library” section handles that aspect.
Maybe we should ask the team to test the feature and give some feedback, and that would be cool to double check with Rux as well

NinjaPedroX:
> Yay 🎶
> 
> When I saw the library design on Figma, I was thinking of doing a curse forge like experience. Where you can start the game with your profile (which would be game installation for our need)
> i.e: 
> image
> image
> 
> And I feel like it's missing UX from this feature. Currently the feature is +- the same as option page. I think we should be able to start the game from this view, but also be able to switch profile from the LaunchGame view as well, maybe something like that : 
> 
> image
> 
> The question that remains for me is why we would keep the UI of the options page if the new “Library” section handles that aspect.
> Maybe we should ask the team to test the feature and give some feedback, and that would be cool to double check with Rux as well

Sounds good, I'm down to implement whatever the team/Rux recommends. I agree that having the game selection feature in two different views is a bit redundant.

If we decide to do it where we can start the game in Library View, would we need the start game buttons in Launch Game View anymore? I'm not sure how that would work.

Measurity:
This PR and master should be connection tested as we had trouble connecting over WAN.

As for the design, game installations are now configurable in both "Library View" and "Options View". I think it would be best to merge these two into the "Play game" view as a combobox instead of the simple "Play Multiplayer" button there's now.

As for "recent servers", I think we should add this to "Servers" view with a new tab "Recently Visited" and a "Self host" tab. Then we can remove the "Library" view and simplify launcher navigation in my opinion.

Thoughts?

PR #2715, opened 2026-04-25
Changed port lifetime to max

### LINKED ISSUE #1595: Fallback UPnP port forwarding to permanent ports if router does not support expiration
Exception message:
`Error 725: OnlyPermanentLeasesSupported`

Reported by tutbrush on discord.

Measurity:
Possible fix for port forward rule (via UPnP) not applying because it immediately expires if bad router defaults.

### DIFF
--- a/Nitrox.Model/Helper/NatHelper.cs
+++ b/Nitrox.Model/Helper/NatHelper.cs
@@ -23,20 +23,27 @@ public static class NatHelper
         }
     }).ConfigureAwait(false);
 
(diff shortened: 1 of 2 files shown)


Coding-Hen (commented):
LGTM though will this mean each start will port forward with a new description will that be a problem if one doesn't clear down properly?

Measurity:
Description isn't used when deleting port forward rule as message gets encoded without this data. Behavior is unchanged.

dartasen (approved):

PR #2721, opened 2026-05-07
Added SteamOS support for starting servers

Measurity:
If user doesn't have `~/.dotnet/dotnet` binary, following error will be shown:
image

### DIFF
--- a/Nitrox.Launcher/GlobalStatic.cs
+++ b/Nitrox.Launcher/GlobalStatic.cs
@@ -1,5 +1,6 @@
 using System;
 using System.IO;
+using System.Threading;
 using Avalonia.Controls;
 using Nitrox.Model.Platforms.OS.Shared;
 
@@ -7,6 +8,9 @@ namespace Nitrox.Launcher;
 
 internal static class GlobalStatic
 {
+    private const string SERVER_FILE_NAME_NO_EXT = "Nitrox.Server.Subnautica";
+    private static long isSteamOs = -1;
+
     public static bool IsDesignMode => Design.IsDesignMode;
 
     /// <inheritdoc cref="ProcessEx.OpenUri" />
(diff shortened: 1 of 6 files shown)


Nitrox.Launcher/GlobalStatic.cs
```diff
@@ -44,13 +48,51 @@ public static bool TryDeleteDirectory(string directory, bool recursive = false)
...
+    public static string GetServerFileName()
     {
-        string serverExeName = "Nitrox.Server.Subnautica.exe";
-        if (!OperatingSystem.IsWindows())
+        if (OperatingSystem.IsWindows())
+        {
+            return $"{SERVER_FILE_NAME_NO_EXT}.exe";
+        }
+
+        return SERVER_FILE_NAME_NO_EXT;
```
Coding-Hen, line 58: Unless I'm missing something aren't these returning the same value? Unless the windows one is wrapped in "

Nitrox.Launcher/GlobalStatic.cs
```diff
@@ -44,13 +48,51 @@ public static bool TryDeleteDirectory(string directory, bool recursive = false)
...
+    }
+
+    public static string GetServerProcessName()
+    {
+        // On Steam Deck, server process name is "dotnet" as it's started through command line.
+        if (IsSteamOs())
```
Coding-Hen, line 64: Is it worth us maybe breaking these out like with the finders so different os can be separated logic rather than all together, then we have easier time adding new/different os flavours if needed

Nitrox.Launcher/GlobalStatic.cs
Measurity: The windows one returns with `.exe` appended. On Linux there's no extension on the executable so it returns the `_NO_EXT` constant value as is.

Nitrox.Launcher/GlobalStatic.cs
Measurity: I don't have anything planned with lots of platforms in this logic. I'll favor keeping it simple to avoid premature API design.

Nitrox.Launcher/GlobalStatic.cs
```diff
@@ -1,12 +1,16 @@
...
 using System.IO;
+using System.Threading;
 using Avalonia.Controls;
 using Nitrox.Model.Platforms.OS.Shared;
 
 namespace Nitrox.Launcher;
 
 internal static class GlobalStatic
 {
+    private const string SERVER_FILE_NAME_NO_EXT = "Nitrox.Server.Subnautica";
```
Jannify, line 11: Would it make sense to use nameof(Server Assembly)?

Nitrox.Launcher/GlobalStatic.cs
```diff
@@ -44,13 +48,51 @@ public static bool TryDeleteDirectory(string directory, bool recursive = false)
...
+
+        return SERVER_FILE_NAME_NO_EXT;
+    }
+
+    public static string GetServerProcessName()
+    {
+        // On Steam Deck, server process name is "dotnet" as it's started through command line.
+        if (IsSteamOs())
+        {
+            return "dotnet";
```
Jannify, line 66: Won't this result in the behaviour that every dotnet application/application started through dotnet is treated as a Nitrox server?

Nitrox.Launcher/GlobalStatic.cs
Coding-Hen: I think maybe keep it this way for the moment then and in future if we add more we can break this out

Coding-Hen (approved):
LGTM CW

Nitrox.Model/Helper/NitroxUser.cs
```diff
@@ -221,4 +214,34 @@ public static string? AssetsPath
...
+                homePath = Environment.GetEnvironmentVariable("HOME");
+            }
+            if (!Directory.Exists(homePath))
+            {
+                throw new DirectoryNotFoundException("User home directory does not exist or is inaccessible");
+            }
+            if (string.IsNullOrWhiteSpace(homePath))
+            {
+                throw new InvalidOperationException("User home directory is not given by the operating system");
+            }
```
Jannify, line 237: Shouldn't this check be moved above the other?

Jannify (commented):
😢

Nitrox.Launcher/GlobalStatic.cs
Measurity: Yes, however it also checks for process id in cases where this is used.

Nitrox.Launcher/GlobalStatic.cs
Measurity: There are [tools](https://github.com/amis92/csharp-source-generators#statically-typed-resources--configurations) to generate this for us. I think we should implemented this everywhere in a separate PR.

Nitrox.Model/Helper/NitroxUser.cs
Measurity: I'll ask Ninja to move this code up.

PR #2723, opened 2026-05-09
Avalonia 12

dartasen:
https://github.com/AvaloniaUI/Avalonia/releases

## TODO
- [ ] Wait for Avalonia to fix our "Notifcations" tab tooltip appearing as a top-left white box on screen.
- [ ] Fix SVG icons appearing black in UI (should be themed as white by default)


Nitrox.Model/Platforms/OS/Shared/ProcessEx.cs
```diff
@@ -233,7 +233,7 @@ public static IEnumerable<T> GetProcessesByName<T>(string procName, Func<Process
             try
             {
                 result = selector(procEx);
-                if (result is not Process or ProcessEx)
+                if (result is not Process)
```
Measurity: We can remove this function anyway as it's not used.

Though `ProcessEx` cannot inherit `Process` so I think the two checks were necessary before.

Nitrox.Launcher/App.axaml.cs
```diff
@@ -41,7 +37,7 @@ internal class App : Application
     public static App Instance;
 
     /// <summary>
-    ///     If true, allows duplicate instances of the app to be active.
+    /// If true, allows duplicate instances of the app to be active.
     /// </summary>
     internal static bool allowInstances;
```
Measurity, line 42: We should decide on which style of code doc we want, with tabs or no because right now my IDE is configured with tabs.

Nitrox.Launcher/ViewModels/Abstract/IServerViewModel.cs
```diff
@@ -0,0 +1,6 @@
+namespace Nitrox.Launcher.ViewModels.Abstract;
+
+/// <summary>
+///     Marker interface for view models that represent a server-related view, used for nav highlighting.
+/// </summary>
+internal interface IServerViewModel;
```
Measurity, line 6: Nice one :)

Nitrox.Launcher/Views/MainWindow.axaml
```diff
@@ -451,7 +356,6 @@
                                     HorizontalAlignment="Right"
                                     VerticalAlignment="Top">
                                     <Svg
-                                        Css="* { fill: #fff; stroke: #fff; }"
                                         Path="/Assets/Icons/close.svg"
                                         VerticalAlignment="Top"
                                         Width="10" />
```
Measurity, line 391: This SVG change makes notification icons look black:

Image

Nitrox.Launcher/Nitrox.Launcher.csproj
```diff
@@ -46,14 +43,16 @@
...
         </PackageReference>
+        <PackageReference Include="ServiceScan.SourceGenerator">
+            <PrivateAssets>all</PrivateAssets>
+            <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+        </PackageReference>
+        <PackageReference Include="Svg.Controls.Skia.Avalonia" />
+        <PackageReference Include="Xaml.Behaviors.Avalonia" />
     </ItemGroup>
     
     <!-- Temp fix for Avalonia 11.3 bug: https://github.com/AvaloniaUI/Avalonia/discussions/18971#discussioncomment-13533852 -->
```
Measurity: This workaround can be removed now, it's included by Avalonia nuget package.

Measurity (changes requested):

Nitrox.Launcher/Models/Converters/BitmapAssetValueConverter.cs
```diff
@@ -1,13 +1,13 @@
-﻿using System;
+using System;
 using System.Globalization;
 using Avalonia.Media.Imaging;
 using Nitrox.Launcher.Models.Utils;
 
 namespace Nitrox.Launcher.Models.Converters;
 
-public class BitmapAssetValueConverter : Converter<BitmapAssetValueConverter>
+public sealed class BitmapAssetValueConverter : Converter<BitmapAssetValueConverter>
```
Measurity: We can make all the converter classes `internal` as well.

Nitrox.Model/Platforms/OS/Shared/ProcessEx.cs
dartasen: Maybe a bug from an analyzer : 
image

Nitrox.Launcher/Assets/Icons/warning.svg
Measurity, line 1: I prefer keeping the fill color black so it's not invisible when viewing the file in a diff

dartasen:
On hold for now, waiting for at least 12.1. This version is too buggy atm

Measurity:
FYI: [Lastest Avalonia releases](https://github.com/AvaloniaUI/Avalonia/releases)

PR #2725, opened 2026-05-10
Fix oxygen tanks losing oxygen on reconnect

### LINKED ISSUE #2266: Oxygen tank is empty in inventory when reconnecting

misterbubb:
Fixes #2266

## Testing
1. Craft oxygen tank and equip it
2. Unequip it
3. Disconnect and reconnect
4. Verify tank retains oxygen value


Nitrox.Model.Subnautica/DataStructures/GameLogic/Entities/Metadata/OxygenMetadata.cs
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    public OxygenMetadata(float oxygenAvailable)
+    {
+        OxygenAvailable = oxygenAvailable;
+    }
+
+    public override string ToString()
+    {
+        return $"[OxygenMetadata OxygenAvailable: {OxygenAvailable}]";
```
Measurity: Use `nameof` to ease with future refactoring work

NitroxClient/GameLogic/Spawning/Metadata/Processor/OxygenMetadataProcessor.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        Oxygen oxygen = gameObject.GetComponent<Oxygen>();
+
+        if (oxygen)
+        {
+            oxygen.oxygenAvailable = metadata.OxygenAvailable;
+        }
+        else
+        {
+            Log.Warn($"[OxygenMetadataProcessor] Could not find Oxygen component on {gameObject.name}");
```
Measurity: Use `nameof` to ease with future refactoring work

NitroxClient/GameLogic/Spawning/Metadata/Processor/OxygenMetadataProcessor.cs
```diff
@@ -0,0 +1,22 @@
...
+public class OxygenMetadataProcessor : EntityMetadataProcessor<OxygenMetadata>
+{
+    public override void ProcessMetadata(GameObject gameObject, OxygenMetadata metadata)
+    {
+        Oxygen oxygen = gameObject.GetComponent<Oxygen>();
+
+        if (oxygen)
+        {
+            oxygen.oxygenAvailable = metadata.OxygenAvailable;
+        }
```
Measurity: Invert if to turn it into a [Guard clause](https://en.wikipedia.org/wiki/Guard_(computer_science))

NitroxClient/MonoBehaviours/OxygenTankBroadcaster.cs
```diff
@@ -0,0 +1,89 @@
...
+
+    public void Update()
+    {
+        time += Time.deltaTime;
+
+        if (time >= BROADCAST_INTERVAL)
+        {
+            time = 0;
+            SyncOxygenTanks();
+        }
```
Measurity: Player oxygen is already synced so constantly updating the tank value isn't necessary. If we can sync the value when an oxygen tank is dropped or moved into a non-player inventory/container , this should be enough and save on CPU cycles.

Measurity (changes requested):

Measurity (changes requested):
Entity metadata updates now appear on every frame when playing with 2 players. Looks like oxygen add / remove are called too often and not just for oxygen tanks.

NitroxPatcher/Patches/Dynamic/Oxygen_AddOxygen_Patch.cs
```diff
@@ -0,0 +1,35 @@
...
+        {
+            return;
+        }
+
+        // Only sync if oxygen level changed by at least 1 unit (avoid spam from continuous small changes)
+        if (Math.Abs(Math.Floor(__instance.oxygenAvailable) - Math.Floor(__state)) > 0.0 &&
+            __instance.TryGetIdOrWarn(out NitroxId id))
+        {
+            Resolve<Entities>().EntityMetadataChanged(__instance, id);
+        }
```
Measurity: I think this is not the right way to sync the oxygen.

1. Doing a `GetComponent` every frame is to be avoided.
2. We're still updating the oxygen value more than necessary.

Oxygen tanks only change value when the player actively consumes it afaik. So we should find the place where the player unequips to oxygen tank and only then sync the latest value with the server.

Measurity (approved):
LGTM and works in game. No spammy network traffic either GJ!

Nitrox.Model.Subnautica/DataStructures/GameLogic/Entities/Metadata/EntityMetadata.cs
```diff
@@ -45,6 +45,7 @@ namespace Nitrox.Model.Subnautica.DataStructures.GameLogic.Entities.Metadata
     [ProtoInclude(86, typeof(GenericConsoleMetadata))]
     [ProtoInclude(87, typeof(BlueprintHandTargetMetadata))]
     [ProtoInclude(88, typeof(PrecursorDisableGunTerminalMetadata))]
+    [ProtoInclude(89, typeof(OxygenMetadata))]
```
dartasen, line 48: I think it should be added in world save unit tests

NitroxPatcher/Patches/Dynamic/Equipment_RemoveItem_Patch.cs
```diff
@@ -12,6 +13,13 @@ public static void Postfix(Equipment __instance, InventoryItem __result)
         if (__result != null)
         {
             Resolve<EquipmentSlots>().BroadcastUnequip(__result.item, __instance.owner);
+
+            // Sync oxygen tank metadata when unequipped
+            if (__result.item.TryGetComponent(out Oxygen oxygen) && 
+                __result.item.TryGetIdOrWarn(out NitroxId id))
+            {
+                Resolve<Entities>().EntityMetadataChanged(oxygen, id);
+            }
```
dartasen: Such code should be located in the Client imo, to avoid bloating patch code, what do you think ?

dartasen (approved):
Thank you 😄

Coding-Hen (approved):
LGTM CW

PR #2656, opened 2026-01-21
Asyncify command and packet handling

Measurity:
- [x] Use `SessionId` instead of `string` (player name) to refer to a connected player in all code
- [x] Refactor disconnect data handling using internal event. Implement via `ISessionCleaner` interface
- [x] Convert commands to an `async` system
- [x] New command system based on types with attribute-based auth, descriptions and command name auto completion
- [x] Convert packets to an `async` system
- [x] Persist config changes to server.cfg file (on save) - see `PersistNitroxSerializableConfigService`
- [x] Fix RNG not being deterministic (move away from using `GetHashCode` calls)

PR #2726, opened 2026-05-12
Prevent duplicate entries in ScannerComplete

### LINKED ISSUE #2666: It's possible for BaseRoom (and other) tech type to reset on server restart, likely conflict when both BaseRoomFragment and BaseRoom are scanned

1.8.1.0 (Latest release)

### Details about version

_No response_

Windows

Steam

### Describe the issue.

After completing a scanned item that has fragments, it's possible for save to get in a state where every server restart the tech type will reset. Rescanning this tech type does mark it "completed" on client. But a server restart will require rescanning every time.

It's possible that modded clients can mess up this server state. But regardless if user error, server should not ever get in a corrupted state.

#### Things to check
- Race condition in saving scanned tech type on server: Check why it's possible for the same "tech type" to be in the same scanner data lists on server and prevent it.
- Verify server logic of both fragment and "one shot" scanned tech type, like `BaseRoom` and `BaseRoomFragment`.

Image

_No response_

### Relevant log output

_No response_

misterbubb:
Fixes #2666

## Problem
Fully scanned tech types could reset on server restart, requiring players to rescan them after each restart.

## Root Cause
In #2656, duplicate prevention was added to all `PdaManager` methods except `UpdateEntryUnlockedProgress`. This allowed duplicate entries in `ScannerComplete`, corrupting the save state and causing tech types to reset.

Additionally, a client-side check was added that blocked ALL completion packets from being sent, not just duplicates, because it checked after the vanilla game had already added the item to the complete list.

## Solution
- **Server-side**: Added duplicate checking to `UpdateEntryUnlockedProgress` when adding to `ScannerComplete`, matching the pattern used in other `PdaManager` methods
- **Client-side**: Removed the problematic check that was blocking all completion packets. The `alreadyComplete` parameter already prevents duplicate packets correctly

The fix prevents duplicate entries from race conditions, client desync, or malformed packets.

## Changes
- Added `Contains` check before adding to `ScannerComplete` in `PdaManager.UpdateEntryUnlockedProgress`
- Removed client-side check that was blocking all completion packets in `PDAScanner_Scan_Patch`

### DIFF
--- a/Nitrox.Server.Subnautica/Models/GameLogic/PdaManager.cs
+++ b/Nitrox.Server.Subnautica/Models/GameLogic/PdaManager.cs
@@ -153,7 +153,14 @@ internal sealed class PdaManager(ILogger<PdaManager> logger) : ISummarize
             if (fullyResearched)
             {
                 PdaState.ScannerPartial.RemoveAllFast(techType, static (entry, toRemove) => entry.TechType.Equals(toRemove));
-                PdaState.ScannerComplete.Add(techType);
+                if (!PdaState.ScannerComplete.Contains(techType))
+                {
+                    PdaState.ScannerComplete.Add(techType);
+                }
+                else
+                {
+                    logger.ZLogDebug($"Tech type [{techType.Name}] is already in {nameof(PdaState.ScannerComplete)}, skipping duplicate add");
+                }
             }
             else
             {
--- a/Nitrox.Server.Subnautica/Models/GameLogic/PdaManager.cs
+++ b/Nitrox.Server.Subnautica/Models/GameLogic/PdaManager.cs
@@ -153,7 +153,10 @@ internal sealed class PdaManager(ILogger<PdaManager> logger) : ISummarize
             if (fullyResearched)
             {
                 PdaState.ScannerPartial.RemoveAllFast(techType, static (entry, toRemove) => entry.TechType.Equals(toRemove));
-                PdaState.ScannerComplete.Add(techType);
+                if (!PdaState.ScannerComplete.Contains(techType))
+                {
+                    PdaState.ScannerComplete.Add(techType);
+                }
             }
             else
             {
--- a/Nitrox.Server.Subnautica/Models/GameLogic/PdaManager.cs
+++ b/Nitrox.Server.Subnautica/Models/GameLogic/PdaManager.cs
@@ -153,14 +153,7 @@ internal sealed class PdaManager(ILogger<PdaManager> logger) : ISummarize
             if (fullyResearched)
             {
                 PdaState.ScannerPartial.RemoveAllFast(techType, static (entry, toRemove) => entry.TechType.Equals(toRemove));
-                if (!PdaState.ScannerComplete.Contains(techType))
-                {
-                    PdaState.ScannerComplete.Add(techType);
-                }
-                else
-                {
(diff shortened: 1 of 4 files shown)


Nitrox.Server.Subnautica/Models/GameLogic/PdaManager.cs
```diff
@@ -153,7 +153,14 @@ public void UpdateEntryUnlockedProgress(NitroxTechType techType, int unlockedAmo
...
             {
                 PdaState.ScannerPartial.RemoveAllFast(techType, static (entry, toRemove) => entry.TechType.Equals(toRemove));
-                PdaState.ScannerComplete.Add(techType);
+                if (!PdaState.ScannerComplete.Contains(techType))
+                {
+                    PdaState.ScannerComplete.Add(techType);
+                }
+                else
+                {
+                    logger.ZLogDebug($"Tech type [{techType.Name}] is already in {nameof(PdaState.ScannerComplete)}, skipping duplicate add");
```
Coding-Hen: Not sure we need this log. I think personally we are okay to silently skip it

Coding-Hen (changes requested):
This is a good guard against this but I wonder if we should be tackling this from the other side of ensuring we don't end up with the ability to have 2 identical trying to add to the list in the first place

Measurity (changes requested):
`PDAScanner.complete` is added to before the check happens. Now no completed packets get received by server at all.

Nitrox.Server.Subnautica/Models/GameLogic/PdaManager.cs
```diff
@@ -153,7 +153,10 @@ public void UpdateEntryUnlockedProgress(NitroxTechType techType, int unlockedAmo
             if (fullyResearched)
             {
                 PdaState.ScannerPartial.RemoveAllFast(techType, static (entry, toRemove) => entry.TechType.Equals(toRemove));
-                PdaState.ScannerComplete.Add(techType);
+                if (!PdaState.ScannerComplete.Contains(techType))
+                {
+                    PdaState.ScannerComplete.Add(techType);
+                }
```
Measurity: We can change `ScannerComplete` to a HashSet<T> to avoid duplicate entries. But have to check if this is maintains backwards compat with the save file format.

Measurity (approved):
LGTM

Nitrox.Server.Subnautica/Models/GameLogic/Unlockables/PdaStateData.cs
```diff
@@ -59,7 +59,7 @@ internal sealed record PdaStateData
     ///     Fully unlocked PDA entries
     /// </summary>
     [DataMember(Order = 7)]
-    public List<NitroxTechType> ScannerComplete { get; } = [];
+    public HashSet<NitroxTechType> ScannerComplete { get; } = [];
```
Coding-Hen, line 62: Have we tested this to ensure backwards compatibility in the save?

Coding-Hen (approved):
LGTM CW, as long as it's not a breaking change to save files. Otherwise we should add a shim

Nitrox.Server.Subnautica/Models/GameLogic/Unlockables/PdaStateData.cs
Measurity: Yes I've tested it. Works on older saves, new scans also unlock and save properly (also same JSON format as before).

PR #2730, opened 2026-05-14
Use named pipes for GRPC on Windows for faster IPC

### LINKED ISSUE #2651: Server output in launcher takes ~2 seconds to appear on Windows
## Context

Due to changes made in the PR below, the server and launcher now use socket-based GRPC to communicate and see logs. On Linux, connecting to a GRPC socket is quite fast (~200ms) but on Windows it takes 10x as long.

- https://github.com/SubnauticaNitrox/Nitrox/pull/2314

## How to fix

**Only for Windows**, the (GRPC) socket transport should be changed to a (GRPC) named pipe transport. 

On Linux, named pipes are limited in functionality (no bi-directional communication) so it should remain socket-based GRPC.

> [!NOTE]
> The Nitrox Launcher is the GRPC server here and Nitrox Server are the GRPC client. This is because the Laucher can manage multiple Nitrox Servers.

[See the MSDN documentation on how to setup a GRPC connection via named pipes.](https://learn.microsoft.com/en-us/aspnet/core/grpc/interprocess-namedpipes#server-configuration)

misterbubb:
Fixes #2651

Changes:
- Use named pipes for GRPC communication on Windows instead of sockets
- Keep socket-based GRPC on Linux (named pipes have limited functionality there)
- Improves server output connection time from ~2 seconds to ~200ms on Windows

Testing:
- Tested on Windows (server output appears almost immediately in launcher but could use more testing)

### DIFF
--- a/Nitrox.Server.Subnautica/Services/ServersManagementService.cs
+++ b/Nitrox.Server.Subnautica/Services/ServersManagementService.cs
@@ -1,6 +1,8 @@
 using System.Buffers;
 using System.IO;
+using System.IO.Pipes;
 using System.Linq;
+using System.Net.Http;
 using System.Threading.Channels;
 using Grpc.Core;
 using Grpc.Net.Client;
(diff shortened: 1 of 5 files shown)


Nitrox.Server.Subnautica/Services/ServersManagementService.cs
```diff
@@ -40,47 +46,105 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
...
+                        pushLogsTask = CreateLoopingTask(PushLogsAsync, api, stoppingToken);
+                    }
                 }
             }
             catch (Exception ex)
             {
                 if (!ShouldIgnoreException(ex))
                 {
-                    logger.ZLogTrace($"{ex.Message}");
+                    logger.ZLogError(ex, $"Error in ServersManagementService");
```
Measurity: This is not an error but expected to happen occasionally. Should remain trace as it's connection related and not interesting for user but only for developers.

Measurity (approved):
Needs testing on Windows.

LGTM and works on Linux.

dartasen (approved):

PR #2738, opened 2026-05-16
Follow the XDG spec on Linux/MacOS (Users must move save files to new place when merged)

### LINKED ISSUE #2654: Follow XDG spec on Linux
### Describe the issue

https://specifications.freedesktop.org/basedir/latest/

If the `--data-path` isn't specified, we should follow XDG spec by reading the XDG environment variables (see spec) or fallback to the default paths for those variables.

BeefyAndTheDucks:
This resolves issue #2654 

Any feedback on my implementation is appreciated.

### DIFF
--- a/Nitrox.Launcher/Models/Design/ServerEntry.cs
+++ b/Nitrox.Launcher/Models/Design/ServerEntry.cs
@@ -83,7 +83,7 @@ internal sealed partial class ServerEntry : ObservableObject
     public partial int MaxPlayers { get; set; } = serverDefaults.MaxConnections;
 
     [ObservableProperty]
-    public partial string? Name { get; set; }
+    public required partial string Name { get; set; } = "";
 
     [ObservableProperty]
     public partial string? Password { get; set; }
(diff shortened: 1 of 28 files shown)


Measurity (approved):
Made a few changes to centralize into NitroxDirectory type to reduce complexity of NitroxUser. I kept your XDG spec implementation.

dartasen:
As a (minimal) migration path, I think it would be great if we could add a warning on top of server pages for Linux players that may have saves in an old data path

BeefyAndTheDucks:
I've added the ugliest warning ever for when a user might need to migrate. Someone with more UI skills than me can make it look a bit better :)

Nitrox.Launcher/Models/Design/ServerEntry.cs
```diff
@@ -334,7 +334,7 @@ public async Task StopAsync()
     [RelayCommand(CanExecute = nameof(CanOpenSaveFolder))]
     public void OpenSaveFolder()
     {
-        OpenDirectory(Path.Combine(KeyValueStore.Instance.GetSavesFolderDir(), Name!));
+        OpenDirectory(Path.Combine(KeyValueStore.Instance.GetSavesPath(), Name!));
```
dartasen: Not related to that PR, but I can't help noticing the usage of the damn-it operator 👀

Nitrox.Launcher/ViewModels/MainWindowViewModel.cs
```diff
@@ -105,6 +106,20 @@ IKeyValueStore keyValueStore
                 LauncherNotifier.Info("You're now using Nitrox DEV build");
             }
 
+            // Only on Linux we allow changing save paths via XDG spec.
+            if (OperatingSystem.IsLinux())
```
dartasen: Note: MacOS should work with the XDG spec as well

Measurity: Added support for macOS now but is untested.

dartasen (approved):
LGTM for MacOS

PR #2740, opened 2026-05-17
Restore Nautilus GameInput compatibility

mindless2831:
## Summary

Restores compatibility between Nitrox and newer Nautilus builds that use Nautilus's Subnautica GameInput integration.

Nautilus 1.0.0-pre.45+ introduced a new GameInput/custom button system. With Nitrox, this caused the game to reach the title/menu scene but fail to populate the expected menu/UI state, making Nitrox effectively incompatible with newer Nautilus builds.

This PR applies the focused GameInput/keybind compatibility work from JackNytely's PR #2663, excluding the unrelated launcher/server changes from that PR.

Full credit for the Nautilus GameInput compatibility work goes to JackNytely and PR #2663. This PR separates the input compatibility pieces into a smaller, focused change set.

## What changed

- Makes Nitrox's GameInput button registration compatible with Nautilus's custom GameInput button handling.
- Ensures Nitrox buttons are safely added without assuming Nitrox is the only system extending GameInput.Button / GameInput.AllActions.
- Adds duplicate-safe handling around GameInput action registration.
- Preserves support for Nautilus's new input labels, keyboard shortcuts, and assignment UI.
- Fixes a server-side spawning exception caused by UwePrefab being declared as a readonly record struct while spawning code expects to mutate Probability.

## UwePrefab fix

During multiplayer world sync, the server repeatedly threw:

System.MissingMethodException: Method not found:

'Void Nitrox.Model.Subnautica.DataStructures.GameLogic.Entities.UwePrefab.set\_Probability(Single)'.

### DIFF
--- a/Nitrox.Launcher/Program.cs
+++ b/Nitrox.Launcher/Program.cs
@@ -75,7 +75,7 @@ internal static class Program
 
                 try
                 {
-                    return Assembly.LoadFile(dllPath);
+                    return Assembly.LoadFrom(dllPath);
                 }
                 catch
                 {
@@ -88,7 +88,14 @@ internal static class Program
                 cache[args.Name] = assembly = ResolveFromLib(args.Name);
                 if (assembly == null && !args.Name.Contains(".resources"))
                 {
-                    cache[args.Name] = assembly = Assembly.Load(args.Name);
+                    try
+                    {
+                        cache[args.Name] = assembly = Assembly.Load(args.Name);
+                    }
+                    catch
+                    {
+                        cache[args.Name] = assembly = null;
+                    }
                 }
             }
 
--- a/Nitrox.Model.Subnautica/DataStructures/GameLogic/Entities/UwePrefab.cs
+++ b/Nitrox.Model.Subnautica/DataStructures/GameLogic/Entities/UwePrefab.cs
@@ -1,3 +1,3 @@
 namespace Nitrox.Model.Subnautica.DataStructures.GameLogic.Entities;
 
-public readonly record struct UwePrefab(string ClassId, int Count, float Probability, bool IsFragment);
+public record struct UwePrefab(string ClassId, int Count, float Probability, bool IsFragment);
--- a/Nitrox.Model.Subnautica/DataStructures/GameLogic/Entities/UwePrefab.cs
+++ b/Nitrox.Model.Subnautica/DataStructures/GameLogic/Entities/UwePrefab.cs
@@ -1,3 +1,3 @@
 namespace Nitrox.Model.Subnautica.DataStructures.GameLogic.Entities;
 
-public record struct UwePrefab(string ClassId, int Count, float Probability, bool IsFragment);
+public readonly record struct UwePrefab(string ClassId, int Count, float Probability, bool IsFragment);
--- a/NitroxClient/MonoBehaviours/Gui/Input/KeyBindings/Actions/ChatKeyBindingAction.cs
+++ b/NitroxClient/MonoBehaviours/Gui/Input/KeyBindings/Actions/ChatKeyBindingAction.cs
@@ -5,7 +5,7 @@ namespace NitroxClient.MonoBehaviours.Gui.Input.KeyBindings.Actions;
 
 public class ChatKeyBindingAction : KeyBinding
 {
-    public ChatKeyBindingAction() : base("Nitrox_Settings_Keybind_OpenChat", "y") { }
+    public ChatKeyBindingAction() : base("Nitrox_Settings_Keybind_OpenChat", "y", "dpad/up") { }
 
     public override void Execute(InputAction.CallbackContext _)
     {
--- a/NitroxClient/MonoBehaviours/Gui/Input/KeyBindings/Actions/DiscordFocusBindingAction.cs
+++ b/NitroxClient/MonoBehaviours/Gui/Input/KeyBindings/Actions/DiscordFocusBindingAction.cs
@@ -6,7 +6,7 @@ namespace NitroxClient.MonoBehaviours.Gui.Input.KeyBindings.Actions;
 
 public class DiscordFocusBindingAction : KeyBinding
 {
-    public DiscordFocusBindingAction() : base("Nitrox_Settings_Keybind_FocusDiscord", "i") { }
+    public DiscordFocusBindingAction() : base("Nitrox_Settings_Keybind_FocusDiscord", "i", "dpad/down") { }
 
     public override void Execute(InputAction.CallbackContext _)
     {
--- a/NitroxPatcher/Patches/Persistent/Enum_GetValues_Patch.cs
+++ b/NitroxPatcher/Patches/Persistent/Enum_GetValues_Patch.cs
@@ -6,7 +6,9 @@ using NitroxClient.MonoBehaviours.Gui.Input;
 namespace NitroxPatcher.Patches.Persistent;
 
 /// <summary>
-/// Specific patch for GameInput.Button enum to return also our own values.
+/// Extends Enum.GetValues() to include Nitrox buttons, which is required for the rebinding UI to recognize them as valid.
+/// Duplicate checking ensures compatibility when Nautilus or other mods also extend the enum.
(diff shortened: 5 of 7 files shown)


Nitrox.Launcher/Program.cs
```diff
@@ -88,7 +88,14 @@ private static class AssemblyResolver
...
                 if (assembly == null && !args.Name.Contains(".resources"))
                 {
-                    cache[args.Name] = assembly = Assembly.Load(args.Name);
+                    try
+                    {
+                        cache[args.Name] = assembly = Assembly.Load(args.Name);
+                    }
+                    catch
+                    {
+                        cache[args.Name] = assembly = null;
```
Coding-Hen, line 97: If the try fails will assembly not still be null and therefore the cache will be null for that entry too already by the previous assignment making this redundant?

Coding-Hen (changes requested):

Nitrox.Launcher/Program.cs
mindless2831: It does not appear so as this fixes the Uwe exception that is thrown during multiplayer sync by that exactly problem.

Nitrox.Model.Subnautica/DataStructures/GameLogic/Entities/UwePrefab.cs
```diff
@@ -1,3 +1,3 @@
 namespace Nitrox.Model.Subnautica.DataStructures.GameLogic.Entities;
 
-public readonly record struct UwePrefab(string ClassId, int Count, float Probability, bool IsFragment);
```
dartasen, line 3: Could you please open a separate PR for that ? 🙏

If you're able to provide more context/logs as well, as I don't seem to have any errors related to that on my end

Nitrox.Launcher/Program.cs
Coding-Hen: I'm referring to line 88, does this not set assembly to null if the method call returns null, then used on 89 to get this code path, assuming that the try above exits to here in an exception it will still be null? And therefore not need resetting here

mindless2831: I apologize. I was taking about a different piece of code entirely in a completely different file. This is why you should never answer questions about code you've been working at all day at 3 am when trying to go to sleep lol. I will check this out right now and see what's up, but i totally think you're right. I probably left that when I was trying else and never removed the try catch block.

mindless2831: Ok, you are totally right and I am dumb. Since `cache[args.Name] = assembly = ResolveFromLib(args.Name)` already stores null when ResolveFromLib returns null, setting `cache[args.Name] = assembly = null` again in the catch is redundant. I fixed it and just return null. Update has been pushed.

Nitrox.Model.Subnautica/DataStructures/GameLogic/Entities/UwePrefab.cs
```diff
@@ -1,3 +1,3 @@
...
+public record struct UwePrefab(string ClassId, int Count, float Probability, bool IsFragment);
```
Measurity: This change doesn't look like it has a purpose atm. Let's undo.

Nitrox.Launcher/Program.cs
```diff
@@ -75,7 +75,7 @@ private static class AssemblyResolver
 
                 try
                 {
-                    return Assembly.LoadFile(dllPath);
+                    return Assembly.LoadFrom(dllPath);
```
Measurity, line 78: Think we want to keep this `LoadFile` so that we get a chance to resolve dependencies ourselves. Otherwise it might load dependencies that cause (hidden) incompatibilities by loading slightly different assembly versions.

[See MSDN docs for more info.](https://learn.microsoft.com/en-us/dotnet/api/system.reflection.assembly.loadfile?view=netframework-4.8.1&viewFallbackFrom=net-10.0#remarks)

> Use the [LoadFile](https://learn.microsoft.com/en-us/dotnet/api/system.reflection.assembly.loadfile?view=netframework-4.8.1) method to load and examine assemblies that have the same identity, but are located in different paths. [LoadFile](https://learn.microsoft.com/en-us/dotnet/api/system.reflection.assembly.loadfile?view=netframework-4.8.1) does not load files into the load-from context, and does not resolve dependencies using the load path, as the [LoadFrom](https://learn.microsoft.com/en-us/dotnet/api/system.reflection.assembly.loadfrom?view=netframework-4.8.1) method does. [LoadFile](https://learn.microsoft.com/en-us/dotnet/api/system.reflection.assembly.loadfile?view=netframework-4.8.1) is useful in this limited scenario because [LoadFrom](https://learn.microsoft.com/en-us/dotnet/api/system.reflection.assembly.loadfrom?view=netframework-4.8.1) cannot be used to load assemblies that have the same identities but different paths; it will load only the first such assembly.

dartasen: Also, what does this code fix ? Could you please provide more context

Nitrox.Model.Subnautica/DataStructures/GameLogic/Entities/UwePrefab.cs
mindless2831: This change fixes a UWE exception that is thrown in the server log window due to it being mutable.

Nitrox.Model.Subnautica/DataStructures/GameLogic/Entities/UwePrefab.cs
mindless2831: I seriously thought about adding that in a different one and should have done it, but was such a small change just to get rid of the UWE exception in the server command window that a lot of useless noise. I can create another PR and separate it once I upload the server close fix here in a minute.

Nitrox.Launcher/Program.cs
mindless2831: That concern makes sense, and I agree we do not want the launcher accidentally resolving incompatible dependency versions.

The reason I changed this is that the LoadFile behavior appears to be the source of the runtime mismatch we were seeing in Release builds. In practice, it can load an assembly outside the normal load context, which means another assembly with the same identity can still be loaded elsewhere. That was causing type/API-surface mismatches, including the UwePrefab/Nitrox.Model.Subnautica MissingMethodException during world sync.

So the issue was not that we needed less control over dependency loading; it was that LoadFile gave us too much isolation and allowed duplicate/mismatched assemblies with the same identity to exist at runtime.

LoadFrom still loads from the explicit path selected by this resolver, so we are still preferring the intended lib/launcher-local assembly path. The important difference is that it participates in the load-from context, so subsequent binds to the same assembly identity resolve consistently instead of creating a hidden duplicate assembly situation.

That said, I understand the concern about broadening dependency resolution. If preferred, I can adjust this to be more targeted, but the purpose of the change is specifically to prevent the Release build from loading mismatched Nitrox assemblies and producing runtime MissingMethodExceptions.

Nitrox.Model.Subnautica/DataStructures/GameLogic/Entities/UwePrefab.cs
mindless2831: Agreed. I removed this from this PR so the current PR stays focused.

I originally included it because I was seeing a server-side sync exception involving `UwePrefab.Probability` during multiplayer world sync, and removing `readonly` cleared that runtime exception in my local testing. I’ll open a separate PR with the exact context/logs for that change.

mindless2831:
I split the Release launcher dependency-loading fix into its own PR: #2749.

This PR currently builds on top of that work, so some shared Release/lib-loading changes may appear here until #2749 is merged. Once #2749 lands, I can rebase/update this branch so this PR only shows the Nautilus/GameInput compatibility changes.

The dependency is intentional because the Release/lib-loading fix is needed for clean Release-build testing: without it, existing servers may not show and new server creation can fail/crash due to mismatched dependency loading from `lib`.

mindless2831:

Nitrox.Launcher/Program.cs
Coding-Hen: Has this been changed it looks the same as before to me? We are still resetting it to null?

PR #2735, opened 2026-05-15
Fixed EGS killing the game because it's not authenticated

Measurity:
**Nvm: the issue was EGS platform wasn't detected because the file structure changed. Closing this PR.**

___

> [!IMPORTANT]
> Needs testing with EGS!

This likely solves the issue, but not sure yet. The `-epicuserid=0` was added to fix the timecapsules submit. But timecapsules are a degraded feature of Subnautica so fixing the crashing is more important.

In Discord we ask users to follow this guide as a workaround:

> 1. If you're using the **Epic Games** version and Subnautica crashes when you open it through Nitrox, refer to the following:
> 2. In the Nitrox launcher, click on the **Options** tab.
> 3. Now in the **Subnautica Launch Arguments** section, add the following:
>     ```
>     -EpicPortal -vrmode none
>     ```
> See if it works. If it doesn't, ask in the ⁠[💡help-channel💡](https://discord.com/channels/525437013403631617/525438387105038346) .
>
> _Credits to @BairWithMe for finding this fix_

PR #2742, opened 2026-05-17
Fix EGS Discovery

### LINKED ISSUE #123: [Misc] Disable ability to pause while in multiplayer.

dartasen:
## Context

Context : [Discord](https://discord.com/channels/525437013403631617/1504387511139700876)
Supersedes #2735 

## Description of Change

`OwnsGame` was checking inside .egstore folder for a determined number of files, however it seems to be unstable across updates because this folder might contain download cache, dead files, ...

However it (should) always contain the game manifest at least. Which is what this PR is looking for.

We could go way deeper in this check though, this file is a custom binary format, but it can be easily read with some tools like [https://github.com/meszmate/manifest](https://github.com/meszmate/manifest). The minimal check would be to check the magic number at the beginning of each manifest (Epic manifest files seems to always start with `0x44BEC00C`.  We could also check the entries to find AppId et AppName in the manifest
⚠️ However that would make us depend on the binary format, which is undocumented so it can be changed (unlikely but still)

## PR Checklist

- [ ] Has tests for the changes
- [x] Has a linked issue
- [x] Has been tested in-game
- [x] Has been tested on Windows & macOS

---

### DIFF
--- a/Nitrox.Model/Platforms/Store/EpicGames.cs
+++ b/Nitrox.Model/Platforms/Store/EpicGames.cs
@@ -1,4 +1,4 @@
-using System.Collections.Generic;
+using System;
 using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
@@ -17,13 +17,16 @@ public sealed class EpicGames : IGamePlatform
     public bool OwnsGame(string gameDirectory)
     {
         string path = Path.Combine(gameDirectory, ".egstore");
-        if (!Directory.Exists(path))
+        
+        try
         {
+            return Directory.EnumerateFiles(path, "*.manifest", SearchOption.TopDirectoryOnly).Any();
+        }
+        catch (Exception ex)
+        {
+            Log.Error(ex);
             return false;
         }
-
-        IEnumerable<string> manifestFiles = Directory.EnumerateFiles(path, "*.manifest", SearchOption.TopDirectoryOnly);
-        return manifestFiles.Any();
     }
 
     public static async Task<ProcessEx> StartGameAsync(string pathToGameExe, string launchArguments)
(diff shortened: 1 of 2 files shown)


Nitrox.Model/Platforms/Store/EpicGames.cs
```diff
@@ -15,7 +17,13 @@ public sealed class EpicGames : IGamePlatform
...
     {
         string path = Path.Combine(gameDirectory, ".egstore");
-        return Directory.Exists(path) && Directory.GetFiles(path).Length > 1;
+        if (!Directory.Exists(path))
+        {
+            return false;
+        }
+
+        IEnumerable<string> manifestFiles = Directory.EnumerateFiles(path, "*.manifest", SearchOption.TopDirectoryOnly);
+        return manifestFiles.Any();
```
Measurity: For IO it's better to let the file access fail and catch it as there are many things that can go wrong (even the directory being removed _just after_ the Directory.Exists check, although unlikely).

```suggestion
        string path = Path.Combine(gameDirectory, ".egstore");
        try
        {
            return Directory.EnumerateFiles(path, "*.manifest", SearchOption.TopDirectoryOnly).Any();
        }
        catch (Exception ex)
        {
            Log.Error(ex);
            return false;
        }
```

dartasen: > even the directory being removed just after the Directory.Exists check, although unlikely

Thank's to all the years working on Nitrox, I know that this will happen for at least 1 user 😂

Measurity (commented):
Let's fix `Discord.OwnsGame` as you found and we discussed on Discord.

```cs
return File.Exists(Path.Combine(Path.GetDirectoryName(Path.GetFullPath(gameDirectory)) ?? "..", "journal.sqlite"));
```

Nitrox.Model/Platforms/Store/EpicGames.cs
```diff
@@ -1,4 +1,7 @@
+using System;
+using System.Collections.Generic;
```
Measurity: `using System.Collections.Generic;` is redundant.

Measurity (approved):
LGTM

PR #2747, opened 2026-05-18
Fix launcher server stop command

mindless2831:
## Summary

Fixes a regression where the Nitrox Launcher Stop button could enter the stopping state but fail to gracefully shut down an externally launched server. The server process could remain running until the console window was closed manually.

This also fixes the Visual Studio Release deployment layout for the launcher output so net10/root-lib assemblies are not overwritten by net472-compatible assets.

## Background

This issue appears related to prior work in JackNytely's PR #2663, which included a commit titled "Fix server shutdown hanging and keybind rebinding being non-functional" (`3b956de`). That PR also touched Release build dependency layout behavior.

Credit to @JackNytely for the prior investigation and attempted server shutdown/build-layout work in PR #2663.

## Why the Release/lib deployment fix is included here

The build/deployment fix is included in this PR because it is integral to the Stop button fix, not a separate cosmetic or packaging-only cleanup.

The launcher Stop button now sends `quit` through the active server management command path. That path depends on the external server successfully connecting back to the launcher through the MagicOnion/gRPC management stream.

During testing, the Stop command still failed even after the launcher logic was corrected because the Visual Studio Release output could place incompatible assemblies in root `lib`. In particular:

- root `lib/Grpc.Core.Api.dll` could be the lower-TFM asset instead of the netstandard2.1 asset required by the net10/MagicOnion path;
- root `lib/Nitrox.Model.dll` could be the net472 build instead of the net10 build.

That caused runtime `MissingMethodException`s in the server management stream, preventing `ServersManagement.OnConnected` from completing and leaving the launcher's command queue with no active reader. In that state, the Stop button could queue `quit`, but the command was never forwarded to the server.

Because the Stop button fix requires the management stream to be functional in normal Visual Studio Release output, the root `lib` asset correction is part of the same functional fix.

The intended deployment split remains:

- root `lib`: net10/server-facing assets;
- `lib/net472`: net472/game/patcher-facing assets.

## What changed

- Changed launcher Stop behavior so it sends `quit` through the active server command queue and waits for the server process to exit.
- Avoided using CTS cancellation as the mechanism that sends the quit command.
- Prevented `ResetCtsAsync` from replacing `CommandQueue` while the management command loop may still be reading it.
- Added a process-exit monitor so the launcher updates correctly when an external server console is closed manually.
- Pinned gRPC package versions used by MagicOnion.
- Added direct gRPC package references where needed so solution/Visual Studio builds do not resolve stale transitive versions.
- Fixed Visual Studio Release launcher output so:
  - root `lib/Nitrox.Model.dll` is the net10 build;
  - root `lib/Grpc.Core.Api.dll` is the netstandard2.1 asset;
  - `lib/net472` remains the net472-compatible location for game/patcher assets.

## Tested

Tested with a clean Visual Studio Release build copied to a deployment folder.

Verified:

- `lib/Nitrox.Model.dll` is the net10 build.
- `lib/net472/Nitrox.Model.dll` remains the net472 build.
- `lib/Grpc.Core.Api.dll` matches the NuGet `netstandard2.1` asset hash.
- External server starts from the launcher.
- Stop button sends `quit` through the management command path.
- Server shuts down cleanly.
- Launcher returns to Start after shutdown.
- Manually closing the external server window updates launcher state correctly.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -30,6 +30,10 @@
         <PackageVersion Include="dnlib" Version="4.4.0" />
         <PackageVersion Include="FluentAssertions.Analyzers" Version="[0.34.1]" />
         <PackageVersion Include="FluentAssertions" Version="[7.2.0]" />
+        <PackageVersion Include="Grpc.Core.Api" Version="2.66.0" />
+        <PackageVersion Include="Grpc.Net.Client" Version="2.66.0" />
+        <PackageVersion Include="Grpc.Net.Common" Version="2.66.0" />
+        <PackageVersion Include="Grpc.AspNetCore.Server" Version="2.66.0" />
         <PackageVersion Include="HanumanInstitute.MvvmDialogs.Avalonia" Version="2.1.0" />
         <PackageVersion Include="HarmonyX" Version="2.10.0" />
         <PackageVersion Include="JetBrains.Annotations" Version="2024.3.0" />
--- a/Nitrox.Launcher/Models/Design/ServerEntry.cs
+++ b/Nitrox.Launcher/Models/Design/ServerEntry.cs
@@ -182,7 +182,7 @@ internal sealed partial class ServerEntry : ObservableObject
         if (Process?.Id != processId)
         {
             await ResetCtsAsync();
-            Process = ServerProcess.Start(Path.Combine(KeyValueStore.Instance.GetSavesFolderDir(), Name), cts, ShouldAttachAsEmbedded(), processId);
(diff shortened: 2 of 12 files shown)


mindless2831:
This PR depends on #2749.

The Stop button fix requires the launcher/server management stream to connect reliably in Release output. During testing, the Stop logic could queue the shutdown command correctly, but the command was never forwarded when the Release build loaded mismatched root `lib` assemblies.

#2749 fixes the foundational Release `lib` dependency-loading issue. Once #2749 lands, I can rebase/update this branch so this PR stays focused on the server Stop command behavior and only carries the additional shutdown-specific changes.

mindless2831:

Measurity:
Is this PR still applicable now that the lib resolving issue has been resolved?

Measurity:
Closing for now

PR #2749, opened 2026-05-18
Fix Release launcher dependency loading from lib

mindless2831:
## Summary

Fixes Release launcher dependency loading so the launcher/server side resolves the correct assemblies from the Release output layout.

This PR is specifically about the **Release build/package output**, not Debug. In the broken Release output, the launcher/server path could load stale or incompatible assemblies from the wrong deployed location. That caused runtime `MissingMethodException`s even though the DLL files existed.

## Problem

The failure was not a simple missing-file issue. The runtime was finding assemblies, but some of them were the wrong version / wrong target asset for the code that was calling them.

The concrete mismatches observed were:

### `Grpc.Core.Api.dll`

The stale/bad loaded version was:

- `Grpc.Core.Api.dll` version `2.57.0`

The dependent library involved was:

- `MagicOnion 7.0.6`

The concrete missing method was:

- `Grpc.Core.IAsyncStreamWriter<T>.WriteAsync(T, CancellationToken)`

This means the runtime had loaded a gRPC API assembly, but not one with the API surface expected by the newer MagicOnion dependency.

### `Nitrox.Model.dll`

The launcher/server path also resolved a stale or wrong `Nitrox.Model.dll`.

The expected methods included:

- `Nitrox.Model.Serialization.NitroxConfig.Load(string, ILogger)`
- `Nitrox.Model.Serialization.NitroxConfig.Load<T>(string, ILogger)`

The failure mode was a `MissingMethodException` because the assembly that got loaded did not contain the overloads expected by the current launcher/server code.

## Release layout detail

The important layout issue is that the root `lib` side needs the current net10/server-facing assets, while `lib/net472` needs to keep the legacy/net472-facing assets.

The layout must not collapse those into one effective resolution pool where the launcher/server can accidentally load stale net472-era assets or incompatible dependency versions.

The observed distinction included:

- correct root `lib` asset: `Grpc.Core.Api.dll` from the newer `netstandard2.1` dependency asset path;
- correct root/server-facing `Nitrox.Model.dll`;
- separate legacy `lib/net472/Nitrox.Model.dll` for the net472-facing side.

The practical problem was that the Release deployment/layout allowed the launcher/server runtime path to resolve incompatible assemblies, producing `MissingMethodException`s instead of clear file-not-found errors.

## What this PR changes

This PR fixes the Release dependency layout/resolution behavior so the launcher/server side loads the intended assemblies from the Release output.

[...]

PR #2750, opened 2026-05-18
Fix UwePrefab probability setter

mindless2831:
## Summary

Fixes a server-side world sync / spawning runtime exception involving `UwePrefab.Probability`.

## Background

During multiplayer world sync and later cell / batch loading, the server repeatedly threw this exception:

> System.MissingMethodException: Method not found: 'Void Nitrox.Model.Subnautica.DataStructures.GameLogic.Entities.UwePrefab.set_Probability(Single)'.

The failure occurred in the entity spawning path:

- `BatchEntitySpawner.SpawnEntitiesUsingRandomDistributionAsync(...)`
- `BatchEntitySpawner.SpawnEntitiesAsync(...)`
- `BatchEntitySpawner.LoadBatchInternalAsync(...)`
- `WorldEntityManager.LoadAndRegisterBatchInternalAsync(...)`
- `CellVisibilityChangedProcessor.Process(...)`
- `LiteNetLibServer.ProcessPacket(...)`

## Why

`UwePrefab` is currently declared as a `readonly record struct`, but the server-side spawning / random distribution path expects a writable `Probability` API surface.

At runtime, the compiled caller attempts to use `set_Probability(Single)`, but that setter is not available with the current readonly struct shape. This causes a `MissingMethodException` during world sync / spawning.

Changing `UwePrefab` to a non-readonly `record struct` restores the expected setter-compatible API surface.

## What changed

- Changed `UwePrefab` from `readonly record struct` to `record struct`.

## Dependency / testing note

This was tested on top of the Release launcher dependency-loading fix because my current mod / test setup requires that fix to create and start servers reliably in a clean Release environment.

This change is otherwise isolated to `UwePrefab`.

## Tested

Tested using a fresh server / world after clearing and rebuilding the relevant prefab placeholder cache.

Before this change:

- The server repeatedly threw `MissingMethodException: UwePrefab.set_Probability(Single)`.
- The exception occurred during `BatchEntitySpawner`, `WorldEntityManager.LoadAndRegisterBatchInternalAsync`, and `CellVisibilityChangedProcessor` processing.

After this change:

- Created and started a fresh server / world.
- `PrefabPlaceholderGroupsResource` rebuilt successfully.
- Player joined and initial multiplayer sync completed.
- Initial spawn batches completed successfully.
- Traveled more than 200m to force additional cell / batch loading.
- Additional `BatchEntitySpawner` batches spawned successfully.
- No `UwePrefab`, `set_Probability`, or `MissingMethodException` error appeared during the same reproduction path.

### DIFF
--- a/Nitrox.Model.Subnautica/DataStructures/GameLogic/Entities/UwePrefab.cs
+++ b/Nitrox.Model.Subnautica/DataStructures/GameLogic/Entities/UwePrefab.cs
@@ -1,3 +1,3 @@
 namespace Nitrox.Model.Subnautica.DataStructures.GameLogic.Entities;
 
-public readonly record struct UwePrefab(string ClassId, int Count, float Probability, bool IsFragment);
+public record struct UwePrefab(string ClassId, int Count, float Probability, bool IsFragment);


mindless2831:
This PR depends on #2749 for clean Release-build server creation/testing.

The code change here is intentionally isolated to `UwePrefab`. The before/after testing was performed on top of the Release launcher dependency-loading fix because that fix was needed in my setup to create/start servers reliably and reproduce the world sync/spawning path.

mindless2831:

Nitrox.Launcher/Program.cs
```diff
@@ -88,7 +88,14 @@ private static class AssemblyResolver
...
                 {
-                    cache[args.Name] = assembly = Assembly.Load(args.Name);
+                    try
+                    {
+                        cache[args.Name] = assembly = Assembly.Load(args.Name);
+                    }
+                    catch
+                    {
+                        return null;
+                    }
```
dartasen: Still unrelated to the UwePrefab change 😄

mindless2831: Ugh, I am dumb, that was an accidental stacked-branch artifact from testing this fix on top of the Release launcher dependency-loading branch.

I rebased this PR onto current `master` and force-pushed it, so this PR should now contain only the `UwePrefab` change.

dartasen (approved):

Measurity:
Why is the compiler generating an impossible set_Probability call? Are we serializing UwePrefab and if so where?

Also would be nice to have a comment about it to avoid someone adding a `readonly` back. But ideally, we should get an error at compile time. 

When I revert this change and build a release, I don't get this missing method exception.

Something weird is happening here...

Issue #2501, opened 2025-10-11
Docker or podman image for NitroxServer-Subnautica


Kiaos:
### Describe the issue

I am in the process of making a dockerfile and trying to make a docker container that runs the server.

this is what i have so far.

#Dockerfile

``` # Base image: .NET 9.0 runtime for Linux
FROM mcr.microsoft.com/dotnet/runtime:9.0

# Set working directory
WORKDIR /app

# Download Nitrox 1.8.0.0
ARG NITROX_VERSION=1.8.0.0
ADD https://github.com/SubnauticaNitrox/Nitrox/releases/download/${NITROX_VERSION}/Nitrox_${NITROX_VERSION}_linux_x64.zip /tmp/nitrox.zip

# Install unzip, extract Nitrox, flatten linux-x64 folder
RUN apt-get update && apt-get install -y unzip \
    && unzip /tmp/nitrox.zip -d /app/Nitrox \
    && mv /app/Nitrox/linux-x64/* /app/Nitrox/ \
    && rmdir /app/Nitrox/linux-x64 \
    && rm /tmp/nitrox.zip \
    && apt-get clean && rm -rf /var/lib/apt/lists/*

# Expose Nitrox server ports
EXPOSE 11000/tcp
EXPOSE 11000/udp

# Set timezone
ENV TZ=America/Denver

# Set working directory to Nitrox folder
WORKDIR /app/Nitrox

# Run the server
ENTRYPOINT ["./NitroxServer-Subnautica"]


```

#podman-compose.yml

```
version: "3.8"

services:
  nitrox:
    build:
      context: .
      args:
        NITROX_VERSION: 1.8.0.0
    container_name: nitrox-server
    restart: unless-stopped
    ports:
      - "11000:11000/tcp"
      - "11000:11000/udp"
    environment:
      - TZ=America/Denver
      - SUBNAUTICA_PATH=/home/user/.local/share/Steam/steamapps/common/Subnautica
      - NITROX_CONFIG_PATH=/home/user/.config/Nitrox
    volumes:
      # Mount Nitrox config from host
      - ${NITROX_CONFIG_PATH}:/home/user/.config/Nitrox:Z

      # Mount Subnautica installation from host
      - ${SUBNAUTICA_PATH}:/home/user/.local/share/Steam/steamapps/common/Subnautica:Z
```

This is how far i got. Obviously my yml is different here cause im working with my own file system and server. But this is an example.

[...]

PR #2754, opened 2026-05-21
Compile with Docker

lammmab:
## Linked Issues / Context

Shadows #2501

## Description of Change

This adds compilation of Nitrox with Docker. This could help with DevOps and is great for containerization. The setup avoids "hacks" so this one is kind of a "one-and-done" type deal where this likely won't need modification again in the future. I added support for Nitrox.Launcher + Nitrox.Server

## Validation
Follow the Docker instructions in the [Documentation PR](https://github.com/SubnauticaNitrox/Documentation/pull/19) that should be merged along with this.

## PR Checklist

- [x] Has a linked issue
- [x] Works for cross-platform Launcher


lammmab:
I ended up going with extract-launcher.sh/bat for the Launcher as it retains cross-platform compatibility and is easy to use, especially since each script is like 10 lines. The .bat version will need testing by someone with a Windows machine. Otherwise, this PR is ready.

guffelman:
lgtm, tested on Linux. I suggest merge.

docker/Dockerfile.Launcher
```diff
@@ -0,0 +1,20 @@
...
+
+WORKDIR /src
+COPY . .
+
+RUN --mount=type=bind,from=subnautica,target=/subnautica \
+    SUBNAUTICA_INSTALLATION_PATH=/subnautica \
+    dotnet publish Nitrox.Launcher/Nitrox.Launcher.csproj \
+      -c Release \
+      -r ${RID} \
+      --self-contained ${SELF_CONTAINED} \
```
Coding-Hen, line 16: What is the reason for needing to specify the flag here, is it not false by default?

Additionally with scratch being used below and false being set here would it not require .net installed for the application to run?

docker/Dockerfile.Launcher
```diff
@@ -0,0 +1,20 @@
+FROM mcr.microsoft.com/dotnet/sdk:10.0 AS build
+
+ARG SELF_CONTAINED=false
+ARG RID=linux-x64
+
+RUN apt-get update && apt-get install -y mono-devel && rm -rf /var/lib/apt/lists/*
```
Coding-Hen, line 6: I'm not sure I understand removing the app lists here when we are not using this layer for runtime

docker/Dockerfile.Server
```diff
@@ -0,0 +1,25 @@
...
+FROM mcr.microsoft.com/dotnet/sdk:10.0 AS build
+
+WORKDIR /src
+COPY . .
+
+RUN --mount=type=bind,from=subnautica,target=/subnautica \
+    SUBNAUTICA_INSTALLATION_PATH=/subnautica \
+    dotnet publish Nitrox.Server.Subnautica/Nitrox.Server.Subnautica.csproj \
+    -c Release \
+    -r linux-x64 \
```
Coding-Hen, line 11: Are we enforcing linux-x64 here but the launcher doesn't have too?

Coding-Hen (changes requested):

docker/Dockerfile.Launcher
lammmab: Docker best practice, it's not as necessary for the launcher but still good hygiene

docker/Dockerfile.Launcher
lammmab: If it werent added then you cant control it without modifying the Dockerfile, no? Most C# programs expect .NET runtime installed, not requiring it is nonstandard and results in a larger build. Could use a note though

docker/Dockerfile.Server
lammmab: the server container runs on Linux, so `linux-x64` is hardcoded and correct

Measurity (changes requested):
Someone already made Nitrox docker compatible: https://github.com/Cuteminded/Nitrox-Server-Docker - Would you mind adjusting the PR to be more like this project but without the python / custom UI?

Can you tell me the goal of having Nitrox launcher in the docker compose file? What are users expected to do with the launcher?

lammmab:
Unsure what exactly you are referring to when you say to make it more like the other project - in what aspect?
And end-users are not expected to use the launcher dockerfile (and shouldn't). This is moreso for internal developers. For example, CI/CD if applicable, or fixing potential environment specific compilation issues. The server dockerfile would be specifically for VPS usage by those who understand what they're doing, but compiling from source there likely shouldn't be the default (which requires changes to the setup).

I would make these adjustments, and can eventually, but as of now I don't have the time. Sometime in the future I can come back to this if someone else doesn't pick up where I left off.

Jannify:
The launcher currently doesn't work without a display output (already tested this when making the AUR package). And since we don't have tests for the launcher or a mocking environment for the GUI I don't see the need for having it in the docker (for now).

dartasen:
Closing for now for inactivity. Feel free to reopen

PR #2757, opened 2026-05-21
Fix #2698: add circuit breaker to NTP sync to stop offline log spam

### LINKED ISSUE #2698: Implement a CircuitBreaker policy for NtpSync
When playing Nitrox without any internet connection (on a LAN), the NtpSync will spam server logs periodically

We should have a circuit breaker policy that stops/holds the sync when we fail to reach Ntp servers multiple time.

SkiLark1:
## Linked Issues / Context

Fixes #2698

## Description of Change

When a server runs without internet access (e.g. on a LAN), `NtpSyncer` retried all 6 NTP services every 60 seconds indefinitely, logging an error for each failed service. That spams the server logs forever.

This adds a circuit-breaker policy and fixes the underlying retry behavior. Three related changes:

1. **`NtpSyncer` — failover logs demoted to `Debug`.** Failing over between NTP services is normal operation (especially offline), not an error, so the two per-service `LogError` calls are now `LogDebug`. This removes the spam from production (`Info`+) logs entirely.

2. **`NtpSyncer.Setup` now resets `IsComplete`/`OnlineMode`.** This is the key fix: `Complete()` latched `IsComplete = true`, and `Setup()` never reset it. So after the first sequence completed, every subsequent `Complete()` on a retry round was a no-op — the finish callback was never invoked again, and the previous round's `NetManager` (`ntp`) leaked instead of being disposed. Resetting in `Setup` makes the syncer genuinely reusable across retry rounds (which `TimeService` already relied on).

3. **`TimeService` drives retries through a new `CircuitBreaker`.** After 3 consecutive failed rounds it logs a single informative line and widens the retry interval from 60s to 30min. The long interval doubles as a half-open probe, so the server still recovers automatically if it later gains connectivity — it just stops wasting resources and log lines while offline.

Adds `CircuitBreaker` (`Nitrox.Model.Networking`) — a small, thread-safe, timer-free failure counter — with unit tests.

## Validation

Tested locally on Windows, Nitrox InDev master:

1. Started a server with **no internet connection**.
2. Confirmed the NTP failover lines are now `Debug` level (gone from default `Info` logs).
3. Confirmed that after 3 failed rounds (~3 minutes) a single line appears:
   `Could not reach any NTP server after 3 attempts; assuming no internet connection and slowing NTP retries to every 1800s. This is expected on offline/LAN servers.`
4. Confirmed the retry cadence then drops from every 60s to every 30 min.

`CircuitBreaker` is covered by unit tests (`PatchesTranspilerTest`-style sanity is N/A; see `Nitrox.Test/Model/Networking/CircuitBreakerTest.cs`): opens exactly at the threshold, reports the opening transition once, and resets on success.

## PR Checklist

- [x] Has tests for the changes (`CircuitBreakerTest`)
- [x] Has a linked issue (#2698)
- [x] Has been tested in-game (server run offline)
- [ ] Has been tested on Windows/Linux/macOS (Windows only)

---

### DIFF
--- a/Nitrox.Model/Networking/CircuitBreaker.cs
+++ b/Nitrox.Model/Networking/CircuitBreaker.cs
@@ -1,0 +1,59 @@
+namespace Nitrox.Model.Networking;
+
+/// <summary>
+///     Minimal circuit breaker that counts consecutive failures and "opens" once a threshold is reached,
+///     signalling callers to back off. A single success closes it again.
+/// </summary>
+/// <remarks>
+///     Timing (how long to stay open, when to probe again) is intentionally left to the caller so this type
+///     stays free of timers and trivially testable. Thread-safe so it can be shared across the retry timer
+///     and network callback threads.
+/// </remarks>
+public sealed class CircuitBreaker(int failureThreshold)
+{
+    private readonly object gate = new();
+    private int consecutiveFailures;
+
+    public bool IsOpen
+    {
+        get
+        {
+            lock (gate)
+            {
+                return consecutiveFailures >= failureThreshold;
+            }
+        }
+    }
+
+    /// <summary>
+    ///     Resets the failure count, closing the breaker.
+    /// </summary>
+    public void RecordSuccess()
+    {
+        lock (gate)
+        {
+            consecutiveFailures = 0;
+        }
+    }
+
+    /// <summary>
+    ///     Records a failure.
+    /// </summary>
+    /// <returns>
+    ///     <see langword="true" /> only on the failure that opens the breaker (i.e. the threshold is reached
+    ///     exactly now), so callers can react once. Subsequent failures while already open return
+    ///     <see langword="false" />.
+    /// </returns>
+    public bool RecordFailure()
+    {
+        lock (gate)
+        {
+            if (consecutiveFailures >= failureThreshold)
+            {
+                return false;
+            }
+            consecutiveFailures++;
+            return consecutiveFailures >= failureThreshold;
+        }
+    }
+}
--- a/Nitrox.Server.Subnautica/Models/GameLogic/TimeService.cs
+++ b/Nitrox.Server.Subnautica/Models/GameLogic/TimeService.cs
@@ -31,6 +31,18 @@ internal sealed class TimeService(IPacketSender packetSender, NtpSyncer ntpSynce
     /// </summary>
     private const int NTP_RETRY_INTERVAL_SECONDS = 60;
 
+    /// <summary>
+    ///     Number of consecutive failed retry rounds before the circuit breaker opens and we assume there is
+    ///     no NTP connectivity (e.g. an offline/LAN server).
+    /// </summary>
+    private const int NTP_FAILURE_THRESHOLD = 3;
+
+    /// <summary>
+    ///     Retry interval in seconds used once the circuit breaker has opened. Slows futile attempts right down
+    ///     while still acting as a half-open probe so the server recovers if it later gains connectivity.
+    /// </summary>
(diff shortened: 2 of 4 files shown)


Nitrox.Server.Subnautica/Models/GameLogic/TimeService.cs
```diff
@@ -31,6 +31,18 @@ internal sealed class TimeService(IPacketSender packetSender, NtpSyncer ntpSynce
...
+    ///     Number of consecutive failed retry rounds before the circuit breaker opens and we assume there is
+    ///     no NTP connectivity (e.g. an offline/LAN server).
+    /// </summary>
+    private const int NTP_FAILURE_THRESHOLD = 3;
+
+    /// <summary>
+    ///     Retry interval in seconds used once the circuit breaker has opened. Slows futile attempts right down
+    ///     while still acting as a half-open probe so the server recovers if it later gains connectivity.
+    /// </summary>
+    private const int NTP_OFFLINE_RETRY_INTERVAL_SECONDS = 1800;
```
Measurity, line 44: Thoughts on using [Exponential backoff](https://en.wikipedia.org/wiki/Exponential_backoff) here?

SkiLark1: Good idea, exponential backoff fits better here. The main win is recovery: with a fixed 30-min interval, if the server gains connectivity shortly after the breaker opens it waits up to 30 min to notice, whereas backoff (60s doubling up to a 30-min cap) recovers within a few minutes while still ramping down to almost no overhead when genuinely offline. It also drops the arbitrary 1800 constant.

Plan: keep the CircuitBreaker for the threshold and the single "assuming offline" log, and drive the retry interval off its failure count with capped exponential backoff. I'll also fix the log line since it currently states a fixed interval. One question though: do you want that line at Information (it's expected on LAN) or Warning (degraded time accuracy)? I'll default to Information unless you'd prefer otherwise.

dartasen: I second this, connectivity retry is commonly implemented using exponential backoff algorithm.

> One question though: do you want that line at Information (it's expected on LAN) or Warning (degraded time accuracy)? I'll default to Information unless you'd prefer otherwise.

@SkiLark1 Since it can cause some desync, I think it does suit for a warning message so it's worth for the user to look at.

Nitrox.Model/Networking/CircuitBreaker.cs
```diff
@@ -0,0 +1,59 @@
...
+public sealed class CircuitBreaker(int failureThreshold)
+{
+    private readonly object gate = new();
+    private int consecutiveFailures;
+
+    public bool IsOpen
+    {
+        get
+        {
+            lock (gate)
```
Coding-Hen, line 21: Not against gate here but in other places in our code base we use lock rather than gate so might be more consistent to rename to something along those lines. lockObject and xLock have been used in other places

Nitrox.Model/Networking/CircuitBreaker.cs
```diff
@@ -0,0 +1,59 @@
...
+    public bool RecordFailure()
+    {
+        lock (gate)
+        {
+            if (consecutiveFailures >= failureThreshold)
+            {
+                return false;
+            }
+            consecutiveFailures++;
+            return consecutiveFailures >= failureThreshold;
```
Coding-Hen, line 56: Can this method not be rewritten to save the additional check, this way you will always record the failure as the method name implies and the code should handle the same and return false up to the point
```suggestion
            consecutiveFailures++;
            return consecutiveFailures > failureThreshold;
```

dartasen:
Closing for inactivity

PR #2755, opened 2026-05-21
Fix scanner room upgrades after base resync

deepvelocity:
Fixes #2696

## Description of Change

Uses https://github.com/SubnauticaNitrox/Nitrox/pull/2746 and https://github.com/SubnauticaNitrox/Nitrox/pull/2719 as a basis to write a solution, taking the comments into account.
 
Scanner room upgrades are persisted as child `InventoryItemEntity` instances of `MapRoomEntity`, but the client restore path did not restore them idempotently.

This change restores scanner room upgrade children when restoring a map room and clears the existing local map room upgrade container first, preventing duping chips with base re-syncs.

## Tested (Windows only)

- re-login
- server restart
- forced base re-sync
- tested with 0, 1, 2, 3, and 4 chips present

Result: expected number of upgrades was found in every scenario.

Issue #2410, opened 2025-08-13
Add scanner rooms sync


SpaceMonkeyy86:
### Describe the issue

Scanners rooms aren't synced at all : 

- [ ] Different resources highlighted for each player
- [ ] Ghost ping for some player
- [ ] Camera aren't sync
- [ ] Upgrades disappearing upon relog

Crapy:
Can confirm all still happens in latest master

PR #2758, opened 2026-05-22
Fix scanner room upgrade sync and persistence

### LINKED ISSUE #2696: Scanner room losing upgrades on server restart

1.8.1.0 (Latest release)

### Details about version

Latest Nitrox 1.8.1.0 

Windows

Steam

### Describe the issue.

The scanner rooms lose the upgrade chips when the server is restarted.

Create scanner room chip
place in scanner room
log out
log in
check scanner room

### Relevant log output

_No response_

mindless2831:
## Summary

Fixes scanner room upgrade sync and persistence.

Fixes #2696.
Contributes to #2410.

## Problem

Scanner room upgrade modules were not reliably synced across clients.

There were two related problems:

1. During live scanner room construction, the scanner room `MapRoomFunctionality` could fail to receive the same Nitrox ID on all clients.
2. Scanner room upgrade modules were not reliably persisted/restored as scanner room child entities across server restart/rejoin.

The live-sync failure caused the upgrade reparent packet to target a parent/container ID that did not exist on the other client. The result was that Player 1 could install a scanner room upgrade, but Player 2 would not see it visually or in the scanner room upgrade inventory.

The persistence failure caused installed scanner room upgrades to disappear after server restart/rejoin.

## What changed

This PR fixes both the live-sync and persistence paths.

### Live scanner room ID sync

When a scanner room is built, the client normally tries to find the new `MapRoomFunctionality` by cell. In the live multiplayer build path, that cell lookup can fail on other clients even though the scanner room object exists locally.

This PR adds a fallback for that case:

- if the cell lookup fails,
- and there is exactly one unregistered `MapRoomFunctionality` under the target base,
- assign the shared scanner room Nitrox ID to that object.

That ensures the scanner room upgrade inventory owner/container exists under the same ID on all clients before upgrade reparent packets arrive.

### Upgrade persistence

This PR also persists installed scanner room upgrades as child `InventoryItemEntity` entries on the `MapRoomEntity`.

On restore/resync:

- the scanner room receives its saved Nitrox ID;
- existing local scanner room upgrade container contents are cleared;
- saved upgrade chip child entities are spawned back into the scanner room upgrade container.

This makes installed scanner room upgrades survive restart/rejoin.

## Relationship to #2755

This PR uses the same general restore/resync idea from #2755 for scanner room upgrade chips, but also fixes the live first-time scanner room ID mismatch that caused upgrade reparenting to fail before rejoin/resync.

With this change, scanner room upgrades synced correctly immediately when installed, and also persisted correctly after server restart/rejoin.

## Testing

Tested with two players connected.

### Live sync test

1. Player 1 built a scanner room while Player 2 was present.
2. Player 1 installed a scanner room upgrade.
3. Player 2 immediately saw the installed module.
4. Player 2 could also see the installed module in the scanner room upgrade inventory.

Result: passed.

### Persistence test

1. Scanner room upgrade was left installed.
2. Both players logged out.
3. Server was stopped.
4. Server was restarted.
5. Both players rejoined.

Result: the scanner room upgrade was still installed and visible to both players.

### Additional observation

This fixes scanner room upgrade live sync and persistence. Broader scanner room behavior under #2410 is still not fully synced yet, including scan target/state, scanner room fabricator activity, animations/graphics, and camera-related behavior. Those should be handled separately after the scanner room identity/persistence foundation is correct.

### DIFF
--- a/NitroxClient/Communication/Packets/Processors/BuildingResyncProcessor.cs
+++ b/NitroxClient/Communication/Packets/Processors/BuildingResyncProcessor.cs
@@ -157,7 +157,7 @@ internal sealed class BuildingResyncProcessor(Entities entities, EntityMetadataM
             switch (childEntity)
             {
                 case MapRoomEntity mapRoomEntity:
-                    yield return InteriorPieceEntitySpawner.RestoreMapRoom(@base, mapRoomEntity);
+                    yield return InteriorPieceEntitySpawner.RestoreMapRoom(@base, mapRoomEntity, entities);
                     break;
                 case BaseLeakEntity baseLeakEntity:
                     yield return entities.SpawnEntityAsync(baseLeakEntity, true);
--- a/NitroxClient/GameLogic/Bases/BuildUtils.cs
+++ b/NitroxClient/GameLogic/Bases/BuildUtils.cs
@@ -1,13 +1,15 @@
 using System.Collections.Generic;
-using NitroxClient.GameLogic.Settings;
-using NitroxClient.GameLogic.Spawning.Bases;
-using NitroxClient.GameLogic.Spawning.Metadata;
-using NitroxClient.MonoBehaviours;
+using System.Linq;
 using Nitrox.Model.DataStructures;
 using Nitrox.Model.Subnautica.DataStructures.GameLogic;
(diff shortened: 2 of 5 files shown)


dartasen:
Duplicate of #2755

PR #2759, opened 2026-05-22
Fix scanner room fabricator sync

mindless2831:
## Summary

Fixes scanner room fabricator sync.

Depends on #2758.
Contributes to #2410.

## Problem

The normal fabricator sync path works because normal fabricators have a shared Nitrox ID that can be used for crafting metadata updates.

The scanner room fabricator did not sync because its `CrafterLogic` did not have a reliable shared Nitrox ID. As a result, crafting from the scanner room fabricator only appeared on the crafting player's client. Other players did not see the crafting animation, the finished product, or visible fabricator activity.

## What changed

Scanner room fabricator crafting now uses the parent scanner room `MapRoomFunctionality` Nitrox ID as its crafter sync ID.

For normal fabricators, behavior is unchanged: they continue to use the `CrafterLogic` object's own Nitrox ID.

For scanner room fabricators:

- the crafter patches detect when the `CrafterLogic` is under a `MapRoomFunctionality`;
- if so, they use the parent scanner room's shared Nitrox ID;
- existing crafter metadata sync then applies the crafting state to the `CrafterLogic` found under that scanner room.

This avoids introducing a separate child entity just for the scanner room fabricator and reuses the scanner room identity established by #2758.

## Why this depends on #2758

This fix relies on all clients having the same `MapRoomFunctionality` Nitrox ID for the scanner room.

That scanner room ID consistency is fixed by #2758. Without that foundation, the scanner room fabricator would not have a reliable shared parent ID to use for metadata sync.

## Testing

Tested with two players connected.

1. Player 1 crafted a scanner room camera from the scanner room fabricator.
2. Player 2 saw the fabricator activity/product.
3. Player 2 crafted from the scanner room fabricator.
4. Player 1 saw the fabricator activity/product.

Result: scanner room fabricator crafting synced correctly from both sides.

## Remaining scanner room work

This PR only fixes scanner room fabricator crafting sync.

Broader scanner room sync work under #2410 still remains.

### DIFF
--- a/NitroxClient/GameLogic/Bases/BuildUtils.cs
+++ b/NitroxClient/GameLogic/Bases/BuildUtils.cs
@@ -1,13 +1,15 @@
 using System.Collections.Generic;
-using NitroxClient.GameLogic.Settings;
-using NitroxClient.GameLogic.Spawning.Bases;
-using NitroxClient.GameLogic.Spawning.Metadata;
-using NitroxClient.MonoBehaviours;
+using System.Linq;
 using Nitrox.Model.DataStructures;
 using Nitrox.Model.Subnautica.DataStructures.GameLogic;
 using Nitrox.Model.Subnautica.DataStructures.GameLogic.Bases;
 using Nitrox.Model.Subnautica.DataStructures.GameLogic.Entities;
 using Nitrox.Model.Subnautica.DataStructures.GameLogic.Entities.Bases;
+using NitroxClient.GameLogic.Helper;
+using NitroxClient.GameLogic.Settings;
+using NitroxClient.GameLogic.Spawning.Bases;
+using NitroxClient.GameLogic.Spawning.Metadata;
+using NitroxClient.MonoBehaviours;
 using UnityEngine;
 
 namespace NitroxClient.GameLogic.Bases;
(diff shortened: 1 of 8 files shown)


NitroxClient/GameLogic/Bases/BuildUtils.cs
```diff
@@ -160,14 +162,23 @@ baseGhost is BaseAddModuleGhost moduleGhost && moduleGhost.modulePrefab ||
         if (isMapRoomGhost)
         {
             MapRoomFunctionality mapRoomFunctionality = baseGhost.targetBase.GetMapRoomFunctionalityForCell(face.Value.cell);
+            bool hasParentBase = constructableBase.GetComponentInParent<Base>(true);
```
Coding-Hen, line 165: Is there any value to moving this out? It's now just doing an assignment even if it is never used

NitroxClient/GameLogic/Bases/BuildUtils.cs
```diff
@@ -223,10 +234,55 @@ public static Int3 GetMapRoomFunctionalityCell(BaseGhost baseGhost)
...
+        if (candidates.Count == 1)
+        {
+            return candidates[0];
+        }
+
+        Log.Warn($"Could not uniquely identify unregistered MapRoomFunctionality under {targetBase.gameObject.GetFullHierarchyPath()}. Found {candidates.Count} candidates.");
+
+        foreach (MapRoomFunctionality candidate in candidates)
+        {
+            Log.Warn($"Unregistered MapRoomFunctionality candidate: {candidate.gameObject.GetFullHierarchyPath()}");
```
Coding-Hen, line 282: Do we want these as warns? Would they be better as debugs if we are trying to work out that functionality we are missing? Not sure they are providing too much use to the user

NitroxClient/GameLogic/Spawning/Bases/InteriorPieceEntitySpawner.cs
```diff
@@ -169,14 +170,51 @@ public static InteriorPieceEntity From(IBaseModule module, EntityMetadataManager
...
         NitroxEntity.SetNewId(mapRoomFunctionality.gameObject, mapRoomEntity.Id);
+
+        List<Entity> upgradeChips = mapRoomEntity.ChildEntities
+                                                 .OfType<InventoryItemEntity>()
+                                                 .ToList<Entity>();
+
+        if (upgradeChips.Count > 0)
+        {
+            ClearMapRoomUpgradeContainer(mapRoomFunctionality);
+            yield return entities.SpawnBatchAsync(upgradeChips, true);
```
Coding-Hen, line 191: If we clear and then spawn these each time are we not going to end up broadcasting these and potentially end up with race conditions on other players? Additionally if we are say just having to add one additional chip won't we end up clearing them all and restoring, is it maybe better to work out a delta?

NitroxPatcher/Patches/Dynamic/GhostCrafter_OnCraftingBegin_Patch.cs
```diff
@@ -27,4 +27,15 @@ public static void Postfix(GhostCrafter __instance, TechType techType, float dur
             Resolve<SimulationOwnership>().RequestSimulationLock(crafterLogicId, SimulationLockType.EXCLUSIVE);
         }
     }
+
+    private static bool TryGetCrafterSyncId(CrafterLogic crafterLogic, out NitroxId crafterId)
```
Coding-Hen, line 31: Do we want to have this method duplicated? Can we have it shared between them, maybe we have a helper class for patches, to remove duplication and forgetting to make changes in three places

Coding-Hen (changes requested):

PR #2764, opened 2026-05-25
Add player ping system with middle mouse button

### LINKED ISSUE #2763: Pinging System
### Describe the issue

A pinging system that lets players use the middle mouse button to ping locations for teammates to see.

misterbubb:
Closes #2763

Implements a multiplayer ping system that allows players to mark locations for teammates to see.

## Features
- Middle mouse button creates location pings
- Pings use player's character color
- Auto-despawn after 10 seconds
- Maximum 3 active pings per player
- 0.5s cooldown to prevent spam

## Testing
- [x] Pings created with middle mouse button
- [x] Pings use player colors
- [x] 3 ping limit enforced per player
- [x] Pings auto-despawn after 10 seconds
- [x] Network sync works across players

### DIFF
--- a/NitroxClient/MonoBehaviours/PlayerPingManager.cs
+++ b/NitroxClient/MonoBehaviours/PlayerPingManager.cs
@@ -1,0 +1,172 @@
+using System;
+using System.Collections.Generic;
+using Nitrox.Model.Core;
+using Nitrox.Model.DataStructures;
+using Nitrox.Model.DataStructures.Unity;
+using NitroxClient.Communication.Abstract;
+using NitroxClient.GameLogic;
+using NitroxClient.MonoBehaviours.Gui.InGame;
+using NitroxModel.Packets;
+using UnityEngine;
+
+namespace NitroxClient.MonoBehaviours;
+
+public class PlayerPingManager : MonoBehaviour
+{
+    private const float MAX_PING_DISTANCE = 1000f;
+    private const float MIN_PING_DISTANCE = 3f;
+    private const float MIN_PING_COOLDOWN = 0.5f;
+    private const int MAX_ACTIVE_PINGS = 3;
+    
+    private static PlayerPingManager instance;
+    
+    private IPacketSender packetSender;
+    private LocalPlayer localPlayer;
+    private float lastPingTime;
+    private Camera mainCamera;
+    private readonly Dictionary<SessionId, List<PlayerPing>> playerPings = new();
+    private readonly List<PlayerPing> localPlayerPings = new();
+
+    public void Awake()
+    {
+        instance = this;
+        packetSender = NitroxServiceLocator.LocateService<IPacketSender>();
+        localPlayer = this.Resolve<LocalPlayer>();
+    }
+
+    private void Update()
+    {
+        if (!mainCamera)
+        {
+            mainCamera = Camera.main;
+        }
+
+        if (Input.GetMouseButtonDown(2) && CanCreatePing())
+        {
+            TryCreatePing();
+        }
+    }
+
+    private bool CanCreatePing()
+    {
+        if (Time.time - lastPingTime < MIN_PING_COOLDOWN)
+        {
+            return false;
+        }
+
+        if (!Player.main || !AvatarInputHandler.main.IsEnabled())
+        {
+            return false;
+        }
+
+        return true;
+    }
+
+    private void TryCreatePing()
+    {
+        if (!mainCamera)
+        {
+            return;
+        }
+
+        Ray ray = mainCamera.ScreenPointToRay(Input.mousePosition);
+        int layerMask = ~(LayerMask.GetMask("UI", "Ignore Raycast"));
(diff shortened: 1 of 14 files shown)


NitroxClient/MonoBehaviours/PlayerPingManager.cs
```diff
@@ -0,0 +1,172 @@
...
+        packetSender = NitroxServiceLocator.LocateService<IPacketSender>();
+        localPlayer = this.Resolve<LocalPlayer>();
+    }
+
+    private void Update()
+    {
+        if (!mainCamera)
+        {
+            mainCamera = Camera.main;
+        }
```
Measurity: I think it's less CPU cycles to move this if check into `TryCreatePing` because it happens less often.

NitroxClient/MonoBehaviours/PlayerPingManager.cs
```diff
@@ -0,0 +1,172 @@
...
+
+        if (Input.GetMouseButtonDown(2) && CanCreatePing())
+        {
+            TryCreatePing();
+        }
+    }
+
+    private bool CanCreatePing()
+    {
+        if (Time.time - lastPingTime < MIN_PING_COOLDOWN)
```
Measurity: Nit: `Time.time` is affected by Unity `timeScale`. `Time.realTimeSinceStartup` can be used for consistent time which is what I think we want here.

NitroxClient/MonoBehaviours/PlayerPingManager.cs
```diff
@@ -0,0 +1,172 @@
...
+        {
+            PlayerPing oldestPing = pings[0];
+            if (oldestPing)
+            {
+                Destroy(oldestPing.gameObject);
+            }
+            pings.RemoveAt(0);
+        }
+        
+        PlayerManager playerManager = NitroxServiceLocator.LocateService<PlayerManager>();
```
Measurity: Use `this.Resolve<PlayerManager>` which performs better.

Measurity (commented):
Minor code review only comments

NitroxClient/MonoBehaviours/PlayerPingManager.cs
```diff
@@ -0,0 +1,172 @@
...
+        }
+
+        if (!mainCamera)
+        {
+            return;
+        }
+
+        Ray ray = mainCamera.ScreenPointToRay(Input.mousePosition);
+        int layerMask = ~(LayerMask.GetMask("UI", "Ignore Raycast"));
+        RaycastHit[] hits = Physics.RaycastAll(ray, MAX_PING_DISTANCE, layerMask);
```
Measurity: Looks like the terrain physics aren't loaded about 50M+ from the player. Would it be possible to do a "raycast" on the terrain voxel data if no Unity physics hits?

NitroxClient/MonoBehaviours/PlayerPingManager.cs
```diff
@@ -0,0 +1,172 @@
...
+            TryCreatePing();
+        }
+    }
+
+    private bool CanCreatePing()
+    {
+        if (Time.realtimeSinceStartup - lastPingTime < MIN_PING_COOLDOWN)
+        {
+            return false;
+        }
```
Measurity, line 68: As a QoL we could queue the last ping and show it after the cooldown (or reset if user requested another ping).

This trick is common to make inputs feel more responsive. I'm fine with doing this in another PR though.

Measurity (approved):
Works well on my end.

Would it be feasible to add a sound to the ping? Would be cool also to have the entity it hits in the ping label.

misterbubb:
> Works well on my end.
> 
> Would it be feasible to add a sound to the ping? Would be cool also to have the entity it hits in the ping label.

I thought about adding a sound yes, any ideas for the sounds? same with the entity ping.

Measurity:
For sound, maybe the cyclops sonar ping? Or play the last part of the scanner sound if we can do this.

And for entity, we can try get the techtype and then its name. Or if no hits, the biome name at the position of the ping.

Measurity:
It would be nice to have feedback when no raycast hits succeed as well. Either sound or visual cue.

NitroxClient/MonoBehaviours/PlayerPingManager.cs
misterbubb: do we really need this? The current implementation only pings on physics hits, which seems reasonable since players would typically ping things within the loaded range.

Measurity: No there's no need and why I approved the PR as it is.

Measurity:
Handled the review comments except for raycast on terrain. The game doesn't have a function for this so would have to calculate terrain octree SDF intersections manually.

dartasen (approved):
Works great for me IG.

Only thing I noticed is that the ping doesn't follow the entity pinged (i.e: fish). But it's not blocking for that PR

Great work 😄

PR #2765, opened 2026-05-26
Fix player SubRootId/ParentId mismatch within escape pods

### LINKED ISSUE #123: [Misc] Disable ability to pause while in multiplayer.

hakonw:
## Linked Issues / Context

Hi! 

I read in the [newsletter](https://nitroxblog.rux.gg/2025/10/08/nitrox-1-8-0-0/) that the mac version is lacking, so i thought i would try to help!

-------

On mac, i had an issue where the player nr 2 glitches though the floor on joining + cannot see other users.   (See video)

https://github.com/user-attachments/assets/9685c212-dce7-4e2d-8a42-d2e27b6b6737

## Description of Change

I assume this doesnt happen on windows/linux, and something special on mac.

With extra logging, on player pos, i got a string like: 
```
[20:27:14.510] PlayerMovement session=1 pos=[-38.17284, 1.902705, -311.7547] bodyActive=True bodyInHierarchy=True parent=<none>
```
Where the parent was none, when it should have been the pod. 

I tried a fix where i delayed messages from/to the client when a user was joining, but that did not fix the issue.
I assume there are multiple similar issues (like the issue where a person is holding an item when a user joins), but i am not familiar enough with the code and could not find a root cause.

So i then tried a simpler fix to reparent the user, and it worked.

## Validation

Validated by running it multiple times, and trying to get the fault again.

Tested on my mac m4 pro 48gb

Video:
https://github.com/user-attachments/assets/0019d505-54d7-43bd-8130-4d64f450a841

## PR Checklist

- [ ] Has a linked issue

Dont have a windows/linux pc available ;/

---

I am unfamiliar with the codebase and used claude as a pair programmer while debugging to try to figure out the fault.

### DIFF
--- a/Nitrox.Server.Subnautica/Models/Commands/QueryCommand.cs
+++ b/Nitrox.Server.Subnautica/Models/Commands/QueryCommand.cs
@@ -1,4 +1,5 @@
 using System.ComponentModel;
+using System.Linq;
 using System.Text;
 using Nitrox.Model.DataStructures;
 using Nitrox.Model.DataStructures.GameLogic;
@@ -11,11 +12,11 @@ using Nitrox.Server.Subnautica.Models.GameLogic.Entities;
 namespace Nitrox.Server.Subnautica.Models.Commands;
 
 [RequiresPermission(Perms.ADMIN)]
-internal sealed class QueryCommand(EntityRegistry entityRegistry, SimulationOwnershipData simulationOwnershipData, ILogger<QueryCommand> logger) : ICommandHandler<NitroxId>
+internal sealed class QueryCommand(EntityRegistry entityRegistry, SimulationOwnershipData simulationOwnershipData, PlayerManager playerManager) : ICommandHandler<NitroxId>
 {
     private readonly EntityRegistry entityRegistry = entityRegistry;
+    private readonly PlayerManager playerManager = playerManager;
     private readonly SimulationOwnershipData simulationOwnershipData = simulationOwnershipData;
-    private readonly ILogger<QueryCommand> logger = logger;
 
     [Description("Query the entity associated with the given NitroxId")]
     public async Task Execute(ICommandContext context, [Description("NitroxId of an entity")] NitroxId entityId)
@@ -41,7 +42,8 @@ internal sealed class QueryCommand(EntityRegistry entityRegistry, SimulationOwne
                 builder.AppendLine("   └ Child");
                 builder.AppendLine($"     └ Type: {childEntity.GetType().Name}");
                 builder.AppendLine($"     └ Id: {childEntity.Id}");
-                builder.AppendLine($"     └ Metadata: {childEntity.Metadata?.ToString() ?? "none"}");
+                builder.AppendLine($"     └ TechType: {childEntity.TechType}");
+                builder.AppendLine($"     └ Metadata: {childEntity.Metadata?.ToString() ?? "<null>"}");
                 builder.AppendLine($"     └ Children: {childEntity.ChildEntities.Count}");
             }
         }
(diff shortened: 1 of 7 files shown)


dartasen:
Thank you ! 

Are you please able to test on 1.8.1.0 release branch to see if this a regression from InDev 1.9 or it's something that never worked for you in the past ?

hakonw:
> Are you please able to test on 1.8.1.0 release branch to see if this a regression from InDev 1.9 or it's something that never worked for you in the past ?

Hi!

Checked out release/1.8.1.0, and that release had the same issue.

New world, launching via "Instant launch world 2 player". Second player to join cannot see player 1, and falls though the floor of the pod, followed by walking on the ocean floor.

Nitrox.Server.Subnautica/Models/Packets/Processors/EscapePodChangedPacketProcessor.cs
```diff
@@ -1,11 +1,13 @@
...
+using Nitrox.Server.Subnautica.Models.Packets.Core;
 
 namespace Nitrox.Server.Subnautica.Models.Packets.Processors;
 
-internal sealed class EscapePodChangedPacketProcessor : IAuthPacketProcessor<EscapePodChanged>
+internal sealed class EscapePodChangedPacketProcessor(EntityRegistry entityRegistry) : IAuthPacketProcessor<EscapePodChanged>
 {
     public async Task Process(AuthProcessorContext context, EscapePodChanged packet)
     {
+        entityRegistry.ReparentEntity(context.Sender.GameObjectId, packet.EscapePodId.OrNull());
```
dartasen, line 10: This line looks good to me as a fix, this is similar to what we're doing for SubRootChangedPacketProcessor server-side.

Currently investigating what could cause one to fall of the Escape pod (Unable to reproduce on my MacBook M2 Pro which is quite slow for game). Likely to be a race condition caused by the difference on Mac on the filesystem that could cause the world streamer to have different timings

dartasen:
@hakonw Heyyy
I've done some changes on your branch, if you could validate that it's still working on your side, that would be awesome ! 😄 

At least it does fix our internal server data being wrong about player parent while using escape pods. If it doesn't fix your player falling through the escape pod, we might need to edit the InitialSyncPlayerProcessor client side to hold the client in place until the terrain and entities are fully loaded.

Measurity (approved):

Nitrox.Server.Subnautica/Models/Commands/WhoisCommand.cs
```diff
@@ -12,13 +12,16 @@ internal sealed class WhoisCommand : ICommandHandler<Player>
...
+        builder.AppendLine($"Id: {targetPlayer.Entity.Id}");
         builder.AppendLine($"SessionId: {targetPlayer.SessionId}");
         builder.AppendLine($"Role: {targetPlayer.Permissions}");
+        builder.AppendLine($"Gamemode: {targetPlayer.GameMode}");
         builder.AppendLine($"Position: {targetPlayer.Position.X}, {targetPlayer.Position.Y}, {targetPlayer.Position.Z}");
         builder.AppendLine($"Oxygen: {targetPlayer.Stats.Oxygen}/{targetPlayer.Stats.MaxOxygen}");
         builder.AppendLine($"Food: {targetPlayer.Stats.Food}");
         builder.AppendLine($"Water: {targetPlayer.Stats.Water}");
         builder.AppendLine($"Infection: {targetPlayer.Stats.InfectionAmount}");
+        builder.AppendLine($"In precursor: {targetPlayer.InPrecursor}");
```
Coding-Hen, line 24: Is this actually useful to log in a who is? Maybe for debugging but normal play I'm not so sure knowing someone is in the precursor is useful

Nitrox.Server.Subnautica/Models/Player.cs
```diff
@@ -156,11 +179,6 @@ public override string ToString()
             return $"[Player - SessionId: {Id}, Name: {Name}, Perms: {Permissions}, Position: {Position}]";
         }
 
-        private bool Equals(Player other)
```
Coding-Hen, line 159: Do we need to move this method for the sale of it? Maybe revert this change?

Nitrox.Server.Subnautica/Models/Commands/QueryCommand.cs
```diff
@@ -56,6 +58,36 @@ public async Task Execute(ICommandContext context, [Description("NitroxId of an
...
+                builder.AppendLine($" └ Stats: {serverPlayer.Stats}");
+                builder.AppendLine($" └ DisplaySurfaceWater: {serverPlayer.DisplaySurfaceWater}");
+                builder.AppendLine($" └ Position: {serverPlayer.Position}");
+                builder.AppendLine($" └ LastStoredPosition: {serverPlayer.LastStoredPosition?.ToString() ?? "<null>"}");
+                builder.AppendLine($" └ SubRootId: {serverPlayer.SubRootId.OrNull()?.ToString() ?? "<null>"}");
+                builder.AppendLine($" └ LastStoredSubRootID: {serverPlayer.LastStoredSubRootID.OrNull()?.ToString() ?? "<null>"}");
+
+                if (entity.ParentId != serverPlayer.SubRootId.OrNull())
+                {
+                    builder.AppendLine("⚠ ParentId doesn't match SubRootId");
```
Coding-Hen, line 81: Is this useful to have printed out? I assume someone seeing the two output will be able to see that?

Nitrox.Server.Subnautica/Models/Player.cs
```diff
@@ -12,14 +12,37 @@ namespace Nitrox.Server.Subnautica.Models
...
+        private ushort sessionId;
+
         public ThreadSafeList<NitroxTechType> UsedItems { get; }
         public Optional<NitroxId>[] QuickSlotsBindingIds { get; set; }
 
         public PlayerContext? PlayerContext { get; set; }
         public PeerId Id { get; init; }
-        public SessionId SessionId { get; set; }
+
+        public SessionId SessionId
```
Coding-Hen, line 27: I'm not that familiar with interlocks, why is this change here required for this escape pod change?

Coding-Hen (dismissed):

Nitrox.Server.Subnautica/Models/Player.cs
Coding-Hen: Ignore this read the commit, it's to make it thread safe which makes sense.

Nitrox.Server.Subnautica/Models/Player.cs
Measurity: It's because I ran a format on this code. Private methods should be sorted close to the bottom following code style.

Nitrox.Server.Subnautica/Models/Commands/QueryCommand.cs
dartasen: Query is not a command accessible by regular players, and mostly use by us to ask the server to print some internal datas. So yes we need as much infos as possible, one could add more fields if they wanted to

dartasen (approved):

hakonw:
> @hakonw Heyyy I've done some changes on your branch, if you could validate that it's still working on your side, that would be awesome ! 😄

@dartasen  It works here!

>  If it doesn't fix your player falling through the escape pod, we might need to edit the InitialSyncPlayerProcessor client side to hold the client in place until the terrain and entities are fully loaded.

Most of the times when i tested before, player 2 could move around in the pod for a few second before falling though the floor, and usually it didn't happen instantly like in the video.

Thanks for all the help!

dartasen:
@hakonw Awesome 😄  !
Thank's for the help on MacOS, please let us know if you encounter any weird issues. Launcher side it should be working, but I haven't work on the InGame part for now !

PR #2771, opened 2026-05-31
Fix base leaks not syncing after repair

### LINKED ISSUE #2769: Base leaks are not synced

1.8.1.0 (Latest release)

### Details about version

_No response_

Windows

Steam

### Describe the issue.

Base is leaking upon server restart. Fixed from player 1. Then once player 2 joins base is still flooding for them. 

Started playing the day after playing for a few days. I joined and my base integrity is around 32%. I have 2 leaks. I fix the leaks. Player 2 then joined. The base was leaking for them. They then repaired the same leaks and the base is fixed for them. Now once we receive a hole in the base we both have to fix that exact same spot. Have tried the base resync. Once we both log off and join back again the base is still leaking. Same 2 holes. Any help? 

### Relevant log output

[server\[PooleFam\]-20260529.log](https://github.com/user-attachments/files/28414821/server.PooleFam.-20260529.log)

[game\[Jonathan\]-20260529.log](https://github.com/user-attachments/files/28414823/game.Jonathan.-20260529.log)

misterbubb:
Closes #2769

## Description
Fixes base leaks not syncing properly between players and persisting after server restarts.

## Problem
When Player 1 repaired base leaks, Player 2 would still see the same leaks when joining. Additionally, after server restart, the repaired leaks would reappear for all players. This was caused by leak entities being destroyed from memory but not persisted to disk until the next autosave cycle.

## Solution
**Server-side:**
- Added debounced save trigger in `LeakRepairedProcessor` that queues a save within 5 seconds of leak repair
- Ensures leak deletions are persisted immediately, preventing reappearance after crashes or player joins
- Debouncing prevents excessive saves when multiple leaks are repaired rapidly

**Client-side:**
- Added defensive health check in `BaseLeakEntitySpawner` to skip spawning fully repaired leaks (Health >= 100)
- Provides additional safety layer for edge cases where save hasn't persisted yet

## Testing
Test 1: Player 1 repairs leaks → Player 2 joins → No leaks appear
Test 2: Player 1 repairs leaks → Server restart → No leaks reappear

### DIFF
--- a/Nitrox.Server.Subnautica/Models/Packets/Processors/LeakRepairedProcessor.cs
+++ b/Nitrox.Server.Subnautica/Models/Packets/Processors/LeakRepairedProcessor.cs
@@ -1,17 +1,50 @@
 using Nitrox.Server.Subnautica.Models.GameLogic.Entities;
 using Nitrox.Server.Subnautica.Models.Packets.Core;
+using Nitrox.Server.Subnautica.Services;
 
 namespace Nitrox.Server.Subnautica.Models.Packets.Processors;
 
-sealed class LeakRepairedProcessor(WorldEntityManager worldEntityManager) : IAuthPacketProcessor<LeakRepaired>
+internal sealed class LeakRepairedProcessor(WorldEntityManager worldEntityManager, SaveService saveService, ILogger<LeakRepairedProcessor> logger) : IAuthPacketProcessor<LeakRepaired>
 {
+    private static readonly Lock saveDebounceLock = new();
+    private static DateTime lastSaveRequest = DateTime.MinValue;
+    private static readonly TimeSpan saveDebounceInterval = TimeSpan.FromSeconds(5);
+
     private readonly WorldEntityManager worldEntityManager = worldEntityManager;
+    private readonly SaveService saveService = saveService;
+    private readonly ILogger<LeakRepairedProcessor> logger = logger;
 
     public async Task Process(AuthProcessorContext context, LeakRepaired packet)
     {
         if (worldEntityManager.TryDestroyEntity(packet.LeakId, out _))
         {
             await context.SendToOthersAsync(packet);
+            
+            // Trigger a debounced save to persist the leak removal
+            // This ensures that if the server crashes or a player joins before the next autosave,
+            // the repaired leak won't reappear
+            await TriggerDebouncedSaveAsync();
+        }
+    }
+
+    private async Task TriggerDebouncedSaveAsync()
+    {
+        bool shouldTriggerSave = false;
+        
+        lock (saveDebounceLock)
+        {
+            DateTime now = DateTime.UtcNow;
+            if (now - lastSaveRequest >= saveDebounceInterval)
+            {
+                lastSaveRequest = now;
+                shouldTriggerSave = true;
+            }
+        }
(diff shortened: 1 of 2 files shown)


Nitrox.Server.Subnautica/Models/Packets/Processors/LeakRepairedProcessor.cs
```diff
@@ -1,17 +1,50 @@
...
+                lastSaveRequest = now;
+                shouldTriggerSave = true;
+            }
+        }
+
+        if (shouldTriggerSave)
+        {
+            logger.ZLogDebug($"Queueing save after base leak repair to ensure persistence");
+            await saveService.QueueActionAsync(SaveService.ServiceAction.SAVE);
         }
```
Measurity, line 48: Leak repair shouldn't trigger a save, it's not related to world saving. 

Data loss on crash isn't preventable and users can already change the save interval to reduce data loss.

Measurity (changes requested):
> Provides additional safety layer for edge cases where save hasn't persisted yet

This shouldn't happen. Server uses latest data from memory on joins. Save files are only used when the server starts up. If not, we have a design problem (race condition or otherwise) and should fix this instead.

Nitrox.Server.Subnautica/Models/Packets/Processors/LeakRepairedProcessor.cs
misterbubb: Wait yeah your right, I confirmed the entity removal is working correctly. closing this PR now, sorry!

NitroxClient/GameLogic/Spawning/Bases/BaseLeakEntitySpawner.cs
```diff
@@ -32,6 +32,14 @@ protected override bool SpawnSync(BaseLeakEntity entity, TaskResult<Optional<Gam
             return true;
         }
 
+        // Skip spawning leaks that are already at full health (fully repaired)
+        // This handles edge cases where a leak was repaired but the save hasn't persisted yet
+        if (entity.Health >= 100f)
+        {
+            Log.Debug($"[{nameof(BaseLeakEntitySpawner)}] Skipping spawn of fully repaired leak {entity.Id} (Health: {entity.Health})");
+            return true;
```
Measurity, line 40: This might still be worth having though apart from other changes.

jonathandwaynepoole796-art: Can you walk me through what i need to do to get this fixed?

dartasen: @jonathandwaynepoole796-art I think the first part is to understand what's happening, by testing and looking at the data stored server-side (either by debugging it or looking at the json files)

* When a leak is created, which entity is created, are the informations correct ? (health, position, rotation, ...)
* When the same leak is repaired, what happen to the data (is the entity deleted, have weird data, ...) ?
* When a client repairs the leak, do we emit a packet that will be processed correctly by the server
* ...

With those informations it's easier to create a chirurgical fix

Maybe @misterbubb have more informations

PR #2777, opened 2026-06-03
Fix captured log redaction and summary command NRE

### LINKED ISSUE #2778: Whois command improvements
### Describe the issue

Currently Whois command takes a mandatory player name as argument.
Whois command without any arguments should display the current player informations

Measurity:
## Preview of server log file

image

Any suggestions for log header are welcome


Nitrox.Server.Subnautica/Models/Logging/ZLogger/WindowConsoleMode.cs
```diff
@@ -0,0 +1,21 @@
...
+internal static class WindowsConsoleMode
+{
+    [DllImport("kernel32.dll")]
+    internal static extern IntPtr GetStdHandle(int nStdHandle);
+
+    [DllImport("kernel32.dll", SetLastError = true)]
+    internal static extern bool GetConsoleMode(IntPtr handle, out int mode);
+
+    [DllImport("kernel32.dll", SetLastError = true)]
+    internal static extern bool SetConsoleMode(IntPtr handle, int mode);
```
dartasen: We should use LibraryImport on #if NET to source generate p/invoke functions at compile time instead of at runtime with DllImport

Nitrox.Server.Subnautica/Models/Logging/Scopes/AtomicScope.cs
```diff
@@ -0,0 +1,44 @@
...
+    {
+        entries.Add(new QueuedEntry(entry, processor));
+        return true;
+    }
+
+    public void Dispose()
+    {
+        lock (Locker)
+        {
+            Task.Delay(100).Wait();
```
dartasen: Async/await inside lock 👀

Afaik this has been forbidden (if using await keyboard) to avoid deadlocks.
Is it worth moving to IAsyncDisposable with a SemaphoreSlim ?

Nitrox.Server.Subnautica/Extensions/LoggingBuilderExtensions.cs
```diff
@@ -72,12 +67,71 @@ public ILoggingBuilder AddNitroxLogging()
...
+        private ILoggingBuilder AddNitroxAtomicZLoggerConsole(Action<ZLoggerConsoleOptions, IServiceProvider> configure)
+        {
+            builder.Services.AddSingleton<ILoggerProvider, ZLoggerAtomicConsoleLoggerProvider>((Func<IServiceProvider, ZLoggerAtomicConsoleLoggerProvider>) (serviceProvider =>
+            {
+                ZLoggerConsoleOptions options = new();
+                configure(options, serviceProvider);
+                if (options.ConfigureEnableAnsiEscapeCode && RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                {
+                    WindowsConsoleMode.TryEnableVirtualTerminalProcessing();
+                }
```
dartasen, line 115: Even with the [doc](https://learn.microsoft.com/en-us/windows/console/setconsolemode), not sure to get what this is about (so I can test it)

Nitrox.Server.Subnautica/Models/Commands/Core/CommandRegistryService.cs
```diff
@@ -262,4 +200,68 @@ private void RegisterHandler(CommandHandlerEntry handler)
...
+                {
+                    continue;
+                }
+                Type[] genericArgsOnConverter = converterInterface.GetGenericArguments();
+                if (genericArgsOnConverter is not { Length: 2 })
+                {
+                    continue;
+                }
+                Type toType = genericArgsOnConverter[1];
+                if (!ArgConverterLookup.TryGetValue(toType, out List<ArgConverterInfo> registeredConverters))
```
dartasen, line 255:
```suggestion
                if (!ArgConverterLookup.TryGetValue(toType, out List<ArgConverterInfo>? registeredConverters))
```

Nitrox.Server.Subnautica/Models/Commands/FileCommand.cs
```diff
@@ -0,0 +1,48 @@
...
+            catch (Exception ex)
+            {
+                logger.ZLogError(ex, $"Failed to open latest log file");
+            }
+            return Task.CompletedTask;
+        }
+        catch (Exception exception)
+        {
+            return Task.FromException(exception);
+        }
```
dartasen, line 33: the try catch-ception 👀

Nitrox.Server.Subnautica/Models/Commands/FileCommand.cs
```diff
@@ -0,0 +1,48 @@
...
+                {
+                    logger.ZLogInformation($"Log file is unavailable");
+                    return Task.CompletedTask;
+                }
+                using Process proc = Process.Start(new ProcessStartInfo
+                {
+                    FileName = lastWriteLogFilePath,
+                    UseShellExecute = true,
+                    Verb = "open"
+                });
```
dartasen: Could we use the helpers inside ProcessEx when we need to use `Verb = open ?`, I guess we have few commands that does that

Nitrox.Server.Subnautica/Models/Logging/Middleware/Core/ILoggerMiddleware.cs
```diff
@@ -43,4 +30,27 @@ static void ExecuteNext(ref Context context)
...
     }
+
+    public ref struct Context
+    {
+        public Context()
+        {
+        }
+
+        public IBufferWriter<byte> Writer { get; init; } = null!;
+        public IZLoggerEntry Entry { get; init; } = null!;
```
dartasen: dammit operator 👀

Nitrox.Server.Subnautica/Models/Logging/Scopes/CaptureScope.cs
```diff
@@ -3,23 +3,40 @@
...
     /// </summary>
     public void Capture(string log) => logs.Add(log);
 
-    public void Dispose() => InnerDisposable?.Dispose();
+    public override string ToString() => string.Join("", Logs).TrimEnd('\n');
+
+    public void Dispose()
+    {
+        InnerDisposable?.Dispose();
+        Logger?.ZLogInformation($"{this}");
```
dartasen, line 40: `$"{this}"` 👀

Nitrox.Server.Subnautica/Extensions/LoggingBuilderExtensions.cs
Measurity: It's code copied from ZLogger because it was internal but I needed a new ZLoggerConsole implementation.

Nitrox.Server.Subnautica/Models/Commands/Core/CommandRegistryService.cs
Measurity: This change will put a warning that `registeredConverters` can be null. But it can't as it would be an empty array if not found. Not sure what to best C# idiom is here.

image

Nitrox.Server.Subnautica/Models/Logging/Scopes/CaptureScope.cs
Measurity: Interpolated string will by default call `.ToString` on objects that's why this works.

dartasen: Yup, but wouldn't that allocate a InterpolatedStringHandler (or whatever handler that can be overriden by a 3rd party library) compared to toString ?

Measurity: InterpolatedStringHandler requires the `$""` always, can't pass in a string. That's why it's done this way.

Nitrox.Server.Subnautica/Models/Logging/Scopes/AtomicScope.cs
Measurity: Good idea. Changed and still works from testing:
image

Nitrox.Server.Subnautica/Models/Logging/ZLogger/WindowConsoleMode.cs
Measurity: Made the changes but without #if NET as server isn't cross compiled

dartasen (approved):

PR #2784, opened 2026-06-11
fix remote players becoming permanently invisible due to NaN velocity

hakonw:
## Linked Issues / Context

- This issue was only on my mac, and not on the other persons windows pc (may be unrelated, but something to note)
- Played for a few hours, and got these warnings when the other person disappeard. 
- Mostly happen when we entered a habitat but also happened while swimming

```
[18:16:51.906] [ERR-UNITY] Infinity or NaN floating point numbers appear when calculating the transform matrix for a Collider. Scene hierarchy path "silje"
...
[19:17:37.047] [ERR-UNITY] Infinity or NaN floating point numbers appear when calculating the transform matrix for a Collider. Scene hierarchy path "silje/player_view/signal_silje"
...
[18:47:32.547] [ERR-UNITY] Infinity or NaN floating point numbers appear when calculating the transform matrix for a Collider. Scene hierarchy path "Landscape/Global Root/Base(Clone)/silje"
...
[19:23:31.791] [ERR-UNITY] Infinity or NaN floating point numbers appear when calculating the transform matrix for a Collider. Scene hierarchy path "EscapePod-MainPrefab/EscapePod/silje"
```

## Description of Change

- guard for nan values

## Validation

- added temporary logging, and entered/exited buildings a few times. 
- in 5 min i got this, BUT the person did not despawn
```
[21:24:08.268] [WRN] NaN rotation on "Player1": local rotation (0.0, -0.2, 0.0, -1.0), incoming remote rotation (0.0, 0.2, 0.0, 1.0), delta angle 360, delta axis (NaN, Infinity, NaN), incoming angular velocity (0.0, 0.0, 0.0)
[21:24:12.284] [WRN] NaN rotation on "Player1": local rotation (0.0, -0.4, 0.0, 0.9), incoming remote rotation (0.0, 0.4, 0.0, -0.9), delta angle 360, delta axis (NaN, Infinity, NaN), incoming angular velocity (0.0, 0.0, 0.0)
[21:26:01.568] [WRN] NaN rotation on "Player1": local rotation (0.0, -1.0, 0.0, 0.2), incoming remote rotation (0.0, 1.0, 0.0, -0.2), delta angle 360, delta axis (NaN, Infinity, NaN), incoming angular velocity (0.0, 0.0, 0.0)
```

## PR Checklist

- [ ] Has a linked issue

---

Pair programmed with ai

### DIFF
--- a/NitroxClient/GameLogic/MovementHelper.cs
+++ b/NitroxClient/GameLogic/MovementHelper.cs
@@ -36,7 +36,7 @@ namespace NitroxClient.GameLogic
 
             float distance = difference.magnitude;
 
-            if (distance > 20f)
+            if (float.IsNaN(distance) || distance > 20f)
             {
                 // This should be a one-off teleport.
                 gameObject.transform.position = remotePosition;
@@ -58,6 +58,12 @@ namespace NitroxClient.GameLogic
             // We get an infinite axis in the event that our rotation is already aligned.
             if (float.IsInfinity(axis.x))
             {
+                return angularVelocty;
+            }
+
+            if (float.IsNaN(angle) || float.IsNaN(axis.x))
+            {
+                gameObject.transform.rotation = remoteRotation;
                 return angularVelocty;
             }
 
--- a/NitroxClient/GameLogic/MovementHelper.cs
+++ b/NitroxClient/GameLogic/MovementHelper.cs
@@ -36,6 +36,7 @@ namespace NitroxClient.GameLogic
 
             float distance = difference.magnitude;
 
+            // NaN guard to recover if distance becomes invalid 
             if (float.IsNaN(distance) || distance > 20f)
             {
                 // This should be a one-off teleport.
@@ -61,6 +62,7 @@ namespace NitroxClient.GameLogic
                 return angularVelocty;
             }
 
+            // Guard for NaN when remoteRotation and gameobjects rotation are parallel (witnessed during macOS habitat transition)
             if (float.IsNaN(angle) || float.IsNaN(axis.x))
             {
                 gameObject.transform.rotation = remoteRotation;


Coding-Hen (commented):
Whilst I see no issues with this request really if it does solve the issue, I think it would be better to understand what is causing the magnitude to be nan and fix the root cause of there is one.

The fact this only seemingly effects mac as well feels wrong me to, if it's a logic but surely it should effect all platforms.

Have you done any digging to understand how this is set and what might have caused it?

hakonw:
> Whilst I see no issues with this request really if it does solve the issue, I think it would be better to understand what is causing the magnitude to be nan and fix the root cause of there is one.

Yeah, i tried finding the real root cause but didn't.
In the debug logging i added, both the local and incoming rotation were valid, and the remote vector and the inverse of local vector were identical. This is the reason why delta.ToAngleAxis divided by 0.

Another possible guard would be to check if the incoming vectors are equal/parallell, but the NaN is a more general safeguard

> The fact this only seemingly effects mac as well feels wrong me to, if it's a logic but surely it should effect all platforms.

I agree, but that wasn't what I observed when playing. I also checked the issues quickly, and didn't find any open issue regarding vanishing players.

> Have you done any digging to understand how this is set and what might have caused it?

A bit unsure what more to check. The inputs "looks" valid, but the issue can be higher up

@Coding-Hen

Measurity (commented):
If we can add a quick comment to both `float.IsNaN` usages that it's mostly for macOS and when transitioning a sub, then PR LGTM.

hakonw:
@Measurity Added comments.

I also checked again, and could not trigger the distance NaN guard. 
I can remove it, but imo it doesn't hurt to be there

Measurity (approved):
I'm fine with the guard. Thanks for this PR!

dartasen (approved):

PR #2789, opened 2026-06-16
[Steam] Make libraryfolders.vdf parsing whitespace-tolerant

vvitovec:
## Purpose
Make Steam library discovery handle valid `libraryfolders.vdf` key/value lines when they are separated by whitespace other than tabs.

Follow-up from maintainer review: this is not a new Steam schema. The existing parser already handles the nested `path` format when Steam writes the file with tabs. This PR is narrowed to the smaller parser robustness issue.

## What changed
- Keeps discovery focused on `SteamFinder`; removed the separate macOS Steam executable lookup from this PR.
- Parses quoted key/value pairs with general whitespace instead of tab-only matching.
- Keeps legacy top-level numeric library entries, but ignores nested numeric app IDs under `apps`.
- Adds focused tests for space-separated entries, nested `apps` entries, and relative `libraryfolders.vdf` input.

## Validation
- `git diff --check`
- Not run locally: this environment still has no .NET SDK installed.

### DIFF
--- a/Nitrox.Model/Platforms/Discovery/InstallationFinders/SteamFinder.cs
+++ b/Nitrox.Model/Platforms/Discovery/InstallationFinders/SteamFinder.cs
@@ -1,5 +1,7 @@
 using System;
+using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Runtime.InteropServices;
 using System.Text.RegularExpressions;
 using Nitrox.Model.Platforms.Discovery.InstallationFinders.Core;
@@ -50,7 +52,7 @@ public sealed class SteamFinder : IGameFinder
         return Ok(path);
     }
 
-    private static string GetSteamPath()
+    internal static string GetSteamPath()
     {
         // OSX: Steam dynamic data isn't near the steam exe. Because it can't (or isn't supposed to) write anything inside application bundle.
         if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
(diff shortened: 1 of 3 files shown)


dartasen:
Do you have a `libraryfolders.vdf` sample that contains the new format by any chance ?

vvitovec:
Yes. The shape I was targeting is the current Steam `libraryfolders.vdf` format where each library is a numbered object and the path is stored under a nested `path` key, for example:

```vdf
"libraryfolders"
{
    "0"
    {
        "path" "/Users/example/Library/Application Support/Steam"
        "label" ""
        "contentid" "123456789"
        "totalsize" "0"
        "update_clean_bytes_tally" "0"
        "time_last_update_verified" "0"
        "apps"
        {
            "264710" "1234567890"
        }
    }
    "1"
    {
        "path" "/Volumes/Games/SteamLibrary"
        "apps"
        {
            "264710" "1234567890"
        }
    }
}
```

The added `SteamFinderTest.SearchAllInstallations_ShouldParseModernLibraryFoldersVdf` uses a minimized version of that structure. If you prefer, I can trim this PR further to just parser support and leave the extra macOS Steam executable lookup for a separate PR.

dartasen:
I don't really get what scenarios our current parser doesn't handle, if you could give more insights on that.
The libraryfolder.vdf format didn't change since 2021 where they introduced the format you seems to speak about. Unless I'm misunderstanding something, we already support this format

image

vvitovec:
You are right that the earlier wording was too broad. The existing parser already handles the nested `path` format when the file is written in the usual tab-separated shape.

I narrowed this PR to the smaller case it actually changes: quoted VDF key/value lines separated by general whitespace rather than tabs. While doing that, I also addressed the review issues by removing the extra primary-library fallback, avoiding nested `apps` IDs as candidate library paths, and dropping the separate macOS Steam executable lookup from this PR.

So the PR is now just parser robustness plus focused tests. If that still feels too marginal, I am fine closing it and keeping the macOS work focused on the native app-bundle path changes.

vvitovec:
Closing this to keep the macOS stack focused. The narrowed whitespace-tolerance change is valid as parser hardening, but it is marginal and not necessary for the native macOS app-bundle path work. Thanks for the correction on the earlier wording; #2790 is the focused macOS path PR now.

PR #2790, opened 2026-06-16
[MacOS] Normalize native app bundle paths

vvitovec:
## Summary
- Normalize native macOS Steam game paths so selecting the `Subnautica` folder or `Subnautica.app` resolves to `Subnautica.app/Contents`.
- Keep the existing host executable layout checks for non-macOS paths.
- Keep the new layout detection internals private, so callers still use `HasValidGameFolder`, `HasGameExecutable`, and `NormalizeGamePath`.

## Context
This PR is intentionally limited to the native macOS app-bundle layout. I dropped Wine fallback/emulation work from this stack after maintainer feedback, because the macOS game build is native.

## Validation
- `git diff --check`
- Not run locally: .NET tests, because `dotnet` is not installed in this local environment.

### DIFF
--- a/Nitrox.Launcher/ViewModels/OptionsViewModel.cs
+++ b/Nitrox.Launcher/ViewModels/OptionsViewModel.cs
@@ -84,6 +84,7 @@ internal partial class OptionsViewModel(IKeyValueStore keyValueStore, StorageSer
 
     private static void SetTargetedSubnauticaPath(string path)
     {
+        path = GameInstallationHelper.NormalizeGamePath(path, GameInfo.Subnautica);
         if (!Directory.Exists(path))
         {
             return;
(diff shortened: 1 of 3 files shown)


vvitovec:
Rebased this onto current master and narrowed it to native macOS app-bundle path normalization only. I also kept the layout/executable lookup details private so the public helper surface stays small. No Wine fallback or emulation support is included here, following the feedback on #2794.

vvitovec:
The PR check appears to have been cancelled while it was still queued for the self-hosted runner. I tried to rerun it from my side, but GitHub requires repository admin rights for that. Could a maintainer rerun the check when a runner is available?

Measurity:
We don't have anyone hosting the runner. I'm cancelling all the jobs until we can host it again. Don't have an ETA.

PR #2791, opened 2026-06-16
[codex] Guard unsupported native macOS multiplayer

vvitovec:
## Purpose
Avoid patching or injecting the native macOS Subnautica client until native multiplayer runtime support exists. This keeps native singleplayer launch available but fails multiplayer clearly.

## What changed
- Skips patch removal for native macOS singleplayer launches.
- Shows an explicit unsupported dialog before multiplayer setup on native macOS layouts.
- Keeps the change local to launcher behavior; no Wine support is introduced.

## Validation
- `git diff --check`
- `dotnet build Nitrox.Launcher/Nitrox.Launcher.csproj --no-restore` could not be run here because `dotnet` is not installed.

### DIFF
--- a/Nitrox.Launcher/ViewModels/LaunchGameViewModel.cs
+++ b/Nitrox.Launcher/ViewModels/LaunchGameViewModel.cs
@@ -82,7 +82,10 @@ internal partial class LaunchGameViewModel(DialogService dialogService, ServerSe
                 return;
             }
 
-            NitroxEntryPatch.Remove(NitroxUser.GamePath);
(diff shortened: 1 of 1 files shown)


dartasen:
> Avoid patching or injecting the native macOS Subnautica client until native multiplayer runtime support exists

Why should we disable it ? It does work without any emulation

vvitovec:
Thanks, that answers the premise of this split. I opened this because I thought native macOS injection was still unsupported, but if it works without emulation then this guard would be wrong. Closing this slice.

dartasen:
@vvitovec Well you can give it a try, if it does not work on your mac we can investigate that 😄 

image

PR #2792, opened 2026-06-16
[codex] Send admin login permission updates directly

vvitovec:
## Purpose
Ensure the player who logs in as admin receives the permission-change packet directly after the server updates their permissions.

## What changed
- Stores the new permission in a local variable.
- Sends `PermsChanged` to the player session via `SendAsync` instead of replying with the packet payload.

## Validation
- `git diff --check`
- `dotnet test` could not be run here because `dotnet` is not installed.


dartasen:
Already fixed by #2786

PR #2793, opened 2026-06-16
[codex] Add LiteNetLib connection toggles

vvitovec:
## Purpose
Make client connection behavior easier to diagnose and tune in constrained runtime environments without changing defaults for normal users.

## What changed
- Adds environment-variable toggles for IPv6, native sockets, manual polling mode, CRC fallback, and connection timeout.
- Logs connection start/finish and disconnect reasons to help diagnose failed joins.
- Retries without the CRC packet layer only when explicitly enabled.

## Validation
- `git diff --cached --check` before commit
- `dotnet build NitroxClient/NitroxClient.csproj --no-restore` could not be run here because `dotnet` is not installed.

### DIFF
--- a/NitroxClient/Communication/NetworkingLayer/LiteNetLib/LiteNetLibClient.cs
+++ b/NitroxClient/Communication/NetworkingLayer/LiteNetLib/LiteNetLibClient.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Buffers;
+using System.Diagnostics;
 using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


dartasen:
Not sure about this one, I think I'll close this one. Thank's for your time anyways

Thoughts @Measurity ?

Measurity:
Yep. Having these options is another vector of connection problems between client and server (if they both have different options enabled). Not worth it at this time until we get feedback from users.

vvitovec:
Thanks, that makes sense. I agree this is not worth adding as a user/config surface without concrete user reports, especially if mismatched client/server options can create new connection failure modes. I will not pursue this direction further for the macOS compatibility split.

PR #2795, opened 2026-06-16
[codex] Document macOS launcher experiments

vvitovec:
## Purpose
Move the macOS launcher notes and helper script out of runtime code changes so maintainers can review documentation/tooling separately.

## What changed
- Adds `docs/macos-launcher.md` with native macOS and Wine experiment notes.
- Adds `scripts/macos-launcher.sh` as a local helper script for building/running the launcher flow.

## Validation
- `git diff --cached --check` before commit
- `bash -n scripts/macos-launcher.sh`
- `dotnet`-based validation could not be run here because `dotnet` is not installed.

### DIFF
--- a/docs/macos-launcher.md
+++ b/docs/macos-launcher.md
@@ -1,0 +1,115 @@
+# macOS launcher support
+
+This document tracks the current macOS state of the Nitrox launcher. It is not a claim of full native macOS Nitrox client support.
+
+## Prerequisites
+
+- macOS 12 or newer.
+- .NET 10 SDK.
+- Git.
+- Steam for native macOS Subnautica detection.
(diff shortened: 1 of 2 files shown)


dartasen:
Thank's !

But our technical documentation belongs to another repository.
AI generated documents aren't useful on their own especially with unnecessary contexts, which are more likely to reduce the models' performance.
If we were to give agents some tools to help on Nitrox, we would need to go down the road of skills and MCP (i.e: [https://github.com/pardeike/DecompilerServer](https://github.com/pardeike/DecompilerServer)). But that would need to be hand-crafted, not AI generated.
Closing for now.

vvitovec:
Thanks for clarifying. Understood on the documentation repository and on avoiding generated docs/tooling here. I’ll keep any future follow-up focused on small code changes or hand-crafted project-specific tooling only when there is a clear maintainer ask.

PR #2803, opened 2026-07-01
Fix console input corruption under non-native terminal front-ends (e.g. PufferPanel)

### LINKED ISSUE #2616: Using PufferPanel: Nitrox Server starts printing nonsense when I use any command

1.8.0.1 (Latest release)

### Details about version

_No response_

Linux

Steam

### Describe the issue.

If I type any command in it starts printing nonsense. For example if I write 'help' this happens:

`leh ppleh hhpleh eehpleh llehpleh pplehpleh //plehpleh hh/plehpleh eeh/plehpleh lleh/plehpleh ppleh/plehpleh ��pleh/plehpleh --�pleh/plehpleh ��-�pleh/plehpleh SS�-�pleh/plehpleh ]]S�-�pleh/plehpleh tt]S�-�pleh/plehpleh ��t]S�-�pleh/plehpleh rr�t]S�-�pleh/plehpleh ��r�t]S�-�pleh/plehpleh \\�r�t]S�-�pleh/plehpleh tt\�r�t]S�-�pleh/plehpleh ��t\�r�t]S�-�pleh/plehpleh t[05:41:21.249] World state saved [05:41:21.531] World backed up t�t\�r�t]S�-�pleh/plehpleh ��t�t\�r�t]S�-�pleh/plehpleh JJ�t�t\�r�t]S�-�pleh/plehpleh SSJ�t�t\�r�t]S�-�pleh/plehpleh ]]SJ�t�t\�r�t]S�-�pleh/plehpleh t[DAEMON] Server was told to stop t]SJ�t�t\�r�t]S�-�pleh/plehpleh ��t]SJ�t�t\�r�t]S�-�pleh/plehpleh tt�t]SJ�t�t\�r�t]S�-�pleh/plehpleh fft�t]SJ�t�t\�r�t]S�-�pleh/plehpleh ��ft�t]SJ�t�t\�r�t]S�-�pleh/plehpleh JJ�ft�t]SJ�t�t\�r�t]S�-�pleh/plehpleh SSJ�ft�t]SJ�t�t\�r�t]S�-�pleh/plehpleh ]]SJ�ft�t]SJ�t�t\�r�t]S�-�pleh/plehpleh tt]SJ�ft�t]SJ�t�t\�r�t]S�-�pleh/plehpleh ��t]SJ�ft�t]SJ�t�t\�r�t]S�-�pleh/plehpleh ���t]SJ�ft�t]SJ�t�t\�r�t]S�-�pleh/plehpleh t[DAEMON] Server killed`

Everything still works completely fine I just can't use the console ALSO I forgot to mention I am using PufferPanel 3 R16 and I am running Ubuntu Server 24.04.3 and I have to kill it because even the 'quit' command doesn't work. 
Just checked and It works fine in SSH.

1. Start the server through PufferPanel.
2. Type in any command.
3. It starts printing nonsense.

### Relevant log output

I couldn't find one

SmokerDoker:
## Linked Issues / Context

Fixes #2616

## Description of Change

When running the Nitrox server through PufferPanel (which relays console input/output through its own web-based terminal wrapper instead of a real interactive PTY), typing any console command produced garbled, repeating, partially-reversed output (e.g. typing `help` echoed back mangled fragments), and even `quit` stopped working, requiring the process to be killed. The same server worked fine when connected to over SSH.

**Root cause:** `ConsoleInputService` implements its own line editor (arrow keys, backspace, history, tab) and repeatedly reads the `Console.CursorLeft` **getter** to track where the caret currently is. On Unix, .NET implements that getter by writing an ANSI "Device Status Report" escape sequence (`\x1b[6n`) to stdout and then reading the terminal's position response back from the *same* stdin stream that `Console.ReadKey()` uses to read key presses.

- Over a real terminal (SSH), that query/response round-trip resolves instantly and correctly.
- Under PufferPanel's console wrapper (not a genuine PTY), that escape response either isn't relayed back correctly or gets interleaved with real key presses in the same input buffer. `Console.ReadKey()` then misreads those stray escape/response bytes as literal typed characters, which is exactly what produced the corrupted, growing, reversed-looking output reported in the issue — and eventually desynced the input state badly enough that `quit` no longer registered.

**Fix:** `ConsoleInputService` now tracks the caret position itself in a local variable instead of reading it back from the terminal. `Console.CursorLeft` is only ever *written* to (to move the visible cursor for redraws), never read — a write-only operation that doesn't require any response from the terminal. This removes the stdin round-trip entirely, so the behavior no longer depends on whether the underlying terminal front-end correctly relays ANSI cursor-position queries.

## Validation

I verified the refactor line-by-line against the original implementation to confirm identical logic/semantics (every prior read of `Console.CursorLeft` is replaced by an equivalent read of the new locally-tracked caret index, with the index updated at every point where the terminal cursor would previously have moved). I was not able to build/run the project in the environment this change was written in.

**This should still be tested on a real PufferPanel setup (or similar non-PTY console wrapper) before merging**, to confirm the fix resolves the reported behavior in practice — normal SSH/interactive console usage should also be re-checked (arrow keys, backspace, tab, history) to make sure nothing regressed there.

## PR Checklist

- [x] Has a linked issue

---

### DIFF
--- a/Nitrox.Server.Subnautica/Services/ConsoleInputService.cs
+++ b/Nitrox.Server.Subnautica/Services/ConsoleInputService.cs
@@ -44,6 +44,10 @@ internal sealed class ConsoleInputService(CommandService commandService, IPacket
     private async Task HandleInputAsync(CancellationToken cancellationToken)
     {
         StringBuilder inputLineBuilder = new();
+        // Tracks the caret position ourselves instead of reading it back from Console.CursorLeft: on Unix, that getter
+        // round-trips an ANSI cursor-position query through the same stdin stream that key presses are read from, which
+        // some non-native terminal front-ends (e.g. PufferPanel's console) don't relay correctly, corrupting input.
+        int caretIndex = 0;
 
         while (!cancellationToken.IsCancellationRequested)
         {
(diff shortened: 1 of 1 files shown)


Measurity (approved):
LGTM, thanks for the fix!

dartasen:
@SmokerDoker If you're able to confirm the fix, that would be awesome. Otherwise the code lgtm

PR #2804, opened 2026-07-01
Reduce movement sync bandwidth by throttling player movement packets

### LINKED ISSUE #2705: Reduce network bandwidth requirement by optimizing movement sync
Movement data is synced at a rate of 30/s. This can be reduced by using velocity vectors instead of positions. If remote player is not moving the characters by some threshold, no movement data should be transferred either.

With the change, the movement packet guarantee should be set to "last packet only" so that when player stops, the player won't continue moving on remote machines.

Occasionally the position should be resynced to avoid inaccuracies (due to latency or physics) to accumulate. But this should only happen if the player isn't standing still.

SmokerDoker:
## Linked Issues / Context

Addresses #2705

## Description of Change

Movement data was being synced far more often than necessary: `PlayerMovementBroadcaster` sent a `PlayerMovement` packet on every single `Update()` frame, uncapped, regardless of whether the player was actually moving. Velocity was already part of the packet, so the main remaining bandwidth win was to stop sending when nothing meaningful changed.

**`PlayerMovementBroadcaster`** now mirrors the threshold/safety-window pattern already used for vehicles (`WatchedEntry`):
- Packets are capped at the existing 30 Hz rate (`MovementBroadcaster.BROADCAST_PERIOD`) instead of sending every rendered frame.
- Packets are skipped entirely while the position/rotation delta since the last sent packet stays below a small threshold.
- Once the player comes to a full stop (velocity below a small threshold), exactly **one** final packet is sent so remote clients settle on the stopped state, then broadcasting stops completely until the player moves again.
- If the player has residual velocity without net displacement past the threshold yet (e.g. oscillating in place), a periodic resync (5s window + short safety grace period, same constants as `WatchedEntry`) still corrects any drift accumulating on remote machines — but only while not standing still, per the issue's request.

**`PlayerMovement`**'s delivery method changes from `UNRELIABLE_SEQUENCED` to `RELIABLE_ORDERED_LAST` ("last packet only", but guaranteed delivery — same as already used for `PlayerStats`), so that final "stopped" packet can't silently get dropped, which would otherwise leave remote players sliding indefinitely.

**`RemotePlayer.UpdatePosition`** derives its velocity-correction time window from the actual elapsed time since the previous packet instead of assuming a fixed `Time.fixedDeltaTime` cadence, since packets can now arrive at irregular intervals (or not at all while idle).

Additionally, while play-testing this change, we found that `MovementHelper.GetCorrectedVelocity`/`GetCorrectedAngularVelocity` compute a "keep nudging towards target" velocity that's only ever exactly zero by coincidence — under the old every-frame-packet model this self-corrected within one frame, but once a stopped player stops sending packets entirely, any left-over non-zero result from the last correction got applied by physics forever, causing remote players to drift or spin in place indefinitely after coming to rest. `RemotePlayer.UpdatePosition` now recognizes a (near) zero incoming velocity as an explicit "I've stopped" signal and snaps directly to the reported position/rotation with both velocities hard-zeroed in that case, instead of running the generic corrective-velocity math.

## Validation

Manually tested with 2 clients (host + client) over an extended session:
- Normal walking/swimming movement: smooth, no jitter or rubber-banding, indistinguishable from before the change.
- Coming to a stop: remote player correctly settles and stays still (this required the `RemotePlayer.UpdatePosition` fix above after an initial test revealed the endless-drift/spin bug described there).
- One remaining minor cosmetic artifact: turning to face a new direction and then immediately stopping can cause the remote view to keep rotating for a couple of milliseconds (network latency for that final packet) before snapping to the correct final orientation. This is standard extrapolation/latency behavior and was judged not worth the added physics-interaction complexity (temporarily toggling kinematic state, coroutine-driven interpolation) to smooth out further.

Given this touches core movement netcode, further testing (in-game, different network conditions, vehicles/base entry edge cases) by maintainers/other players before merging is still recommended.

## PR Checklist

- [x] Has a linked issue

---

### DIFF
--- a/Nitrox.Model.Subnautica/Packets/PlayerMovement.cs
+++ b/Nitrox.Model.Subnautica/Packets/PlayerMovement.cs
@@ -15,7 +15,7 @@ public class PlayerMovement : Movement
         Velocity = velocity;
         BodyRotation = bodyRotation;
         AimingRotation = aimingRotation;
-        DeliveryMethod = NitroxDeliveryMethod.DeliveryMethod.UNRELIABLE_SEQUENCED;
+        DeliveryMethod = NitroxDeliveryMethod.DeliveryMethod.RELIABLE_ORDERED_LAST;
         UdpChannel = UdpChannelId.MOVEMENTS;
     }
 
--- a/NitroxClient/GameLogic/RemotePlayer.cs
+++ b/NitroxClient/GameLogic/RemotePlayer.cs
@@ -68,6 +68,7 @@ public class RemotePlayer : INitroxPlayer
     private const float STOPPED_VELOCITY_THRESHOLD = 0.1f;
 
     private float lastPositionUpdateTime = -1f;
+    private bool previousUpdateWasStop;
 
     public RemotePlayer(PlayerContext playerContext, PlayerModelManager playerModelManager, PlayerVitalsManager playerVitalsManager, FMODWhitelist fmodWhitelist)
     {
@@ -165,8 +166,15 @@ public class RemotePlayer : INitroxPlayer
         SetPilotingChair(null);
 
         // Movement packets no longer arrive at a fixed cadence (idle players send none at all, moving players are throttled),
-        // so the correction has to be based on how long it's actually been since the last packet instead of a fixed timestep.
-        float correctionTime = lastPositionUpdateTime >= 0f ? Mathf.Max(Time.time - lastPositionUpdateTime, Time.fixedDeltaTime) : Time.fixedDeltaTime;
+        // so the correction is normally based on how long it's actually been since the last packet instead of a fixed timestep.
+        // Exception: if the player was standing still until now, that idle time doesn't represent any real drift to catch
+        // up on (the remote transform was already snapped exactly to rest by the hasStopped branch below). Using the full
+        // idle duration here would divide the tiny position delta of the first "moving again" packet by that whole
+        // duration, producing a near-zero corrective velocity instead of the real reported speed - which looks like a
+        // brief stutter/delay right when the player starts moving again after standing still for a while.
+        float correctionTime = (lastPositionUpdateTime >= 0f && !previousUpdateWasStop)
+            ? Mathf.Max(Time.time - lastPositionUpdateTime, Time.fixedDeltaTime)
+            : Time.fixedDeltaTime;
         lastPositionUpdateTime = Time.time;
 
         AnimationController.AimingRotation = aimingRotation;
@@ -178,6 +186,7 @@ public class RemotePlayer : INitroxPlayer
         // result (e.g. from residual jitter) would otherwise keep getting integrated by physics forever, causing the
         // remote player to drift or spin in place indefinitely instead of actually coming to rest.
         bool hasStopped = velocity.sqrMagnitude < STOPPED_VELOCITY_THRESHOLD * STOPPED_VELOCITY_THRESHOLD;
+        previousUpdateWasStop = hasStopped;
         AnimationController.Velocity = hasStopped ? Vector3.zero : MovementHelper.GetCorrectedVelocity(position, velocity, Body, correctionTime);
 
         // If in a subroot the position will be relative to the subroot
--- a/NitroxClient/GameLogic/RemotePlayer.cs
+++ b/NitroxClient/GameLogic/RemotePlayer.cs
@@ -61,6 +61,12 @@ public class RemotePlayer : INitroxPlayer
 
     public CyclopsPawn Pawn { get; set; }
 
+    /// <remarks>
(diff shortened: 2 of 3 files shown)


Measurity (changes requested):
This feature will be really helpful for people. Thanks for taking the time to fix it! 

There's some delay when moving the player for the first time. Would it be possible to not have thresholds and detect mouse movement / WASD input instead? As the trigger to send data?

SmokerDoker:
Thanks for the feedback! I looked into it, and I don't think switching the trigger to raw WASD/mouse input is the right fix, for a couple of reasons:

Not all movement is input-driven — water currents, falling, knockback from creatures, and moving while inside a moving vehicle/sub wouldn't trigger a send at all if we only listened for key/mouse input, so remote players would appear frozen during all of that instead of just idle.
Mouse movement as a trigger is really just another threshold, one step earlier in the pipeline — you'd still need to filter out tiny/incidental mouse jitter, or bandwidth usage would go up instead of down.
I did track down what's actually causing the delay though: it's a bug on the receiving side, not the trigger logic. When a player starts moving again after standing still for a while, the very first "moving" packet's velocity correction was being divided by the entire idle duration instead of a normal short interval, since no packets were sent while idle. That produced a near-zero corrective velocity for one frame (instead of the real reported speed), which is exactly the stutter/delay you saw — it then self-corrected on the next packet ~30ms later.

Fixed in the latest commit: the remote player now remembers when it was left standing still, and uses a normal correction window for the packet that resumes movement instead of the stale idle time. Should be gone now — let me know if you still see it after this.

Measurity (approved):
Thanks! Works well here, really close to previous behavior but way less network usage.

dartasen (approved):
Will need an IG test with the team

dartasen (changes requested):
Tested the PR with 6 people today

Cause some regressions inside Subroots : 
[https://streamable.com/zh3d9v](https://streamable.com/zh3d9v)

But locally with two players it's pretty easy to reproduce the regression : 
Enter a base with one player => the second player will see the player far away from it's real location

SmokerDoker:
@dartasen Okay thanks for the test, im gonna take a look to see what causes this.

SmokerDoker:
@dartasen  Thanks for the detailed report — found the cause, and it explains exactly why it only happened inside bases.

`PlayerMovementBroadcaster.Update()` converts the local player's position to base-local space (subtract + un-rotate), but then immediately called `subRootTransform.TransformPoint(currentPosition)` on the result — which transforms a local point *back* to world space. That mathematically undoes the conversion that was just done, so the broadcaster was actually sending **world position** over the network. `RemotePlayer.UpdatePosition` on the receiving side, however, only applies its local→world conversion for `SubRoot.isBase`, expecting a base-local position — so it converted the already-world position a second time, producing a badly wrong result (hence players flying off / disappearing) whenever someone was inside a base. Vehicles (Cyclops) were never affected since the receiver doesn't apply that conversion for non-base subroots.

This coordinate mismatch has actually been sitting in the code since August 2024 (an unrelated commit re-added the position-subtraction lines without removing the leftover `TransformPoint` call), but it went unnoticed because movement packets used to be sent every single frame — any bad correction got silently overwritten by the next packet a frame later. The new threshold-based throttling in this branch sends corrective packets far less often, so a single bad computation now produces a visible, lasting jump instead of a one-frame flicker that nobody could ever see. That's also why it never showed up in my own solo testing.

Fixed in `3c0c0bbf`: restricted the local-space conversion to `SubRoot.isBase` (mirroring the receiver's own condition) and dropped the erroneous `TransformPoint` call. Position/rotation for a base are now genuinely base-local; everything else (Cyclops etc.) is unaffected.

Could you have your 6-person group re-test the same base scenario when you get a chance? Would be great to confirm this actually resolves it in practice.

SpaceMonkeyy86:
IMO not sending any movement packets at all while a player isn't moving is bad for stability. If anything were to push a remote player object around, or move it in any way, the player's position would be desynced until that player started moving on their end again. A common approach is to send movement packets at a much slower rate (say 1/s) when not moving, so any desync can be corrected quickly.

SmokerDoker:
> IMO not sending any movement packets at all while a player isn't moving is bad for stability. If anything were to push a remote player object around, or move it in any way, the player's position would be desynced until that player started moving on their end again. A common approach is to send movement packets at a much slower rate (say 1/s) when not moving, so any desync can be corrected quickly.

Fixed in f4173991: instead of going fully silent after the "stopped" packet, the broadcaster now keeps the existing MAX_TIME_WITHOUT_BROADCAST resync running while idle (5s window + safety grace) — the same low-rate idle correction WatchedEntry already does for parked vehicles. That's ~1 packet every 5s while standing still rather than 1/s, which keeps player and vehicle netcode consistent and is still far below the old per-frame rate. Any drift now self-corrects within ~5s even if the player never moves.

PR #2805, opened 2026-07-01
Remove server session bloat & fix player name already reserved bug

Measurity:
No more redundant `correlationId` or `reservationId` tracking. Both are merged into `SessionId`.

## TODO
- [x] Nitrox.Test code changes


Nitrox.Server.Subnautica/Models/Commands/ConfigCommand.cs
```diff
@@ -18,6 +18,7 @@ public async Task Execute(ICommandContext context)
         {
             // TODO: Handle this case to generate config?
             await context.ReplyAsync("No configuration file exists");
+            return;
```
dartasen, line 21: 👁️

Nitrox.Model/Packets/CorrelatedPacket.cs
```diff
@@ -1,10 +0,0 @@
-﻿using System;
-
-namespace Nitrox.Model.Packets;
-
-// TODO: Refactor this away. Use SessionId instead of correlationId.
```
dartasen, line 5: Maybe we should do a ~spring~ summer cleanup of TODOs and create issues for the one we wanna keep

dartasen (approved):
Can't test IG, but otherwise code lgtm

PR #2806, opened 2026-07-02
Fix remote player equipment (e.g. rebreather) not showing in sessions with 3+ players

### LINKED ISSUE #2359: Recheck player armors mapping

SmokerDoker:
Fixes #2359

## Problem

Since #2635, `RegisterEquipmentVisibilityHandler` is correctly called when each remote player spawns, yet the issue persisted: in sessions with multiple players (the reporters were four), some players are missing equipment on their model — most noticeably the rebreather, since its absence on the head is immediately obvious, whereas e.g. a wrong glove barely stands out. Testing with only one remote player (2 players total) never reproduced the bug — which is exactly what masked it during the earlier local verification.

## Root Cause

`PlayerModelManager` is registered in Autofac as `InstancePerLifetimeScope` (`ClientAutoFacRegistrar.cs`), meaning **a single instance is injected into every `RemotePlayer` object for the whole session** (`PlayerManager.cs:76`).

`PlayerModelManager` used to store the equipment visibility handlers in a single mutable field:

```csharp
private List<IEquipmentVisibilityHandler> equipmentVisibilityHandlers;

public void RegisterEquipmentVisibilityHandler(GameObject playerModel)
{
    equipmentVisibilityHandlers = new List<IEquipmentVisibilityHandler> { ... };
}
```

Each of these handlers (`ScubaSuitVisibilityHandler`, etc.) captures a reference to the `playerModel` GameObject it was constructed with via closure. Because `equipmentVisibilityHandlers` is a **shared** field, every remote player that (re)spawns (`RemotePlayer.InitializeGameObject` → `RegisterEquipmentVisibilityHandler`) overwrites that list for **every** player in the session.

Result: once `RemotePlayer.UpdateEquipmentVisibility(...)` is called for player A, the update actually lands on the model of whichever player was registered last (e.g. player D) — not on player A's own model. With exactly two players (1 local + 1 remote) there is never a second registration to overwrite the first, so the bug is invisible there. It only becomes visible from the third player onward, and reproducibly depends on join order — which explains why targeted testing with only a couple of people never caught it, while it showed up reliably in the real 4-player playtest.

## Fix

`equipmentVisibilityHandlers` is no longer kept on the shared `PlayerModelManager`; it now belongs to each individual `RemotePlayer`:

- `PlayerModelManager.RegisterEquipmentVisibilityHandler(...)` now returns the handler list instead of caching it internally.
- `PlayerModelManager.UpdateEquipmentVisibility(...)` is now a stateless static method that applies a given handler list.
- `RemotePlayer` holds its own `equipmentVisibilityHandlers` list and passes it in on every equipment update.

This makes it impossible for one remote player's spawn to overwrite another's handlers.

---

- [x] Has a linked issue

### DIFF
--- a/NitroxClient/GameLogic/RemotePlayer.cs
+++ b/NitroxClient/GameLogic/RemotePlayer.cs
@@ -6,6 +6,7 @@ using NitroxClient.GameLogic.HUD;
 using NitroxClient.GameLogic.PlayerLogic;
 using NitroxClient.GameLogic.PlayerLogic.PlayerModel;
 using NitroxClient.GameLogic.PlayerLogic.PlayerModel.Abstract;
+using NitroxClient.GameLogic.PlayerLogic.PlayerModel.Equipment.Abstract;
 using NitroxClient.MonoBehaviours;
 using NitroxClient.MonoBehaviours.Cyclops;
 using NitroxClient.MonoBehaviours.Gui.HUD;
@@ -32,6 +33,7 @@ public class RemotePlayer : INitroxPlayer
     private readonly PlayerModelManager playerModelManager;
     private readonly PlayerVitalsManager playerVitalsManager;
     private readonly FMODWhitelist fmodWhitelist;
+    private List<IEquipmentVisibilityHandler> equipmentVisibilityHandlers;
 
     public PlayerContext PlayerContext { get; }
     public GameObject? Body { get; private set; }
@@ -100,7 +102,7 @@ public class RemotePlayer : INitroxPlayer
 
         CoroutineUtils.StartCoroutineSmart(playerModelManager.AttachPing(this));
         playerModelManager.BeginApplyPlayerColor(this);
-        playerModelManager.RegisterEquipmentVisibilityHandler(PlayerModel);
+        equipmentVisibilityHandlers = playerModelManager.RegisterEquipmentVisibilityHandler(PlayerModel);
         SetupBody();
         SetupSkyAppliers();
         SetupPlayerSounds();
@@ -394,7 +396,7 @@ public class RemotePlayer : INitroxPlayer
 
     public void UpdateEquipmentVisibility(List<TechType> equippedItems)
     {
-        playerModelManager.UpdateEquipmentVisibility(new ReadOnlyCollection<TechType>(equippedItems));
+        PlayerModelManager.UpdateEquipmentVisibility(equipmentVisibilityHandlers, new ReadOnlyCollection<TechType>(equippedItems));
     }
 
     /// <summary>
--- a/NitroxClient/GameLogic/RemotePlayer.cs
+++ b/NitroxClient/GameLogic/RemotePlayer.cs
@@ -1,5 +1,4 @@
 using System.Collections.Generic;
-using System.Collections.ObjectModel;
 using Nitrox.Model.Core;
 using Nitrox.Model.DataStructures.GameLogic;
 using NitroxClient.GameLogic.HUD;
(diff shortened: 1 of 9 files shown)


dartasen:
While doing playlists with the team, the issue was sometimes hard to reproduce but we know it exists.
Did you find any reproduction steps I can follow to validate the fix ?

SmokerDoker:
I have wrote it down somewhere let me just find it

SmokerDoker:
Repro steps, derived from the root cause:

Minimum 3 players total (not necessarily 4 — that's just what the original reporters happened to have). The bug is client-side: it triggers on any client that has 2+ other players loaded as RemotePlayers, so all 3 participants can independently hit it on their own screen.

Player A is already in-game.

Player B joins. On any client that now sees B, PlayerModelManager's shared handler list gets registered pointing at B's model.

Player C joins after B. On any client that now sees both, C's spawn overwrites that same shared field to point at C's model instead — B's handler reference is now stale.

Trigger: have B (the earlier-registered, now-stale player) change something visible — cleanest is taking Rebreather off/on.

On a client that can see both B and C: the update doesn't land on B's model (rebreather fails to show correctly on B), because the shared field currently points at C.

NitroxClient/GameLogic/RemotePlayer.cs
```diff
@@ -394,7 +396,7 @@ public void UpdateAnimationAndCollider(AnimChangeType type, AnimChangeState stat
 
     public void UpdateEquipmentVisibility(List<TechType> equippedItems)
     {
-        playerModelManager.UpdateEquipmentVisibility(new ReadOnlyCollection<TechType>(equippedItems));
+        PlayerModelManager.UpdateEquipmentVisibility(equipmentVisibilityHandlers, new ReadOnlyCollection<TechType>(equippedItems));
```
dartasen: Do you think we can avoid to reallocate the whole collection every time someone changes some piece of equipment ?

EDIT: Probably out of scope of this PR, there's quite a lot to refactor inside VisibilityHandlers which is quite old code (2019 lol)

NitroxClient/GameLogic/RemotePlayer.cs
```diff
@@ -32,6 +33,7 @@ public class RemotePlayer : INitroxPlayer
     private readonly PlayerModelManager playerModelManager;
     private readonly PlayerVitalsManager playerVitalsManager;
     private readonly FMODWhitelist fmodWhitelist;
+    private List<IEquipmentVisibilityHandler> equipmentVisibilityHandlers;
```
dartasen:
```suggestion
    private List<IEquipmentVisibilityHandler> equipmentVisibilityHandlers = [];
```

SmokerDoker: Good catch, applied — equipmentVisibilityHandlers is now initialized to [] instead of null.

NitroxClient/GameLogic/RemotePlayer.cs
SmokerDoker: Yeah, this one turned out to be avoidable without touching the old VisibilityHandler internals: all six handlers only ever call .Contains(...) on the collection, so there's no need for the ReadOnlyCollection<TechType> wrapper at all. I changed the signature (interface + all handlers + both UpdateEquipmentVisibility methods) to take IReadOnlyList<TechType> instead, so the underlying List<TechType> from PlayerMetadataProcessor is now passed straight through — no allocation on every equipment update anymore. List<T> still gets recognized as ICollection<T> by Enumerable.Contains, so there's no behavior change, just one fewer allocation per update.
▎
▎ Agreed the bigger VisibilityHandler refactor (2019-era code) is out of scope here.

dartasen (approved):

PR #2809, opened 2026-07-03
Fix battery duplication when picking up a batteryless tool (#2668)

### LINKED ISSUE #2668: Battery duplication issue

SmokerDoker:
## Summary

Fixes #2668 — battery-powered tools (Flashlight, Scanner, Seaglide, etc.) duplicated a fully-charged battery whenever an emptied tool was dropped and picked back up.

## Root cause

`Items.ConvertToInventoryItemEntity` unconditionally called `BatteryChildEntityHelper.TryPopulateInstalledBattery`, which always created a brand-new default (fully charged) `InstalledBatteryEntity` for any object with an `EnergyMixin`, regardless of whether a battery was actually installed. Since Nitrox disables Subnautica's own default-battery auto-spawn (`EnergyMixin_SpawnDefaultAsync_Patch`), this was needed for freshly crafted items — but it was also wrongly applied to already-known items being picked back up, silently re-filling any empty battery slot.

## Related prior attempt (#2752)

A previous fix (#2752) addressed the basic case but was closed unmerged: [@Measurity](https://github.com/Measurity) found that it broke the cross-player case — if Player A drops a tool with a battery installed and Player B picks it up, Player B would receive the tool without its battery.

That regression came from a second, separate gap: `Items.Dropped()` never captured the currently-installed battery at all when converting an item to a `WorldEntity`, and the receiving client's `InstalledBatteryEntity` spawn (for a battery coming from another player) is asynchronous, so a pickup performed in that narrow window would see an empty `EnergyMixin.batterySlot`.

## Fix

- `Items.ConvertToInventoryEntityUntracked` now distinguishes between genuinely new objects (e.g. freshly crafted — still allowed a default battery) and already-known objects (dropped/picked up before — only reflect their real current battery state, never assume one).
- `Items.Dropped()` now also reflects the tool's actual currently-installed battery (if any) when converting it to a `WorldEntity`, instead of silently losing that information on drop.
- `BatteryChildEntityHelper.TryPopulateInstalledBattery` now checks the live `EnergyMixin.batterySlot.storedItem` first and, if present, reuses that battery's real id/tech type/charge instead of fabricating a new default one. It only falls back to creating a default battery for genuinely new objects with no battery at all.
- Added a small pending-battery registry (`BatteryChildEntityHelper` + `InstalledBatteryEntitySpawner`) that closes the short window between an `InstalledBatteryEntity` being received over the network and its backing GameObject finishing its (async) spawn into `EnergyMixin.batterySlot`. This is what makes the cross-player pickup case (Measurity's finding above) work correctly.
- Added a local-only `EnergyMixin_OnRemoveItem_Patch` that clears that pending-battery bookkeeping when a battery is actually removed from a slot, so a later pickup can't resurrect a battery that isn't there anymore. This patch does not broadcast anything over the network.

## Test plan

Tested manually with 2 Windows clients:
- [x] Craft a battery-powered tool, remove its battery, drop it, pick it back up → stays batteryless (original exploit fixed).
- [x] Craft a new battery-powered tool → still receives its intended default battery (no regression).
- [x] Player A drops a tool with a battery installed, Player B picks it up → Player B correctly receives the tool with its battery (Measurity's regression fixed).

### DIFF
--- a/NitroxClient/GameLogic/Helper/BatteryChildEntityHelper.cs
+++ b/NitroxClient/GameLogic/Helper/BatteryChildEntityHelper.cs
@@ -1,10 +1,12 @@
 using System;
 using System.Collections.Generic;
+using NitroxClient.GameLogic.Spawning.Metadata;
 using NitroxClient.MonoBehaviours;
 using Nitrox.Model.Core;
 using Nitrox.Model.DataStructures;
 using Nitrox.Model.Subnautica.DataStructures.GameLogic;
 using Nitrox.Model.Subnautica.DataStructures.GameLogic.Entities;
+using Nitrox.Model.Subnautica.DataStructures.GameLogic.Entities.Metadata;
 using UnityEngine;
 using UWE;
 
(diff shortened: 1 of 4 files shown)


NitroxClient/GameLogic/Helper/BatteryChildEntityHelper.cs
MistaOmega, line 1: Documentation throughout this MR is crazy, there's a lot of the "what" when we just need the "why"

I would make the not-so-crazy assumption that all this documentation is AI generated. When using AI for this stuff, adding some restrictions into whatever configuration file it has can help with this OTT comment style AI agents tend to do.

MistaOmega:
Tested functionality and matched @SmokerDoker's results
Tested with 2 clients on the same Linux machine

SpaceMonkeyy86 (approved):
LGTM CW and works when testing

NitroxClient/GameLogic/Helper/BatteryChildEntityHelper.cs
SmokerDoker: Yeah its AI generated haha. But yeah no problem I will optimize my prompts next time.

PR #2811, opened 2026-07-03
Sync base flood state so reconnecting players don't see it restart

### LINKED ISSUE #2808: FloodState is not synchronized correctly between players

### Details about version

029d320

Windows

Steam

### Describe the issue.

There is no synchronization for the base flooding state when a player disconnects and reconnects. The FloodState starts from zero again for the reconnected player, while the player who stayed in the session continues to see the actual flood progression of the base. This causes both players to see different water levels for the same base.

When a player reconnects, the flooding state is not restored from the current server state. Instead, the rejoined player sees the base as if flooding had restarted, while the remaining player sees the already progressed or fully flooded state. The expected behavior would be that reconnecting players receive the current flood state immediately and see the same base state as everyone else.

1. Join a session with a base that can flood.

2. Let the base flood for a while.

3. Disconnect a player.

4. Reconnect the same player.

5. Compare the flood state between the reconnecting player and the player who stayed connected.

### Relevant log output

_No response_

SmokerDoker:
## Summary

Fixes #2808 — reconnecting players saw their base's flooding restart from empty, while players who stayed connected kept seeing the actual, progressed water level.

## Root cause

`BaseFloodSim`'s per-cell water levels are purely local simulation state on each client - there was no synchronization at all. `BuildingResyncProcessor` rebuilds the base from scratch on reconnect (as already established for the related leak-sync fix in #2769), but never restored any water level afterward, so the newly-rebuilt `BaseFloodSim` always started at zero and had to gradually re-simulate flooding from there.

## Fix

- New `BaseFloodMetadata` entity metadata type carrying the base's flood grid, using `BaseFloodSim`'s own `OnProtoSerialize`/`OnProtoDeserialize` methods (the same ones vanilla Subnautica uses for its own save-game persistence) - no reflection needed.
- A throttled broadcast (`BaseFloodSim_Update_Patch`) periodically shares the current flood state, reusing the existing `EntityMetadataChangedThrottled` mechanism also used for e.g. battery charge and sub name color.
- The processor only applies a received update if it represents *at least as much* flooding as what's already simulated locally, and only if the grid size still matches the base's current shape. This prevents an already-actively-simulating client from having its progress periodically reset backwards by another player's (possibly slightly stale) broadcast, while still letting a freshly (re)joining client - whose own simulation starts empty - catch up to the real state.
- `BuildingResyncProcessor.OverwriteBase` now explicitly applies the base's stored metadata after rebuilding it, since that path (unlike the normal entity spawn path) doesn't do so automatically.

## Test plan

Tested manually with 2 clients:
- [x] Base floods normally for players who stay connected (no regression - an earlier iteration of this fix caused flooding to stall entirely; fixed by the "only apply if it represents more flooding" check above).
- [x] Player disconnects while a base is actively flooding, reconnects later, and now sees the same progressed water level as the player who stayed connected, instead of it restarting from empty.
- [x] FloodState survives both reconnects and full server restarts on both clients - whether the water level is currently rising or falling, it stays in sync between both clients throughout.

### DIFF
--- a/NitroxClient/GameLogic/Spawning/Metadata/Extractor/BaseFloodMetadataExtractor.cs
+++ b/NitroxClient/GameLogic/Spawning/Metadata/Extractor/BaseFloodMetadataExtractor.cs
@@ -1,0 +1,14 @@
+using Nitrox.Model.Subnautica.DataStructures.GameLogic.Entities.Metadata;
+using NitroxClient.GameLogic.Spawning.Metadata.Extractor.Abstract;
+
+namespace NitroxClient.GameLogic.Spawning.Metadata.Extractor;
+
+public class BaseFloodMetadataExtractor : EntityMetadataExtractor<BaseFloodSim, BaseFloodMetadata>
+{
+    public override BaseFloodMetadata Extract(BaseFloodSim entity)
+    {
+        // The serializer argument isn't actually used by BaseFloodSim's implementation; it only (re)computes flatValueGrid from the current simulation state.
+        entity.OnProtoSerialize(null);
+        return new(entity.flatValueGrid);
+    }
+}
--- a/NitroxClient/GameLogic/Spawning/Metadata/Processor/BaseFloodMetadataProcessor.cs
+++ b/NitroxClient/GameLogic/Spawning/Metadata/Processor/BaseFloodMetadataProcessor.cs
@@ -1,0 +1,52 @@
+using Nitrox.Model.Subnautica.DataStructures.GameLogic.Entities.Metadata;
+using NitroxClient.GameLogic.Spawning.Metadata.Processor.Abstract;
+using UnityEngine;
+
+namespace NitroxClient.GameLogic.Spawning.Metadata.Processor;
+
+public class BaseFloodMetadataProcessor : EntityMetadataProcessor<BaseFloodMetadata>
+{
+    public override void ProcessMetadata(GameObject gameObject, BaseFloodMetadata metadata)
+    {
+        BaseFloodSim baseFloodSim = gameObject.GetComponent<BaseFloodSim>();
+
+        if (!baseFloodSim)
+        {
+            Log.Error($"Could not find BaseFloodSim on {gameObject.name}");
+            return;
+        }
+
+        float[] incoming = metadata.FlatValueGrid;
+        // The base's geometry may have changed (or not finished (re)building yet) since this snapshot was captured elsewhere,
+        // in which case the array no longer maps to the same cells. Applying it regardless would throw inside BaseFloodSim's
+        // own restore logic (index mismatch) and permanently break its Update() loop (no more leaks/flooding) from then on.
+        if (incoming == null || incoming.Length != baseFloodSim.shape.Size)
+        {
+            return;
+        }
+
+        // Refresh flatValueGrid from the currently-simulated state so we can compare fairly before deciding to overwrite it.
+        baseFloodSim.OnProtoSerialize(null);
+        if (Sum(incoming) <= Sum(baseFloodSim.flatValueGrid))
+        {
+            // The local simulation is already at least as flooded as this (possibly network-delayed) snapshot. Skip applying it,
+            // otherwise a client that's already actively simulating this base would keep getting its progress reset backwards
+            // by other players' periodic updates. This still lets a freshly (re)joined client - whose own simulation starts
+            // from empty - catch up to the real, more-flooded state.
+            return;
+        }
+
+        baseFloodSim.flatValueGrid = incoming;
+        baseFloodSim.OnProtoDeserialize(null);
+    }
+
+    private static float Sum(float[] values)
+    {
+        float total = 0f;
+        foreach (float value in values)
+        {
+            total += value;
+        }
+        return total;
+    }
+}
--- a/NitroxPatcher/Patches/Dynamic/BaseFloodSim_Update_Patch.cs
+++ b/NitroxPatcher/Patches/Dynamic/BaseFloodSim_Update_Patch.cs
@@ -1,0 +1,24 @@
+using System.Reflection;
+using NitroxClient.GameLogic;
+using Nitrox.Model.DataStructures;
+
+namespace NitroxPatcher.Patches.Dynamic;
+
+/// <summary>
+/// Periodically (throttled) broadcasts a base's current flood/water-level state, so that it can be restored for
+/// players who (re)join later instead of them seeing the base's interior flooding restart from empty.
+/// </summary>
+public sealed partial class BaseFloodSim_Update_Patch : NitroxPatch, IDynamicPatch
+{
+    public static readonly MethodInfo TARGET_METHOD = Reflect.Method((BaseFloodSim t) => t.Update());
+
+    public static void Postfix(BaseFloodSim __instance)
+    {
+        if (!__instance.IsInitialized() || !__instance.baseComp || !__instance.baseComp.TryGetIdOrWarn(out NitroxId id))
+        {
+            return;
+        }
+
+        Resolve<Entities>().EntityMetadataChangedThrottled(__instance, id, 2f);
+    }
+}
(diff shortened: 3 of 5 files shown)


NitroxClient/GameLogic/Spawning/Metadata/Processor/BaseFloodMetadataProcessor.cs
```diff
@@ -0,0 +1,52 @@
...
+        // in which case the array no longer maps to the same cells. Applying it regardless would throw inside BaseFloodSim's
+        // own restore logic (index mismatch) and permanently break its Update() loop (no more leaks/flooding) from then on.
+        if (incoming == null || incoming.Length != baseFloodSim.shape.Size)
+        {
+            return;
+        }
+
+        // Refresh flatValueGrid from the currently-simulated state so we can compare fairly before deciding to overwrite it.
+        baseFloodSim.OnProtoSerialize(null);
+        if (Sum(incoming) <= Sum(baseFloodSim.flatValueGrid))
```
SpaceMonkeyy86, line 30: This logic seems incorrect since we shouldn't really care about whether higher/lower water levels are more correct, just about making sure everyone has the same water level.

NitroxClient/GameLogic/Spawning/Metadata/Processor/BaseFloodMetadataProcessor.cs
```diff
@@ -0,0 +1,52 @@
...
+            // by other players' periodic updates. This still lets a freshly (re)joined client - whose own simulation starts
+            // from empty - catch up to the real, more-flooded state.
+            return;
+        }
+
+        baseFloodSim.flatValueGrid = incoming;
+        baseFloodSim.OnProtoDeserialize(null);
+    }
+
+    private static float Sum(float[] values)
```
SpaceMonkeyy86, line 43: LINQ already has a `Sum()` method, so you should just be able to import `System.Linq` and do `array.Sum()`.

NitroxPatcher/Patches/Dynamic/BaseFloodSim_Update_Patch.cs
```diff
@@ -0,0 +1,24 @@
...
+    public static readonly MethodInfo TARGET_METHOD = Reflect.Method((BaseFloodSim t) => t.Update());
+
+    public static void Postfix(BaseFloodSim __instance)
+    {
+        if (!__instance.IsInitialized() || !__instance.baseComp || !__instance.baseComp.TryGetIdOrWarn(out NitroxId id))
+        {
+            return;
+        }
+
+        Resolve<Entities>().EntityMetadataChangedThrottled(__instance, id, 2f);
```
SpaceMonkeyy86, line 22: Unrelated to your code, but I think there might be a bug in this function that recomputes the entity metadata every time, even if it isn't needed. With your metadata extractor and running every frame, this might cause some game lag; you may want to verify this isn't the case

NitroxPatcher/Patches/Dynamic/BaseFloodSim_Update_Patch.cs
```diff
@@ -0,0 +1,24 @@
...
+/// Periodically (throttled) broadcasts a base's current flood/water-level state, so that it can be restored for
+/// players who (re)join later instead of them seeing the base's interior flooding restart from empty.
+/// </summary>
+public sealed partial class BaseFloodSim_Update_Patch : NitroxPatch, IDynamicPatch
+{
+    public static readonly MethodInfo TARGET_METHOD = Reflect.Method((BaseFloodSim t) => t.Update());
+
+    public static void Postfix(BaseFloodSim __instance)
+    {
+        if (!__instance.IsInitialized() || !__instance.baseComp || !__instance.baseComp.TryGetIdOrWarn(out NitroxId id))
```
SpaceMonkeyy86, line 17: We should also check for a lock on the base simulation here

NitroxClient/GameLogic/Spawning/Metadata/Extractor/BaseFloodMetadataExtractor.cs
```diff
@@ -0,0 +1,14 @@
+using Nitrox.Model.Subnautica.DataStructures.GameLogic.Entities.Metadata;
+using NitroxClient.GameLogic.Spawning.Metadata.Extractor.Abstract;
+
+namespace NitroxClient.GameLogic.Spawning.Metadata.Extractor;
+
+public class BaseFloodMetadataExtractor : EntityMetadataExtractor<BaseFloodSim, BaseFloodMetadata>
```
SpaceMonkeyy86, line 6: The patch to BaseFloodSim sends a metadata update for the object with the Base component, but this type definition assumes the entity is the object with the BaseFloodSim component. Even if those are actually the same object, it would be worth clearing up this inconsistency in the code. You have two options:
1. Use the existing entity with the Base component. In that case I would rename `BaseFloodMetadata` to `BaseMetadata` (making sure that entity didn't already have any metadata on it), and you would need to change this type signature to be `EntityMetadataExtractor<Base, BaseMetadata>` and find the BaseFloodSim component from the Base component in the metadata extractor.
2. Create a new child entity based on the BaseFloodSim component (assuming BaseFloodSim is on a different object than Base). If this entity doesn't already get tagged on its own, you would need to add code for that. Then change the patch to use `__instance.gameObject` instead of `__instance.baseComp.gameObject`.

SpaceMonkeyy86 (commented):
It seems the main use case for this sync is on reconnect. If so, you might want to consider just moving this info into InitialSync and just letting each client handle the water level individually (assuming that was working properly already).

SmokerDoker:
Thanks for the review — reworked the whole thing along the lines you suggested. Summary of what changed in the follow-up commit:

**Moved to a (re)join-only seed instead of live sync.** The server now stores `BaseFloodMetadata` but no longer relays it to already-connected clients — `EntityMetadataUpdateProcessor` early-returns for that type after saving it. Connected clients each run `BaseFloodSim` locally; its inputs are the base's leaks / hull-cell health, which we already sync, so two clients that are both actually simulating a base stay consistent. The only real gap was the (re)join case (base fully flooded for everyone else, starts from empty for the joiner), and that's what the stored snapshot now seeds via the normal initial-sync path and `BuildingResyncProcessor`.

**Dropped the `Sum()` comparison and the helper.** You were right that it's the wrong question. Since the processor now only runs on the spawn / building-resync paths (local sim freshly built = empty), it just adopts the server's grid unconditionally. The only remaining guard is the grid-size check against a geometry change, which would otherwise throw inside `RestoreSerializedWaterLevels` and permanently break `Update()`.

**LINQ `Sum()`** — moot now, the helper is gone.

**Per-frame metadata recompute.** Fixed in `BaseFloodSim_Update_Patch`:
- Gate on `tIsLeaking()` first — allocation-free, so a dry base (the common case) costs nothing.
- Throttle the whole-grid serialization to once per 2s per base (`ConditionalWeakTable` keyed on the sim, weak keys) instead of relying on the send-side throttle while still extracting every frame.

**Lock / resync check.** The patch now bails on `!InitialSyncCompleted`, `PacketSuppressor<EntityMetadataUpdate>.IsSuppressed`, and `BuildingHandler.Main.Resyncing` — during a rebuild the sim is torn down and anything read there is transient.

**Entity type inconsistency.** Went with option 1: `BaseFloodMetadataExtractor` is now `EntityMetadataExtractor<Base, BaseFloodMetadata>` and pulls `BaseFloodSim` off the `Base`; the patch broadcasts on `baseComp`. Kept the name `BaseFloodMetadata` since it's specifically the flood grid — happy to rename to `BaseMetadata` if you'd rather have the generic slot.

**Known trade-off (documented in the patch's XML doc):** a client that had a base dormant (far away / low LOD) and then teleports or fast-travels to it can briefly show a stale water level until its local sim catches up. Flood events are fast and normally repaired immediately (the base is unusable otherwise), so this window is small and self-correcting — chose to accept it rather than add a relay or broadcaster election.

PR #2814, opened 2026-07-04
Fix typo in failure message

vvitovec:
## Summary
- fix a typo in the English failure localization string

## Checks
- jq empty Nitrox.Assets.Subnautica/LanguageFiles/en.json
- git diff --check

Full .NET tests were not run because dotnet is not installed on this machine.


Measurity:
Added this to en language on Weblate. Thanks for reporting.

Localization files cannot be edited locally or weblate will go out of sync.

PR #2818, opened 2026-07-05
Fix player-join color-swap hitch (#2168)

### LINKED ISSUE #2168: Freezes when someone is joining the game
### Describe the issue

When a player joins a server, other people already connected will experience a lag peak  shortly before the message "{PLAYER} joined ..."

The issue should originate in this file : [PlayerJoinedMultiplayerSessionProcessor](https://github.com/SubnauticaNitrox/Nitrox/blob/07cdb19aa336288c54ded766c96570e60e369d8d/NitroxClient/Communication/Packets/Processors/PlayerJoinedMultiplayerSessionProcessor.cs)

SmokerDoker:
## Summary

Fixes #2168. 

Every already-connected client experienced a lag spike whenever a
new player joined, caused by `PlayerModelManager` setting up the joining
player's clothing/equipment colors: ~18 `Graphics.Blit` + `ReadPixels` GPU
readback calls ran synchronously on the main thread, once per `RemotePlayer`
spawn, because `Renderer.material` auto-instantiates a fresh material per
player and the color-swap task creation ran fully synchronously before any
work was backgrounded to the `ThreadPool`.

- **Cache the base pixel readback per source texture**
  (`RendererExtensions.Clone()`): every remote player model clones the same
  prototype prefab, so the source `Texture2D` for a given body part is the
  same asset for every player. The expensive readback is now cached and only
  runs once per texture for the whole session instead of once per join — no
  behavioral/visual change, same pixel data, just fetched once.
- **Spread the one remaining first-time cost across frames**
  (`ColorSwapAsyncOperation.BeginColorSwapOverFrames()`): used only by the
  remote-player-join path (`PlayerModelManager.ApplyPlayerColor`), creates one
  `IColorSwapManager` task per yielded frame instead of all ~10 in one frame,
  so even the very first join in a session doesn't stall a single frame. The
  existing synchronous `BeginColorSwap()` used for local-player init (during
  the loading screen in `Multiplayer.cs`) is untouched, since a stall there
  isn't user-facing.

A third, larger option (replacing `ReadPixels` with `AsyncGPUReadback` to
remove the GPU stall entirely, including the very first occurrence) was
investigated and confirmed technically feasible on the shipped Subnautica
Unity 2019.4.36f1 assemblies, but was intentionally not implemented here — see
discussion on #2168. Left as a documented follow-up if this turns out to not
be enough in practice.

## Test plan

- [x] `dotnet build NitroxClient/NitroxClient.csproj` — builds with 0 errors
- [ ] Manual test: join a dedicated/public server with a second client and
      confirm no perceptible hitch on join for already-connected clients
- [ ] Manual test: start a fresh 2-person co-op session and confirm the first
      join doesn't produce a single noticeable frame hitch

### DIFF
--- a/NitroxClient/GameLogic/PlayerLogic/PlayerModel/ColorSwap/ColorSwapAsyncOperation.cs
+++ b/NitroxClient/GameLogic/PlayerLogic/PlayerModel/ColorSwap/ColorSwapAsyncOperation.cs
@@ -46,7 +46,8 @@ namespace NitroxClient.GameLogic.PlayerLogic.PlayerModel.ColorSwap
         {
             if (taskCount >= 0)
             {
-                throw new InvalidOperationException("This operation has already been started.");
+                Log.Error("This operation has already been started.");
+                return this;
             }
 
             List<Action<ColorSwapAsyncOperation>> tasks = colorSwapManagers
@@ -70,7 +71,8 @@ namespace NitroxClient.GameLogic.PlayerLogic.PlayerModel.ColorSwap
         {
             if (taskCount >= 0)
             {
-                throw new InvalidOperationException("This operation has already been started.");
+                Log.Error("This operation has already been started.");
+                yield break;
             }
 
             taskCount = 0;
--- a/NitroxClient/GameLogic/PlayerLogic/PlayerModel/ColorSwap/ColorSwapAsyncOperation.cs
+++ b/NitroxClient/GameLogic/PlayerLogic/PlayerModel/ColorSwap/ColorSwapAsyncOperation.cs
@@ -64,8 +64,8 @@ namespace NitroxClient.GameLogic.PlayerLogic.PlayerModel.ColorSwap
         /// <summary>
         /// Same as <see cref="BeginColorSwap"/>, but creates one manager's task per yielded frame instead of all of
         /// them in one frame. Each <see cref="IColorSwapManager.CreateColorSwapTask"/> call can involve an expensive,
-        /// main-thread-only GPU texture readback (see <see cref="Extensions.RendererExtensions.Clone"/>), so spreading
-        /// the creation over several frames turns a single noticeable hitch into several smaller, imperceptible ones.
+        /// main-thread-only GPU texture readback (see <see cref="Extensions.RendererExtensions.GetSourcePixels(Texture2D)"/>),
+        /// so spreading the creation over several frames turns a single noticeable hitch into several smaller, imperceptible ones.
         /// </summary>
         public IEnumerator BeginColorSwapOverFrames()
         {
@@ -74,6 +74,9 @@ namespace NitroxClient.GameLogic.PlayerLogic.PlayerModel.ColorSwap
                 Log.Error("This operation has already been started.");
                 yield break;
             }
+
+            IEnumerable<Texture2D> sourceTextures = colorSwapManagers.SelectMany(manager => manager.GetSourceTextures(nitroxPlayer));
+            yield return sourceTextures.PrewarmSourcePixelsAsync();
 
             taskCount = 0;
             foreach (IColorSwapManager manager in colorSwapManagers)
@@ -95,6 +98,25 @@ namespace NitroxClient.GameLogic.PlayerLogic.PlayerModel.ColorSwap
             }
 
             colorSwapManagers.ForEach(manager => manager.ApplyPlayerColor(texturePixelIndexes, nitroxPlayer));
+        }
+
+        /// <summary>
+        /// Same as <see cref="ApplySwappedColors"/>, but applies one manager's final textures per yielded frame
+        /// instead of all managers (up to ~18 texture uploads) in a single frame.
+        /// </summary>
+        public IEnumerator ApplySwappedColorsOverFrames()
+        {
+            if (taskCount != 0)
+            {
+                Log.Error("Colors must be swapped before the changes can be applied to the player model.");
+                yield break;
+            }
+
+            foreach (IColorSwapManager manager in colorSwapManagers)
+            {
+                manager.ApplyPlayerColor(texturePixelIndexes, nitroxPlayer);
+                yield return null;
+            }
         }
 
         private void ExecuteTask(object state)
--- a/NitroxClient/GameLogic/PlayerLogic/PlayerModel/ColorSwap/ColorSwapAsyncOperation.cs
+++ b/NitroxClient/GameLogic/PlayerLogic/PlayerModel/ColorSwap/ColorSwapAsyncOperation.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
@@ -13,6 +14,7 @@ namespace NitroxClient.GameLogic.PlayerLogic.PlayerModel.ColorSwap
         private readonly IEnumerable<IColorSwapManager> colorSwapManagers;
         private readonly Dictionary<string, Color[]> texturePixelIndexes;
         private int taskCount = -1;
+        private volatile bool tasksCreated;
 
         public ColorSwapAsyncOperation(INitroxPlayer nitroxPlayer, IEnumerable<IColorSwapManager> colorSwapManagers)
         {
@@ -37,7 +39,7 @@ namespace NitroxClient.GameLogic.PlayerLogic.PlayerModel.ColorSwap
 
         public bool IsColorSwapComplete()
         {
-            return taskCount == 0;
+            return tasksCreated && taskCount == 0;
         }
 
         public ColorSwapAsyncOperation BeginColorSwap()
@@ -52,9 +54,35 @@ namespace NitroxClient.GameLogic.PlayerLogic.PlayerModel.ColorSwap
                 .ToList();
 
             taskCount = tasks.Count;
+            tasksCreated = true;
             tasks.ForEach(task => ThreadPool.QueueUserWorkItem(ExecuteTask, task));
 
             return this;
+        }
+
+        /// <summary>
+        /// Same as <see cref="BeginColorSwap"/>, but creates one manager's task per yielded frame instead of all of
+        /// them in one frame. Each <see cref="IColorSwapManager.CreateColorSwapTask"/> call can involve an expensive,
+        /// main-thread-only GPU texture readback (see <see cref="Extensions.RendererExtensions.Clone"/>), so spreading
+        /// the creation over several frames turns a single noticeable hitch into several smaller, imperceptible ones.
+        /// </summary>
+        public IEnumerator BeginColorSwapOverFrames()
+        {
+            if (taskCount >= 0)
+            {
+                throw new InvalidOperationException("This operation has already been started.");
+            }
+
+            taskCount = 0;
+            foreach (IColorSwapManager manager in colorSwapManagers)
+            {
+                Action<ColorSwapAsyncOperation> task = manager.CreateColorSwapTask(nitroxPlayer);
+                Interlocked.Increment(ref taskCount);
+                ThreadPool.QueueUserWorkItem(ExecuteTask, task);
+                yield return null;
+            }
+
+            tasksCreated = true;
         }
 
         public void ApplySwappedColors()
--- a/NitroxClient/GameLogic/PlayerLogic/PlayerModel/PlayerModelManager.cs
+++ b/NitroxClient/GameLogic/PlayerLogic/PlayerModel/PlayerModelManager.cs
@@ -97,6 +97,6 @@ public class PlayerModelManager
 
         yield return swapOperation.BeginColorSwapOverFrames();
         yield return new WaitUntil(() => swapOperation.IsColorSwapComplete());
-        swapOperation.ApplySwappedColors();
+        yield return swapOperation.ApplySwappedColorsOverFrames();
     }
 }
--- a/NitroxClient/GameLogic/PlayerLogic/PlayerModel/PlayerModelManager.cs
+++ b/NitroxClient/GameLogic/PlayerLogic/PlayerModel/PlayerModelManager.cs
@@ -95,8 +95,6 @@ public class PlayerModelManager
     {
         ColorSwapAsyncOperation swapOperation = new(player, colorSwapManagers);
 
-        // Spread the (potentially expensive, first-time-per-texture) task creation across frames so that setting up
-        // a newly joined remote player's clothing colors doesn't stall already-connected clients for a single frame.
         yield return swapOperation.BeginColorSwapOverFrames();
         yield return new WaitUntil(() => swapOperation.IsColorSwapComplete());
         swapOperation.ApplySwappedColors();
--- a/NitroxClient/GameLogic/PlayerLogic/PlayerModel/PlayerModelManager.cs
+++ b/NitroxClient/GameLogic/PlayerLogic/PlayerModel/PlayerModelManager.cs
@@ -95,7 +95,9 @@ public class PlayerModelManager
     {
         ColorSwapAsyncOperation swapOperation = new(player, colorSwapManagers);
 
-        swapOperation.BeginColorSwap();
+        // Spread the (potentially expensive, first-time-per-texture) task creation across frames so that setting up
+        // a newly joined remote player's clothing colors doesn't stall already-connected clients for a single frame.
+        yield return swapOperation.BeginColorSwapOverFrames();
         yield return new WaitUntil(() => swapOperation.IsColorSwapComplete());
         swapOperation.ApplySwappedColors();
     }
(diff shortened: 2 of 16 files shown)


NitroxClient/GameLogic/PlayerLogic/PlayerModel/ColorSwap/ColorSwapAsyncOperation.cs
```diff
@@ -52,11 +54,37 @@ public ColorSwapAsyncOperation BeginColorSwap()
...
+        /// them in one frame. Each <see cref="IColorSwapManager.CreateColorSwapTask"/> call can involve an expensive,
+        /// main-thread-only GPU texture readback (see <see cref="Extensions.RendererExtensions.Clone"/>), so spreading
+        /// the creation over several frames turns a single noticeable hitch into several smaller, imperceptible ones.
+        /// </summary>
+        public IEnumerator BeginColorSwapOverFrames()
+        {
+            if (taskCount >= 0)
+            {
+                throw new InvalidOperationException("This operation has already been started.");
+            }
```
Measurity, line 76: I see other code here also did a throw but would go with a log & return here. And change old one too.

Game client is susceptible to invalid state if errors are thrown due to not handling errors gracefully with try...catch.

NitroxClient/GameLogic/PlayerLogic/PlayerModel/PlayerModelManager.cs
```diff
@@ -95,7 +95,9 @@ private static IEnumerator ApplyPlayerColor(INitroxPlayer player, IEnumerable<IC
     {
         ColorSwapAsyncOperation swapOperation = new(player, colorSwapManagers);
 
-        swapOperation.BeginColorSwap();
+        // Spread the (potentially expensive, first-time-per-texture) task creation across frames so that setting up
+        // a newly joined remote player's clothing colors doesn't stall already-connected clients for a single frame.
```
Measurity: These comments are quite wordy and not sure if it's true?

"Prevents stalling already connected clients for a single frame"

Testing on my side it still stutters when clients join while already playing.

I'm fine with just removing these comments.

Measurity (changes requested):

NitroxClient/GameLogic/PlayerLogic/PlayerModel/ColorSwap/ColorSwapAsyncOperation.cs
SmokerDoker: Good catch, fixed in the latest commit.

NitroxClient/GameLogic/PlayerLogic/PlayerModel/PlayerModelManager.cs
SmokerDoker: I removed these comments, but i will look into it myself. The caching should resolve this issue but if that doesnt work properly then maybe we should really think about option 3 - AsyncGPUReadback.

SmokerDoker:
@Measurity AsyncGPUReadback (option 3) could be added on top of the current fix, since it complements rather than replaces the existing approach. Given that it may fail on older GPUs, I’d keep the current path as a fallback. If you’re okay with that, I can implement it and commit it to this branch, as that would resolve the issue more properly

Measurity:
@SmokerDoker

I would be happy with that if you have the time.

SmokerDoker:
@Measurity 

# Fix player-join color-swap hitch (#2168 / PR #2818) — Debugging Summary

Branch: `fix/player-join-color-swap-hitch`
Latest commit: `997bd37e` — "Eliminate double texture upload and add async readback for color swap"

## PR context

PR #2818 fixes #2168: every already-connected client experienced a lag spike whenever a new player
joined, caused by `PlayerModelManager` setting up the joining player's clothing/equipment colors.
Root cause per the original PR description: ~18 `Graphics.Blit` + `ReadPixels` GPU readback calls ran
synchronously on the main thread, once per `RemotePlayer` spawn, because `Renderer.material`
auto-instantiates a fresh material per player and the color-swap task creation ran fully synchronously
before any work was backgrounded to the ThreadPool.

Original fix (commit `5979700`, already in the PR before this session):
- **Cache the base pixel readback per source texture** (`RendererExtensions.Clone()`): every remote
  player model clones the same prototype prefab, so the source `Texture2D` for a given body part is the
  same asset for every player. The readback now only runs once per texture for the whole session instead
  of once per join.
- **Spread the one remaining first-time cost across frames** (`ColorSwapAsyncOperation.BeginColorSwapOverFrames()`):
  creates one `IColorSwapManager` task per yielded frame instead of all ~10 in one frame, so even the very
  first join in a session doesn't stall a single frame.
- A third, larger option — replacing `ReadPixels` with `AsyncGPUReadback` to remove the GPU stall entirely,
  including the very first occurrence — was investigated and confirmed technically feasible on the shipped
  Subnautica Unity 2019.4.36f1 assemblies, but intentionally **not** implemented in the first commit, left
  as a documented follow-up.

## Review round 1 (Measurity)

1. `if (taskCount >= 0) throw new InvalidOperationException(...)` in `BeginColorSwapOverFrames()` (and the
   pre-existing `BeginColorSwap()`) should log and return cleanly instead of throwing — an invalid state
   shouldn't be able to crash the client.
2. The comments added to `PlayerModelManager.cs` claiming the fix "prevents stalling already connected
   clients for a single frame" were too strong — Measurity still saw stuttering when testing with two
   clients.

**Fix (commit `20c037e1`):**
- Both `throw` guards replaced with `Log.Error(...)` + `return`/`yield break`.
- The overstated comments removed.

## First real-world test: "lag is still there, caching doesn't seem to help"

Test setup: two Subnautica instances (Test Player 1 & 2), local server, client restarts between some
tests. Observation: the lag was still noticeably present on essentially every join, whether it was the
first join in a session or a repeat join of the same player.

### Root-cause analysis (code reading, no logs yet)

Going through every `IColorSwapManager` implementation (`DiveSuitColorSwapManager`,
`ScubaTankColorSwapManager`, etc.) revealed that each manager did **two full texture uploads**
(`SetPixels()` + `Apply()`) per body-part texture, not one:

1. **Upload #1** in `CreateColorSwapTask()` via `renderer.material.ApplyClonedTexture()` — created a new,
   readable clone with the **unswapped** pixels and immediately uploaded it to the GPU, purely so
   `GetMainTexturePixels()` could be called afterwards.
2. **Upload #2** in `ApplyPlayerColor()` via `UpdateMainTextureColors()` — wrote the **swapped** pixels
   again via `SetPixels()` + `Apply()` onto the same texture.

Upload #1 was completely wasted — its result was immediately overwritten by upload #2. The existing cache
(`sourceTexturePixelCache`) only saved the **readback** portion (Blit + ReadPixels), which was only ever
needed once per texture in the first place. The two `SetPixels()+Apply()` calls (CPU→GPU upload, including
mipmap regeneration) still ran unchanged on **every single join** — that cost was never part of what got
cached.

### Fix round 2 (part of final commit `997bd37e`)

**`RendererExtensions.cs`:**
- `Clone()` (texture-based, mutated the material) replaced with `GetSourcePixels(this Texture2D)` — returns
  only a `Color[]` (a private copy, from cache or freshly read), **without** touching the material. The
  scratch texture used for the readback is no longer `Apply()`'d (saves an unnecessary upload/mipmap
  rebuild even on a cache miss) and is destroyed right after.
- `ApplyClonedTexture()` and `GetMainTexturePixels()` removed — no longer needed.
- `UpdateMainTextureColors()` now builds the final texture directly from the already-swapped pixels and
  does a **single** `SetPixels()+Apply()` call.
- New: `ExtractBlock()` / `InsertBlock()` — plain array arithmetic for sub-regions (the
  `RadiationSuitColorSwapManager` leg/feet/belt-block case), without needing a texture for it.

**All 10 `ColorSwapManager` implementations:** `ApplyClonedTexture(); ... GetMainTexturePixels();` →
`GetSourcePixels();`. No intermediate upload anymore.

**`RadiationSuitColorSwapManager`** (special block-based case): instead of patching 3 blocks one-by-one
onto a pre-cloned texture, it now fetches the full pixel array once, patches all 3 blocks into it in
memory, then does a single final `UpdateMainTextureColors()` call.

**Result:** one GPU upload per texture instead of two.

## Second test round: added diagnostic logging

To verify the effect, temporary `Log.Info("[ColorSwapPerf] ...")` lines were added:
- Cache hit/miss plus Blit+ReadPixels duration in `GetSourcePixels(Texture2D)`.
- `SetPixels()+Apply()` duration in `UpdateMainTextureColors()`.
- Total duration of `PlayerModelManager.ApplyPlayerColor()`.

Methodological note that came out of the test setup: the server console and the client log are **two
separate processes** — client-side logging (our `[ColorSwapPerf]` lines) never shows up in the server log.
Similarly, Visual Studio's debug console usually only shows the launcher process, not the separately
spawned Subnautica client processes. The reliable source was the log file under `%AppData%\Nitrox\logs\`,
one file per player (`game[SaveName][PlayerName]-Date.log`).

### Findings from the logs (test 1, after fix round 2)

- Caching worked correctly: each client's own local player (synchronous `BeginColorSwap()` path during
  loading) warmed the cache for all standard equipment textures (`player_02_*`), so **every** later remote
  spawn (regardless of which player) showed exclusively `CACHE HIT`.
- BUT: `TOTAL ApplyPlayerColor for 'Test Player 1' took 5915,25ms` — despite **100% cache hits**! Not a
  single Blit+ReadPixels stall, yet almost 6 seconds total.
- Cause: `PlayerModelManager.ApplyPlayerColor()` called `swapOperation.ApplySwappedColors()` right after
  the `WaitUntil` — that iterates **synchronously over all ~10 managers (~18 textures) in a single frame**,
  with no `yield return` in between at all. A frame with ~581ms of `SetPixels()+Apply()` work is a genuine
  hard hitch — this explained the "hard, long stutter" the user was still seeing.
- The bulk of the remaining 5-6 second total came from **unrelated** join synchronization (inventory/entity
  sync packets: "Received: Added item ... to container ..."), which happened to overlap in time with the
  color-swap coroutine and loaded the main thread concurrently — not part of the color-swap code itself.

## Fix round 3 (also part of commit `997bd37e`)

1. **Spread `ApplySwappedColors()` across frames:** new method
   `ColorSwapAsyncOperation.ApplySwappedColorsOverFrames()` applies one manager per frame instead of all at
   once. `PlayerModelManager.ApplyPlayerColor()` now uses it for the remote-join path. The synchronous
   `ApplySwappedColors()` is left untouched for the local player (loading screen, not user-facing).
2. **AsyncGPUReadback pre-warm** — this is the "option 3" flagged as a follow-up in the original PR
   description, now implemented at Measurity's request (see PR conversation: "I would be happy with that if
   you have the time"), even though by this point the logs showed the readback itself was rarely the
   bottleneck anymore. New method `IColorSwapManager.GetSourceTextures()` — every manager reports which
   textures it needs without reading them. `RendererExtensions.PrewarmSourcePixelsAsync()` collects every
   not-yet-cached texture across all managers, kicks off an `AsyncGPUReadback` request for each (in
   parallel, non-blocking), and waits frame-by-frame instead of blocking the main thread.
   `BeginColorSwapOverFrames()` calls this before the actual task-creation loop. The existing synchronous
   Blit+ReadPixels path in `GetSourcePixels()` is kept as-is and used as an explicit fallback if a given
   `AsyncGPUReadback` request reports an error (per Measurity's ask to keep the current path as a fallback
   for GPUs where async readback might fail).
   - Risk to flag: Direct3D/Metal/consoles can return GPU readback rows in a different order (top-to-bottom)
     than Unity's `GetPixels()/SetPixels()` convention (bottom-to-top). Guarded via
     `SystemInfo.graphicsUVStartsAtTop` with a row flip when needed — visually verified by the user
     (both test players' colors correct, no flip artifacts).
   - Added a `Unity.Collections` assembly reference to `NitroxClient.csproj` (needed for
     `NativeArray<Color32>`, returned by `AsyncGPUReadbackRequest.GetData<T>()`).

### Verification round 3

Re-tested with the same `[ColorSwapPerf]` logs: apply lines now spread over ~1 second across multiple
frames instead of one single 581ms block. The largest single-frame chunk dropped to ~200-215ms (Reinforced
/ Still Suit, two 2048x2048 textures per manager call). No `AsyncGPUReadback failed` messages. Colors for
both test players confirmed correct (green/green). The async readback path itself was never exercised
under a real cache-miss scenario during a remote join in these tests, since each client's own local player
always warms the cache first — but it ran without errors throughout.

## Separate bug found: `FootstepPacketProcessor` NullReferenceException

During the second test run (Test Player 1 was actively walking around inside the base, right where Test
Player 2 was joining), a wave of ~20 repeated exceptions showed up in Test Player 2's log:

```
[ERR] Error trying to process packet [FootstepPacket: SessionId: 1, AssetIndex: METAL]
System.NullReferenceException: Object reference not set to an instance of an object
  at FMODUnity.RuntimeUtils.To3DAttributes (UnityEngine.GameObject go)
  at NitroxClient.Communication.Packets.Processors.FootstepPacketProcessor.Process(...)
```

### Root cause

`FootstepPacketProcessor.Process()`
(`NitroxClient/Communication/Packets/Processors/FootstepPacketProcessor.cs`) called
`player.Value.Body.To3DAttributes()` without checking whether `Body` had actually been assigned yet.

`RemotePlayer.Body` is `GameObject?` and is only assigned inside `RemotePlayer.InitializeGameObject()`.
`PlayerManager.Create()` registers the `RemotePlayer` in `sessionsById` **before** that happens — so there
is a window where `remotePlayerManager.Find(sessionId)` returns a player with `HasValue == true` whose
`Body` is still `null`. If a `FootstepPacket` for that exact session arrives during that window (because
the other player is actively walking), `To3DAttributes()` crashes on `null`.

This race most likely already existed in the code independently of the color-swap fix, but became more
visible due to the now longer join-processing window (more frames from the frame-spreading changes above)
widening the time during which `Body` can still be null. The repeated exceptions themselves also cost extra
CPU time and likely explain some of the outlier apply times seen in that particular test run.

### Fix

Guard added in `FootstepPacketProcessor.cs`, matching the pattern already used elsewhere in the codebase
(e.g. `PlayerHeldItemChangedProcessor`: skip cleanly instead of crashing):

```diff
- if (player.HasValue)
+ if (player.HasValue && player.Value.Body)
```

No dedicated GitHub issue for this bug — found while testing this PR and fixed directly alongside it since
it's small and related (triggered/surfaced by the same join process).

## Cleanup before committing

All temporary `[ColorSwapPerf]` diagnostic logs (`Log.Info` + `Stopwatch` measurements) were removed again
so no permanent per-join log overhead remains in production code. The `Log.Error` call for the actual
`AsyncGPUReadback` error case (with fallback to the synchronous readback) was kept, since that's a genuine
error condition, not just diagnostics.

## Files changed (commit `997bd37e`)

- `NitroxClient/Extensions/RendererExtensions.cs`
- `NitroxClient/GameLogic/PlayerLogic/PlayerModel/Abstract/IColorSwapManager.cs`
- `NitroxClient/GameLogic/PlayerLogic/PlayerModel/ColorSwap/ColorSwapAsyncOperation.cs`
- `NitroxClient/GameLogic/PlayerLogic/PlayerModel/ColorSwap/DiveSuitColorSwapManager.cs`
- `NitroxClient/GameLogic/PlayerLogic/PlayerModel/ColorSwap/FinColorSwapManager.cs`
- `NitroxClient/GameLogic/PlayerLogic/PlayerModel/ColorSwap/RadiationHelmetColorSwapManager.cs`
- `NitroxClient/GameLogic/PlayerLogic/PlayerModel/ColorSwap/RadiationSuitColorSwapManager.cs`
- `NitroxClient/GameLogic/PlayerLogic/PlayerModel/ColorSwap/RadiationSuitVestColorSwapManager.cs`
- `NitroxClient/GameLogic/PlayerLogic/PlayerModel/ColorSwap/RadiationTankColorSwapManager.cs`
- `NitroxClient/GameLogic/PlayerLogic/PlayerModel/ColorSwap/RebreatherColorSwapManager.cs`
- `NitroxClient/GameLogic/PlayerLogic/PlayerModel/ColorSwap/ReinforcedSuitColorSwapManager.cs`
- `NitroxClient/GameLogic/PlayerLogic/PlayerModel/ColorSwap/ScubaTankColorSwapManager.cs`
- `NitroxClient/GameLogic/PlayerLogic/PlayerModel/ColorSwap/StillSuitColorSwapManager.cs`
- `NitroxClient/GameLogic/PlayerLogic/PlayerModel/PlayerModelManager.cs`
- `NitroxClient/Communication/Packets/Processors/FootstepPacketProcessor.cs`
- `NitroxClient/NitroxClient.csproj`

## Open items / possible follow-ups

- The `AsyncGPUReadback` path was never exercised under a genuine cache-miss scenario during a remote join
  (only the synchronous fallback path, via the local player's own login, actually went through
  Blit+ReadPixels in testing). Worth verifying with test players wearing different equipment so their
  textures actually differ.
- Reinforced Suit and Still Suit (2048x2048, two textures per manager) remain the most expensive single
  items at apply time (~150-215ms per frame slot). Further tuning (e.g. spreading per-texture instead of
  per-manager) is possible if this still shows up as a noticeable stutter in practice.

PR #2819, opened 2026-07-05
Fixed log file location while logging in wine

Measurity:
Passed `NITROX_HOST_HOME` and `--nitrox-host-home <path>` to game process so that logs can be written to home of OS user instead of wine home user.

This should be tested on Windows if game still loads as normal and not throws an error in player.log.

### DIFF
--- a/Nitrox.Model/Core/NitroxEnvironment.cs
+++ b/Nitrox.Model/Core/NitroxEnvironment.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
 using System.Runtime.InteropServices;
+using Nitrox.Model.Platforms.OS.Windows.Internal;
 
 namespace Nitrox.Model.Core;
 
@@ -13,9 +14,8 @@ public static class NitroxEnvironment
 {
     private static bool hasSet;
     private static Assembly? executingAssembly;
-
     private static string? appName;
-
+    private static bool? isWine;
     private static Assembly ExecutingAssembly => executingAssembly ??= Assembly.GetExecutingAssembly();
     public static string ReleasePhase => IsReleaseMode ? "Alpha" : "InDev";
     public static Version Version => ExecutingAssembly.GetName().Version ?? new Version(1, 0);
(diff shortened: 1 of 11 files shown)


Nitrox.Model/Platforms/OS/Windows/Internal/Win32Native.cs
```diff
@@ -237,6 +249,10 @@ private enum UIContext
     [DllImport("user32.dll", EntryPoint = "SetWindowLongPtr")]
     internal static extern IntPtr SetWindowLongPtr64(HandleRef hWnd, int nIndex, long dwNewLong);
 
+    [return: MarshalAs(UnmanagedType.LPTStr)]
+    [DllImport("ntdll.dll", EntryPoint = "wine_get_version")]
+    private static extern string WineGetVersion();
```
dartasen, line 258: #if NET
LibraryImport 

? :eyes:

Nitrox.Model/Core/NitroxEnvironment.cs
```diff
@@ -142,6 +142,28 @@ public static string[] CommandLineArgs
...
+        {
+            if (isWine.HasValue)
+            {
+                return isWine.Value;
+            }
+            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                isWine = false;
+                return false;
+            }
```
dartasen, line 160: Always super counterintuitive that this API is working under wine emulation, had to read multiple times xD

Nitrox.Model/Helper/NitroxDirectory.cs
```diff
@@ -190,6 +161,41 @@ protected string GetAsNitroxPath(string path)
...
+                    {
+                        if (char.IsLetter(result[0]) && result[1] == ':')
+                        {
+                            result = $"Z{result[1..]}";
+                        }
+                        else
+                        {
+                            result = $"Z:{result}";
+                        }
+                    }
```
dartasen, line 188: oh lord

Measurity: Joy of reinventing C# for Unix because wine layer :clown_face: ...

Nitrox.Model/Platforms/OS/Windows/Internal/Win32Native.cs
Measurity: ~~I think it's not worth the duplicate code for marginal performance improvement. `NitroxEnvironment.IsWine` caches the result already.~~ I submit.

Would be nice to have a tool (source generator) to generate the old .NET and modern .NET pinvoke syntax though.

dartasen (approved):

PR #2820, opened 2026-07-05
Fix VS Code launch config JSON

vvitovec:
## Summary
- remove the trailing comma from `.vscode/launch.json` so it parses as standard JSON
- keep the existing launch configurations unchanged

## Validation
- `for f in .vscode/*.json; do jq empty "$f"; done`
- `git diff --check`

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -27,6 +27,6 @@
             "osx": {
                 "program": "${workspaceFolder}/Nitrox.Server.Subnautica/bin/Debug/net10.0/Nitrox.Server.Subnautica"
             }
-        },
+        }
     ]
 }


Measurity:
Hello @vvitovec thanks for helping out with Nitrox.

Would you be willing to explain the _why_ in your PRs? 
 - Did you get an error before?
 - Is it a nice to have or is it more important to you?
 - etc.

I ask because I noticed many of your PRs are generated by AI and maybe the code change ideas are also from AI. While AI can be useful to get inspiration, it doesn't necessarily understand the needs/wants of the target audience. In addition, any code change can introduce bugs or introduce tech debt (especially with AI which can hallucinate).

If you prefer to use AI as you do then I'm afraid to say we don't think this way is compatible with Nitrox development.

I hope you understand and tell me if I'm off base on something. Thanks.

Coding-Hen:
I agree with the above.

I do agree though also this is technically invalid json at least according to the spec but many parsers usually have compatibility to handle this

vvitovec:
Thanks for the direct feedback. You are not off base; I should have made the motivation explicit and been more selective about what is worth opening as a PR.

For this one specifically: I found it while validating the VS Code JSON with `jq`; it is standard-JSON cleanup, not a user-facing Nitrox bug. Given that it is low impact and VS Code can tolerate it, I agree it is not worth maintainer time, so I am closing it.

Going forward I will only open Nitrox PRs when I can explain the concrete problem, why it matters to the project/users, and how I validated it. For the existing macOS PR (#2790), the motivation is native macOS Steam installs exposing `Subnautica.app` while launcher path validation needs to normalize that to the app `Contents` layout; that PR is intentionally scoped to that path-normalization issue only.

PR #2824, opened 2026-07-05
Fixed server not waiting for actual save data before saving

### LINKED ISSUE #2823: Server save behavior being inconsistent
* Create a fresh server

Image

Notice that Aurora is not exploded

* Stop the server using red button

* Relaunch the server

Image

Aurora is exploded 🥲 

Note that if you use save command before the first quit, it will work as intended

Measurity:
Closes #2823

### DIFF
--- a/Nitrox.Server.Subnautica/Models/Serialization/World/WorldService.cs
+++ b/Nitrox.Server.Subnautica/Models/Serialization/World/WorldService.cs
@@ -13,7 +13,6 @@ using Nitrox.Server.Subnautica.Models.GameLogic.Entities;
 using Nitrox.Server.Subnautica.Models.GameLogic.Entities.Spawning;
 using Nitrox.Server.Subnautica.Models.GameLogic.Players;
 using Nitrox.Server.Subnautica.Models.GameLogic.Unlockables;
-using Nitrox.Server.Subnautica.Models.Helper;
 using Nitrox.Server.Subnautica.Models.Serialization.SaveDataUpgrades;
 using Nitrox.Server.Subnautica.Services;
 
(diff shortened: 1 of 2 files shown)


Nitrox.Server.Subnautica/Models/Serialization/World/WorldService.cs
```diff
@@ -313,38 +320,45 @@ void EnsureChildrenTransformAreParented(WorldEntity entity)
...
+            storyManager.AuroraCountdownTimeMs = pWorldData.WorldData.GameData.StoryTiming.AuroraCountdownTime ?? storyManager.GenerateDeterministicAuroraTime(seed);
+            storyManager.AuroraWarningTimeMs = pWorldData.WorldData.GameData.StoryTiming.AuroraWarningTime ?? timeService.GameTime.TotalMilliseconds;
+            // +27 is from CrashedShipExploder.IsExploded, -480 is from the default time (see TimeService)
+            storyManager.AuroraRealExplosionTime = TimeSpan.FromSeconds(pWorldData.WorldData.GameData.StoryTiming.AuroraRealExplosionTime ?? storyManager.AuroraCountdownTimeMs * 0.001 + 27 - TimeService.DEFAULT_STARTING_GAME_TIME_SECONDS);
+        }
+        finally
         {
-            playerManager.AddSavedPlayer(player);
+            hasFinishedLoadingTcs.SetResult(true);
         }
```
dartasen, line 361: We're silently swallowing any exceptions, is it wanted ?

I guess this piece of code if somehow responsible of the aurora suddenly being exploded (as described in the issue)

Measurity: Changed it to `TrySetResult`.

> I guess this piece of code if somehow responsible of the aurora suddenly being exploded (as described in the issue)

Yes. It's because server now is thread safe so services can be accessed at any time even if no data. Previously I added code to immediately save the server on startup which meant that saves files contained `0` values...

dartasen (approved):
Works for me on Mac

PR #2825, opened 2026-07-06
Fixes #2168 ("Freezes when someone is joining the game").

### LINKED ISSUE #2168: Freezes when someone is joining the game

SmokerDoker:
## Summary

Fixes https://github.com/SubnauticaNitrox/Nitrox/issues/2168

Every already-connected client used to hitch whenever a new remote player's clothing/equipment colors were
set up, because recoloring was fundamentally CPU-bound: read the base texture's pixels back from the GPU,
loop over every pixel in C# doing an HSV-range-based hue swap, then re-upload a freshly baked `Texture2D` to
the GPU — repeated for each of the ~10 equipment pieces, per player.

A previous pass (merged into this branch) tried to fix this by caching the pixel readback per source texture,
spreading the first-time work across frames, and prewarming reads with `AsyncGPUReadback` instead of a
blocking `ReadPixels`. That cut the hitch to roughly 1/3 of the original — better, but the CPU roundtrip
itself (readback → per-pixel loop → re-upload) was still there and still cost time on every join. None of the
caching/threading approaches actually removed the underlying bottleneck, they just hid parts of it.

## What changed

Recoloring is now done entirely on the GPU:

- A new HLSL shader (`NitroxUnity/Assets/colorswap/Recolor.shader`) ports the exact HSV range-check/replace
  logic that used to live in `HsvSwapper`/`HueSwapper`/`HueSaturationVibrancySwapper` (same
  `Color.RGBToHSV`/`HSVToRGB` math, same inclusive range semantics), driven by per-region shader properties
  (up to 4 regions per texture, needed for `RadiationSuitColorSwapManager`'s leg/feet/belt sub-rects).
- `GpuRecolorer.Recolor(...)` does a single `Graphics.Blit(source, renderTexture, recolorMaterial)` per
  texture and returns the result, which gets assigned straight to `material.mainTexture`. No more
  `ReadPixels`/`GetPixels`/`SetPixels`/`Apply()` anywhere in the color-swap path.
- All 10 `*ColorSwapManager` classes, `IColorSwapManager`, and `PlayerModelManager`'s orchestration were
  simplified accordingly — the old two-phase "create task on a background thread, then apply" split only
  existed to hide CPU pixel-loop latency, which no longer exists.
- Removed the now-dead CPU pipeline: `HsvSwapper`, `ColorValueRange`, the `Strategy/` swappers,
  `ColorSwapAsyncOperation`, and the pixel-readback/caching helpers in `RendererExtensions`.
- Added `PlayerColorRenderTextures`, a small `MonoBehaviour` that tracks the `RenderTexture`s allocated for a
  player's model and releases them in `OnDestroy()`, since RenderTextures (unlike Texture2D) are a scarce GPU
  resource that needs an explicit `Release()` to avoid leaking VRAM across join/leave cycles.
- Ships the shader as a new `colorswap` AssetBundle via the existing `AssetBundleLoader` pattern, built the
  same way as the other UI bundles (`Nitrox → Build AssetBundles` in the Unity Editor).

The Marmoset player-model shader/material setup (`FixRebreatherMaterials`, shader keywords, etc.) is
completely untouched — the new shader only ever produces a `RenderTexture` that gets dropped into the
existing `_MainTex`/`_SpecTex` slots, it never touches lighting.

## Two small adjacent bugs found and fixed while testing this

1. **Remote player equipment invisible until an equip/unequip event, in sessions with 2+ remote players.**
   `PlayerModelManager` is registered as a shared instance, and it stored each player's equipment-visibility
   handlers on itself instead of per-player — so the most recently (re)spawned remote player's handlers
   silently replaced the previous one's, and updates meant for player A were applied to player B's model
   instead. Fixed by returning the handler list to the caller (`RemotePlayer` now keeps its own).
2. **`FootstepPacketProcessor` `NullReferenceException`** when a footstep packet for a session arrives before
   that remote player's `Body` GameObject has been assigned (a pre-existing race, surfaced more often by the
   longer join-processing window from the caching work above). Guarded with the same
   `player.HasValue && player.Value.Body` pattern already used elsewhere (e.g.
   `PlayerHeldItemChangedProcessor`).

## Testing done so far

- Manual 2-client test: join hitch is completely gone (previously ~1/3 of the original hitch even with
  caching, now imperceptible).
- Equipment colors visually checked against the previous CPU-baked look — correct as far as tested.

## Still needs testing before merge

This has only been verified with 2 clients and a single player color so far. Before merging, please test more
thoroughly:

- [ ] 3+ simultaneous remote players joining/leaving repeatedly (VRAM check: `RenderTexture` cleanup via
      `PlayerColorRenderTextures` should prevent leaks, but worth confirming over a longer session).
- [ ] Several different player colors, especially near the per-equipment hue/saturation/vibrancy range
      boundaries (e.g. dive suit hue 5°-45°, radiation tank hue 0°-85°) to rule out off-by-one edge artifacts
      in the GPU range check vs. the old CPU one.
- [ ] Rebreather/radiation suit specifically, since those combine multiple materials and (for the radiation
      suit) multiple regions on one texture.


SmokerDoker:
I also took two screenshots showing players with different colors and equipment using the new shader. I hope this helps.

Player with color cyan
Player with color red_2

PR #2826, opened 2026-07-07
Removed protobuf save option from server

Measurity:
Should be tested with Visual Studio.

Removed ProtoBuf save files as they cannot be upgraded and required server code access to work (as launcher needed to parse the save files).

## How to test

1. Start launcher
2. Try start server
3. It should start without errors

## Fix sourced from
- https://redirect.github.com/dotnet/msbuild/issues/1916#issuecomment-2031413812

### DIFF
--- a/Nitrox.Server.Subnautica/Models/GameLogic/Players/PersistedPlayerData.cs
+++ b/Nitrox.Server.Subnautica/Models/GameLogic/Players/PersistedPlayerData.cs
@@ -1,6 +1,5 @@
 using System.Collections.Generic;
 using System.Linq;
-using System.Runtime.Serialization;
 using Nitrox.Model.Core;
 using Nitrox.Model.DataStructures;
 using Nitrox.Model.DataStructures.GameLogic;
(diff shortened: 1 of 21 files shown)


Coding-Hen:
Is it worth us having one final upgrade to move the remaining files from protobuf over to json if they are currently using them as a one way save upgrade?

Do we want to remove the serialisation properties on the classes as well if they won't be needed anymore?

Measurity:
I don't expect non-tech inclined users to switch to protobuf. Though we should add a release note that users should switch to json on the old version before updating Nitrox.

I'll see which attributes I can remove from the classes. I think there is some overlap between json <-> protobuf attributes.

Measurity:
Removed attributes from server project serializable objects only. The rest might be used in the game and it's too much testing to do in this PR.

Tested with a 1.8.1.0 large save.

Coding-Hen (approved):
LGTM

Nitrox.Server.Subnautica/Models/GameLogic/Players/PersistedPlayerData.cs
```diff
@@ -9,61 +8,43 @@
 
 namespace Nitrox.Server.Subnautica.Models.GameLogic.Players;
 
-[DataContract]
 internal sealed class PersistedPlayerData
 {
-    [DataMember(Order = 1)]
     public string Name { get; set; }
 
-    [DataMember(Order = 2)]
```
dartasen, line 18: I think that the serializer is using DataMember as a fallback, isn't it ?

Measurity: Yes, but should be fine:

**MessagePack** only uses it for `Order`, but we don't use it for server types.
**NewtonsoftJSON** only uses it for `Name`, but we don't specify it anywhere. It will use property name as fallback.

image

dartasen (approved):

dartasen:
Works on VS 26

PR #2839, opened 2026-08-16
Allow quitting to main menu

### LINKED ISSUE #2545: Change quit button to not close the game
### Describe the issue

I am wondering if the quit button can be made to not close the game, but return the user back to the main menu. NinjaPedro in the Discord said the reason it quits the game, is because it causes bugs if a user left one server, and joined another. So maybe that can be fixed.

Image

MistaOmega:
## Linked Issues / Context

Fixes: https://github.com/SubnauticaNitrox/Nitrox/issues/2545

## Description of Change

Changes a number of GameObjects and behaviours to discard and destroy correctly on game quit, to allow rejoining from the main menu

## Validation

1. Enter game
2. Do stuff
3. Quit game
4. *should arrive in main menu*
5. Go back into game
6. profit??

## PR Checklist

- [x] Has a linked issue
- [] Has been tested in-game (if applicable) || tests were complete from original PR before patch reapply, less diverse checks have been done on this new branch

---


NitroxClient/GameLogic/InitialSync/StoryGoalInitialSyncProcessor.cs
```diff
@@ -151,7 +153,12 @@ private static void SetScheduledGoals(InitialPlayerSync packet)
 
         // We don't want any scheduled goal we add now to be executed before initial sync has finished, else they might not get broadcasted
         StoryGoalScheduler.main.paused = true;
-        Multiplayer.OnLoadingComplete += () => StoryGoalScheduler.main.paused = false;
+        if (!storyGoalHandlerRegistered)
+        {
+            Multiplayer.OnLoadingComplete += OnLoadingCompleteUnpauseStoryGoals;
+            Multiplayer.OnAfterMultiplayerEnd += CleanupStoryGoalHandler;
+            storyGoalHandlerRegistered = true;
+        }
```
Jannify: What is the reason that we need to manually clean up the registered actions? Shouldn't they be cleared when recreating the LifetimeScope? I imagine the Multiplayer.cs still hold NRE references but they could then be cleared no? Having a general solution sound safe in my hand then to remember to clear these states in the future (when coding).

NitroxClient/GameLogic/Bases/BuildingHandler.cs
```diff
@@ -416,4 +416,15 @@ public void AskForResync()
...
+
+    protected override void OnSessionEnd()
+    {
+        if (Main == this)
+        {
+            Main = null;
+        }
+        BuildQueue?.Clear();
+        BasesCooldown?.Clear();
+        Operations?.Clear();
```
Jannify: Same here. ~~clearing static fields is fine but maybe we should re-initialze all MBs created by Multiplayer.InitMonoBehaviours()~~ I see you are doing that, is that not enough?

NitroxClient/GameLogic/InitialSync/StoryGoalInitialSyncProcessor.cs
MistaOmega: I don't think the lifetime scope will clean these, because OnLoadingComplete and OnAfterMultiplayerEnd aren't scoped as a dependency of anything, they're just static events on a type 

in my head the rule going forward would make sense to be any handler with per-session state added to these two events needs to unsubscribe itself on OnAfterMultiplayerEnd. NitroxSessionBehaviour does this automatically for monos for plain classes we use the pattern seen here

NitroxClient/GameLogic/Bases/BuildingHandler.cs
MistaOmega: This is extra-defensive, yes. I'm happy for you to make the judgement call here, because as far as we're concerned these clear when the session ends (I.E I've left the game) so they'll be freshly remade on the rejoin.

NitroxClient/GameLogic/InitialSync/StoryGoalInitialSyncProcessor.cs
Jannify: Yeah. My point is only that this solution leaves the danger that someone forgets one "-=", if we on the other hand jus do a clear all solution we will never have this problem. Maybe something like answer 1 or 3 from https://stackoverflow.com/questions/91778/how-to-remove-all-event-handlers-from-an-event can work.

NitroxClient/GameLogic/Bases/BuildingHandler.cs
Jannify: I would only leave the static fields there. We expect the session and all its objects to be cleared and if not we want to get the error/message asap.

NitroxClient/GameLogic/InitialSync/StoryGoalInitialSyncProcessor.cs
MistaOmega: preferred 3 here. Added a type to subscribe to, with a registry to allow a single command clear all

PR #2842, opened 2026-08-21
Changes to broadcasting health and death updates

### LINKED ISSUE #123: [Misc] Disable ability to pause while in multiplayer.

### LINKED ISSUE #2510: Creatures won't die for one player

1.8.0.0 (Latest release)

### Details about version

_No response_

Windows

Steam

### Describe the issue.

Me and my friend are going to kill all leviathans in the game, and when we were killing Reapers, it died for my friend, but for me it just freeze for like 5 seconds, and start swimming to the end of the map or to some place very far, i did not tried to follow it but that happened 2 times till now, we just killed 2 and both have bugged in that exactly way for me.
One note: if i stasis it when it will try to swim to somewhere far, it will just start rounding that place for only 3 meters away and will not try to attack, but it will stay there, don't know if it vanish when restarting the server. 
And when server was resetted it was happening to my friend too but with Ghosts, not tested with Reapers.

(Just to record and i don't think that counts as more than 1 issue in a thread (if i does, i am really sorry))

I THINK (not confirmed) i saw a Reaper appearing and vanishing in a random spot of the map, but maybe i was wrong and was just a... idk, a Bone Shark or something else maybe, but if i had to say i think they are not respawning as intended when you kill then.

I am from Brazil btw, sorry a lot for the bad english...

Well, i just tried to kill the Reaper and Ghost with my friend, i was the host, i don't know if that makes the bug happen or not.

### Relevant log output

Nothing about in the logs, tried to search but i did not founded a single line of it or about, i think it don't even store in the log the Reaper or Ghost kill

MistaOmega:
## Linked Issues / Context

Fixes #2510 

## Description of Change

Adds a collection of changes to allow the syncing of creature deaths to clients from the server

## Validation

1. Start a server
2. join with 2 clients
3. kill the creature together (both do damage)
4. Observe the creature dies on both clients

## PR Checklist

- [x] Has a linked issue

---

Specifically scanning through decompiled functions, which surfaced a bug with how we used liveMixin.TakeDamage

### DIFF
--- a/NitroxClient/GameLogic/Spawning/Metadata/Extractor/LiveMixinMetadataExtractor.cs
+++ b/NitroxClient/GameLogic/Spawning/Metadata/Extractor/LiveMixinMetadataExtractor.cs
@@ -1,0 +1,20 @@
+using Nitrox.Model.Subnautica.DataStructures.GameLogic.Entities.Metadata;
+using NitroxClient.GameLogic.Spawning.Metadata.Extractor.Abstract;
+
+namespace NitroxClient.GameLogic.Spawning.Metadata.Extractor;
+
+/// <summary>
+/// Vehicles/cyclops already sync health through their own metadata.
+/// </summary>
+public class LiveMixinMetadataExtractor : EntityMetadataExtractor<LiveMixin, LiveMixinMetadata>
+{
+    public override LiveMixinMetadata Extract(LiveMixin liveMixin)
+    {
+        if (Resolve<LiveMixinManager>().IsWhitelistedUpdateType(liveMixin))
+        {
+            return null;
+        }
+
+        return new(liveMixin.health);
+    }
+}
--- a/NitroxPatcher/Patches/Dynamic/LiveMixin_TakeDamage_Patch.cs
+++ b/NitroxPatcher/Patches/Dynamic/LiveMixin_TakeDamage_Patch.cs
@@ -88,15 +88,23 @@ public sealed partial class LiveMixin_TakeDamage_Patch : NitroxPatch, IDynamicPa
 
     private static void BroadcastDefaultTookDamage(LiveMixin liveMixin)
     {
-        // Let others know if we have a lock on this entity
-        if (liveMixin.TryGetIdOrWarn(out NitroxId id) && Resolve<SimulationOwnership>().HasAnyLockType(id))
+        if (!liveMixin.TryGetIdOrWarn(out NitroxId id))
         {
-            Optional<EntityMetadata> metadata = Resolve<EntityMetadataManager>().Extract(liveMixin.gameObject);
+            return;
+        }
 
-            if (metadata.HasValue)
-            {
-                Resolve<Entities>().BroadcastMetadataUpdate(id, metadata.Value);
-            }
+        // Unlike vehicles (gated by ShouldApplyNextHealthUpdate above), creatures apply every attacker's damage locally with no lock check
+        // so every attacker should also broadcast the result.
+        if (Resolve<LiveMixinManager>().IsWhitelistedUpdateType(liveMixin) && !Resolve<SimulationOwnership>().HasAnyLockType(id))
+        {
+            return;
+        }
+
+        Optional<EntityMetadata> metadata = Resolve<EntityMetadataManager>().Extract(liveMixin.gameObject);
+
+        if (metadata.HasValue)
+        {
+            Resolve<Entities>().BroadcastMetadataUpdate(id, metadata.Value);
         }
     }
 }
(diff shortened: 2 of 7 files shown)


NitroxClient/GameLogic/Spawning/Metadata/Extractor/LiveMixinMetadataExtractor.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace NitroxClient.GameLogic.Spawning.Metadata.Extractor;
+
+/// <summary>
+/// Vehicles/cyclops already sync health through their own metadata.
+/// </summary>
+public class LiveMixinMetadataExtractor : EntityMetadataExtractor<LiveMixin, LiveMixinMetadata>
+{
+    public override LiveMixinMetadata Extract(LiveMixin liveMixin)
+    {
+        if (Resolve<LiveMixinManager>().IsWhitelistedUpdateType(liveMixin))
```
SpaceMonkeyy86, line 13: The `IsWhitelistedUpdateType()` method only allows updates for vehicles. With this PR, this should be expanded to include regular creatures, or maybe the method is no longer needed and should be removed.

NitroxPatcher/Patches/Dynamic/LiveMixin_TakeDamage_Patch.cs
```diff
@@ -88,15 +88,23 @@ private static bool HandlePvP(LiveMixin liveMixin, GameObject dealer, float dama
...
-        if (liveMixin.TryGetIdOrWarn(out NitroxId id) && Resolve<SimulationOwnership>().HasAnyLockType(id))
+        if (!liveMixin.TryGetIdOrWarn(out NitroxId id))
         {
-            Optional<EntityMetadata> metadata = Resolve<EntityMetadataManager>().Extract(liveMixin.gameObject);
+            return;
+        }
+
+        // Unlike vehicles (gated by ShouldApplyNextHealthUpdate above), creatures apply every attacker's damage locally with no lock check
+        // so every attacker should also broadcast the result.
+        if (Resolve<LiveMixinManager>().IsWhitelistedUpdateType(liveMixin) && !Resolve<SimulationOwnership>().HasAnyLockType(id))
```
SpaceMonkeyy86, line 98: Shouldn't the first value be inverted, as we only want to broadcast update types which are whitelisted? Also, I do not understand how the comment applies to this code.

NitroxClient/GameLogic/Spawning/Metadata/Extractor/LiveMixinMetadataExtractor.cs
MistaOmega: This function excludes vehicles, as they do as you say

NitroxPatcher/Patches/Dynamic/LiveMixin_TakeDamage_Patch.cs
MistaOmega: Vehicles only broadcast when they hold the lock, but creates don't have that. So if I invert it creatures wouldn't broadcast

NitroxClient/GameLogic/Spawning/Metadata/Extractor/LiveMixinMetadataExtractor.cs
```diff
@@ -0,0 +1,20 @@
+using Nitrox.Model.Subnautica.DataStructures.GameLogic.Entities.Metadata;
+using NitroxClient.GameLogic.Spawning.Metadata.Extractor.Abstract;
+
+namespace NitroxClient.GameLogic.Spawning.Metadata.Extractor;
+
+/// <summary>
+/// Vehicles/cyclops already sync health through their own metadata.
+/// </summary>
+public class LiveMixinMetadataExtractor : EntityMetadataExtractor<LiveMixin, LiveMixinMetadata>
```
SpaceMonkeyy86, line 9: This would cause every object with a LiveMixin component to receive this metadata. Many LiveMixins have their own special handling for health (e.g. fires which use a LiveMixin to track how much they have already been doused), so this would have sweeping unintended side effects. I would at least limit it to objects with a Creature component

tornac1234: You're right but there's another catch, some Creatures have their own metadata for other behaviours.
To me, we can't have this work in before we can allow for multiple metadatas on one entity

MistaOmega: Drafting whilst I look into the above as a separate PR 
been doing this morning but forgot to comment / draft here kekw

PR #2845, opened 2026-08-22
Fix PipeSurfaceFloater (Floating Air Pump) losing oxygen supply on reconnect/reload

Gugarion:
## Summary
- `PipeSurfaceFloater.deployed` is a purely client-local runtime flag that vanilla only ever sets via the original placer's own tool-use callback (`OnToolUseAnim`) — but `PipeSurfaceFloater` doesn't derive from `PlaceTool`, so it's never covered by this spawner's existing `PlaceTool` handling.
- `PipeSurfaceFloater.GetProvidesOxygen()` returns `false` unconditionally while `!deployed`, and the entire `OxygenPipe` chain rooted at this floater ultimately asks `GetRoot().GetProvidesOxygen()` — so the whole chain silently stops providing oxygen for every client except the original placer's own still-running session, i.e. on any reconnect, and even for the original placer once this GameObject's cell unloads and reloads.
- `deployed` also gates `WorldForces.handleGravity`, so an undeployed floater drifts instead of staying anchored.
- Any `GlobalRootEntity` of this `TechType` reaching this spawner is, by construction, already placed (see `Items.cs`'s `Dropped()` checks), so it's always correct to force `deployed = true` and the matching kinematic/physics state here.
## Mods I've been using
- Nitrox 1.8.1.0
- https://github.com/Okabintaro/SubmersedVR/releases - Sumbersed [0.2.0]
- https://github.com/TheGeeks0424/BuddySystem - Buddy System [1.10.6]
- https://www.nexusmods.com/site/mods/202 - Subnautica Support [3.3.3]
- https://www.nexusmods.com/subnautica/mods/1108 - Tobey's BepInEx Pack for Subnautica [5.4.23-pack.3.1.1]
## Test plan
- [x] Placed a Floating Air Pump, connected an oxygen pipe chain to it, disconnected and reconnected to the server, confirmed oxygen supply continued working (previously stopped).
- [x] Confirmed a second player joining after the pump was placed sees it correctly providing oxygen immediately, without needing the original placer present.
- [x] Confirmed the pump stays anchored (doesn't drift) after reconnect.

### DIFF
--- a/NitroxClient/GameLogic/Spawning/WorldEntities/GlobalRootEntitySpawner.cs
+++ b/NitroxClient/GameLogic/Spawning/WorldEntities/GlobalRootEntitySpawner.cs
@@ -75,6 +75,23 @@ public class GlobalRootEntitySpawner : SyncEntitySpawner<GlobalRootEntity>
         {
             PlacedWorldEntitySpawner.AdditionalSpawningSteps(gameObject);
         }
+
+        // PipeSurfaceFloater (Floating Air Pump) is in SubnauticaMap.GLOBAL_ROOT_TECH_TYPES, so it always
+        // spawns through this generic path -- but "deployed" is a purely client-local runtime flag vanilla
+        // only ever sets via the original placer's own OnToolUseAnim() tool-use callback (PipeSurfaceFloater
+        // doesn't derive from PlaceTool, so the block above never covers it). PipeSurfaceFloater.GetProvidesOxygen()
+        // returns false unconditionally while !deployed, and the entire OxygenPipe chain rooted at this floater
+        // (OxygenPipe.GetProvidesOxygen() ultimately asks GetRoot().GetProvidesOxygen()) silently stops
+        // providing oxygen as a result -- for every client except the original placer's own still-running
+        // session, i.e. any reconnect, and even the original placer after this GameObject's cell unloads and
+        // reloads. deployed also gates WorldForces.handleGravity, so an undeployed floater drifts instead of
+        // staying anchored. Any GlobalRootEntity of this TechType reaching this spawner is, by construction
+        // (Items.cs.Dropped()), already placed -- so it's always correct to force this here.
+        if (gameObject.TryGetComponent(out PipeSurfaceFloater pipeSurfaceFloater))
+        {
+            pipeSurfaceFloater.deployed = true;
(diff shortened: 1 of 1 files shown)


NitroxClient/GameLogic/Spawning/WorldEntities/GlobalRootEntitySpawner.cs
```diff
@@ -75,6 +75,23 @@ private void SetupObject(GlobalRootEntity entity, GameObject gameObject)
...
+        // spawns through this generic path -- but "deployed" is a purely client-local runtime flag vanilla
+        // only ever sets via the original placer's own OnToolUseAnim() tool-use callback (PipeSurfaceFloater
+        // doesn't derive from PlaceTool, so the block above never covers it). PipeSurfaceFloater.GetProvidesOxygen()
+        // returns false unconditionally while !deployed, and the entire OxygenPipe chain rooted at this floater
+        // (OxygenPipe.GetProvidesOxygen() ultimately asks GetRoot().GetProvidesOxygen()) silently stops
+        // providing oxygen as a result -- for every client except the original placer's own still-running
+        // session, i.e. any reconnect, and even the original placer after this GameObject's cell unloads and
+        // reloads. deployed also gates WorldForces.handleGravity, so an undeployed floater drifts instead of
+        // staying anchored. Any GlobalRootEntity of this TechType reaching this spawner is, by construction
+        // (Items.cs.Dropped()), already placed -- so it's always correct to force this here.
```
tornac1234: I agree that this change (deployed = true) is good, but the comments isn't entirely true.

PipeSurfaceFloater.deployed is set to true as soon as you drop it under water (under y = -0.F) (in FixedUpdate)

The real issue is when the floater spawns above y = -0.5f, because then it no longers gets ``deployed = true``, and then if its children spawn onto it (``children.Count > 0``) (before the floater sinks to under -0.5f), it will be set to kinematic (by ``UpdateRigidbody``) and then it will be stuck upon y = -0.5f. Result: ``GetRoot`` will return ``null`` forever, and ``GetProvidesOxygen`` will also return ``false`` forever.

tornac1234 (approved):
Thank you for finding about this @Gugarion !

dartasen (approved):

PR #2846, opened 2026-08-22
Add DiveReel (Pathfinder Tool) multiplayer sync

Gugarion:
## Summary
Adds full multiplayer support for the DiveReel (Pathfinder Tool), previously entirely single-player-only — other players never saw a reel trail placed by anyone else.

- **Packets** (`Nitrox.Model.Subnautica`): node placed, nodes reset, and an initial-sync packet for late-joining players.
- **Server**: `DiveReelNodeTracker` records each player's currently-placed node positions, relays placements/resets live to other connected players, sends a late-joining player everyone else's existing trails, and re-broadcasts a rejoining player's own trail to others (the tracker is intentionally never cleared on disconnect — a reconnect shouldn't lose the player's own trail, but already-connected clients have no other path to get it back, since `PlayerJoinedMultiplayerSession` carries no DiveReel data).
- **Client**: processors apply incoming packets by spawning/clearing `DiveReelNodeMarkers` — decorative, non-interactive visual markers (tagged via `NitroxDiveReelNodeMarkerTag` so a separate patch can tell them apart from the local player's own real nodes, since both use the exact same prefab) placed at each remote node position, tinted to that player's login color with a directional arrow matching the reel's actual placement.
- `DiveReelNode_Start_LocalTint_Patch` tints the local player's own placed nodes to their own login color too, matching the remote-marker tinting, since vanilla's rendering otherwise always uses a fixed default color for the local player.
- Spoofed-`PlayerId` packets are rejected server-side (a client claiming a different `PlayerId` than the one it authenticated with is dropped with a warning, never trusted).

**Note on scope:** developed and tested primarily via VR play (SubmersedVR + Nitrox), but nothing here is VR-specific — it syncs identically for any client, VR or flat/gamepad, with no reference to or dependency on any VR mod. The visual markers are likely most useful in VR specifically (e.g. via BuddySystem, where spatial trail-following reads more naturally), but the sync mechanics themselves are universal.

**Note on target branch:** targets `release/1.8.1.0` (the current latest release) rather than `master` — `master` has since undergone a substantial server-side architectural refactor (async packet sending, DI-based construction replacing the old Autofac registrar/`WorldPersistence` pattern) that this PR's server-side pieces were not written against and have not been tested against. Happy to help port this to current `master` conventions in a follow-up once there's time to properly test against the new server architecture.
## Mods I've been using
- Nitrox 1.8.1.0
- https://github.com/Okabintaro/SubmersedVR/releases - Sumbersed [0.2.0]
- https://github.com/TheGeeks0424/BuddySystem - Buddy System [1.10.6]
- https://www.nexusmods.com/site/mods/202 - Subnautica Support [3.3.3]
- https://www.nexusmods.com/subnautica/mods/1108 - Tobey's BepInEx Pack for Subnautica [5.4.23-pack.3.1.1]
## Test plan
- [x] Two players placed DiveReel nodes independently; each saw the other's trail appear live, correctly positioned and oriented.
- [x] A third player joining after both trails existed received both via initial sync.
- [x] A player disconnecting and reconnecting kept their own trail, and other already-connected clients correctly saw it re-appear.
- [x] Resetting a reel correctly cleared that player's markers for everyone else.
- [x] Confirmed a client sending a packet with a spoofed `PlayerId` is rejected server-side.
- [x] Extended multiplayer play session — no marker leaks, desyncs, or stale trails observed.

### DIFF
--- a/Nitrox.Model.Subnautica/Packets/DiveReelNodePlaced.cs
+++ b/Nitrox.Model.Subnautica/Packets/DiveReelNodePlaced.cs
@@ -1,0 +1,25 @@
+using System;
+using Nitrox.Model.DataStructures.Unity;
+using Nitrox.Model.Packets;
+
+namespace Nitrox.Model.Subnautica.Packets;
+
+[Serializable]
+public class DiveReelNodePlaced : Packet
+{
+    public ushort PlayerId { get; }
+    public NitroxVector3 Position { get; }
+    public bool IsFirst { get; }
+
+    public DiveReelNodePlaced(ushort playerId, NitroxVector3 position, bool isFirst)
+    {
+        PlayerId = playerId;
+        Position = position;
+        IsFirst = isFirst;
+    }
+
+    public override string ToString()
+    {
+        return $"[{nameof(DiveReelNodePlaced)} PlayerId: {PlayerId}, Position: {Position}, IsFirst: {IsFirst}]";
+    }
+}
--- a/Nitrox.Model.Subnautica/Packets/DiveReelNodesInitialSync.cs
+++ b/Nitrox.Model.Subnautica/Packets/DiveReelNodesInitialSync.cs
@@ -1,0 +1,22 @@
+using System;
+using System.Collections.Generic;
+using Nitrox.Model.DataStructures.Unity;
+using Nitrox.Model.Packets;
+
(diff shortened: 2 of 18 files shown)


dartasen:
Thank's for the PR
This should probably target main rather than the release branch

PR #2642, opened 2026-01-11
Player Animation Sync

misterbubb:
Closes #2287 and #2639

Re-enables the PlayerCinematicController patches that were disabled in 1.8.

## Changes
- Added `InCinematic` flag to `RemotePlayer` to prevent movement packet race conditions during animations
- Fixed moonpool docking animation (proper vehicle position interpolation)
- Fixed undocking replaying enter animation (set animator state before cinematic)
- Fixed `MultiplayerCinematicReference` throwing exceptions (now logs warnings)
- Added gun terminal animation parameter sync (first use, cured state)
- Added cinematic lock system
- Added defensive checks for race conditions when players join mid-cinematic
- Added bed multiplayer sync with animation support

## Testing
- Needs loads of testing, things got quite messy.

PR #2859, opened 2026-08-25
Player Animation Sync - with fixes

MistaOmega:
PR Co-authored by @misterbubb 

## Linked Issues / Context

Contains 1 additional commit atop the changes in #2642 with issues outlined within that PR addressed
I tried to direct this to @misterbubb 's nitrox directly but could not :( 

## Description of Change

This is the same change as #2642 but with the following changes: 
I've reworked some of the Bed animation works, so now two people can simultaneously access a 2 sided bed
The cyclops bug should be resolved
and the animations for the pod are synced nicely, properly dismantle when players leave mid-flow and are player, not session specific

## Validation

Test collection thus far:

Enter 2 sided bed with 2 people - success, sleeps through night, can see anim plays into bed
Enter and exit 2 sided bed - success
Enter 2 sided bed and leave mid-anim - success, player disappears, bed can be accessed by another 2 players
Try to control cyclops with 2 people - success, can't access with 2 players
Try ladder pod animation with multiple people, success all people can see a player do the anim, and all players can do the anim
above, but for exit-from-bottom pod animation, success

## PR Checklist

- [x] Has a linked issue

---


MistaOmega:
Odd test to fail @Measurity (you're on the git blame) - Flakiness?

MistaOmega:
Some things I noticed in general that I'm unsure of if they've been around prior to the sync getting disabled in 1.8

Changing direction when swimming is initially kinda choppy, as the legs move, to do with packets sent for updates maybe? unsure.

Cyclops movement seems to have no animation sync, or it... kinda does? the character's rig seems to almost stutter / shudder a little whilst moving

PR #2864, opened 2026-08-28
Improve vehicle movement desync by using a RingBuffer

### LINKED ISSUE #123: [Misc] Disable ability to pause while in multiplayer.

tornac1234:
## Linked Issues / Context

When using a time skip command ("day" or "night"), there is a slight chance that for a certain time period (defined by the LatencyUpdatePeriod nitrox setting), a vehicle will stop moving at all.

Let's see an example (Player A, Player B and server):
- Player A sends "day" command
- Player A sends VehicleMovement packet with current time (time = 1000s)
- Server receives the time change command and sends a TimeChange Packet to Player B (time=2000s)
- Server transmits VehicleMovement packet to Player B
- Player B receives time change packet (new time=2000s)
- Player B receives VehicleMovement packet (it contains time=1000s), latency = current (2000s) - packet time (1000s) (and MaxAllowedLatency is set to be = latency), so the packet will be treated after a while... (the vehicle stops moving as soon as it has no VehicleMovement treated in a little while)
- Player B keeps receiving VehicleMovement packets, and after LatencyUpdatePeriod has passed, the MaxAllowedLatency is finally lowered to a regular threshold (the vehicle finally starts moving again)

## Description of Change

- [x] Use a brand new implemented RingBuffer to remove heap allocations at each VehicleMovement packet received (no more flooded GC)
- [x] Implement a time change detection system in TimeManager and read its values in MovementReplicator to act accordingly when encountering the edge case mentioned above

## Validation

- [x] Tested in game by spamming the "day" command while driving a Cyclops and see if the other player ever sees the Cyclops freeze. (it no longer does)

## PR Checklist

- ~~Has a linked issue~~
- ~~Has been tested on Windows/Linux/macOS (if applicable) (for cross-platform code)~~

---

The test was written by AI and I corrected it afterwards.
I also used AI to check if my RingBuffer was correct (it corrected some mistakes from me).

### DIFF
--- a/Nitrox.Model/DataStructures/ExpandableRingBuffer.cs
+++ b/Nitrox.Model/DataStructures/ExpandableRingBuffer.cs
@@ -1,4 +1,6 @@
 using System;
+using System.Collections;
+using System.Collections.Generic;
 
 namespace Nitrox.Model.DataStructures;
 
@@ -6,28 +8,34 @@ namespace Nitrox.Model.DataStructures;
 /// A dynamically expanding, circular first-in, first-out array that reduces memory allocations by reusing slots.
 /// Automatically doubles its capacity when full.
 /// </summary>
-public class RingBuffer<T>
+public class ExpandableRingBuffer<T> : IEnumerable<T>, IReadOnlyCollection<T>
 {
     private T[] buffer;
     private int capacity;
-    public int Head { get; private set; }
-    public int Tail => (Head + Count) % capacity;
+    private int Head { get; set; }
+    private int Tail => (Head + Count) % capacity;
     public int Count { get; private set; }
 
     /// <summary>
-    /// Gets or sets the element at the wrapped index.
+    /// Gets or sets the element at the index of the buffer (0 is first element).
     /// </summary>
     public T this[int index]
     {
-        get => buffer[index % capacity];
-        private set => buffer[index % capacity] = value;
+        get
+        {
+            if (index < 0 || index >= Count)
+            {
+                throw new ArgumentOutOfRangeException(nameof(index));
+            }
+            return buffer[(Head + index) % capacity];
+        }
     }
 
     public T First => Count > 0 ? buffer[Head] : throw new InvalidOperationException("Buffer is empty");
 
     public T Last => Count > 0 ? buffer[(Head + Count - 1) % capacity] : throw new InvalidOperationException("Buffer is empty");
 
-    public RingBuffer(int initialCapacity = 64)
+    public ExpandableRingBuffer(int initialCapacity = 64)
     {
         if (initialCapacity <= 0)
         {
@@ -40,11 +48,23 @@ public class RingBuffer<T>
 
     public void Expand()
     {
-        int newCapacity = capacity * 2;
+        int newCapacity = checked(capacity * 2);
         T[] newBuffer = new T[newCapacity];
-        for (int i = 0; i < Count; i++)
+
+        // example: [val, val, default, default, default, val, val, val]
+        //          left chunk                             right chunk
+        // NB: Head is at the beginning of the right chunk, Tail is at the end of the left chunk
+
+        // elements from Head to the end of the array
+        int rightChunkLength = Math.Min(Count, capacity - Head);
+
+        Array.Copy(buffer, Head, newBuffer, 0, rightChunkLength);
+
+        // in case there are elements from the beginning of the array to some extent
+        int leftChunkLength = Count - rightChunkLength;
+        if (leftChunkLength > 0)
         {
-            newBuffer[i] = buffer[(Head + i) % capacity];
+            Array.Copy(buffer, 0, newBuffer, rightChunkLength, leftChunkLength);
         }
 
         capacity = newCapacity;
@@ -89,7 +109,22 @@ public class RingBuffer<T>
             throw new InvalidOperationException("Can't remove an element from an empty ring buffer");
         }
 
+        buffer[(Head + Count - 1) % capacity] = default; // avoid memory leaks
         Count--;
-        buffer[Tail] = default; // avoid memory leaks
     }
+
+    public bool IsEmpty()
+    {
+        return Count == 0;
+    }
+
+    public IEnumerator<T> GetEnumerator()
+    {
+        for (int i = 0; i < Count; i++)
+        {
+            yield return buffer[(Head + i) % capacity];
+        }
+    }
+
+    IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
 }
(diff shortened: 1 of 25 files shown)


tornac1234:
Thanks Copilot reviews, made me realize the problem was actually server-side (I'll still refactor the client-side so the variable names are more explicit)

tornac1234:
> Copying what I wrote on discord
It seems that during the PR #2314, TimeKeeper was modified to become TimeService. In the process, you the intent behind RealTimeElapsed was removed, which was that it would never be modified and would only track server active time. Thus this metric would be useful when we would need a static timeline, for example when we sync movement.
Client-side kept that intent but server-side would serve data without that meaning.

The corrections applied now is that ActiveTime and GameTime are once again decoupled so that ActiveTime is on its own static timeline and movement sync can be based on it without fear that it'd be modified.

Nitrox.Server.Subnautica/Services/TimeService.cs
```diff
@@ -41,28 +41,28 @@ internal sealed class TimeService(IPacketSender packetSender, NtpSyncer ntpSynce
     private readonly Stopwatch stopWatch = new();
```
Measurity, line 42: Mind to move `TimeService` to the Nitrox.Server.Subnautica.Services namespace?

Measurity (approved):
Works for me in game

Nitrox.Model/DataStructures/RingBuffer.cs
```diff
@@ -0,0 +1,95 @@
+using System;
+
+namespace Nitrox.Model.DataStructures;
+
+/// <summary>
+/// A dynamically expanding, circular first-in, first-out array that reduces memory allocations by reusing slots.
+/// Automatically doubles its capacity when full.
+/// </summary>
+public class RingBuffer<T>
```
dartasen: Usually for data structures like that you want to implements big boss contracts like ICollection<T>, IEnumerable<T>, ... So you can use all the c# goodies like linq, foreach, ...

Nitrox.Model/DataStructures/RingBuffer.cs
```diff
@@ -0,0 +1,95 @@
...
+            throw new ArgumentOutOfRangeException(nameof(initialCapacity), initialCapacity, "Initial capacity must be greater than zero.");
+        }
+
+        capacity = initialCapacity;
+        buffer = new T[capacity];
+    }
+
+    public void Expand()
+    {
+        int newCapacity = capacity * 2;
```
dartasen:
```suggestion
        int newCapacity = checked(capacity * 2);
```

Nitrox.Model/DataStructures/ExpandableRingBuffer.cs
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        buffer[Head] = default; // avoid memory leaks
+        Head = (Head + 1) % capacity;
+        Count--;
+    }
+
+    public void RemoveLast()
+    {
+        if (Count == 0)
```
dartasen, line 107: Guess we could expose things like IsEmpty, IsFull

Nitrox.Server.Subnautica/Services/TimeService.cs
```diff
@@ -82,17 +82,12 @@ public TimeSpan ActiveTime
...
-        stopWatch.Reset();
-    }
-
     /// <summary>
     ///     Set current time depending on the current time in the day (replication of SN's system, see DayNightCycle.cs
     ///     commands for more information).
     /// </summary>
     /// <param name="type">Time to which you want to get to.</param>
-    public void ChangeTime(StoryManager.TimeModification type)
+    public async Task ChangeTime(StoryManager.TimeModification type)
```
dartasen: Same for all methods in this class that were turned async : You should add Async suffix at the end of the method

Nitrox.Model/DataStructures/ExpandableRingBuffer.cs
```diff
@@ -0,0 +1,95 @@
...
+        Count = 0;
+        Array.Clear(buffer, 0, buffer.Length);
+    }
+
+    public void Add(T item)
+    {
+        if (Count == capacity)
+        {
+            Expand();
+        }
```
dartasen, line 87: I don't really get why a ring buffer would want to automatically expand beyond the original capacity. If we reach the capacity, it should override an already existing element ?

Nitrox.Server.Subnautica/Models/GameLogic/SleepManager.cs
```diff
@@ -50,31 +49,26 @@ private bool AreAllPlayersInBed()
...
         int totalPlayers = playerManager.GetConnectedPlayers().Count;
-        packetSender.SendPacketToAllAsync(new SleepStatusUpdate(sessionIdsInBed.Count, totalPlayers));
+        await packetSender.SendPacketToAllAsync(new SleepStatusUpdate(sessionIdsInBed.Count, totalPlayers));
     }
 
-    private void StartSleep()
+    private async Task StartSleep()
     {
         isSleepInProgress = true;
+        sessionIdsInBed.Clear();
```
dartasen, line 61: Why is it cleared at the beginning of the method ?

Nitrox.Model/DataStructures/ExpandableRingBuffer.cs
tornac1234: I could probably rename it ExpandableRingBuffer because the point of this one is it doesn't lose data.
In the case of movements, we never want to drop any data because it would mean we would just no longer be able to reproduce movements at all as long as the movement input is bigger than expected (which we don't have much control on).

Nitrox.Server.Subnautica/Models/GameLogic/SleepManager.cs
tornac1234: since the function happens whatever it takes, it doesn't matter where it goes ig

Nitrox.Model/DataStructures/ExpandableRingBuffer.cs
```diff
@@ -1,33 +1,41 @@
...
 /// </summary>
-public class RingBuffer<T>
+public class ExpandableRingBuffer<T> : IEnumerable<T>, IReadOnlyCollection<T>
 {
     private T[] buffer;
     private int capacity;
-    public int Head { get; private set; }
-    public int Tail => (Head + Count) % capacity;
+    private int Head { get; set; }
+    private int Tail => (Head + Count) % capacity;
```
Jannify, line 16: Do we want the Tail to point to the last element + 1? My gut says it should point to the last element like Head points to the first. If we make this change we can also replace a few instances with Tail in the other methods.

tornac1234: From all online sources, Tail is always at the position where elements are added

Jannify: Then ignore my comment xD

dartasen (approved):

PR #2867, opened 2026-08-29
Fix LiveMixin entity destruction not syncing

### LINKED ISSUE #1701: LiveMixin is not synced for most of the instances
### Version
Master commit 9a028b24f42465c8dacfcdfdcf08dde58bf4f40e - Additional help information (#1690)

### Bug
When you destroy something like mushrooms with a Knife or a Prawn, it only happens on your client

### Expected
The destruction of entities should be synced

### Steps to reproduce
Get two players around some mushrooms (the pickupable ones), make one player hit the mushrooms with a knife.
They will disappear on its screen, but not on the other players' ones.
The same issue will happen if you use a prawn with a drill arm on the mushrooms.

### Other details:
It may have something to do with `LiveMixin_TakeDamage_Patch`

misterbubb:
## Linked Issues / Context

Fixes #1701

## Description of Change

Entities destroyed with knife or other damage sources were only visible on the local client. Other clients would still see the entity alive (mushrooms, coral, creatures).

## Validation

I did minor testing with 2 clients that destruction now syncs correctly for:
- Harvestables (purple mushrooms, coral plates)
- Passive fish (bladderfish, hoopfish)
- Aggressive creatures (stalker, biter)

All entity types now correctly disappear on remote clients when destroyed.

## PR Checklist

- [x] Has a linked issue
- [x] Has been tested in-game
- [x] Has been tested on Windows

### DIFF
--- a/NitroxPatcher/Patches/Dynamic/LiveMixin_Kill_Patch.cs
+++ b/NitroxPatcher/Patches/Dynamic/LiveMixin_Kill_Patch.cs
@@ -29,6 +29,8 @@ public sealed partial class LiveMixin_Kill_Patch : NitroxPatch, IDynamicPatch
         }
 
         if (__instance.destroyOnDeath || 
+            __instance.broadcastKillOnDeath || 
+            __instance.passDamageDataOnDeath || 
             Resolve<LiveMixinManager>().ShouldBroadcastDeath(__instance))
         {
             Resolve<IPacketSender>().Send(new EntityDestroyed(objectId));
--- a/NitroxPatcher/Patches/Dynamic/LiveMixin_Kill_Patch.cs
+++ b/NitroxPatcher/Patches/Dynamic/LiveMixin_Kill_Patch.cs
@@ -18,17 +18,20 @@ public sealed partial class LiveMixin_Kill_Patch : NitroxPatch, IDynamicPatch
             return;
         }
 
-        // We don't broadcast if we don't have objectId or if the object is whitelisted,
-        // in which case kill broadcast is managed differently
-        if (!__instance.TryGetNitroxId(out NitroxId objectId) ||
-            Resolve<LiveMixinManager>().IsWhitelistedUpdateType(__instance))
+        if (!__instance.TryGetNitroxId(out NitroxId objectId))
         {
             return;
         }
-        
-        // Some objects don't have destroyOnDeath but we still need to broadcast the death
-        // (because the destruction is managed by another script)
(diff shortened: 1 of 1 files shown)


MistaOmega:
LGTM, tested on two linux clients. 

Two noticed things worth covering elsewhere
Things like mushroom breaks don't trigger the sound / particle effects for all players nearby
same for damage from an acid mushroom. close by players unaffected, minus the one who broke the mushroom

misterbubb:
> LGTM, tested on two linux clients.
> 
> Two noticed things worth covering elsewhere Things like mushroom breaks don't trigger the sound / particle effects for all players nearby same for damage from an acid mushroom. close by players unaffected, minus the one who broke the mushroom

thanks for testing, those are separate missing features in Nitrox that would need their own implementation. Issues should probably be opened for those.

MistaOmega:
> > LGTM, tested on two linux clients.
> > Two noticed things worth covering elsewhere Things like mushroom breaks don't trigger the sound / particle effects for all players nearby same for damage from an acid mushroom. close by players unaffected, minus the one who broke the mushroom
> 
> thanks for testing, those are separate missing features in Nitrox that would need their own implementation. Issues should probably be opened for those.

Yeah agreed, I'll make the issues

NitroxPatcher/Patches/Dynamic/LiveMixin_Kill_Patch.cs
```diff
@@ -18,17 +18,20 @@ public static void Postfix(LiveMixin __instance)
...
-        if (__instance.destroyOnDeath || Resolve<LiveMixinManager>().ShouldBroadcastDeath(__instance))
+
+        if (Resolve<LiveMixinManager>().IsWhitelistedUpdateType(__instance))
+        {
+            return;
+        }
+
+        if (__instance.destroyOnDeath || 
+            __instance.broadcastKillOnDeath || 
+            __instance.passDamageDataOnDeath ||
```
tornac1234: How do you know these two flags (broadcastKillOnDeath and passDamageDataOnDeath) have the meaning of destroying the entity ?
I'm pretty sure in the LiveMixin code, the only part destroying immediately the entity is destroyOnDeath, the two other flags being only there to trigger a message broadcast ("OnKill")

misterbubb: broadcastKillOnDeath is the default (true for all fish, creatures, mushrooms) even when destroyOnDeath is false. So these flags correctly identify entities that need to send EntityDestroyed for multiplayer sync, even if the GameObject destruction happens differently.

PR #2872, opened 2026-08-30
Fix WaitScreen progress never reaching 100%

dartasen:
## Description of Change

Since 2.0 the waitscreen is no longer visible during game load, and most initial sync step processors are overriding this method. But always good to have code working 😄 

image

## PR Checklist

- [ ] Has a linked issue

---

### DIFF
--- a/NitroxClient/GameLogic/InitialSync/Abstract/InitialSyncProcessor.cs
+++ b/NitroxClient/GameLogic/InitialSync/Abstract/InitialSyncProcessor.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections;
 using System.Collections.Generic;
 using Nitrox.Model.Subnautica.Packets;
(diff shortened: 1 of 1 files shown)


NitroxClient/GameLogic/InitialSync/Abstract/InitialSyncProcessor.cs
```diff
@@ -15,7 +15,7 @@ public virtual IEnumerator Process(InitialPlayerSync packet, WaitScreen.ManualWa
         for (int i = 0; i < Steps.Count; i++)
         {
             yield return Steps[i](packet);
-            waitScreenItem.SetProgress((float)i / Steps.Count);
+            waitScreenItem.SetProgress(i + 1, Steps.Count);
```
Coding-Hen, line 18: Rather than +1 can't we just initialise it at 1 in the loop and make it <=? Otherwise we are +1 making it a little unclear why

Coding-Hen (approved):
Other than the minor comment this looks good to me. Better to have to working as you say

NitroxClient/GameLogic/InitialSync/Abstract/InitialSyncProcessor.cs
dartasen: Since arrays starts at 0 in C#, I don't think it would fit to start at one and add a i - 1 for the array access. It's just a basic currentStep / totalSteps computation that should be self explanatory

Coding-Hen: Agreed, it was just a thought, let's keep as is

Measurity (approved):

PR #2873, opened 2026-08-30
Fix Pipe Surface Floater's movements

tornac1234:
## Linked Issues / Context

- [x] Pipe surface floater used to be highly inconsistant with how they would be fully deployed and surface correctly.
- [x] RemotelyControlled entities would be teleported once around 0, 0, 0 at first, which would cause some unexpected errors

image

## Description of Change

- [x] Reworked RemotelyControlled so it's smoother for both entities and swimming creatures (in particular, reaper leviathans used to teleport once every few meters)
- [x] Fix RemotelyControlledPipeFloater to properly detect surfacing and act accordingly when it should be static
- [x] Remove useless GC allocations in EntityPositionBroadcaster by reworking a good part of it

## Validation

To observe the actual PipeSurfaceFloater fix:
- connect with Player A and Player B
- type subnautica command ``item pipesurfacefloater`` to get one
- throw it with one player, observe its movement with the other player
- when it's surfaced, make sure neither player can move it by pushing it. also try stepping on it to make sure that it's at the exact same place for both players

To observe Reaper Leviathan's movements improvements (as a side effect from the whole rework):
- connect with Player A and Player B
- type subnautica command ``spawn reaperleviathan``
- observer that it no longer glitches every few meters

## PR Checklist

- ~~Has tests for the changes (if applicable)~~
- ~~Has a linked issue~~
- ~~Has been tested on Windows/Linux/macOS (if applicable) (for cross-platform code)~~

---

AI was used to review myself before opening this PR


tornac1234:
I'll send a commit with slight corrections (fixing the test too) once #2845 is merged

NitroxClient/MonoBehaviours/EntityPositionBroadcaster.cs
```diff
@@ -38,73 +66,132 @@ public void Update()
...
+        }
+
+        if (entityObject.TryGetComponent(out RemotelyControlled remotelyControlled))
         {
             Object.Destroy(remotelyControlled);
         }
     }
 
-    public static void StopWatchingEntity(NitroxId id)
+    private void CheckEntities()
```
Jannify, line 143: Minor: Would like a comment what this method checks

Jannify, line 143: This seems like a heavy method. Do we want/need to call it every update?

NitroxClient/GameLogic/MovementHelper.cs
```diff
@@ -32,22 +32,13 @@ public static void Stop(GameObject go)
         public static Vector3 GetCorrectedVelocity(Vector3 remotePosition, Vector3 remoteVelocity, GameObject gameObject, float correctionTime)
```
Jannify: Would change input param from GameObject to Vector3 to make clear we don't alter the state. Or what is your opinion?

NitroxClient/MonoBehaviours/RemotelyControlled.cs
```diff
@@ -6,82 +6,171 @@ namespace NitroxClient.MonoBehaviours;
...
+            {
+                rigidbody.isKinematic = false;
+            }
+        }
+
+        if (worldForces && !followsSpline)
+        {
+            disabledWorldForces = worldForces.enabled;
+            worldForces.enabled = false;
+        }
```
Jannify, line 44: What is the case/state if this triggers? From just the 3 lines I can't interfere the SN behaivour. Maybe a comment to explain?

NitroxClient/MonoBehaviours/RemotelyControlled.cs
```diff
@@ -6,82 +6,171 @@ namespace NitroxClient.MonoBehaviours;
...
 
-        if ((selfTransform.position - position).sqrMagnitude > 25) // Optimized 5m distance test
+        if (rigidbody)
         {
-            selfTransform.position = position;
-            selfTransform.rotation = rotation;
+            rigidbody.position = position;
+            rigidbody.rotation = rotation;
+            rigidbody.velocity = Vector3.zero;
+            rigidbody.angularVelocity = Vector3.zero;
```
Jannify: Will this result in abrupt velocity changes (20m/s -> Packet loss -> Teleport with 0 m/s -> 20 m/s) and maybe provoke other abnormal behaviour?

Jannify (changes requested):

NitroxClient/MonoBehaviours/EntityPositionBroadcaster.cs
tornac1234: Happens once every 100ms, benchmarks showed it takes 0.0265ms for 245 spline entities and 168 regular entities. Although, the benchmarks also showed that notSpawnedEntityIds was filling up forever so I fixed this issue on the way.

NitroxClient/GameLogic/MovementHelper.cs
tornac1234: You're absolutely right, I also didn't want this function to alter states so I'll change it accordingly

NitroxClient/MonoBehaviours/RemotelyControlled.cs
tornac1234: Probably very few things go that fast or anywhere near that speed, but with one update every 0.1s, it means the object has moved 2 meters so even if you miss one your two packets, the speed correction system (adjusted velocity) should be able to cover up a few packet lost for creatures.

For regular items, they will never go that fast so it's fine to lose as many packets as you want. If you really have that bad of a connection, you can't really expect much from a multiplayer game ...

NitroxClient/MonoBehaviours/RemotelyControlled.cs
tornac1234: Will add this comment

```
// - non spline entities generally are dropped items/utility (like a flare or a pipe surface floater)
// we don't want the environment to interfere with their movement since the simulating player is already broadcasting movement
// that was affected by those interactions
// - spline entities (Creatures) follow movements based on the local game's calculations so they need to be affected by gravity
// to not stay blocked in the air (for leviathans and other fishes jumping out of the water)
// NB: WorldForces is responsible for applying gravity, currents, explosions
```

tornac1234:
Added a commit to answer the requested changes. Also deprecated MultiplayerVehicleControl for good since it was entirely replaced by MovementReplicator.

Nitrox.Model.Subnautica/Packets/SpawnEntities.cs
```diff
@@ -4,46 +4,37 @@
...
-        // Constructor for serialization. 
-        public SpawnEntities(List<Entity> entities, List<SimulatedEntity> simulations, List<AbsoluteEntityCell> spawnedCells, bool forceRespawn)
-        {
-            Entities = entities;
-            Simulations = simulations;
-            SpawnedCells = spawnedCells;
-            ForceRespawn = forceRespawn;
-        }
+    // Constructor for serialization. 
+    public SpawnEntities(List<Entity> entities, List<SimulatedEntity> simulations, List<AbsoluteEntityCell> spawnedCells, bool forceRespawn)
```
dartasen, line 33:
```suggestion
    public SpawnEntities(List<Entity> entities, List<SimulatedEntity> simulatedEntities, List<AbsoluteEntityCell> spawnedCells, bool forceRespawn)
```
To match new constructor parameter

Nitrox.Model/Extensions/DictionaryExtensions.cs
```diff
@@ -30,6 +30,30 @@ public static void RemoveWhere<TKey, TValue, TParameter>(this IDictionary<TKey,
...
+            {
+                dictionary.Remove(toRemove[i]);
+            }
+        }
+        finally
+        {
+            ArrayPool<TKey>.Shared.Return(toRemove, true);
+        }
+    }
+
```
dartasen, line 56: Could we add unit tests for that extension ? 🙏

NitroxClient/MonoBehaviours/RemotelyControlledPipeFloater.cs
```diff
@@ -1,22 +1,27 @@
...
-    public new void Awake()
+    private bool positioned;
+
+    public void SetPositioned()
     {
-        base.Awake();
-        pipeSurfaceFloater = gameObject.GetComponent<PipeSurfaceFloater>();
+        positioned = true;
+        rigidbody.isKinematic = true;
+        rigidbody.interpolation = UnityEngine.RigidbodyInterpolation.None;
```
dartasen, line 14: Maybe we should document the fact that isKinematic false will cause pipe to appear broken

NitroxClient/MonoBehaviours/EntityPositionBroadcaster.cs
```diff
@@ -38,73 +69,142 @@ public void Update()
...
-    public static void StopWatchingEntity(NitroxId id)
+    /// <summary>
+    /// For each tracked entity, ensures it stays in the right HashSet/Dictionary depending on its state.
+    /// Either the entity has not spawned (<see cref="notSpawnedEntityIds"/>) or it follows a spline (<see cref="splineEntities"/>)
+    /// or in the default case (<see cref="regularEntities"/>).
+    /// </summary>
+    private void CheckEntities()
+    {
+        // when fishes die, they're only a corpse and their swim behaviour stops functioning
+        splineEntities.RemoveWhere(pair =>
```
dartasen, line 146: I find the usage of RemoveWhere in this class really weird, we're adding side effects to a boolean predicate

NitroxPatcher/Patches/Dynamic/Vehicle_OnPilotModeEnd_Patch.cs
```diff
@@ -12,11 +11,6 @@ public sealed partial class Vehicle_OnPilotModeEnd_Patch : NitroxPatch, IDynamic
     public static void Prefix(Vehicle __instance)
     {
         Resolve<Vehicles>().BroadcastOnPilotModeChanged(__instance.gameObject, false);
-        // Fixes instances of vehicles stuck on nothing by forcing the workaround (let another player enter and leave the vehicle)
-        if (__instance.TryGetComponent(out MultiplayerVehicleControl mvc))
-        {
-            mvc.Exit();
-        }
```
dartasen, line 19: 😱

Nitrox.Server.Subnautica/Models/GameLogic/Entities/EntitySimulation.cs
```diff
@@ -130,6 +131,17 @@ public bool TryAssignEntityToPlayer(Entity entity, Player player, bool shouldEnt
         return false;
     }
 
+    /// <summary>
+    /// Forcefully assign an entity to a player, revoking any previous ownership.
+    /// </summary>
+    public SimulatedEntity AssignEntityToPlayer(Entity entity, Player player, bool shouldEntityMove)
+    {
+        simulationOwnershipData.RevokeOwnerOfId(entity.Id);
+        simulationOwnershipData.TryToAcquire(entity.Id, player, DEFAULT_ENTITY_SIMULATION_LOCKTYPE);
```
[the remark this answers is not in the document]
tornac1234: The point of EntitySpawnedByClient is that we trust the client to spawn things from their "initiative" so acquiring forcefully is on purpose (for example when you pickup) because the spawned entity is also a reimplacement for the already existing one (previously an item in inventory but now an item in water which needs simulation).
I'll create a simulationOwnershipData.ForceAcquire to make the purpose clear and intended though.

NitroxPatcher/Patches/Dynamic/Vehicle_OnPilotModeEnd_Patch.cs
tornac1234: MVC is deprecated but it was using some utilities that were modified in this PR so instead of updatint it, I just deleted it

NitroxClient/MonoBehaviours/EntityPositionBroadcaster.cs
tornac1234: If you can suggest a better name for RemoveWhere please, I've been thinking but it's really hard. Also I really think the performance gain from it is worth it. So either we rename it or we copy paste the code twice 😩

PR #2877, opened 2026-08-30
Allow playing through the intro cinematic in singleplayer

SpaceMonkeyy86:
## Description of Change

This PR fixes a bug where skipping waiting for another player to begin the intro sequence would unintentionally skip the intro sequence as well. This is because, when moving to a newer Subnautica version with a new input system, some code was never ported over which properly reset the held time of the Escape key. Since both of these skips involve holding the same key for 1.5s, the code skipped both as soon as that 1.5s time was reached.

I also added a new command-line argument, `--noskipintro`, which allows playing through the intro sequence when not on a release build. The old behavior was to always skip this sequence for development builds, and that is still the default.

Put together, these two changes allow easy testing of the intro sequence, since it is now accessible on a development build and with only one player.

## Validation

1. Create a new server in Survival mode and join it with one player. You should see the text "Waiting for another player to join", which would never appear previously on a development build.
2. Hold Escape to skip this prompt. You should see the text "Unknown World Entertainment and the Nitrox Team present", and keeping the Escape key held down should start to skip this prompt as well. Previously, this text could never appear and the player would instantly get put into regular gameplay in a post-intro escape pod.

## PR Checklist

- [ ] Has a linked issue

---

### DIFF
--- a/NitroxClient/MonoBehaviours/IntroCinematicUpdater.cs
+++ b/NitroxClient/MonoBehaviours/IntroCinematicUpdater.cs
@@ -7,7 +7,7 @@ namespace NitroxClient.MonoBehaviours;
 
 public class IntroCinematicUpdater : MonoBehaviour
 {
-    public static RemotePlayer Partner;
+    public static RemotePlayer? Partner;
     private static Transform modelRoot;
 
     private static readonly Transform playerTransform = Player.main.transform;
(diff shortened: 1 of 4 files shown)


Measurity (approved):

MistaOmega:
"I like it, picasso"
lgtm <3

MistaOmega (approved):

PR #2880, opened 2026-08-31
Fix launcher being unable to find world backups

### LINKED ISSUE #2879: Launcher can't find world backups
### Describe the issue

My server has two backups in the backups folder on disk, but the launcher says "No backups here yet" when I use the "Restore a Backup" option. Restarting the launcher does not fix the problem. This seems to be because the launcher checks for a folder named `Backups`, but the actual folder is named `backups`.

Launcher code to find backups: https://github.com/SubnauticaNitrox/Nitrox/blob/7f7363f3bc283059622e00d0ea954d01a37eef01/Nitrox.Launcher/ViewModels/BackupRestoreViewModel.cs#L80
Server code to set backups folder: https://github.com/SubnauticaNitrox/Nitrox/blob/7f7363f3bc283059622e00d0ea954d01a37eef01/Nitrox.Server.Subnautica/Extensions/ServerStartOptionsExtensions.cs#L67

SpaceMonkeyy86:
## Linked Issues / Context

Fixes #2879 

## Description of Change

The launcher and server used two different names of the world backups folder, so the launcher couldn't find backups made by the server. This PR changes them to use the same name.

## Validation

Create and start a server, and ensure it saves the world at least once. Then find the server in the launcher and click Manage, then click Restore a Backup. You should see at least one backup listed; previously the list was always empty.

## PR Checklist

- [x] Has a linked issue

---

### DIFF
--- a/Nitrox.Launcher/ViewModels/BackupRestoreViewModel.cs
+++ b/Nitrox.Launcher/ViewModels/BackupRestoreViewModel.cs
@@ -77,7 +77,7 @@ public partial class BackupRestoreViewModel : ModalViewModelBase
         {
             yield break;
         }
-        string backupDir = Path.Combine(saveDirectory, "Backups");
+        string backupDir = Path.Combine(saveDirectory, "backups");
         if (!Directory.Exists(backupDir))
         {
             yield break;


Measurity (approved):
Thanks!

Nitrox.Launcher/ViewModels/BackupRestoreViewModel.cs
```diff
@@ -77,7 +77,7 @@ IEnumerable<string> GetBackupFilePaths(string backupRootDir) =>
         {
             yield break;
         }
-        string backupDir = Path.Combine(saveDirectory, "Backups");
+        string backupDir = Path.Combine(saveDirectory, "backups");
```
dartasen, line 80: Maybe we could have a shared constant inside NitroxDirectory or any related class

SpaceMonkeyy86: Not really necessary imo, I don't think the folder name will ever change

PR #2881, opened 2026-09-01
Sync fires

### LINKED ISSUE #2230: Add sync for Fires
### Describe the issue

- [ ] Cyclops
- [ ] Aurora
- [ ] Escapepod
- [ ] Other ?

SpaceMonkeyy86:
## Linked Issues / Context

Fixes #2230 

## Description of Change

This PR aims adds/fixes sync for all types of fire in the game. Most of the code for fire sync was already there, but it contained several logic errors that prevented it from working most of the time or at all. For example, the old code to send FireDoused packets accidentally sent the wrong value for DouseAmount, which was only about 3% of the correct value, so it effectively did nothing. The code to register new Cyclops fires also failed to run because it expected the new fire to already have an entity tag, which it never did, rather than just adding one itself.

Features:
- Sync for the escape pod fire in the intro sequence
  - This fire is not saved to the server's entity registry, and the client will simply delete the fire if the player chooses to skip the intro. This fire is only synced if both players let the intro sequence play out, as otherwise there is no point.
- Sync for fires in the Aurora (no special handling needed, they are regular cell entities)
- Sync for Cyclops fires; they now work roughly the same way as Cyclops damage points

I also refactored the Cyclops damage system to remove lots of unnecessary information. Both damage points and fires still function properly after this change. However, this system should be reworked, since the information isn't persisted on the server yet (no entity system integration). This also applies to Cyclops fires, but those should be reworked at the same time as damage points since the two are so similar. Also, Cyclops health sync does not work properly, as other clients will only register a health update when the pilot sends a metadata update by changing the Cyclops speed or other HUD settings, by exiting pilot mode, etc. I think all of this should be saved for a separate PR.

## Validation

Escape pod fire: create a new server in Survival mode, then join the game with two players. Make sure to use the `--noskipintro` launch argument. Sit through the opening cutscene and the fire should be synced.

Aurora fires: `goto aurora` for both players and try putting out one of the fires.

Cyclops fires: Spawn a Cyclops, then have one player set the engine speed to max and drive around for about 30 seconds; this spawns a fire in the engine room. Then try putting it out with the other player watching.

Cyclops damage points: Make sure both players are in Survival mode. `goto lavacastlebase`, then swim above the structure and spawn a Cyclops (make sure it isn't clipping into the terrain). There may be better spots for this with more room for a Cyclops; the important thing is that you are below the maximum depth of 500m. Start piloting the Cyclops with one player and sit still until a damage point spawns. Then go outside with both players to check if they both see it, and try repairing it.

## PR Checklist

- [x] PR rebased on top of `main` at the time of opening (close enough :p)
- [x] Has a linked issue

---


Nitrox.Model.Subnautica/Packets/FireDoused.cs
```diff
@@ -1,25 +1,29 @@
...
+        Id = id;
+        Health = health;
+        DouseAmount = douseAmount;
     }
+
+    public NitroxId Id { get; }
+    public float Health { get; }
+    public float DouseAmount { get; }
+    [IgnoredMember]
+    public bool IsExtinguished => Health <= 0;
```
Measurity: Thanks 😅

dartasen: Nit: If we could keep properties above the constructor 😄

Nitrox.Model.Subnautica/Packets/CyclopsDamagePointCreated.cs
```diff
@@ -0,0 +1,27 @@
...
+    ///     The created point's index in <see cref="CyclopsExternalDamageManager.damagePoints" />
+    /// </param>
+    public CyclopsDamagePointCreated(NitroxId id, int damagePointIndex)
+    {
+        Id = id;
+        DamagePointIndex = damagePointIndex;
+    }
+
+    public NitroxId Id { get; }
+    public int DamagePointIndex { get; }
```
dartasen: Same, if we could keep properties above the actor

Nitrox.Model.Subnautica/Packets/CyclopsDamagePointCreated.cs
```diff
@@ -0,0 +1,27 @@
...
+
+    public override string ToString()
+    {
+        return $"[CyclopsDamagePointCreated Id: {Id} DamagePointIndex: {DamagePointIndex}]";
+    }
```
dartasen: I think the ToString is already auto-generated by the Packet class

Nitrox.Server.Subnautica/Models/Packets/Processors/FireDousedProcessor.cs
```diff
@@ -1,11 +1,19 @@
...
+internal sealed class FireDousedProcessor(EntityRegistry entityRegistry) : IAuthPacketProcessor<FireDoused>
 {
+    private readonly EntityRegistry entityRegistry = entityRegistry;
+
     public async Task Process(AuthProcessorContext context, FireDoused packet)
     {
+        if (packet.IsExtinguished)
+        {
+            entityRegistry.RemoveEntity(packet.Id);
+        }
```
dartasen, line 15: That's weird we never unregistered those entities in the past 😦

Nitrox.Server.Subnautica/Models/Packets/Processors/SetIntroCinematicModeProcessor.cs
```diff
@@ -32,6 +33,8 @@ public async Task Process(AuthProcessorContext context, SetIntroCinematicMode pa
 
             await context.SendToAllAsync(new SetIntroCinematicMode(allWaitingPlayers[0].SessionId, IntroCinematicMode.START, allWaitingPlayers[1].SessionId));
             await context.SendToAllAsync(new SetIntroCinematicMode(allWaitingPlayers[1].SessionId, IntroCinematicMode.START, allWaitingPlayers[0].SessionId));
+
+            await escapePodManager.SetupIntroSequenceAsync(allWaitingPlayers[0], allWaitingPlayers[1]);
```
dartasen: Nit: I guess we could have a local var to avoid accessing the allWaitingPlayers array every time in this if branch

NitroxClient/Communication/Packets/Processors/FireDousedProcessor.cs
```diff
@@ -1,28 +1,60 @@
...
-            fireGameObject.RequireComponent<Fire>().Douse(packet.DouseAmount);
+            using (PacketSuppressor<FireDoused>.Suppress())
+            {
+                fire.Douse(douseAmount);
+            }
         }
+        else
+        {
+            // Fire health went up
+            fire.livemixin.health = packet.Health;
```
dartasen, line 65: Is it possible ? (besides the desync that could occur)

NitroxClient/GameLogic/Cyclops.cs
```diff
@@ -136,63 +83,48 @@ private void BroadcastDamageState(SubRoot subRoot, Optional<DamageInfo> info)
...
-        /// Get all of the index locations of all the fires on the <see cref="SubRoot"/>. <see cref="SubFire.RoomFire.spawnNodes"/> contains
-        /// a static list of all possible fire nodes.
-        /// </summary>
-        private IEnumerable<CyclopsFireData> GetActiveRoomFires(SubFire subFire)
-        {
-            if (!subFire.subRoot.TryGetIdOrWarn(out NitroxId subRootId))
-            {
-                yield break;
-            }
+            int index = Array.IndexOf(subRoot.damageManager.damagePoints, damagePoint);
```
dartasen, line 109: IndexOf could return a negative result

NitroxClient/GameLogic/Cyclops.cs
```diff
@@ -136,63 +83,48 @@ private void BroadcastDamageState(SubRoot subRoot, Optional<DamageInfo> info)
...
+        // Remake of the StartSystem Coroutine from original player. Some Methods are not used from the original coroutine
+        // For example no temporaryClose as this will be initiated anyway from the originating Player
+        // Also the fire extiguishing will not start cause the initial player is already extiguishing the fires. Else this could double/triple/... the extinguishing
+        private IEnumerator StartFireSuppressionSystem(SubFire fire)
+        {
+            fire.subRoot.voiceNotificationManager.PlayVoiceNotification(fire.subRoot.fireSupressionNotification, false, true);
+            yield return Yielders.WaitFor3Seconds;
+            fire.fireSuppressionActive = true;
+            fire.subRoot.fireSuppressionState = true;
+            fire.subRoot.BroadcastMessage("NewAlarmState", null, SendMessageOptions.DontRequireReceiver);
```
dartasen: Nit: Same for TemporaryLock : Is it possible to use nameof() ?

NitroxPatcher/Patches/Dynamic/uGUI_SceneIntro_IntroSequence_Patch.cs
```diff
@@ -117,6 +117,11 @@ public static void SkipLocalCinematic(uGUI_SceneIntro uGuiSceneIntro, bool wasNe
         Transform introFireHolder = EscapePod.main.transform.Find("Intro");
         if (introFireHolder) // Can be null if called very early
         {
+            if (introFireHolder.TryGetComponentInChildren(out NitroxEntity fireEntity))
```
dartasen:
```suggestion
            if (introFireHolder.TryGetComponentInChildren(out NitroxEntity? fireEntity))
```

NitroxPatcher/Patches/Dynamic/CyclopsExternalDamageManager_CreatePoint_Patch.cs
```diff
@@ -1,26 +1,40 @@
...
-        if (__state)
+        (bool hasLock, List<CyclopsDamagePoint> damagePoints) = __state;
+        if (hasLock)
         {
-            Resolve<Cyclops>().OnCreateDamagePoint(__instance.subRoot);
+            foreach (CyclopsDamagePoint damagePoint in damagePoints)
+            {
+                if (!__instance.unusedDamagePoints.Contains(damagePoint))
+                {
+                    int index = Array.IndexOf(__instance.damagePoints, damagePoint);
```
dartasen, line 32: same index could be negative (in theory)

NitroxPatcher/Patches/Dynamic/CyclopsExternalDamageManager_CreatePoint_Patch.cs
```diff
@@ -1,26 +1,40 @@
...
+
+        // Save the current damage state so we can find what changed in the postfix
+        __state = (hasLock, [.. __instance.unusedDamagePoints]);
 
-        return __state;
+        return hasLock;
     }
 
-    public static void Postfix(CyclopsExternalDamageManager __instance, bool __state)
+    public static void Postfix(CyclopsExternalDamageManager __instance, (bool, List<CyclopsDamagePoint>) __state)
```
dartasen: I guess you could use [__runOriginal ](https://harmony.pardeike.net/articles/patching-injections.html#__runoriginal) to know if the method has been skipped (therefore if the player has a lock or not)

dartasen (commented):
Thank's for the PR :smiling:

Nitrox.Server.Subnautica/Models/Packets/Processors/FireDousedProcessor.cs
SpaceMonkeyy86: Yeah, also I got confused as to why this method had so few usages, but I think most things are just done with `EntityDestroyed`. This packet is an exception since we want to call `Fire.Extinguish()` specifically

NitroxClient/Communication/Packets/Processors/FireDousedProcessor.cs
SpaceMonkeyy86: Yes, fire health will regenerate on its own, so if someone starts extinguishing a fire after it has been regenerating for a while, the other players might have a slightly lower fire health due to packet latency.

NitroxClient/GameLogic/Cyclops.cs
SpaceMonkeyy86: It shouldn't ever fail to find the damage point, since the `damagePoints` list contains a list of all possible damage points and never changes. I'll add a safety check just in case, so we don't send any invalid packets

NitroxPatcher/Patches/Dynamic/uGUI_SceneIntro_IntroSequence_Patch.cs
SpaceMonkeyy86: Why is this needed?

dartasen: with @Measurity we're slowly trying to annotate the codebase against [nullable](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-value-types) warnings.

Currently the codebase is using `<Nullable>annotations</Nullable>` to allow using those annotations and attributes without generating any warnings

NitroxClient/GameLogic/Cyclops.cs
dartasen: Yes maybe we should add a log as well, so we don't silently fail/send invalid data

dartasen:
Tested with 3 players : 

* Seems to work inside Escapepod
* When dousing a aurora fire with multiple people at the same time : 
```txt
[20:20:52.799] [ERR] Error trying to process packet [FireDoused: Id: fed30f06-d935-60a2-d4c6-5ab331b72478, Health: 8,47112, DouseAmount: 0,138892, IsExtinguished: False]
Nitrox.Model.DataStructures.OptionalEmptyException`1[UnityEngine.GameObject]: Optional <T> is empty:
    Game object required from id: fed30f06-d935-60a2-d4c6-5ab331b72478
  at Nitrox.Model.Helper.Validate.IsPresent[T] (Nitrox.Model.DataStructures.Optional`1[T] opt, System.String message) [0x00016] in <ebe7ab6649604432ac61e5e61b62e8f0>:0 
  at NitroxClient.MonoBehaviours.NitroxEntity.RequireObjectFrom (Nitrox.Model.DataStructures.NitroxId id) [0x00014] in <53e056b29a6d468f9143fdf1e4c5044d>:0 
  at NitroxClient.Communication.Packets.Processors.FireDousedProcessor.Process (NitroxClient.Communication.Packets.Processors.Core.ClientProcessorContext context, Nitrox.Model.Subnautica.Packets.FireDoused packet) [0x00007] in <53e056b29a6d468f9143fdf1e4c5044d>:0 
```
* Aurora fires are respawning when we relog
* Cyclops fire sync doesn't seem to work at all

* No sound / SFX when using the fire extinguisher (maybe we could steal @tornac1234's work) which makes the sync a bit sad (this can still be done in a separate PR though)

SpaceMonkeyy86:
Since the temporary cyclops fire sync seems to not be working, I will go ahead and redo cyclops damage sync in this PR so we don't have to test and review a system that will be replaced anyway

SpaceMonkeyy86:
PR is now ready for review, with fully working Cyclops health and damage sync. All the issues from the previous test session have been fixed

PR #2883, opened 2026-09-02
Add ban command (#1734)

### LINKED ISSUE #1734: Add a ban command
We need a ban command for the future.

SmokerDoker:
## Linked Issues / Context

Closes #1734.

## Description of Change

The commands themselves are small; the decisions behind them are the part worth reviewing.

**Bans are purely IP-based.** Player names are client-supplied and free to change, so a name-based ban is theatre — the same person types a different name and walks back in. The IP is the only identifier the server currently holds that a returning griefer can't simply retype. This is deliberately pragmatic rather than airtight: VPN and dynamic-IP bypass are explicitly out of scope, and so is LAN, where NAT makes per-player IPs meaningless and a server password is the right tool anyway (as raised by @Measurity on the issue back in 2022). The goal is to raise the cost of coming back in the common case, not to solve identity. That's also why the name stored on a ban entry is only a display label for `banlist` and is never read during enforcement: someone banned as `Griefer` stays banned as `Griefer2`.

**Rejection runs through the existing reservation handshake** via a new `PLAYER_BANNED` state, instead of dropping the connection. That reuses the same `[Description]` mechanism already behind `AUTHENTICATION_FAILED`, so a banned player is told *why* they can't join instead of hitting a silent failure. The rejection path itself needs no client-side changes; the only client-side addition is the optional ban detail described below.

**The banned player sees the reason and expiry** (@Measurity). The reservation packet carries an optional detail string; the server fills it from the ban entry (`Reason: …`, plus either an expiry timestamp or "permanent") and the join-menu dialog appends it under the generic message. That detail is resolved purely from the connection's own transport source address and is only ever sent back on that same connection — there is no client-supplied IP parameter, so a player cannot query anyone else's ban.

**A ban no longer traps the join menu in a retry loop** (@Measurity). On a `PLAYER_BANNED` rejection the client now tears down the session and returns to the server list, instead of immediately reconnecting and getting rejected again — which re-showed the same dialog and made the "OK" button look like it needed a second click. Recoverable rejections (wrong password, name taken, server full) keep retrying as before.

**Bans persist in their own `Bans.json`, outside the `WorldService` save cycle.** Moderation state isn't world state. Keeping it separate means saves created before this PR keep loading untouched (no file simply means no bans), a server owner can inspect or wipe the ban list without touching the world, and a problem with the ban file can never take the save down with it.

**Temporary bans are included from the start** rather than deferred to a follow-up. "Cool off for a day" is the common moderation case, and adding expiry to a persisted format after the fact means a migration for everyone who already has a ban file. A permanent ban is simply an entry with no expiry.

**Moderators can't ban their peers.** A target with equal or higher permissions is rejected, so mods and admins can't ban each other; the server owner and the console outrank everyone and stay the escalation path for a rogue moderator. Conservative on purpose — loosening this later is a one-character change, tightening it after the fact isn't.

Surface, for reference: `ban <player|ip> [reason] [duration]`, `unban <ip>` and `banlist`, all gated behind `MODERATOR`, following the existing `kick`/`mute` command conventions.

## Validation

- Build and the existing test suite are green (259 tests). No new unit tests: the meaningful behaviour here is the connection-time rejection path, which isn't reachable from the current test setup.
- **Not yet validated in-game end to end.** Reconnect rejection in particular needs two machines on separate networks to test honestly, which I don't have set up at the moment. Specifically unverified: kick-on-ban, the rejection message a banned client actually sees on reconnect (including the reason/expiry text and the return-to-server-list behaviour), expiry rollover, `unban`, and `banlist` output.

I'll update this section once I've run it properly, and I'd rather this not be merged before someone has.

## PR Checklist

- [ ] Has tests for the changes (if applicable) — see Validation
- [x] Has a linked issue
- [ ] Has been tested in-game — see Validation
- [ ] Has been tested on Windows/Linux/macOS

---


Nitrox.Server.Subnautica/Models/GameLogic/Players/Bans/BanManager.cs
```diff
@@ -0,0 +1,132 @@
...
+        return bansByIp.Values.ToList();
+    }
+
+    private void Load()
+    {
+        string filePath = FilePath;
+        if (!File.Exists(filePath))
+        {
+            return;
+        }
```
Measurity: The recommended way is to read the file and handle the exception if not found.

This is because with `File.Exists` it's possible for a file to get added/disappear/locked before reading it.

Nitrox.Server.Subnautica/Models/GameLogic/Players/Bans/BanManager.cs
```diff
@@ -0,0 +1,132 @@
...
+using Nitrox.Server.Subnautica.Models.AppEvents;
+using Nitrox.Server.Subnautica.Models.Serialization;
+
+namespace Nitrox.Server.Subnautica.Models.GameLogic.Players.Bans;
+
+/// <summary>
+///     Tracks IP bans. Keeps its own save file (independent from <see cref="Serialization.World.WorldService" />'s
+///     save/load cycle) so that old saves created before this feature existed keep loading fine.
+/// </summary>
+internal sealed class BanManager(ServerJsonSerializer serializer, IOptions<ServerStartOptions> startOptions, ILogger<BanManager> logger) : IHostedService, ISaveState
```
Measurity: `BanManager` -> `BanService`
and move to `Nitrox.Server.Subnautica/Services` namespace

To keep in line with other types that are also `IHostedService`.

Nitrox.Server.Subnautica/Models/Commands/BanCommand.cs
```diff
@@ -0,0 +1,124 @@
...
+{
+    [GeneratedRegex(@"^(\d+)(s|m|h|d|w)$", RegexOptions.IgnoreCase)]
+    private static partial Regex DurationRegex();
+
+    [Description("Bans a player (online or offline) or a raw IP address, kicking them if currently online")]
+    public async Task Execute(ICommandContext context,
+                              [Description("Player name or IP address")] string target,
+                              [Description("Ban reason")] string reason = "",
+                              [Description("Duration like 30m/12h/7d/2w, omit for permanent")] string duration = "")
+    {
```
Measurity: The command system supports custom argument types like `Player` which will try resolve by name / session id.

(See `PlayerNameToPlayerArgConverter` and `SessionIdToPlayerArgConverter`)

The IP address can be added as a separate `ICommandHandler<IPAddress>` which might need a custom `ArgConverter` to handle `string` -> `IPAddress`.

Doing it this way makes the command system more flexible for future command implementations and reduces complexity of the `BanCommand` implementation.

Nitrox.Server.Subnautica/Models/Commands/UnbanCommand.cs
```diff
@@ -0,0 +1,21 @@
...
+using Nitrox.Server.Subnautica.Models.Commands.Core;
+using Nitrox.Server.Subnautica.Models.GameLogic.Players.Bans;
+
+namespace Nitrox.Server.Subnautica.Models.Commands;
+
+[RequiresPermission(Perms.MODERATOR)]
+internal sealed class UnbanCommand(BanManager banManager) : ICommandHandler<string>
+{
+    [Description("Removes a ban by player name or IP address")]
+    public async Task Execute(ICommandContext context, [Description("Player name or IP address")] string target)
```
Measurity: Same here as with other command. Can use `Player` as parameter and another `ICommandHandler<IPAddress>` to handle IP only unbanning.

Nitrox.Server.Subnautica/Models/GameLogic/Players/Bans/BanManager.cs
```diff
@@ -0,0 +1,132 @@
...
+        string key = Normalize(ip);
+        if (!bansByIp.TryGetValue(key, out entry))
+        {
+            return false;
+        }
+        if (!entry.IsExpired)
+        {
+            return true;
+        }
+        bansByIp.Remove(key);
```
Measurity: Looks like the ban is removed when getting the ban here? Might be leftover from testing.

Measurity (changes requested):
Didn't test yet in-game. Thanks for the work!

Nitrox.Server.Subnautica/Models/GameLogic/Players/Bans/BanManager.cs
```diff
@@ -0,0 +1,132 @@
...
+/// </summary>
+internal sealed class BanManager(ServerJsonSerializer serializer, IOptions<ServerStartOptions> startOptions, ILogger<BanManager> logger) : IHostedService, ISaveState
+{
+    private readonly ThreadSafeDictionary<string, BanEntry> bansByIp = [];
+
+    private string FilePath => Path.Combine(startOptions.Value.GetServerSavePath(), $"Bans{serializer.FileEnding}");
+
+    public Task StartAsync(CancellationToken cancellationToken)
+    {
+        Load();
```
Measurity: This BanManager might be called before the `Load` is done. We can add a `TaskCompletionSource` and signal it after`Load`. Every `public` ban api call will need an `await` on this task before accessing `bansByIp`.

Nitrox.Server.Subnautica/Models/GameLogic/Players/Bans/BanManager.cs
```diff
@@ -0,0 +1,132 @@
...
+            foreach (string key in expiredKeys)
+            {
+                bansByIp.Remove(key);
+            }
+            Save();
+        }
+        return bansByIp.Values.ToList();
+    }
+
+    private void Load()
```
Coding-Hen: Other than manually calling unban or getactivebans with the list command there doesn't seem to be a way to automatically unban after the timeout. Would it make sense to have a check here to unban upon load or a timer to automatically unban during running?

Coding-Hen (commented):
There seems to be a mix of IP and name for banning here, the trygetban only works for IP but might not be for the player name they have which other areas of this code allow. 

Do we want to handle both cases of IP ban and player name ban, a user could just change their name and log back in but we might not want to fully up ban everyone behind an ip

SmokerDoker:
> There seems to be a mix of IP and name for banning here, the trygetban only works for IP but might not be for the player name they have which other areas of this code allow.
> 
> Do we want to handle both cases of IP ban and player name ban, a user could just change their name and log back in but we might not want to fully up ban everyone behind an ip

Also something I thought about. In my opinion I would say we should choose the complete IP Ban not the name or a combination of both, what do you guys think? @Measurity @Coding-Hen

Measurity:
Only IP ban is good I think. We don't have another method yet that can't be easily spoofed.

Nitrox.Server.Subnautica/Models/GameLogic/Players/Bans/BanManager.cs
Measurity: Good idea. Can use a `PeriodicTimer` and set it on the player that will be have their ban expire the earliest.

Just need to account for new bans that might expire sooner, requiring an update to the timer.

Coding-Hen:
Yeah, IP ban seems to be the most pragmatic. In that case do we need to have the IP match based on the name and not just all based on the ip

Nitrox.Server.Subnautica/Models/Commands/BanCommand.cs
```diff
@@ -0,0 +1,117 @@
...
+        List<Player> connectedFromIp = playerManager.GetConnectedPlayers()
+                                                    .Where(player => ip.Equals(sessionManager.GetEndPoint(player.SessionId)?.Address))
+                                                    .ToList();
+
+        if (connectedFromIp.Any(player => context.OriginId == player.SessionId))
+        {
+            await context.ReplyAsync("You can't ban yourself");
+            return;
+        }
+        Player outranking = connectedFromIp.FirstOrDefault(player => context.Permissions <= player.Permissions);
```
Coding-Hen, line 57: Do we want to block admins from blocking other admins? I think this is the right approach just wanted to make sure we don't want to allow it

Nitrox.Server.Subnautica/Models/Commands/BanCommand.cs
```diff
@@ -0,0 +1,117 @@
...
+        {
+            await context.ReplyAsync($"You're not allowed to ban {outranking.Name}");
+            return;
+        }
+
+        label ??= connectedFromIp.Count == 1 ? connectedFromIp[0].Name : null;
+
+        await banService.BanAsync(ip, label, reason, context.OriginName, parsedDuration);
+
+        foreach (Player player in connectedFromIp.Where(player => player.IsOnline))
```
Coding-Hen, line 80: Am I right in thinking this is only going to be players online anyway? We get this list from connected players which I thought is only current online players not all players who have ever connected. If that is the case does this stop us banning any players which aren't currently connected?

Nitrox.Server.Subnautica/Models/Commands/BanCommand.cs
```diff
@@ -0,0 +1,117 @@
...
+        string reasonText = string.IsNullOrEmpty(reason) ? "" : $" - {reason}";
+        string targetText = string.IsNullOrEmpty(label) ? ip.ToString() : $"{label} ({ip})";
+        await context.ReplyAsync($"Banned {targetText} {durationText}{reasonText}");
+    }
+
+    private static bool TryParseDuration(string duration, out TimeSpan? parsed)
+    {
+        if (string.IsNullOrWhiteSpace(duration))
+        {
+            parsed = null;
```
Coding-Hen: Do we need to set null here is it not by default null if not initialised?

Nitrox.Server.Subnautica/Models/Commands/BanCommand.cs
SmokerDoker: Yes and no. The /ban followed by the ip works always but theoretically someone could use your name to play on the Subnautica Server and if he would be banned then it would ban his IP because he is connected not yours, you can connect later on the same account/name with a different IP and it lets you trough. 

However the /ban name bans the ip not the name so a banned player playing as Griefer cannot come back as Griefer 2 for example. 

Hope that makes it clearer.

Nitrox.Server.Subnautica/Models/Commands/BanCommand.cs
SmokerDoker: parsed is an out parameter, not a field — C# doesn't auto-initialize those. The compiler requires definite assignment on every return path (CS0177), so this line is mandatory.

Nitrox.Server.Subnautica/Models/Commands/BanCommand.cs
SmokerDoker: Intentional — you can't ban someone of equal or higher rank. Prevents mods/admins from banning each other; a rogue peer still gets handled by the server owner or console, which outrank everyone. Easy to loosen to < later if we decide equal-rank bans should be allowed.

dartasen:
Thank's for the PR 😄

But the first thing that bothers me is that the PR description is written in German. Not that I hate _German_ and can't use a translator; but that the PR description was written entirely by AI without any human review. But that raises the following question: what about the code, and what about the tests done to ensure the code is right and functional ?

While we don't have anything against usage of AI while developing (as stated in the contributing guidelines), Pull Request have a **`default template`** for a reason, mainly because we're here to explain the context and the "why" of your PR. Not the "what" have been done inside the PR; I can assign copilot for a review and it would still give me a similar description.

SmokerDoker:
Ah, that explains it thanks for flagging. I usually read through my PR description before submitting, but it looks like Claude rewrote it via the GH CLI tool after that, without me noticing, which is how it ended up in German. My mistake for not catching it before submitting.

To be clear on process: I always document what I have and haven't tested directly in the PR description. Some things I can verify locally, others need more setup — the ban command, for example, needs two separate devices on different networks to properly test reconnect/kick behavior, which I don't always have readily available. Either way, testing status is never left ambiguous; it's spelled out in the description itself.

I appreciate the code review either way — catching edge cases I missed is exactly what review is for. I'll fix the PR description now and make sure it follows the template properly going forward.

Measurity:
Made some changes. See commit messages.

On another note: when banned, would be nice if the reason for the ban was displayed in-game when trying to connect to server.

I also noticed I need to click the "OK" button twice before the "ban message window" (see below) disappears.
image

SmokerDoker:
Both points from your review are implemented in 3279d67:

- **Ban reason shown in-game.** The rejected reservation now carries an optional
  detail string — the server fills it with the ban reason and expiry (or
  "permanent") and the join dialog shows it under the generic message. It's
  resolved only from the connecting socket's own transport address and only sent
  back on that same connection, so a player can't read someone else's ban reason.
- **"OK needs two clicks" / retry loop fixed.** A `PLAYER_BANNED` rejection now
  tears down the session and returns to the server list instead of silently
  reconnecting and re-showing the same dialog. Recoverable rejections (wrong
  password, name taken, server full) still retry as before.

Still not verified end-to-end in-game (that needs two machines on separate
networks) — called out in the PR description.

Measurity (approved):
LGTM

Nitrox.Server.Subnautica/Models/Commands/ArgConverters/StringToIPAddressArgConverter.cs
```diff
@@ -0,0 +1,23 @@
...
+namespace Nitrox.Server.Subnautica.Models.Commands.ArgConverters;
+
+/// <summary>
+///     Converts a string to an <see cref="IPAddress" />, if it is a valid IPv4/IPv6 literal.
+/// </summary>
+internal sealed class StringToIPAddressArgConverter : IArgConverter<string, IPAddress>
+{
+    public Task<ConvertResult> ConvertAsync(string value)
+    {
+        if (!IPAddress.TryParse(value, out IPAddress address))
```
dartasen:
```suggestion
        if (!IPAddress.TryParse(value, out IPAddress? address))
```

Nitrox.Server.Subnautica/Models/Commands/BanCommand.cs
```diff
@@ -0,0 +1,90 @@
...
+    : ICommandHandler<Player, string, TimeSpan>, ICommandHandler<IPAddress, string, TimeSpan>
+{
+    [Description("Bans an online player by their current IP address, kicking them")]
+    public async Task Execute(ICommandContext context,
+                              [Description("Player to ban")] Player target,
+                              [Description("Ban reason")] string reason = "",
+                              [Description("Duration like 30m/12h/7d/2w, omit for permanent")]
+                              TimeSpan duration = default)
+    {
+        IPEndPoint endPoint = sessionManager.GetEndPoint(target.SessionId);
```
dartasen:
```suggestion
        IPEndPoint? endPoint = sessionManager.GetEndPoint(target.SessionId);
```

Nitrox.Server.Subnautica/Services/LiteNetLibServerService.cs
```diff
@@ -193,11 +193,12 @@ private void OnConnectionRequest(ConnectionRequest request)
...
-        {
-            request.Reject();
-            return;
-        }
+        // TODO: Enable this check when client can report LiteNetLib rejection message. Then remove check done in PlayerManager.ReservePlayerContext
+        // if (server.ConnectedPeersCount >= options.Value.MaxConnections)
+        // {
+        //     request.Reject();
+        //     return;
+        // }
```
dartasen: We should probably open an issue about it if it's important, otherwise we can just remove the code

Nitrox.Server.Subnautica/Services/BanService.cs
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    /// <summary>
+    ///     Returns extra detail to show a banned player (the ban reason and, for a temporary ban, when it expires),
+    ///     or <see langword="null" /> when the IP has no active ban.
+    /// </summary>
+    public string? GetBanRejectionDetail(IPAddress ip)
+    {
+        loaded.Task.GetAwaiter().GetResult();
+        if (!bansByIp.TryGetValue(ip, out BanEntry entry) || entry.IsExpired)
```
dartasen:
```suggestion
        if (!bansByIp.TryGetValue(ip, out BanEntry? entry) || entry.IsExpired)
```

Nitrox.Server.Subnautica/Services/BanService.cs
```diff
@@ -0,0 +1,231 @@
...
+        await base.StartAsync(cancellationToken);
+    }
+
+    /// <summary>
+    ///     Returns whether the given IP address currently has an active (non-expired) ban.
+    /// </summary>
+    public bool IsBanned(IPAddress ip)
+    {
+        loaded.Task.GetAwaiter().GetResult();
+        return bansByIp.TryGetValue(ip, out BanEntry entry) && !entry.IsExpired;
```
dartasen:
```suggestion
        return bansByIp.TryGetValue(ip, out BanEntry? entry) && !entry.IsExpired;
```

Nitrox.Server.Subnautica/Services/BanService.cs
```diff
@@ -0,0 +1,231 @@
...
+            bansByIp.Remove(entry.IP);
+            logger.ZLogDebug($"A ban expired: {entry}");
+        }
+    }
+
+    private void Load()
+    {
+        try
+        {
+            BanData data = serializer.Deserialize<BanData>(FilePath);
```
dartasen:
```suggestion
            BanData? data = serializer.Deserialize<BanData>(FilePath);
```

dartasen (approved):

PR #2885, opened 2026-09-02
Fixed multiple capture scopes conflicting with each other

### LINKED ISSUE #2884: Summary command doesn't work InGame
Command does display the summary inside the server console, but it's blank for the player who's requesting it

Measurity:
Indirectly fixes the summary command because it caused 2 `CaptureScope`s which conflicted with each other.


dartasen (approved):
image

Works for me

PR #2889, opened 2026-09-04
Fixed task queue memory leak in LiteNetLibServerService

Measurity:
Confirmed that tasks are queued and awaited immediately in the new `TaskQueueService`.

Thanks to @tornac1234 for discovering this memory leak.

### DIFF
--- a/Nitrox.Server.Subnautica/Services/LiteNetLibServerService.cs
+++ b/Nitrox.Server.Subnautica/Services/LiteNetLibServerService.cs
@@ -3,7 +3,6 @@ using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Net;
-using System.Threading.Channels;
 using LiteNetLib;
 using LiteNetLib.Layers;
 using LiteNetLib.Utils;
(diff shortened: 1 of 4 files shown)


Nitrox.Server.Subnautica/Services/LiteNetLibServerService.cs
```diff
@@ -318,7 +314,7 @@ private async Task ProcessPacket(PeerContext peerContext, Packet packet)
             logger.ZLogError(ex, $"Error in packet processor {processor.GetType().Name:@TypeName}");
         }
 
-        static ProcessorTarget GetProcessorTarget(PacketProcessorsInvoker.Entry? processor, SessionId sessionId, PlayerManager playerManager, [NotNullIfNotNull(nameof(player))] out Player? player)
+        static ProcessorTarget GetProcessorTarget(PacketProcessorsInvoker.Entry? processor, SessionId sessionId, PlayerManager playerManager, [NotNullIfNotNull(nameof(player))] out Models.Player? player)
```
dartasen: not related to that PR but the NotNullIfNotNull annotation is weird

Nitrox.Server.Subnautica/Services/TaskQueueService.cs
```diff
@@ -0,0 +1,40 @@
...
+        taskChannel.Writer.TryComplete();
+        await base.StopAsync(cancellationToken);
+    }
+
+    public bool TryQueue(Task task) => taskChannel.Writer.TryWrite(task);
+
+    protected override async Task ExecuteAsync(CancellationToken stoppingToken)
+    {
+        uint processedTasks = 0;
+        // Cancellation token not used because reading will stop on quit due to "TryComplete" in "this.StopAsync".
```
dartasen: Is it wrong if we still pass in the cancellation token to avoid relying on an implementation ?

Nitrox.Server.Subnautica/Services/TaskQueueService.cs
```diff
@@ -0,0 +1,40 @@
...
+        await Task.Delay(200, cancellationToken);
+        taskChannel.Writer.TryComplete();
+        await base.StopAsync(cancellationToken);
+    }
+
+    public bool TryQueue(Task task) => taskChannel.Writer.TryWrite(task);
+
+    protected override async Task ExecuteAsync(CancellationToken stoppingToken)
+    {
+        uint processedTasks = 0;
```
dartasen: do we wanna count the tasks in Release mode for a trace log (thinking of public servers that can run for months, so the count will still overflow at some point (even if we use a bigger data type like long)

Nitrox.Server.Subnautica/Services/TaskQueueService.cs
Measurity: Yes because the token will be cancelled as soon as we stop. It throws errors that packets cannot be queued.

Nitrox.Server.Subnautica/Services/TaskQueueService.cs
```diff
@@ -0,0 +1,40 @@
...
+        // Cancellation token not used because reading will stop on quit due to "TryComplete" in "this.StopAsync".
+        await foreach (Task task in taskChannel.Reader.ReadAllAsync(CancellationToken.None))
+        {
+            unchecked
+            {
+                processedTasks++;
```
Coding-Hen: Not that I am too worried but do we need to worry about roll over?

Nitrox.Server.Subnautica/Services/LiteNetLibServerService.cs
Measurity: It's weird sounding but without it there are wrong null warnings in the calling code.

Nitrox.Server.Subnautica/Services/TaskQueueService.cs
Measurity: Removed the counting

Nitrox.Server.Subnautica/Services/TaskQueueService.cs
Measurity: Went back on this. Stopping the channel in `StoppedAsync` is safer than doing task.delay. So now I do use the cancellation token and process any remaining tasks in `StoppedAsync` as well.

dartasen (approved):

Coding-Hen (approved):

PR #2891, opened 2026-09-05
Fix PlayerHeldItemChanged NRE for items with null inventoryItem

### LINKED ISSUE #2460: PlayerHeldItemChanged NRE
### Describe the issue

```txt
[22:16:00.788] [ERR] Error while processing packet [PlayerHeldItemChanged: PlayerId: 6, ItemId: c054f03f-55c0-431c-819e-711be1dd6034, Type: DRAW_AS_TOOL, IsFirstTime: ]
System.ArgumentNullException: Value cannot be null.
Parameter name: pickupable.inventoryItem
  at NitroxModel.Helper.Validate.NotNull[T] (T o, System.String argumentExpression) [0x00009] in <2164b9cfe671423282f33e40b866e9d7>:0 
  at NitroxClient.Communication.Packets.Processors.PlayerHeldItemChangedProcessor.Process (NitroxModel.Packets.PlayerHeldItemChanged packet) [0x00040] in <d04259b6f32b473bb845b6ee4c9bebf2>:0 
  at NitroxClient.Communication.Packets.Processors.Abstract.ClientPacketProcessor`1[T].ProcessPacket (NitroxModel.Packets.Packet packet, NitroxModel.Packets.Processors.Abstract.IProcessorContext context) [0x00000] in <d04259b6f32b473bb845b6ee4c9bebf2>:0 
  at NitroxClient.MonoBehaviours.Multiplayer+<>c.<ProcessPackets>b__29_1 (NitroxModel.Packets.Packet packet, System.Collections.Generic.Dictionary`2[TKey,TValue] processorCache) [0x0000b] in <d04259b6f32b473bb845b6ee4c9bebf2>:0 
````

Reproduce steps : 
* Join with 2 players
* [Player1] Craft & Drop a vehicle mobile bay
* [Player1] Leave a slot empty in your quickslot bar
* [Player1] Pickup vehicle mobile bay
* [Player2] Shouldn't see player1 with a vehicle bay in his hand
* [Player2] Should see an error log

### PR DESCRIPTION
**Pull request by Malionaro**
## Linked Issues / Context

Fixes #2460

## Description of Change

Remote clients crashed with `ArgumentNullException: pickupable.inventoryItem` in `PlayerHeldItemChangedProcessor` whenever a player held an item that has no inventory wrapper, e.g. the Mobile Vehicle Bay (Constructor), which is a `Pickupable` + `PlayerTool` with null `inventoryItem`.

Changes in `NitroxClient/Communication/Packets/Processors/PlayerHeldItemChangedProcessor.cs`:
- Removed the `Validate.NotNull(pickupable.inventoryItem)` assertion that threw for such items.
- `HOLSTER_AS_TOOL`, `DRAW_AS_ITEM` and `HOLSTER_AS_ITEM` now reparent / show / hide via `Pickupable` directly (`pickupable.Reparent(...)`, `pickupable.SetVisible(...)`) and set layers on `item` instead of going through `pickupable.inventoryItem.item`. For normal items this is equivalent (`inventoryItem.item` is the same `Pickupable`); for items without an inventory wrapper it no longer throws.
- `DRAW_AS_TOOL` needed no change (never touched `inventoryItem`).

This ports the approach of #2591 (closed without merge) onto the current codebase.

## Validation

- `dotnet build NitroxClient`: 0 errors (only pre-existing DIMA001 warnings).
- `dotnet test Nitrox.Test`: 246 passed, 0 failed, 13 skipped (platform-specific).
- In-game validation still pending (needs 2 players, repro steps from #2460):
  1. Join with 2 players.
  2. [Player1] Craft and drop a Mobile Vehicle Bay, leave a quickslot empty, pick the bay up.
  3. [Player2] Before: error log + NRE. After: no error, bay visible in Player1 hand.

## PR Checklist

- [x] PR rebased on top of `master` (actual default branch) at the time of opening
- [x] Has tests for the changes (full existing suite passes with no regressions; no new unit test added as the processor requires the Unity runtime)
- [x] Has a linked issue (Fixes #2460)
- [ ] Has been tested in-game (if applicable) (pending, see validation steps above)
- [ ] Has been tested on Windows/Linux/macOS (if applicable) (pending)

---

🤖 This description was translated with the help of AI.

### DIFF
--- a/NitroxClient/Communication/Packets/Processors/PlayerHeldItemChangedProcessor.cs
+++ b/NitroxClient/Communication/Packets/Processors/PlayerHeldItemChangedProcessor.cs
@@ -42,7 +42,10 @@ internal sealed class PlayerHeldItemChangedProcessor : IClientPacketProcessor<Pl
         Pickupable pickupable = item.GetComponent<Pickupable>();
         Validate.IsTrue(pickupable);
 
-        Validate.NotNull(pickupable.inventoryItem);
+        // NB: Do NOT assert pickupable.inventoryItem here. It is null for items like the
+        // Mobile Vehicle Bay (Constructor), which is a Pickupable+PlayerTool without an
+        // inventory wrapper (see issue #2460). All uses below go through Pickupable directly,
+        // which is equivalent for normal items (inventoryItem.item is the same Pickupable).
 
         ItemsContainer inventory = player.Inventory;
         PlayerTool tool = item.GetComponent<PlayerTool>();
@@ -91,7 +94,8 @@ internal sealed class PlayerHeldItemChangedProcessor : IClientPacketProcessor<Pl
                 {
                     floater.collider.enabled = true;
                 }
-                pickupable.inventoryItem.item.Reparent(inventory.tr);
+                // inventoryItem can be null for items like the Mobile Vehicle Bay (Constructor, #2460)
+                pickupable.Reparent(inventory.tr);
                 foreach (Animator componentsInChild in tool.GetComponentsInChildren<Animator>())
                 {
                     componentsInChild.cullingMode = AnimatorCullingMode.CullUpdateTransforms;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: NitroxClient/Communication/Packets/Processors/PlayerHeldItemChangedProcessor.cs [hunk 1]
```diff
@@ -91,7 +94,8 @@ public Task Process(ClientProcessorContext context, PlayerHeldItemChanged packet
                 {
                     floater.collider.enabled = true;
                 }
-                pickupable.inventoryItem.item.Reparent(inventory.tr);
+                // inventoryItem can be null for items like the Mobile Vehicle Bay (Constructor, #2460)
```
Coding-Hen: These comments don't make sense to me, yes they say what the issue is but with the context of the previous call which used this removed and replaced with the line below inventoryItem isn't here and so this comment fit with the code. 

If we don't need the call anymore and the code below fixes that then I think these comments can be removed in my opinion.

Malionaro: Good point, thanks. I removed the comments. The Reparent and SetVisible calls stand on their own now.

dartasen:
Thank's for the PR. Though when modifying game code,  ensuring it does work in game without breaking anything is the first and only indicator to make a PR ready for review. Mainly because we can't trust an AI model that doesn't know anything about an undocumented proprietary code and even less when combined with the modding layers, architecture and custom logic we're adding on top.

Malionaro:
Understood. I will test the repro in game with two clients and post the results here before this goes anywhere.

PR #2892, opened 2026-09-07
Fixed cases where SessionId was compared with PeerId

### LINKED ISSUE #2287: Benches are broken
### Describe the issue

Fix it

Measurity:
Thanks @dartasen for finding the bug.

Also closes #2287 as it only took one line to fix

### DIFF
--- a/Nitrox.Server.Subnautica/Models/Player.cs
+++ b/Nitrox.Server.Subnautica/Models/Player.cs
@@ -22,8 +22,10 @@ namespace Nitrox.Server.Subnautica.Models
         public Optional<NitroxId>[] QuickSlotsBindingIds { get; set; }
 
         public PlayerContext? PlayerContext { get; set; }
-        public PeerId Id { get; init; }
+        /// <inheritdoc cref="Nitrox.Model.Core.PeerId" />
+        public PeerId PeerId { get; init; }
 
+        /// <inheritdoc cref="Nitrox.Model.Core.SessionId" />
         public SessionId SessionId
         {
             get => Interlocked.CompareExchange(ref sessionId, OFFLINE_SESSION_ID, OFFLINE_SESSION_ID);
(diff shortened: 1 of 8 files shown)


NitroxClient/GameLogic/Bases/BuildingHandler.cs
```diff
@@ -151,7 +151,7 @@ public IEnumerator ProgressConstruction(ModifyConstructedAmount modifyConstructe
             if (modifyConstructedAmount.ConstructedAmount >= 1f)
             {
                 constructable.SetState(true, true);
-                yield return BuildingPostSpawner.ApplyPostSpawner(gameObject, modifyConstructedAmount.GhostId);
+                yield return BuildingPostSpawner.ApplyPostSpawner(constructable.gameObject, modifyConstructedAmount.GhostId);
```
Jannify, line 154: Is this mistakenly included? If not what is the bug this fixes (was not in pr description).

Nitrox.Server.Subnautica/Models/Player.cs
```diff
@@ -176,7 +178,7 @@ public void Teleport(NitroxVector3 destination, Optional<NitroxId> subRootID, IP
 
         public override string ToString()
         {
-            return $"[Player - SessionId: {Id}, Name: {Name}, Perms: {Permissions}, Position: {Position}]";
+            return $"[Player - SessionId: {PeerId}, Name: {Name}, Perms: {Permissions}, Position: {Position}]";
```
Jannify:
```suggestion
            return $"[Player - PeerId: {PeerId}, Name: {Name}, Perms: {Permissions}, Position: {Position}]";
```

NitroxClient/GameLogic/Bases/BuildingHandler.cs
Measurity: Fixes benches, added to PR description. 

It's a clear mistake to call postspawner on BuildingHandler and not on the constructable gameobject. So I added it here instead of opening new PR for one line fix.

dartasen (approved):
Did not test, but the code changed is what I had in mind so lgtm

Jannify (approved):
LGTM CW

PR #2896, opened 2026-09-15
Add PrefabDebugger

dartasen:
## Description of Change

Requested by @tornac1234

Introduces a new debugger that allows to fetch and explore prefabs entries (pulled from `PrefabDatabase` class), giving the opportunity to inspect and play with those without digging in the game files or finding them in game.

Also added a QoL that selects the right tab and opens the SceneDebugger when inspecting a game object (used by SceneExtraDebugger also for ray casting)

image

image

## Features

* List all prefabs available in PrefabDatabase with the following informations : Prefab name, Class ID, TechType (if available)
* Allows to select a prefab from the list and renders it inside a cloned isolated preview object (only prefabs with renderers are displayed, ParticleRenderers aren't simulated and not needed in that PR)
* Allows to instanciate the prefab in front of the player
* Allows to inspect the preview prefab with the SceneDebugger (useful to inspect components for prefabs that can't be instanciated)
* Allows to manage and inspect instanciated prefabs in a different tab

## 3D Preview controls 
- Rotate with right mouse dragging
- Move the prefab in X/Y with left-drag.
- Move it in depth with Ctrl + left-drag
- Zoom with the mouse wheel
- Adjust manually with sliders for: Yaw, Pitch, X position, Y position, Z position, Camera distance 

## PR Checklist

- [ ] Has tests for the changes
- [ ] Has a linked issue
- [x] Has been tested in-game
- [x] Has been tested on Windows

---
🤖 This code was created with the help of AI (especially for the IMGUI part which is a nightmare to write by hand)


tornac1234:
Very cool feature, thanks for implementing it!

A few comments after testing it:
- I usually test Nitrox with a window 1280x768, and the prefab debugger is bigger than that it would seem.
- Would appreciate that "enter" also searchs.
- Also I'd probably invert the mouse scroll direction (scrolling up should zoom I think (as an example google maps does it that way)).
- Is the XYZ part supposed to turn around because it doesn't on my screen

